PR #137, opened 2014-12-06
Added Jack'O'Lantern for next Halloween Mission

hakan42:
Just in time (only another 11 months to go) for the next Mission to the Pumpkin Empire


zer0Kerbal:
https://github.com/KSP-CKAN/NetKAN/pull/7469

PR #226, opened 2014-12-19
Update KerbalAircraftExpansion.netkan

Teonlight:
#224 should be handled like this since the file is included in the KAX download. This also prevents all the firespitter parts from being installed.


Ippo343:
Hi @snjo , it's just the Resources.cfg file that's included in Firespitter (the one that defines FSCoolant).

Ippo343:
Pinging everyone here (@pjf, @snjo, @Teonlight ): what's the status on this? Still blocked?

Ippo343:
Re-poke. @pjf @snjo @Teonlight

Ippo343:
Third (and final) poke: @pjf , @snjo , @Teonlight

PR #463, opened 2015-01-06
Mod oriented tech tree

hakan42:


hakan42:
ckan creation and installation works, but the required TechManager is 0.25 only.

PR #467, opened 2015-01-07
Use the new identifier of the AdvancedJetEngine

hakan42:


hakan42:
Fixes KSP-CKAN/CKAN-meta#217

PR #468, opened 2015-01-07
Added Procedural Airships

hakan42:


hakan42:
With permission from the author: http://forum.kerbalspaceprogram.com/threads/99132-0-90-Procedural-Airships-1-3-dec-20th-2014?p=1654906&viewfull=1#post1654906

PR #470, opened 2015-01-07
HorizontalLandingAid Delete request

SirDiazo:
This is duplicate of RCSLandingAid, please delete.

Also, can you please check your bot or whatever is making these, I make very sure the "Add to CKAN" box is unchecked when I add my mods to kerbal stuff as I want CKAN pointing to my github and the bot messes up the install directory.


hakan42:
netkan file and expanded ckan deleted as requested.

I also added a kerbalstuff resource hint to your RCSLandAid.netkan, so the next one stumbling over this could actually recognize that the mod is actually contained.

Might I humbly suggest, though, that you clean up the entry over at kerbalstuff (for example, "source" points to your LandingHeight repo) to avoid misunderstandings like this?

PR #479, opened 2015-01-08
Netkans for all four different kinds of ATM

hakan42:
This requires https://github.com/KSP-CKAN/CKAN-netkan/pull/30 (asset_name regexp in github releases) to be merged.


AlexanderDzhoganov:
#rebuild

PR #486, opened 2015-01-09
Update SmartParts.netkan

SirDiazo:
Directory in GameData changed as new maintainer is taking over.


AlexanderDzhoganov:
#rebuild

AlexanderDzhoganov:
We have a problem here. The new maintainer has apparently chosen another versioning, this requires that we bump up the epoch part of the version so we don't confuse the two mods. This is doable by hand but we've got no way to tell NetKAN. I will work on a fix until then this PR shouldn't be merged.

PR #510, opened 2015-01-10
Added CrewManifest from sarbians Jenkins server

hakan42:


hakan42:
And the first Jenkins-inflated addon in the main repository :)

PR #511, opened 2015-01-10
Split PlanetShine into core and default config

SpaceTeph:
Preparatory work for other mods that tweak the PlanetShine config - Astronomer's Pack is one of them.

### DIFF
--- a/NetKAN/PlanetShine-Config-Default.netkan
+++ b/NetKAN/PlanetShine-Config-Default.netkan
@@ -1,0 +1,25 @@
+{
+    "spec_version"   : 1,
+    "identifier"     : "PlanetShine-Config-Default",
+    "$kref"          : "#/ckan/kerbalstuff/266",
+    "license"        : "Apache-2.0",
+    "install": [
+        {
+            "file": "GameData/PlanetShine/Config",
+            "install_to": "GameData/PlanetShine"
+        }
+    ],
+    "depends": [
+        {
+            "name": "PlanetShine"
+        }
+    ],
+    "provides": [
+        "PlanetShine-Config"
+    ],
+    "conflicts": [
+        {
+            "name": "PlanetShine-Config"
+        }
+    ]
+}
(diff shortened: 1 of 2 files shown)


hakan42:
#rebuild

hakan42:
Could you provide the pull requests for those splitting jobs as two separate PRs? Then, Jenkins would have a chance to show that the Config part is working, and after merging that, we could re-build the other part...

SpaceTeph:
Well, the issue is, this mod pair can't be automatically checked because it has a circular dependency. Those are apparently common in NetKAN, and they're useful: the config depends on the mod, the mod depends on the config... It means that the config cannot be independently installed without the mod and the other way around. Doesn't that mean that Jenkins will always fail the automated test? If yes, the only way for Jenkins to succeed would be to check the installation of both the mod and its config at the same time with something like

    ckan install -c built/PlanetShine-0.2.2.1.ckan built/PlanetShine-Config-Default-0.2.2.1.ckan

I get your point about using two separate PRs, but if I start by submitting the Config first, it will pass dependency checks but will fail when installing because it contains files that are part of the not-yet-changed PlanetShine mod in the repository...

hakan42:
Yes, I get you... So, if you are confident with this PR, gimme a ping and I'll merge it. Any errors can be repaired in upcoming PRs :smiling_imp:

SpaceTeph:
Then ping @hakan42, because I have tested this with the newly-changed, multiple-files -c option. My only ugly feeling is with the existing installations of the mod - do the clients get a heads-up and the possibility to update the mod when the NetKAN/CKAN data changed, but the mod version itself didn't?

AlexanderDzhoganov:
[*wink* *wink*](https://github.com/KSP-CKAN/CKAN-core/pull/41)

SpaceTeph:
You guys are good XD

SpaceTeph:
Ah, one issue pops up: since the mod name is generated by NetKAN, they have different IDs (PlanetShine vs PlanetShine-Config-Default) but the same display name in the CKAN GUI (PlanetShine). Whoops, didn't think of that.

hakan42:
Yes, just update the "name"  attribute in the netkans. I had the same issue with the BoxSat addon, and seldomly using the GUI, did not discover it until after people started complaining in the forums :blush:

PR #517, opened 2015-01-13
CC-SCANsat netkan had wrong file case

hakan42:


hakan42:
Strangely, Jenkins believes that the build hash for this change contains one billion netkan files. Closing this PR and reworking.

PR #523, opened 2015-01-14
Environmental Visual Enhancement: Core Mod and Config split, act 2

SpaceTeph:
So it seems my previous correction of the EVE split was only partially correct. Not only the directories were inverted, but the entire mod was too. It is not the core mod that needs to be split into HR and LR variants, but rather the default configuration.

This PR makes EnvironmentalVisualEnhancements-HR and -LR into default configuration packages in their respective resolutions, renames EnvironmentalVisualEnhancements-Config-Default into EnvironmentalVisualEnhancements and makes it into the core package. The whole thing now finally makes sense.


SpaceTeph:
I expect the merged build to fail because of the dependency loop between the mod and its config.

PR #524, opened 2015-01-15
Add CC-CP-SCANSat to CKAN

DBT85:
New contract pack for Contract Configurator.

### DIFF
--- a/NetKAN/CC-CP-SCANSat.netkan
+++ b/NetKAN/CC-CP-SCANSat.netkan
@@ -1,6 +1,6 @@
 {
     "spec_version": "1",
-    "identifier": "CC-CP-SCANSat",
+    "identifier": "ContractConfigurator-ContractPack-SCANsat",
     "$kref": "#/ckan/github/DBT85/CC-CP-SCANSat",
     "abstract": "A contract pack containing contracts for SCANSat.",
     "name": "Contract Pack: SCANSat",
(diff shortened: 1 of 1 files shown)


hakan42:
Out of curiosity, how are those different from the pack tattagreis created?

Also, we should come up with a nicer identifier that matches the two others :smile:

DBT85:
Tattagreis pack was only for LoRes Altimetry, had no optional extras like going to Ike, or Gilly or de-orbiting old satellites, no spanky orbit lines and some other bits. He said he wasn't going to maintain his pack as it was just a demo IIRC.

I went for CC (Contract configurator), CP (Contract Pack) and then the mod they work on :P. It just so happened to spell CCCP! 

It goes into the same ContractPacks folder as Nightingales pack, and in CKAN appears as "Contract Pack: ScanSat" just like Nightingales pack.

hakan42:
Ah, o.k., now I have some more insight :smile: 

My main point is about the "identifier" in line 3 though. Could you change that to "ContractConfigurator-ContractPack-SCANsat" (careful, the last "s" lowercased) so it matches https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/ContractConfigurator-SCANsat.netkan and https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/ContractConfigurator-RemoteTech.netkan ?

Then, the generated ckan files would end up in the vicinity of the other CC packages in git, making things easier for me to keep track of... The name in CKAN would not be affected by this, also not how it is displayed in the GUI.

DBT85:
Change to line 3 made as requested. 

now reads

"identifier": "ContractConfigurator-ContractPack-SCANsat",

hakan42:
#rebuild

hakan42:
Here you go. Thank you for your contribution :smile:

PR #529, opened 2015-01-16
Update RealismOverhaul.netkan

Felger:
Add Vostok / Voskhod as suggested mods for Realism Overhaul

### DIFF
--- a/NetKAN/RealismOverhaul.netkan
+++ b/NetKAN/RealismOverhaul.netkan
@@ -67,7 +67,7 @@
 		{ "name" : "SCANsat" },
 		{ "name" : "SovietEngines" },
 		{ "name" : "SpaceFactory-Voskhod-unofficial" },
-		{ "name" : "SpaceFacotry-Vostok-unofficial" },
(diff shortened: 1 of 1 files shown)


hakan42:
According to Jenkins log, module TACLS is not available. Could you re-check please?

Felger:
That's a bit odd, it definitely exists.  It just doesn't yet have 0.90 compatibility:  
https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/TACLS.netkan  
Which shouldn't make a difference for a recommendation, it just won't be able to install it until it's updated.

NetKAN/RealismOverhaul.netkan
```diff
@@ -66,6 +66,8 @@
 		{ "name" : "RocketdyneF-1" },
 		{ "name" : "SCANsat" },
 		{ "name" : "SovietEngines" },
+		{ "name" : "SpaceFactory-Voskhod-unofficial" },
+		{ "name" : "SpaceFacotry-Vostok-unofficial" },
```
Ippo343: Hey @Felger , this looks like a typo to me :)

PR #531, opened 2015-01-16
Add Sers Contract Science Modifier from Kerbal Stuff

todi:


hakan42:
Looks good, thanks for the contribution.

PR #543, opened 2015-01-20
Install deprecation marker file for UniversalStorage

hakan42:


hakan42:
Fixes KSP-CKAN/CKAN-meta#268 , KSP-CKAN/CKAN-meta#269 , KSP-CKAN/CKAN-meta#264

hakan42:
Fixes KSP-CKAN/CKAN-support#65

hakan42:
Build fails because ECLSS is not installed but the rest is fine...

hakan42:
Now our ckans are o.k. but build fails because ```ckan show``` cannot determine which addon to select.

The "show" command needs to be worked on.

PR #544, opened 2015-01-20
Added handmade metadata for the Throttle Controlled Avionics instead of the one made by KerbalStuffBot.

allista:
The suffix "Continued" is not a part of the mod's title.


allista:
Stray commas again =(

PR #558, opened 2015-01-22
Update SETIBalanceMod.netkan

Y3mo:
In preparation for the next update of the SETI-BalanceMod, additional dependencies are included.

Best regards,
Yemo


Y3mo:
This change is obsolete now, a new pull request with an additional dependency will be opened.

PR #569, opened 2015-01-23
Updating RFStockalike name so it appears near RealFuels in the GUI

Raptor831:
Had a request on the KSP forum to change the name so it appears next to RealFuels in the program. Seemed like a good idea.


Ippo343:
Thanks!

PR #570, opened 2015-01-23
Rcs build aid

hakan42:


hakan42:
Fixes KSP-CKAN/CKAN-support#66 , netkan is o.k., but the upstream repo has no releases defined yet.

PR #585, opened 2015-01-26
PartMapper: Fix release status beta -> testing

pjf:
The spec doesn't have a `beta` status as far as I can tell. This sets it
to `testing` (which is in the spec).


hakan42:
#rebuild

PR #603, opened 2015-01-27
Fixes #505: ProceduralParts has a correct version file

### LINKED ISSUE #505: ProceduralParts has broken AVC version file
I just opened up https://github.com/Swamp-Ig/ProceduralParts/issues/97 , but in the meantime, I will disable th $vref line.

If and when that other ticket is worked upon, we can re-enable $vref again...

hakan42:


hakan42:
The ckan generation works, but the "show" at the end fails with multiple search hits.

PR #610, opened 2015-01-27
Update SETIBalanceMod.netkan

Y3mo:
I am the author of the SETI-BalanceMod and these are the new dependencies/recommendations for version 0.6.4 and version 0.7.0.

### DIFF
--- a/NetKAN/SETIBalanceMod.netkan
+++ b/NetKAN/SETIBalanceMod.netkan
@@ -3,7 +3,7 @@
 	"identifier" : "SETI",
 	"$kref" : "#/ckan/kerbalstuff/484",
 	"$vref" : "#/ckan/ksp-avc",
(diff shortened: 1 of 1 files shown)


Ippo343:
Hy @Y3mo , if you are the author I guess you know best.
But anyway, are you sure you have all those dependencies? Remember that your mod won't be installable *at all* unless all the dependencies are available, and users cannot remove any dependency without also removing your mod.

Just a reminder, if you are sure I'll merge :)

Y3mo:
Hey,
thanks for the reminder, I was a bit worried myself when I wrote the dependencies.
Unfortunately they all became necessary.

Thank you for merging!

Ippo343:
The build fails for some weird reason (can't find ContractConfigurator, but it's in the repo). Poking @AlexanderDzhoganov on this for the Jenkins issue.

Aside from that, everything looks good to me, so merging :)
Thanks @Y3mo !

PR #597, opened 2015-01-27
Filled in more fields of the TACLS netkan file.

taraniselsu:
Untested, but hopefully it is correct.

* It is not pulling the version information correctly because of #36.
* I was unable to test using [these instructions](https://github.com/KSP-CKAN/CKAN-support/wiki/Adding-a-mod-to-the-CKAN#testing-your-file). I assume because my mod is already in the "repo"(1). It would be nice if we could test a local copy of the file before sending it off to become the official meta data.

````
(1) The error message:
KSP>ckan-1.5.6.exe install -c TACLS-v0.10.2b.ckan
1) TACLS-Config-RealismOverhaul (TAC Life Support (TACLS) - Realism Overhaul Config)
2) TACLS-Config-Stock (TAC Life Support (TACLS) - stock config)
Enter a number between 1 and 2 (To cancel press "c" or "n".):
2
5008 [1] ERROR CKAN.RelationshipResolver (null) - Assertion failed: Adding TACLS twice in relationship resolution

Unhandled Exception: System.Exception: Exception of type 'System.Exception' was thrown.
   at CKAN.RelationshipResolver.Add(CkanModule module)
   at CKAN.RelationshipResolver..ctor(ICollection`1 modules, RelationshipResolverOptions options, Registry registry, KSPVersion kspversion)
   at CKAN.ModuleInstaller.InstallList(List`1 modules, RelationshipResolverOptions options, NetAsyncDownloader downloader)
   at CKAN.CmdLine.MainClass.Install(InstallOptions options, KSP current_instance, IUser user)
   at CKAN.CmdLine.MainClass.Install(InstallOptions options, KSP current_instance, IUser user)
   at CKAN.CmdLine.MainClass.Main(String[] args)
````

PR #612, opened 2015-01-28
Filled in more fields of the TACLS netkan file.

taraniselsu:
Resubmitted #597

Is that better?


Shirkit:
I'm afraid that if you set to depend on TACLS-Config, and that is set to depend on Realism Overhaul config, and that downloads Real Solar Systems and a lot of other mods, I think that the end result is not the expected one.

PR #627, opened 2015-02-01
Removes default configs so alternate ones can be used.

playful1510:


playful1510:
Ok, I have no idea why this keeps failing.

PR #636, opened 2015-02-03
Added Anomaly Surveyor contract pack.

jrossignol:


hakan42:
The "ckan show" part is slightly but annoyingly broken... One of those days, I need to get around to repair that part.

Apart from that, I can download the netkan, expand it, and successfully install the produced ckan. Therefore, big fat "Merge" button, here I come...

hakan42:
@jrossignol , thank you for your contribution.

PR #663, opened 2015-02-08
Remove KIDS suggestion

NathanKell:
Can anyone tell me why this was added in the first place? It seems to mostly end up with people installing it for no good reason....


pacbard:
I included the suggest line when I submitted the netkan package for KIDS. I didnt/don't have a reason for it other than including a reminder to install KIDS with FAR, but I also understand your point that most people end up installing KIDS without knowing what it does just because it's suggested.

PR #665, opened 2015-02-09
Kopernicus provides binaries at long last

hakan42:


hakan42:
Jenkins is failing because the github releases are still marked as "prerelease".

PR #666, opened 2015-02-09
BDAnimationModules works for 0.90

TeddyDD:


TeddyDD:
@hakan42 I suppose I should change AdjustableLandingGear.netkan too?

hakan42:
@TeddyDD yes, changing the netkan would be the right (and only) way to go.

PR #686, opened 2015-02-13
Put the simplewings netkan in the right directory

hakan42:


JustinKerbice:
Damned ! Shame on me, Messed up with the id (again !).
But I don't understand why license has been changed to WTFPL it looks like the field is free, if not, "restricted" as it was used for my others mods would be better (the same apply to MOVAP).
Also, this file is not (yet) on the main netkan directory, as opposite to MOVAP, created later.

hakan42:
@JustinCerbice I honestly did not think overmuch on the license. Could you please send a pull request fixing it to be as you like it?

PR #689, opened 2015-02-14
vref should be unbroken now

jrossignol:


hakan42:
Thank you for fixing both your avc and the netkan :smile:

PR #736, opened 2015-02-23
Deprecate under InFlightWaypoints name, and move to WaypointManager.

jrossignol:
Praying that everything works out...


hakan42:
This change is merged, but could you please provide one more ckan file (in the CKAN-meta repo) with the special version of "Deprecated" like we did for Universal Storage? This is the only way at the moment to make people uninstall old dlls

PR #738, opened 2015-02-24
NearFuturePropulsion MM patches for nukes

Dazpoet:
Adds the ability to install the optional ModuleManager configs bundled
with NearFuturePropulsion for nuclear engines.


hakan42:
#rebuild

PR #753, opened 2015-02-27
Added FAR/NEAR usability Package

TeddyDD:
closes #748 

I'm not sure about MiniAVC plug-in, should I filter this? I don't think so because is intended to be bundled with mods so I left this..


TeddyDD:
>The following inconsistencies were found:
>FARNEARusabilityPackage requires AerodynamicModel but nothing provides it

How to fix this :question:

AlexanderDzhoganov:
It's ok, this is a known issue with the build system. Merging now.

PR #808, opened 2015-03-06
I Manage CKAN entries manualy

N3h3miah:
There are 2 folders used by several of my mods. I'll create zips and entries for CKAN and manage them manually to prevent unnecessary downloads.


AlexanderDzhoganov:
Sure, but you have to recreate the new .netkan's before I can merge this as users that have already installed will have issues otherwise.

N3h3miah:
That is what I'm currently do :-)

AlexanderDzhoganov:
What do you mean by "manage manually"? The best setup I can see is that you split your common tools in a separate KS mod and point NehemiahScienceCommon to it.

N3h3miah:
There is also a folder with multi purpose parts used by 2 of my mod. Because CKAN does not overwrite folders installing those with CKAN is a problem. So my idea was to upload special CKAN zip for all the mods containing only the parts that are unique for each mod. I would upload them to the existing mods on KS  but use the zips for manual installation containing everything needed as default downloads. The problem here is that NetCAN would pickup the wrong downloads.
To solve this I would create the needed files for the CKAN-repository myself an maintain them and make the use the ckan zps. For the folder with the multi purpose parts I would create a new CKAN entry and dependencies in the 2 mods using them.
Do you see a better solution for this?

AlexanderDzhoganov:
This is definitely not a good idea, manually managing .ckan files needs work for every release and skips a large part of our validation pipeline. I think the current solution is pretty decent, considering this alternative.

AlexanderDzhoganov:
Ok, I see what you mean. There is an error in the current metadata preventing all the mods from installing at once, let me see what I can do about this.

N3h3miah:
When we create a new NetCAN file for the MultiPurposePart folder point to the Kerbal Life Science or the Orbital Material Science mod and declare a dependency to this new entry and update the install instructions to just copy the specific folder would this work?

AlexanderDzhoganov:
I will just add it to the NehemiahScienceCommon package, would that be an issue?

AlexanderDzhoganov:
I submitted #809 with my proposed fix, let's continue the conversation there.

PR #823, opened 2015-03-10
Fix much metadata

pjf:
I've no idea why travis is being so grumpy, but running `prove` in the
NetKAN repo revealed a number of metadata errors. This commit doesn't
fix all of them, but it fixes the easy-to-fix ones. In particular:

* A number of documents have been given a spec `v1.2` or `v1.4` header, as they use newer features.
* A number of files have been renamed to match their identifier.

This doesn't fix hard errors, like missing licenses (which might be okay
if the remote data is fine), or identifiers which don't match the CKAN
spec (which maybe we should just update to allow).


Ippo343:
#rebuild

Dazpoet:
#rebuild

pjf:
#rebuild

Srsly, `build.sh`, you should not be trying to `jsonlint` yourself.

pjf:
#rebuild now that I've fixed `netkan.exe`

pjf:
OMFG. KerbalStuff is down. I've shaved so many yaks, and now this!!? o_O

pjf:
It's back. Quick, #rebuild!

pjf:
Um.. #rebuild again? C'mon, I'm pretty sure this should reach completion if everything is up.

pjf:
So it looks like the test script is installing mods, but then not clearing away the installs afterwards, meaning mods which conflict with each other can be attempted to be installed.

At this point, I'm declaring my changes to be good even though jenkins is failing them, because they're working on my local system, and if I have to shave the yak in the test script I will go insane and I wants to sleep please and thank you.

PR #825, opened 2015-03-11
netkan for http://forum.kerbalspaceprogram.com/threads/92586

spudbean:


pjf:
Not sure why our bot is choking on TweakScale, but this looks good to me. Merging with thanks!

PR #826, opened 2015-03-11
Added Kanadarm to NetKAN

hakan42:


hakan42:
#rebuild

hakan42:
failures are caused by jenkins itself

hakan42:
Oh, it already is there :)

PR #843, opened 2015-03-17
Add LunaticAeronauticsAntygravityRoom.netkan

Dazpoet:
The files in CKAN-meta appear to have been generated prior to .netkan files? Either way they don't work due to the " ' " in the download link and since it's hosted on KS I assume this .netkan file will solve the problem.

Latest version in CKAN-meta is 1.01 and this .netkan will generate 1.11 so I assume there will be no conflict by adding it in this stage?


Dazpoet:
The error in spelling is intentional, that is the name of the mod.

Issue #838, opened 2015-03-15
Add FieldExperience (by Magico13) meta 


mwerle:
Hi,

Any chance of adding this little mod?

KerbalStuff: https://kerbalstuff.com/mod/505/Field%20Experience
Forum thread: http://forum.kerbalspaceprogram.com/threads/107663

Dazpoet:
Addad with https://github.com/KSP-CKAN/NetKAN/pull/850

PR #850, opened 2015-03-18
Add FieldExperience

Dazpoet:
As requested here https://github.com/KSP-CKAN/NetKAN/issues/838


hakan42:
@Dazpoet , are you sure that we want to install the C# source code along with the plugin?

Dazpoet:
@hakan42 I removed it. I didn't know about correct procedure and kept it thinking it was correct. Some more looking about and it seems the correct thing is to not bundle it. Thanks for pointing it out :)

PR #845, opened 2015-03-18
Add Kerbalized Flags and Space Agencies from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of DirtyFace83, to add [Kerbalized Flags and Space Agencies](https://kerbalstuff.com/mod/648/Kerbalized%20Flags%20and%20Space%20Agencies) to CKAN.

Please direct questions about this pull request to [DirtyFace83](https://kerbalstuff.com/profile/DirtyFace83).

PR #851, opened 2015-03-18
Add KerbalizedFlagpack

Dazpoet:
Fixed version of #845


Dazpoet:
Works and installs fine locally for me. Are the problems Jenkins report due to the spaces in the adress?

pjf:
Having two spec versions might have something to do with it, although that's not explicitly clear in the output from the bot...

Dazpoet:
I knew I was tired when I wrote it... but didn't realize I was *that* tired. However it still doesn't sit well with Jenkins for some reason.

AlexanderDzhoganov:
#rebuild

Dazpoet:
I give up!

hakan42:
Don't give up too early, there is still the weekend to work on this :smiling_imp:

Dazpoet:
Now I'm confused, Jenkins explicetly failed that one repeatedly :s

hakan42:
Yes, but it was mostly about our toolchain setup, so I just merged the PR and will see what happens when the daemon sees it. The netkan file can be fixed if and when necessary :smile:

PR #870, opened 2015-03-24
Patch CivilianPopulation with dependencies and filters

Dazpoet:
A user came onto the irc-channel saying this mod didn't seem to install properly. This patch fixes the problem locally for me so let's hope Jenkins approves aswell.


Dazpoet:
#rebuild

PR #878, opened 2015-03-26
Create ModelRocketry.netkan

SuperRedNova:
Since I didn't realize I needed a specific license when I checked the ckan box on kerbalstuff.


SuperRedNova:
realized I had created the file in the wrong directory, also, added the correct licenses instead of relying on what KerbalStuff gives.

PR #883, opened 2015-03-27
Move dockingtube.netkan to the right directory

Dazpoet:


xZise:
A bit additional documentation: The file was added in 3dbc3b1e which references #783. Itself has another commit 1b34f079 which does actually put the file in the correct directory. So I'm not sure if that was intentional by @Ippo343 (I guess not).

PR #660, opened 2015-02-08
Add 64K from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of Paul_Kingtiger, to add [64K](https://kerbalstuff.com/mod/573/64K) to CKAN.

Please direct questions about this pull request to [Paul_Kingtiger](https://kerbalstuff.com/profile/Paul_Kingtiger).

PR #889, opened 2015-03-29
Break RSS into core+plugin

Dazpoet:
Currently we are unable to handle PRs like #811, #660 and #621. After asking the awesome @NathanKell on how to solve this I propose that we break the current RSS install into two parts. ~~The core (seen in this PR) and the config ( #890 ).~~ The drawback of this is of course that all current installs of RSS will fail with the next update and the users will have to reinstall them by doing
```
ckan.exe remove RealSolarSystem
ckan.exe install RSSConfig
```
I admit this is far from optimal but I think it's the best longterm solution to not lock mods depending on RSS out from CKAN. These mods are often bundled with a bunch of mods and seem a perfect fit for the CKAN platform.


NathanKell:
For the record I like this very much. :)

pjf:
Oh! This is where meta-packages come in handy! Rather than doing things as per the attached PR, we make RealSolarSystem a meta-package that depends upon RSS-Config and RSS-Core, but doesn't provide any content itself. Users with RSS already installed will get a bit of disk activity as we flip the file ownerships between packages, but they *shouldn't* have to take any actions aside from the upgrade. (If they don't have the RSS download in their cache it may re-download, but that's not the end of the world.)

Existing dependencies and relationships will continue to work, and end-users needn't be aware of the difference. There should certainly be no need to remove existing packages ask users to manually install new ones.

If the CKAN core can't handle a transition to meta-package form, then that's a bug we should fix. (Although it should be a pretty easy one, since meta-packages are technically easier than real packages in terms of the steps required to handle them.)

~ Paul

Dazpoet:
So basically I should add a metapackage with the identifier RealSolarSystem to CKAN-meta that depend on the new split up version and is otherways identical to the original RSS file deleted in this PR? I guess I could then change the name on #890 to something like 'RealSolarSystem - default config' to avoid confusion about two mods with the same name.

Or am I misunderstanding you?

pjf:
Aye!

* RealSolarSystem depends upon RealSolarSystem-Core
* RealSolarSystem depends upon RealSolarSystem-Config-Default
* RealSolarSystem-Config-Default provides and conflicts with RealSolarSystem-Config
* RealSolarSystem-Config-Default depends upon RealSolarSystem-Core
* 64k depends upon RealSolarSystem-Core provides and conflicts with RealSolarSystem-Config
* (and so on)

RealSolarSystem itself wouldn't really need a version (since it doesn't change), but the spec says it has to have one. We could use `virtual` as the version (which will parse as more recent than whatever RSS has now), and we pop a note in the description that this provides the default install for RSS, and is the equivalent to installing `RealSolarSystem-Config-Default`.

RSS-Core and RSS-Config-Default (or whatever we call them) would both have netkan files, and likely index the same URL (unless RSS starts seeing separate releases for the core vs config).

I totally missed #890, but it looks like it would be RealSolarSystem-Config-Default under the new scheme of things.

In terms of merging, we'd want the new RSS-Core and RSS-Config-Default .netkan files in place first (so we can test them and their packages are being generated), and then the metapackage in second (which then allows users to transition). If we do it the other way around there's a tiny space of time where the metapackage exists but its dependencies do not.

Dazpoet:
Ok so I made https://github.com/KSP-CKAN/CKAN-meta/pull/353 but Jenkins refuse it since `"download"`is a required field. I'm not sure metapackages are supported in this way but then again I also don't know how to check nor even test it...

Dazpoet:
And then I closed https://github.com/KSP-CKAN/CKAN-meta/pull/353 because from what I gather it isn't what we want at all. I must admit I don't know how to go about creating the kind of metapackage wanted.

https://github.com/KSP-CKAN/CKAN-meta/pull/353 would not update every time there's a new release of RSS without manually updating the version (which we most certainly _don't_ want to do).

A question for @NathanKell also, is there a usecase where a user would want to download only RSS-Core or should I add a dependency to RSSConfig so that users downloading Core will get a list of available configs to choose from? Just like RSSTextures work today.

Dazpoet:
#rebuild

Dazpoet:
Updated this PR to no longer remove RealSolarSystem until we can get a working metapackage. I edited RealSolarSystem to conflic with RSSCore (which changed name from RSS-core) and RSSConfig. Jenkins fails due to the download taking to long so someone looking through the files an extra time would be deeply appreciated. 

After this is merged Jenkins should approve of #890 after a rebuild and we can get this ship going :)

I forsee that this PR will break absolutly nothing but it will enable us (me) to keep working towards adding the various other configs in #892 which will make users rejoice!

Dazpoet:
Confirmed on irc that there is a usercase where someone could want only RSSCore, therefore no dependency is needed on RSSConfig in RSSCore

pjf:
Oh gosh, discovered that metapackages don't currently work as I thought they did. They're going to be a high priority on my to-do list.

We can still go down the route of RealSolarSystem depending upon RSS-Core and RSS-Config, although we may need it to install a dummy file (like a readme) until we get virtual packages happening. This is better than it conflicting with those packages, since it means that any other mods that depend upon RSS-Core (but which are compatible with the whole RSS system) can keep that dependency.

In general, if X contains Y and Z that exist as separate packages, then we want X to depend upon Y and Z. If we want X is entirely separate to Y and Z and would break if those were installed, then we'd do conflicts.

Dazpoet:
Ok an update to what I'm doing here while I'm trying to get it to work (sorry for a billion commits but I'm learning *loads* from this).

I've added RSSConfig-default to this PR and closed #890 to be able to have Jenkins test everything together which should now work since none of the mods collide.

RSS has been rewritten to only be the readme file (once I get the regexp right :anger:) and then depends on RSSCore and RSSConfig-default. The latter in turn provides and conflicts with RSSConfig as to keep it working with #892.

Since this project requires a download of RO each time it seems Jenkins just failed due to downloads taking to long so I'll make another commit after this that just comment out RO and then focus can be shifted to the important stuff rather than have Jenkins dl a bunch of extra stuff each time.

Dazpoet:
Ok so extended local testing led me to locate that the problem is not with the files but rather with the dependencies themselves. Most of the RSS stuff is interconnected (for good reasons!) and both RO and RSSTextures are attempting to fix their dependency on RealSolarSystem by installing the current implementation of it available through CKAN which of course conflicts wildly with what I'm trying to achieve here. I've got a working version locally with the only difference being that dependencies where removed.

Probably a lot of the overkill regexp is superfluous aswell so I'll do some more testing locally and fix it... again -.-

Dazpoet:
FINALLY! This version works locally *IF* I manually remove the RO, CustomBiomes-Data-RSS and RSSTexture dependencies from the generated .ckan file for RSSConfig-default. Installing the 3 generated ckan files in the order of Core, Config and RealSolarSystem (aka the order CKAN will install them to satisfy dependencies) works flawlessly. Installing RO, CustomBiomes-Data-RSS and RSSTextures through CKAN *after* these 3 are installed works without a problem.

Everything tested through commandline, which btw immedietly picked up that RealSolarSystem was different on my local machine upon `ckan update` and asked for permission to resolve the issue through reinstallation. This does not appear to be the case for GUI based on https://github.com/KSP-CKAN/CKAN/issues/795 where situations like that leads to crashes and require manual intervention.

pjf:
So, I was working on this earlier today, and was delighted because I thought I had a simple, elegant, and easy solution:

1. RSS-Core consists of `RealSolarSystem.dll` and the icons.
2. RealSolarSystem depends upon RSS-Core, and consists of everything else.

It's simple, it's easily verified that both parts together produce the whole, and it means all the mods out there can depend upon RSS-Core as much as they like. Best of all, the CKAN core *totally* supports it; you simply remove the old `RealSolarSystem` and install the new `RealSolarSystem` along with `RSS-Core`. The `ModuleInstaller.AddRemove()` call lets you do exactly that, as one transaction. It's built exactly for this purpose.

Unfortunately, I don't think any of the clients actually use `AddRemove()`, which is a shame, because it's pretty much made for situations like this. So I also suspect any changes split the package will cause consistency errors if the client tries to install RSS-Core as a separate transaction to upgrading RealSolarSystem. New users will be fine.

At this point I got side-tracked by discovering that libcurl hates Sarbian's SSL certs, so I'm currently working on a new CKAN release so I've got something that can install RSS Textures. My office is currently filled with yak-hair. o_O

pjf:
Uh oh. I don't want to be *that* project maintainer who comes in and says "great idea, here's my totally different solution", but here's what I've got:

`RSS-Core.netkan`:

```
{
    "spec_version"   : "v1.6",
    "$kref"          : "#/ckan/github/NathanKell/RealSolarSystem",
    "$vref"          : "#/ckan/ksp-avc",
    "name"           : "Real Solar System - Core",
    "identifier"     : "RSS-Core",
    "abstract"       : "Core plugin for RealSolarSystem, for the whole RSS experience install Real Solar System",
    "license"        : "CC-BY-NC-SA",
    "release_status" : "stable",
    "depends" : [
        { "name" : "DDSLoader" }
    ],
    "resources" : {
        "homepage" : "http://forum.kerbalspaceprogram.com/threads/55145"
    },
    "install": [
        {
            "file"       : "RealSolarSystem/Plugins/RealSolarSystem.dll",
            "install_to" : "GameData/RealSolarSystem/Plugins",
        },
        {
            "file"       : "RealSolarSystem/Plugins/Icons",
            "install_to" : "GameData/RealSolarSystem/Plugins",
        }
    ]
}
```

```
Real Solar System - Core: Core plugin for RealSolarSystem, for the whole RSS experience install Real Solar System

Module info:
- version:	v8.6.1
- authors:	NathanKell
- status:	stable
- license:	CC-BY-NC-SA

Depends:
- DDSLoader

Provides:
- RSS-Core

Resources:
- homepage: http://forum.kerbalspaceprogram.com/threads/55145
- repository: https://github.com/NathanKell/RealSolarSystem

Showing 11 installed files:
- GameData/RealSolarSystem/Plugins/RealSolarSystem.dll
- GameData/RealSolarSystem/Plugins/Icons
- GameData/RealSolarSystem/Plugins/Icons/eye-highlight.png
- GameData/RealSolarSystem/Plugins/Icons/eye-normal.png
- GameData/RealSolarSystem/Plugins/Icons/info-background.png
- GameData/RealSolarSystem/Plugins/Icons/launch-site-active-texture.png
- GameData/RealSolarSystem/Plugins/Icons/launch-site-highlight.png
- GameData/RealSolarSystem/Plugins/Icons/launch-site-normal.png
- GameData/RealSolarSystem/Plugins/Icons/launch-site-texture.png
- GameData/RealSolarSystem/Plugins/Icons/magnifier-highlight.png
- GameData/RealSolarSystem/Plugins/Icons/magnifier-normal.png
```

`RealSolarSystem.netkan`:

```
{
    "spec_version"   : "v1.6",
    "$kref"          : "#/ckan/github/NathanKell/RealSolarSystem",
    "$vref"          : "#/ckan/ksp-avc",
    "name"           : "Real Solar System",
    "identifier"     : "RealSolarSystem",
    "abstract"       : "Resizes and rearranges the Kerbal system to more closely resemble the Solar System",
    "license"        : "CC-BY-NC-SA",
    "release_status" : "stable",
    "depends" : [
        { "name" : "RSS-Core" },
        { "name" : "RSSTextures" },
        { "name" : "ModuleManager" }
    ],
    "recommends" : [
        { "name" : "CustomBiomes-Data-RSS" },
        { "name" : "RealismOverhaul" },
        { "name" : "Toolbar" }
    ],
    "suggests" : [
        { "name" : "RemoteTech-Config-RSS" }
    ],
    "resources" : {
        "homepage" : "http://forum.kerbalspaceprogram.com/threads/55145"
    },
    "install" : [
        {
            "file"       : "RealSolarSystem",
            "install_to" : "GameData",
            "filter"     : [
                "RealSolarSystem.dll",
                "Icons"
            ]
        }
    ]
}
```

```
Real Solar System: Resizes and rearranges the Kerbal system to more closely resemble the Solar System

Module info:
- version:	v8.6.1
- authors:	NathanKell
- status:	stable
- license:	CC-BY-NC-SA

Depends:
- RSS-Core
- RSSTextures
- ModuleManager

Recommends:
- CustomBiomes-Data-RSS
- RealismOverhaul
- Toolbar

Suggests:
- RemoteTech-Config-RSS

Provides:
- RealSolarSystem

Resources:
- homepage: http://forum.kerbalspaceprogram.com/threads/55145
- repository: https://github.com/NathanKell/RealSolarSystem

Showing 28 installed files:
- GameData/RealSolarSystem
- GameData/RealSolarSystem/Compatibility
- GameData/RealSolarSystem/Compatibility/CustomAsteroids.cfg
- GameData/RealSolarSystem/Compatibility/RealHeat.cfg
- GameData/RealSolarSystem/Compatibility/SCANsat.cfg
- GameData/RealSolarSystem/LaunchSites.cfg
- GameData/RealSolarSystem/Plugins
- GameData/RealSolarSystem/Plugins/PluginData
- GameData/RealSolarSystem/Plugins/PluginData/Bop.obj
- GameData/RealSolarSystem/Plugins/PluginData/Dres.obj
- GameData/RealSolarSystem/Plugins/PluginData/Duna.obj
- GameData/RealSolarSystem/Plugins/PluginData/Eeloo.obj
- GameData/RealSolarSystem/Plugins/PluginData/Eve.obj
- GameData/RealSolarSystem/Plugins/PluginData/Gilly.obj
- GameData/RealSolarSystem/Plugins/PluginData/Ike.obj
- GameData/RealSolarSystem/Plugins/PluginData/Kerbin.obj
- GameData/RealSolarSystem/Plugins/PluginData/Laythe.obj
- GameData/RealSolarSystem/Plugins/PluginData/Minmus.obj
- GameData/RealSolarSystem/Plugins/PluginData/Moho.obj
- GameData/RealSolarSystem/Plugins/PluginData/Mun.obj
- GameData/RealSolarSystem/Plugins/PluginData/Pol.obj
- GameData/RealSolarSystem/Plugins/PluginData/Tylo.obj
- GameData/RealSolarSystem/Plugins/PluginData/Vall.obj
- GameData/RealSolarSystem/Readme_RSS.txt
- GameData/RealSolarSystem/RealSolarSystem.cfg
- GameData/RealSolarSystem/RealSolarSystem.version
- GameData/RealSolarSystem/RealSolarSystemSettings.cfg
- GameData/RealSolarSystem/TextureFix.cfg
```

The main advantage here are obvious correctness. The core obviously consists of the `.dll` and Icons (because there's an install directive for each), and `RealSolarSystem` obviously installs everything except those (because there are explicit filters to remove them).

As far as I can tell, this is working pretty well, and for new users it's great. I'm still worried about the upgrade path, though, so I'm looking to see if I can provide a patch to at least the command-line client (which I'm most familiar with) which will allow users a smoother upgrade.

Getting the CKAN to handle something like this automatically *should* be possible, but it's hard, as it involves relationship resolution smarts. As such, my solution is instead likely to be something that lets the user manually specify the transaction, possibly as an extended syntax to the upgrade command: `ckan.exe upgrade RealSolarSystem +RSS-Core`.

Dazpoet:
I was just writing a comment about a damn near identical soultion, great minds think alike!

RealSolarSystem should provide and conflict with RSSConfig though to be consistent with all of the other alternate configs added in #892.

I think RSS-Core should be called RSSCore since I think there is precedence for that, e.g. Firespitter and I think (hope) Kopernicus.

Dazpoet:
Looking around is hard on mobile but I was wrong since RasterPropMonitor-Core is a thing and even My own PRs in CKAN- meta for SpaceFactoryCE use -Core. We could really use a framework for naming conventions, that is a discussion for another place though.

I can update this PR tonight to reflect the changes discussed above.

pjf:
> RealSolarSystem should provide and conflict with RSSConfig though to be consistent with all of the other alternate configs added in #892.

Oh aye, I agree!

> I think RSS-Core should be called RSSCore since I think there is precedence for that, e.g. Firespitter and I think (hope) Kopernicus.

I'm not particularly flustered with the naming, although more recently we've seen a tendency to use hyphens for readability, eg: `TACLS-Config-*`, `CustomBiomes-Data-*`, `ContractConfigurator-*`, etc etc. There's certainly precedence for *both* naming schemes, and I'm not going to bock on either.

KSP-CKAN/CKAN#809 is also relevant to our discussions here. :)

Dazpoet:
#rebuild

Dazpoet:
#rebuild

Dazpoet:
It works! ...or well Jenkins times out while echoi'ing everything he installed because RO+RSS+RSSTextures wasn't doable in the 8 minute timeframe.

pjf:
Out of curiosity, do you know what happens if we do a `ckan update` (or GUI equivalent) when the fetched metadata contains the new RSS info? I know it *should* engage the upgrade code if the user says "yes, please update", and that *should* do the right thing and install RSS-Core to everything works just fine.

However I haven't had a chance to test it yet, and I'm about to drop and sleep. (Apologies for not having been around much the last few days, my clients seem to be wanting a lot of attention from me!)

Dazpoet:
I must admit I have no idea and my test-setup won't allow me to test something like that since I don't have a custom environment built at home nor do I have the knowledge on how to create it I'm afraid..

I've been busy brea... updating other things in NetKAN while this has been standing still so no sweat :) Would like it done before 1.0 hits though!

Dazpoet:
So I might have found a way to test it myself, I butchered the RealSolarSystem-8.6.1.ckan generated from the .netkan file proposed in this PR locally to be
```
{
    "spec_version": "v1.4",
    "name": "Real Solar System - Default config",
    "identifier": "RealSolarSystem",
    "abstract": "Resizes and rearranges the Kerbal system to more closely resemble the Solar System",
    "license": "CC-BY-NC-SA",
    "release_status": "stable",
    "provides": [
        "RSSConfig"
    ],
    "conflicts": [
        {
            "name": "RSSConfig"
        }
    ],
    "recommends": [
        {
            "name": "Toolbar"
        }
    ],
    "suggests": [
        {
            "name": "RemoteTech-Config-RSS"
        }
    ],
    "resources": {
        "homepage": "http://forum.kerbalspaceprogram.com/threads/55145",
        "repository": "https://github.com/NathanKell/RealSolarSystem"
    },
    "install": [
        {
            "find": "RealSolarSystem",
            "install_to": "GameData",
            "filter_regexp": [
                "RealSolarSystem.dll",
                "RealSolarSystem/Plugins/Icons"
            ]
        }
    ],
    "author": "NathanKell",
    "version": "v8.6.1",
    "download": "https://github.com/NathanKell/RealSolarSystem/releases/download/v8.6.1/RealSolarSystem_v8.6.1.zip",
    "x_generated_by": "netkan",
    "download_size": 15562393,
    "ksp_version": "0.90.0"
}
```
As you can see this one lacks the dependencies of the currently indexed [RealSolarSystem](https://github.com/KSP-CKAN/CKAN-meta/blob/master/RealSolarSystem/RealSolarSystem-v8.6.1.ckan) and as such should trigger a check.

Alas a check was triggered and this is the entirety of the installation terminal, including the installation of the butchered file.

```
O:\Kerbal Space Program>ckan install -c RealSolarSystem-v8.6.1.ckan
About to install...

 * RealSolarSystem v8.6.1 (cached)
 * Toolbar 1.7.8 (cached)

Continue? [Y/N] y


Installing mod "RealSolarSystem v8.6.1"
Installing mod "Toolbar 1.7.8"
Updating registry
Commiting filesystem changes
Rescanning GameData
Done!

O:\Kerbal Space Program>ckan update
Downloading updates...
The following mods have had their metadata changed since last update - RealSolarSystem v8.6.1.
It is advisable that you reinstall them in order to preserve consistency with the repository. Do you wish to reinstall now? [
Y/N] y
8186 [1] WARN CKAN.ModuleInstaller (null) - RealSolarSystem is already at the latest version, reinstalling

Unhandled Exception: CKAN.InconsistentKraken: Exception of type 'CKAN.InconsistentKraken' was thrown.
   at CKAN.SanityChecker.EnforceConsistency(IEnumerable`1 modules, IEnumerable`1 dlls)
   at CKAN.RegistryManager.Save(Boolean enforceConsistency)
   at CKAN.ModuleInstaller.AddRemove(IEnumerable`1 add, IEnumerable`1 remove)
   at CKAN.ModuleInstaller.Upgrade(IEnumerable`1 modules, NetAsyncDownloader netAsyncDownloader)
   at CKAN.Repo.UpdateRegistry(Uri repo, Registry registry, KSP ksp, IUser user, Boolean clear)
   at CKAN.Repo.Update(RegistryManager registry_manager, KSP ksp, IUser user, Boolean clear, Uri repo)
   at CKAN.CmdLine.Update.UpdateRepository(KSP ksp, String repository)
   at CKAN.CmdLine.Update.RunCommand(KSP ksp, Object raw_options)
   at CKAN.CmdLine.MainClass.Main(String[] args)
```

Hopefully the above makes some kind of sense to someone, all I see is a warning about it already being the latest version. I've found no way of installing an earlier version of a butchered .ckan file to try with. I tried `ckan install -c <mod>=<older-version>`but it seems installing older versions with `-c` set isn't supported?

Dazpoet:
So this is funny and sad at the same time but seeing as RSS is switching over to Kopernicus all of this was for naught and is no longer needed. I must say I learned a lot from it though and we squatted some bugs along the way so I scratch it up to a success even though a merge is no longer relevant so I'm closing this and all related PRs.

PR #890, opened 2015-03-29
Break RSS into core+plugin pt2

Dazpoet:
Depends on https://github.com/KSP-CKAN/NetKAN/pull/889

PR #891, opened 2015-03-29
Add 64k.netkan

Dazpoet:
Depends on #889 and #890


Dazpoet:
Screwed this one up, will do a pack instead.

PR #892, opened 2015-03-29
Add rss configs

Dazpoet:
Depends on #889 and #890


Dazpoet:
Merge will supercede #621, #660 and #811

PR #900, opened 2015-04-02
RSS configs for CustomAsteroids and SCANSat

aw94:
Preeeety sure that the chances of these working is ~50%


aw94:
Failed because of [this](https://github.com/aw94/CKAN-core/commit/72577c59d88e29ec1f326cbed58b35921fbb5676) change I made earlier in CKAN-core... Going to see if I can figure out how to make it work.

aw94:
Timed out...

Dazpoet:
Stumbled upon this just now. These files are automatically installed through RSS are they not? Is there a specific reason to break them out of RSS and into their own thing? If so they would need to be filtered out from RSS before these could work. That might be a good thing to add to https://github.com/KSP-CKAN/NetKAN/pull/889 if the need exists. Seeing as we're already mucking about with RSS in there.

aw94:
RSS includes them, but I don't think they're installed to the proper folders at this time. 

I broke them out because I thought that's what's preferred with configs, so if someone wants to use something different with RSS they can. If that's not the case, then I'm fine with including them in RSS.

I think I got the filters right? And yeah, adding this to #889 seems like it could work, since some of the dependencies could have changed

Dazpoet:
.cfg files do not require to be put into specific folders to my knowledge, hence their placement in a unified folder for compatibility. Do correct me on this if I'm wrong!

The implications of #889 and #892 is that RSSConfigs are now bundled with their own compatibility files and I wonder if there is ever a case where a person would create a mod that specifically messes with the compatibility files for a distinct RSSConfig without also making other changes big enough to warrant its own RSSConfig?

After some consideration I find myself somewhat opposed to these changes seeing as they generate additional configurability to a part of a mod I doubt anyone will ever want to configure. If there are indeed a "market" for changed compatibility configs for specific RSSConfigs I'll add it but right now it just seems like adding complexity for the sake of complexity.

Dazpoet:
#rebuild

aw94:
I always had the impression that the location of .cfg files mattered, which is why we have things like the RemoteTech RSS config. It might depend on the mod, though; I believe Contract Configurator works with .cfg files anywhere, while other mods might require files to be in the right place. I'm not sure what the case is for SCANSat and CustomAsteroids.

If the .cfg file locations don't matter, then I agree that this isn't necessary, especially with the new changes. N

Dazpoet:
RemoteTech RSS config exists because it used to be that RemoteTech came with a specific file which RSS had to overwrite. CKAN doesn't support overwriting so instead a workaround using different downloads was created. As for .cfg files mattering in their placement this is true for e.g. part.cfg since those often include hardcoded paths to stuff, it is not true for MM patches though which these .cfg files are.

Dazpoet:
Doublechecking and yes these are indeed MM patches. With that I'll conclude that we can close this since #889 should solve any and all usecases where this would be relevant. So if everything goes to hell in a few months I'll accept blame ;)

A note also: You would have had to change RealSolarSystem.netkan aswell for this change to work since it would in its current state create duplicates of the .cfg files you're putting into other folders due to them depending on RSS :)

aw94:
Ah, ok. Didn't know that. I think you're right in thise case.

As for the .netkan -- oops >___<

PR #911, opened 2015-04-07
RealFuels no longer has a TweakScale directory in the zip

pjf:
This PR makes it index friendly again.


pjf:
I hereby invoke the power of hubris, with the reassurance of jenkins, to merge my own PR.

PR #913, opened 2015-04-07
Add impact mod

tomforwood:
Please can you add my mod to CKAN

### DIFF
--- a/NetKAN/Impact.netkan
+++ b/NetKAN/Impact.netkan
@@ -1,0 +1,9 @@
+{
+    "spec_version" : 1,
+    "identifier"   : "Impact",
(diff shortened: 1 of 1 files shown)


BasharMilesTeg:
Line 7 (license) should have "MIT" instead of "MIT license". The full list of valid licenses can be found https://github.com/KSP-CKAN/CKAN/blob/master/CKAN.schema#L254

Dazpoet:
The "name" line should be the name of the mod as it is displayed in the CKAN GUI. Now it's the same as the abstract which seems unorthodox, is this the way you want it or has there been a misunderstanding?

tomforwood:
Thanks, I think I got confused with the identity field

PR #915, opened 2015-04-08
Update kerbalstuff URL

kikillotest:


Ippo343:
I'm sorry but I have to reject this.
The URL pointing only to the mod identifier will keep working in the future even if the mod is renamed, while this change makes it so that the link might break in the future.

Sorry, but thanks for contributing!

PR #919, opened 2015-04-08
Update SETIBalanceMod.netkan

Y3mo:
Removed 2 dependencies and added them as recommendations, also added 2 more recommendations based on the feedback in the forum thread.

I am the author of the mod and those changes are valid for new version released yesterday.

Thank you very much!


Dazpoet:
I'll merge this based on visual inspection since Jenkins not having time to dl SETI is a known issue since earlier versions.

PR #922, opened 2015-04-09
New file added

linuxgurugamer:
New mod


Dazpoet:
Does this supercede https://github.com/KSP-CKAN/NetKAN/pull/921 ?

linuxgurugamer:
I think so, if I recall, the 922 included the file that 921 had

linuxgurugamer:
I just saw that the GrandTours is now in CKAN, thanks

How would I change the name?  It appears that I didn't put a space 
between "Contract" and "Pack"?

Is it just a matter of updating the netkan file and pushing it?

Thanks


JBB

Dazpoet:
The name is fetched from Kerbalstuff where there is no space. You can hardcode a name into the netkan file using ``` "name" : "<nameyouwant>",```

Should you choose to update the name on KS (if that is even possible?) You'll also need to change the metadata in the netkan file for the kerbal stuff resource entry which now (correctly) points to an adress with no space between contract and pack.

PR #938, opened 2015-04-14
Fixing dependency of KSPAPIExt to avoid AD problems

Dazpoet:
A user whose name I cannot for the life me remember told me about this problem on irc. Installation of the two mods affected by this PR (NF-Propulsion and SmartParts) both give the KSPAPIExtensions.dll which can then be autodetected by CKAN at a later stage and prevent the correct installation of that mod.

Effects of these changes are that people doing `ckan update` will be asked to perform a reinstall and ckan will handle it for them while people using GUI will run into some forced manual intervention upon updating either of these mods, see https://github.com/KSP-CKAN/CKAN/issues/795


Dazpoet:
Ok so I just read http://forum.kerbalspaceprogram.com/threads/81496-0-90-KSPAPIExtensions-V1-7-2-Utilities-for-shared-mod-use-16-Dec?p=1184389&viewfull=1#post1184389 and now I'm not sure this is the right way to go about this.

It would appear that the way KSPAPIExtensions is used/bundled clashes with how CKAN does things. That information is pretty old though but http://forum.kerbalspaceprogram.com/threads/81496-0-90-KSPAPIExtensions-V1-7-2-Utilities-for-shared-mod-use-16-Dec?p=1664750&viewfull=1#post1664750 seems to imply that the situation has not changed.

On another note though it woul also appear that KSPAPIExtensions isn't update all that often https://github.com/Swamp-Ig/KSPAPIExtensions/releases and as such this solution should work until such a time that KSPAPIExtensions start bursting out new releases at a rapid pace and mods depending on it stop working due to not keeping up.

Dazpoet:
Some discussion on irc leads me to the conclusion that this is a bad idea. However the issue of it being Autodetected probably persists... not sure how to solve it but this is apperantly *not* the right way.

PR #939, opened 2015-04-15
Mods by RaiderNick

BasharMilesTeg:
At the request of @NathanKell I have created .netkan files for the 5 mods by @RaiderNick on KS.
The following issues should be resolved before merging in addition to further testing:

Should any of these 'suggest' each other or another mod?

**Skylab (RN_Skylab)**:
Does this really need MM?

**Soviet Probes & Rockets (RN_Soviet_Probes)**:
Should 'Soviet_Probes_TR.cfg' be installed by default? if not does it need MM?
Can CKAN handle Subassemblies?
Do I need to add the FASA Reflection Plugin to CKAN so this can depend on it?

**Salyut Stations & Soyuz Ferries (RN_Salyut)**:
Would it be better to name this RN_Salyut_Soyuz?
Do I need to add the FASA Reflection Plugin to CKAN so this can depend on it?
Do I need to make a 'just the solar panels' variation on KOSMOS for the purposes of this? (and where is KOSMOS in CKAN?)
Do I need to add 'soviet pack plugins' to CKAN so this can depend on it?

**US Probes Pack - Old and New (RN_US_Probes)**:
Do I need to add the FASA Reflection Plugin to CKAN so this can depend on it?


Dazpoet:
#rebuild

Dazpoet:
The problem reported by Jenkins is not reproducable using the latest netkan that works on windows (https://github.com/KSP-CKAN/CKAN-nightly/releases/tag/jenkins-NetKAN-126) but shows up using the latest NetKAN (https://github.com/KSP-CKAN/CKAN-nightly/releases/tag/jenkins-NetKAN-132) under OS X.

Dazpoet:
From my understanding the solar panels you require from KOSMOS are found in the KOSMOS-SSPP pack which is marked down as only compatible with 0.25 both on KS and here on CKAN. Is it working in 0.90 and metadata hasn't caught up or is it truly 0.25 only?

As for the other part of KOSMOS (the -URM package) I just opened #940 to resolve the issue of it not updating properly :+1:

NathanKell:
Parts packs almost never are made obsolete by an update. I would bet that KOSMOS-SSPP works fine on .90.

Dazpoet:
#950 should help you with getting solar panels, from what I gather they are in there?

Dazpoet:
#rebuild

Dazpoet:
Afaik CKAN cannot handle subassemblies since they appear to be stored in the saves folder which isn't one you're allowed to manipulate, see [here](/https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#optional-fields).

RN_Salyut should now be able to depend on KOSMOS-SSPP to get the solar panels. As for the name it seems to have Salyut and Soyuz in it so I think it would not be incorrect to name it RN_Salyut_Soyuz, heck maybe the namechange will shake Jenkins enough to stop being an ass.

FASA reflection plugin appears to be bundled when installing FASA through CKAN. Atleast there's a file named ReflectionPlugin.dll in `FASA/Plugins/ReflectionPlugin.dll`

On a totally unrelated note this PR made me check out these mods and now I want to use them.

Dazpoet:
More testing under OS X made me notice this, which also happens on Jenkins

```
7810 [1] DEBUG CKAN.NetFileCache (null) - Storing https://kerbalstuff.com/mod/79/Salyut Stations %26 Soyuz Ferries/download/0.92
...
7887 [1] DEBUG CKAN.NetFileCache (null) - Checking cache for https://kerbalstuff.com/mod/79/Salyut Stations %2526 Soyuz Ferries/download/0.92
```

It would appear that the file netkan.exe is storing is not identical to what it's searching for. Why this happens I do not know but maybe it can give someone else a lead?

Under OS X the %2526 reported by Jenkins in the second line is instead a %252526. Some quick google gave me [this answer](http://stackoverflow.com/questions/9292865/what-is-url-encoding-2526) which claims this happens when you url-encode something twice. In this case, just like the example one, it appears the "&" sign is encoded twice, or thrice under OS X.

The only other mod I found directly with the "&" is [Tantares](https://github.com/KSP-CKAN/NetKAN/pull/76) and quite a lot has changed since november last year.

Dazpoet:
#rebuild since Jenkins has had some changes and now appear to genereate more information so we can track down this problem.

pjf:
#rebuild, since I can't seem to reproduce this locally. :/

pjf:
Upgraded the version of mono that jenkins uses.

#rebuild

Dazpoet:
#rebuild

pjf:
#rebuild and oh please to be working this time netkan

pjf:
BOOYAH!

pjf:
#rebuild again, this time if all goes well jenkins will pretend we're living in a 1.0.0 paradise.

hakan42:
I believe this will not work because the build.sh change is in master, but this branch was created before... It would need to be rebased first.

Dazpoet:
#rebuild

BasharMilesTeg:
This should be ready to go now, with the exception of HotRockets needing a version bump.

Dazpoet:
Closing on request of author

NathanKell:
Reopened at request of Raidernick (miscommunication re: closing).

Dazpoet:
Does this mean we can also re-open https://github.com/KSP-CKAN/NetKAN/pull/1073 ?

Dazpoet:
@BasharMilesTeg will have to add it and push to this branch then :)

PR #943, opened 2015-04-15
Fix metadata of EngineIgnitor-Unofficial-Repack

Dazpoet:
closes https://github.com/KSP-CKAN/CKAN-meta/issues/337

Since this is a repack by @pjf I'd like him to be able to comment on this PR before we consider merging it. This is the first (only?) unofficial repack I see in CKAN so I'm not sure what the policy are for these.

I can see no reason that this should depend on RealFuels, I assume it is because it is part of RO or something along those lines? I also added the recommended mods from the forum thread for EngineIgnitor as recommends in the metadata.

Proposed changes are based on the issuereport which this PR closes.

### DIFF
--- a/NetKAN/EngineIgnitor-Unofficial-Repack.netkan
+++ b/NetKAN/EngineIgnitor-Unofficial-Repack.netkan
@@ -9,9 +9,11 @@
     "license"        : "MIT",
     "release_status" : "stable",
     "depends" : [
-        { "name" : "RealFuels" },
         { "name" : "ModuleManager" }
     ],
+	"recommends": [ { "name" : "RealFuels" },
+					{ "name" : "KAS" },
+					{ "name" : "RasterPropMonitor" } ],
     "suggests" : [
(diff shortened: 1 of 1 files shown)


Dazpoet:
#rebuild

pjf:
Oh goodness, I can't remember why RealFuels was in there. @NathanKell would know; it *might* be because it needs it for engine configs, or it could be there for no good reason at all.

KAS and RPM should be suggests at most, and could be listed as simply supported. Most of the gameplay aspects are available without them, and both KAS and RPM can use a big chunk of resources.

NathanKell:
It can suggest RealFuels, but it certainly doesn't require RF, nor even recommend.

Dazpoet:
Changed everything to be suggests. Seeing as the modauthor thought KAS and RPM was important enough to build in support for KAS and have stuff created for usage with RPM I think they deserve staying there.

RF and RO stay as suggests since the latter is already connected to about half of CKAN so why not this mod aswell :P

Issue #554, opened 2015-01-22
ImprovedStrategies v0.12 not working if installed from CKAN


vtdmhd:
The CKAN version unzips to gamedata/Strategies instead of gamedata/ImprovedStrategies/Strategies. It gets loaded after SQUAD's strategies.cfg and the original unmodded file gets used.

hakan42:
Actually, now that I think about it, maybe the installation folder should be something like 000_ImprovedStrategies like it is with 000_Toolbar. 
This would make it obvious that the loading order is important. But it would require the ImprovedStrategies project to provide a changed zip file so people installing via CKAN or manually both see the same result.

borsek:
renamed folder, might help, if not, I'll add a file structure - gamedata folder to the zip

PR #944, opened 2015-04-15
Patch ImprovedStrategies metadata

Dazpoet:
closes https://github.com/KSP-CKAN/NetKAN/issues/554

This change should make stuff load in the right order for the mod to work based on the issuereport linked above.


Dazpoet:
#rebuild

PR #947, opened 2015-04-16
Rename 001ControlLock to just ControlLock

pjf:
The CKAN spec specifies that mods *must* start with a letter. Despite
the fact that `netkan.exe` will process this file, it may not in the
future, and at the moment our bot chokes on the validation phase.

This change renames the identifier to one that's CKAN friendly.

Unfortunately, users who already have the mod installed won't be able to
auto-upgrade with the identifier change, but they can uninstall with the old
identifier and install with the new identifer just fine.

### DIFF
--- a/NetKAN/ControlLock.netkan
+++ b/NetKAN/ControlLock.netkan
@@ -2,7 +2,7 @@
 "spec_version" : 1,
 "$kref" : "#/ckan/github/SirDiazo/ControlLock",
 "name" : "Control Lock",
-"identifier" : "001ControlLock",
+"identifier" : "ControlLock",
 "author" : "Diazo",
 "abstract" : "Lock KSP so that when entering text in a text field, you do not pass commands to your vessel.",
 "license" : "public-domain",


pjf:
Attentioning @SirDiazo in on this.

Relates to #641 .

Dazpoet:
Double checked and nothing depend on this here or in CKAN-meta.

Considering the reproduction steps in https://github.com/KSP-CKAN/CKAN/issues/746 and their effect I think the GUI users will run into problems if they run a refresh after this PR is merged. The installed version of Control Lock Will probably be removed from the list of installed mods. With luck it should be AD as the new one, without luck it won't and users attempting to install the new one get hit by a Kraken saying it can't overwrite.

Not saying we should not merge, this does go against the schema and must be adressed but I feel it is good to outlined the risks.

hakan42:
I believe ControlLock was named "001ControlLock" to make sure it is loaded before any other addons. Just like how Toolbar is called "000_ToolBar"

Dazpoet:
Based on Jenkins output it seems to still work with the new identifier and install to GameData/001ControlLock/*, why is this? I thought an install stanza was always required when namn of folder != identifier :O

pjf:
It has an install stanza. I've not changed that at all.

KSP-CKAN/CKAN#746 makes me cry. The CKAN core and command-line client are completely fine with installed mods that aren't in the upstream index; there's no reason why the GUI should insist on them. (Which either means this'll be an easy fix, or agonising because there *is* a reason why things are like this.)

In any case, it looks like this *may* be causing the indexing bot to stick, although I'm not 100% sure on that. I've really got to crash and sleep now, but I expect to check the bot's status in the morning which will tell us for sure.

Dazpoet:
I am an idiot who forgot to scroll through the whole file and just assumed what I saw on screen was everything in there but forgot I could expand.

Atleast the GUI issue is easy to reproduce, let us hope it is also easy to fix!

SirDiazo:
Hmm, this seems like an evolving design oversight/conflict maybe?

Just to confirm the logic at the time that has lead up to this:

1) Control Lock needs to load after the toolbar (if present), but before any other mods in case one of those mods depends on it. Hence the 001 prefix.

2) CKAN guidelines state that the name should be the same as the ModuleManager name (or maybe it was C# assembly name?) so added to NetKan as 001ControlLock

3) Now there is the requirement that mods in the CKAN must start with a letter. In this mod's case, this contradicts the ModuleManager name in point 2 as the mod does in fact start with a number.

Having said that, the entire purpose of this "load first" design was so that other mods could use this mod to lock KSP out while their text entry fields had focus. However, as I added an on-screen button to do so manually, it looks like the preferred route for the player base is that rather then creating the dependency between the mods, it has instead been left up to the player to remember to click said on-screen button.

What this means for the CKAN I don't know, but I can at least confirm the steps that have led to this point.

pjf:
> Hmm, this seems like an evolving design oversight/conflict maybe?

Not quite. The module naming restrictions date back to before the CKAN code was written, they haven't changed since the spec was first drafted. Normally invalid names would never get into the NetKAN directories, because travis would test the identifiers, spec version, and numerous other common mistakes that we see.

Earlier this year, the merge checking was switched from using travis-ci to jenkins, Due to what I suspect is an oversight, the tests themselves (which run with the `prove` command) weren't enabled on the new system, meaning that easy-to-overlook mistakes could sneak through. Re-enabling the old tests is on my to-do list, because they *do* catch a lot of common mistakes we see.

However, the bot which updates `CKAN-meta` fastidiously checks the files it generates against the spec, and refuses to commit if it spots any errors. Since ControlLock has been recently updated, and the resulting file fails to validate against the spec, and CKAN-meta doesn't update. :/

We *can* change the spec, but I'd rather not; insisting identifiers start with a letter means we'll never have a situation where it's ambiguous if a field is a number or an ident.

The good news is that updating the identifier doesn't change where we install the mod, what files are installed, or anything else. It *does* mean any relationships (depends/recommends/suggests/conflicts) need to be updated if any of them refer to this mod, but from @Dazpoet's report I don't believe anything does.

So, I suggest the following plan:

- I push the updated version of `001ControlLock` to CKAN-meta, even though it breaks the spec. The bot doesn't check the files it doesn't change, and technically the core *will* work with the identifier for the moment. That means users can upgrade, and the bot keeps working.
- When we build the next CKAN release, we add a rule to the registry-upgrade code that internally renames `001ControlLock` to `ControlLock`. This only gets run on upgrade, so it's not an extra time burden, and it should provide a seamless identifier upgrade for users.
- At the same time, we update this NetKAN file and the related `CKAN-meta` files to the new identifier. Users of the new client won't know the difference (everything will work as intended), users of the old client *might* see the mod doubled-up, but that will solve itself on upgrade.
- We re-enable the old tests on the NetKAN repo, so we'll never hit this problem in the future. :)

Sound good?

pjf:
I'm proceeding with my plan, not least so I can also get #823 merged, which will re-enable a lot of the metadata checks. I've changes that fix up installed ControlLock files (KSP-CKAN/CKAN-core#98) upon CKAN upgrade.

As such, I'm merging this now so new users get the new identifier. The new CKAN release should be out in a few hours, and existing users will get an automatic ident fix when they open the new client.

Issue #932, opened 2015-04-12
ksp.marce.at has been taken down, mods stuck in limbo


Dazpoet:
Mods created by Marce have been halted and removed from the webpage in the heading. This leaves the mods in limbo with nowhere to be downloaded from. Reason can be found [here](http://forum.kerbalspaceprogram.com/threads/95865-DEV-HALTED-Caelum-Ire-Technologies-%28CIT%29-General-Thread?p=1820738&viewfull=1#post1820738) so I don't expect the mods to return shortly nor at all for the forseeable future.

Mods are not available through KS nor any other alternate mirror that I could find in a short time of searching nor do I think the author would welcome an alternate download solution based on the linked post above.

As we lack a means of depreciation how are cases like these handled?

TeddyDD:
If any external mod depends on Marce's mods we are fu****
If no, I think we can remove packages. I think (but I'm not sure) they just become "unknown" for users that already have them installed :warning: I'm not really sure :warning: 

Also look here https://github.com/KSP-CKAN/CKAN-meta/search?utf8=%E2%9C%93&q=deprecated
ContractConfigurator-SCANsat was just renamed
(btw.  WTF is ``` "version": "Deprecated",``` ?)

pjf:
So, the good news is that as far as I can tell all of Marce's mods have been released under licenses that permit redistribution ([CC-BY-NC-SA-4.0](https://creativecommons.org/licenses/by-nc-sa/4.0/)). That explicitly gives irrevocable permission to modify and distribute the work as long as we comply with the license terms. Licenses are important because they're assurances to users that even if the author changes their mind later on, the users can continue to use, distribute, and maintain the software as required. 

The other good news is that we'll have multiple copies of any mod that permits redistribution as part of the MirrorKAN network, so recovering these mods from the abyss *should* be a straightforward process.

From a technical and legal standpoint, we're 100% fine in making these available from the CKAN, and having new developers take over their maintenance.

The sticky points here are *social* ones, although ones that have been encountered *many* times before by the free software community. Our first steps are gathering information as to what has happened, seeing of Marce has a change of mind, and then assessing the impact and outcomes of our options.

marce155:
I forgot about CKAN, should have notified you guys, sorry.
[...]

PR #951, opened 2015-04-16
Add InfernalRobotics Rework + Model Rework

Dazpoet:
IR-Model-Rework-Utility coming in another PR since it depends on CIT mods we no longer have access to due to https://github.com/KSP-CKAN/NetKAN/issues/932 so that one will probably take a longer time to figure out.


Dazpoet:
A quick rebuild once the rework have their first release should work now. Hoping it's soon because I'm aching for those awesome robotic parts!

Dazpoet:
Needs some rewriting, this should not have it's own identifier but be an update of InfernalRobotics

Dazpoet:
#rebuild

Dazpoet:
Awww Yis!

PR #952, opened 2015-04-16
add Infernal robotics model rework - utility pack

Dazpoet:
Depends on #951 and requires a solution to https://github.com/KSP-CKAN/NetKAN/issues/932 since it depends on those mods. An alternate route might be to have this install and conflict with `CIT-Util` and `ActiveStruts` since it redistributes them within the .zip.

We'll see once #951 is merged how to resolve this, CIT mods might be back on the market by the time IR-Rework have had their first release.


Dazpoet:
#rebuild

Dazpoet:
It seems the bot wont index other IR stuff because this one is missing and this won't resolve through Jenkins because the other mods aren't indexed properly. I'm merging this hoping it'll resolve the issue. In a perfect world everything works after this merge, in the real world I'll probably spend the next hour trying to fix something I didn't forsee.

PR #957, opened 2015-04-17
Add Scatterer.netkan

Dazpoet:
Supercedes #949


Dazpoet:
KerbalStuff feeds `"version" : "whatever"` through the metadata. This is *sure* to create problems down the line.

Note to self to fix that to something less arbitrary,

Dazpoet:
#rebuild

Dazpoet:
That solved itself faster than I thought, merging.

PR #960, opened 2015-04-17
Add KerbalStockLauncherOverhaul.netkan

Dazpoet:
supercedes https://github.com/KSP-CKAN/NetKAN/pull/948

What in the world is Modulefixer.dll? I'm not sure if it will collide with anything else and I don't know which mod it originates from either so I can't really set up a dependency.


Dazpoet:
#rebuild

Dazpoet:
#rebuild

Dazpoet:
I found modulefixer.dll origin [here](http://forum.kerbalspaceprogram.com/threads/63109-0-23-Modulefixer). Based on that thread I have also added conflicts with all USI mods after confirmation by their author.

Awaiting reply from the author of this mod about whether or not this .dll is acctually needed before merging.

Dazpoet:
#rebuild

PR #982, opened 2015-04-20
Add NewHorizons.netkan

Dazpoet:


Dazpoet:
Failed due to timeout, all other PlanetPacks work with the same modification so I'll deem this a success.

PR #988, opened 2015-04-20
add IR Sequencer

erendrake:
I have a question about the `install` node in this file. i am not sure i did it right.


Dazpoet:
To follow some semblance of naming scheme for IR mods would it be ok with you to change the identifier to `InfernalRobotics-Sequencer`?

The supplied install stanza will install to the path GameData/MagicSmokeIndustries/MagicSmokeIndustries/ so you're correct in asking about it :)

The ending comma on line 15 isn't needed and is the reason the bot fails the build.

A better install stanza would probably be

```
"install" : [ { "find" : "IRSequencer",
"install_to" : "GameData/MagicSmokeIndustries" } ] 
```

This will make CKAN able to find the folder even if you happen to release a .zip somewhere down the line with a slightly altered folderstructure, it will search for the topmost folder named IRSequencer. To use the "find" keyword in the install section you need to change spec_version to "v1.4" though. If you want to use `"file"` though the correct change is to have `"install_to" : "GameData"`

From the .zip file it would appear that ModuleManager is a dependency for this mod (seeing as it's bundled) but this is not reflected in the .netkan file.

erendrake:
@Dazpoet I just squashed the fixes and my first commit into a new commit. Is it ok that it failed the build?

Dazpoet:
Taking another look now after some sleep and you are missing the $kref field which tell us where to get stuff. Also you need to but citations around your spec_version, there being none around the 1 is a special case.

PR #992, opened 2015-04-21
Add BaconLabs.netkan

Dazpoet:
closes #918

### DIFF
--- a/NetKAN/BaconLabs.netkan
+++ b/NetKAN/BaconLabs.netkan
@@ -2,7 +2,7 @@
     "$kref": "#/ckan/kerbalstuff/688",
     "spec_version": "v1.4",
     "identifier": "BaconLabs",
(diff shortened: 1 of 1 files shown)


Dazpoet:
Basically https://github.com/KSP-CKAN/CKAN/issues/816 needs some testing. This file is very simple so I'll merge it based on manual inspection and see if something explodes.

##### Motivation
[Tantares](https://github.com/KSP-CKAN/CKAN-meta/tree/master/Tantares) was updated by the netkan-bot 16 days ago so "&" in names worked then. @BasharMilesTeg opened https://github.com/KSP-CKAN/NetKAN/pull/939 10 days ago which didn't work. If merging of this mod doesn't work we've narrowed it down to a 6 day gap to look for changes. Sounds a lot better than shooting in the dark. At worst we get a broken .netkan file which can easily be removed and maybe someone will have to kick the netkan-bot again.

PR #994, opened 2015-04-21
How many IR-Seg PRs can we collect?

Dazpoet:
closes #988 
closes #993


Dazpoet:
notifying @erendrake since this should bring joy :)

erendrake:
joy!

PR #1001, opened 2015-04-23
Initial release

linuxgurugamer:
Shjould work :-)
Copied another which does

### DIFF
--- a/NetKAN/ContractConfigurator-SpriteMissions.netkan
+++ b/NetKAN/ContractConfigurator-SpriteMissions.netkan
@@ -16,7 +16,7 @@
     "install" : [
         {
             "find"       : "SpriteMissions",
-            "install_to" : "GameData/ContractPacks/SpriteMissions"
+            "install_to" : "GameData/ContractPacks/Spacetux/SpriteMissions"
         }
     ]
 }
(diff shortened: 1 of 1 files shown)


Dazpoet:
So this is not supposed to install to `GameData/ContractPacks/Spacetux` like your other contract packs? Just making sure before merging :)

linuxgurugamer:
Ummmm, oops

Yes it is.  I must have make a typo.

Thanks for catching it


JBB

linuxgurugamer:
Fixed, thank you

linuxgurugamer:
This finished this morning, any idea when it will get into CKAN?

Dazpoet:
That something finishes just means that the bot found no errors. To get into CKAN someone has to merge it, something we do manually after making sure it works as it is supposed to. Like in this case I asked about the install stanza just to make sure that it was correct. Then I read the details of the bots build and shortly after writing this post I'll merge it into CKAN.

linuxgurugamer:
Ah, now I understand.

OK, thanks for the explanation.  I'll reduce my noise level :-)

and I did see that it finally got in, I saw it there last night.

Thanks



JBB

Issue #1002, opened 2015-04-23
How do we handle the 1.0 transition for metadata?


Dazpoet:
With (modbreaking) KSP 1.0 around the corner it is somewhat important to get a handle on how much metadata updating will be needed in the near future. By knowing which data is affected by manually set `ksp_version` fields we won't have to check through files when someone ask "why isn't mod X updated to 1.0 in CKAN yet" since if it's on the list it means it requires some manual changes. I did some filtering to get a list of mods in NetKAN with a hardcoded 0.90 in them. This is the list I got

```
ActsEW.netkan                      
ADIOS.netkan                       
AdjustableLandingGear.netkan       
AGExt.netkan                       
AlcubierreStandalone.netkan        
AnimatedDecouplers-x64.netkan      
AnimatedDecouplers-x86.netkan      
AntennaRange.netkan                
AutoAction.netkan                  
BDAnimationModules.netkan          
BetterBuoyancy.netkan              
CommunityResourcePack.netkan       
ContractConfigurator-KerbinSpace...
ControlLock.netkan                 
CrossFeedEnabler.netkan            
CrowdSourcedScience.netkan         
CustomBiomes.netkan                
DeadlyReentry.netkan               
IFILS.netkan                       
Impact.netkan                      
InfernalRobotics.netkan
Karbonite.netkan                   
KarbonitePlus.netkan               
KerbalFlightData.netkan            
KerbalFlightIndicators.netkan      
KerbalWind.netkan                  
KOSMOS-SSPP.netkan                 
kribubble.netkan                   
krispeeder1.netkan                 
KSPAPIExtensions.netkan            
MapResourceOverlay.netkan          
MFS.netkan                         
ModuleRCSFX.netkan                 
NearFutureElectrical.netkan        
NebulaEVAHandrails.netkan          
OblivionAerospace.netkan           
PilotAssistant.netkan              
ProceduralDynamics.netkan          
ProceduralFairings.netkan          
RCSBuildAid.netkan                 
RCSLandAid.netkan                  
RealFuels.netkan                   
Regolith.netkan                    
RemoteTech.netkan                  
SDHI-ServiceModuleSystem.netkan    
SmartParts.netkan                  
SmartStage.netkan                  
Snacks.netkan                      
StageRecovery.netkan               
StockBugFixModules.netkan          
StockDragFix.netkan                
StorkDeliverySystem.netkan         
TestFlight.netkan                  
TWR1.netkan                        
[...]

PR #1003, opened 2015-04-23
Add ksp_version_max field to mods which lack it

Dazpoet:
Don't merge until discussed in https://github.com/KSP-CKAN/NetKAN/issues/1002

After a potential merge, update the list of hardcoded mods in the above discussion.

### DIFF
--- a/NetKAN/AMEG.netkan
+++ b/NetKAN/AMEG.netkan
@@ -6,6 +6,7 @@
     "license"         : "restricted",
     "release_status"  : "stable",
     "ksp_version_min" : "0.23",
+	"ksp_version_max" : "0.90",
   "resources": {
     "homepage": "http://forum.kerbalspaceprogram.com/threads/74309"
 	},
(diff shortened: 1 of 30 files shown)


Dazpoet:
Removed ksp_version fields from 8 mods that are active and correctly indexed on KerbalStuff so having version fields isn't needed

Dazpoet:
Removed some `$vref` that overwrote KS metadata and forced a `ksp_version_min` rather than `ksp_version` field.

Removed Tantares and Tantares LV due to their inability to be re-indexed for now.

This build is now failing due to mods within colliding with eachother and since the only thing changed was version fields I think it should be ready for merging now.

Dazpoet:
1.0 comes out tomorrow. Lots of revision of metadata is still needed and I want netkan-bot to have time to request and fix metadata from KS before authors go ham and start updating stuff. I'm merging this and if everything explodes I'll revert it. Fingers crossed!

PR #1008, opened 2015-04-25
Fix CapCom install location; add recommends

DMagic1:


Dazpoet:
Would you mind terribly much if I asked you to change the identifier to just CapCom? The recommended one is something of a mouthful and since the foldername and modname is CapCom it seems somewhat longwinded to add the description of the mod to the identifier.

Ofcourse if you want to keep it that way, it's your mod :)

DMagic1:
Does it matter if the Kerbal Stuff title is different?

I'll change it if not.

Dazpoet:
It makes no difference what so ever, identifiers are just what we use internally and it won't break anything to change it.

Dazpoet:
Thank you for that, merging now :)

PR #1014, opened 2015-04-26
Revert "Add ksp_version_max field to mods which lack it"

Dazpoet:
Reverts KSP-CKAN/NetKAN#1003

This was a way to sweeping change. Reseting it back and taking another look.


Dazpoet:
Everything "worked" before so this won't break anything else than was already broken. Note to self to fix the ones that were legit broken.

PR #1016, opened 2015-04-27
Add conflict with HotRockets to RealPlume

Dazpoet:


Dazpoet:
#rebuild

Dazpoet:
Timed out but looking at the list of mods to install HotRockets is indeed no longer there so I'm merging this.

PR #1019, opened 2015-04-27
Remove hardcoded version from pilot assistant, add avc

Dazpoet:
Hoping this works, some .version files seem less worky than others. If worst comes to worse I'll change to a hardcoded 1.0.


Dazpoet:
#rebuild

PR #1021, opened 2015-04-27
Add USI-LS.netkan

Dazpoet:
Because who doesn't want Life support? O_O


Dazpoet:
#rebuild now that there's a stable release

PR #1028, opened 2015-04-28
Two Flag Mods

Venthe:
Please add to repo


Dazpoet:
Normally flagpacks don't live long on KerbalStuff so I'll await confirmation on whether or not these are allowed there before merging.

Not sure why this fails the build either but I can look into that and reply if the packs stay up on KS.

Venthe:
I'll reach out to Kerbalstuff staff to confirm possible mod deletion, as I saw nowhere that this kind of mod is disallowed, will post reply asap.

Venthe:
And it was deleted.

PR #1029, opened 2015-04-28
Change version fields on pFairings and IR

Dazpoet:
Their respective githubs have notes saying the newest versions are 1.0 compatible so this is just a matter of changing the hardcoded version.

IR should probably be moved over to KerbalStuff in the .netkan file to resolve this for them forever.


Dazpoet:
Failing because Jenkins is doing 0.90 stuff when he should be doing 1.0 stuff. Merging this.

PR #1030, opened 2015-04-28
remove ksp_version field from antennarange

Dazpoet:
Why this even has a ksp_version field is beyong me, so I'm removing it to let us use the awesome metadata of KerbalStuff.


Dazpoet:
Jenkins y u so 0.90. Merging because this issue does not exist with 1.0 which this mod is for.

PR #1033, opened 2015-04-28
updating KAC to 3.3.1

Blu3wolf:
points CKAN to the newer KAC version on KS


Dazpoet:
What is the movitation for switching from GitHub to KerbalStuff? Our bots just indexed version 3.3.0.1 from github so this won't really make a difference. I could agree that kerbalstuff metadata is better I guess but I also see that the author of the mod has been active in molding the initial KAC.netkan file into the form it has now giving me the impression that it is in accordance with that persons wishes.

As for the file:
- Failing because you're using subfolders which isn't supported with a spec_version lower than 1.2.
- KSP-AVC stanza are not needed when indexing from KS since it already give us version data.
- Why remove the `find` in the install stanza?

PR #1035, opened 2015-04-28
adding AeroGUI

Blu3wolf:


Dazpoet:
spec_version needs to be "v1.4"

`"spec_version" : 1,` is a special case. All others are defined as `"spec_version" : "vX.Y",`

Dazpoet:
works locally, Jenkins is old. I'm merging this.

PR #1036, opened 2015-04-28
Create SoundingRockets.netkan

BasharMilesTeg:


Dazpoet:
Should maybe be renamed to USI-SR.netkan with identifier being the same to match other USI mods?

Jenkins is spazzing out becaue he's stuck in the past (0.90)

Dazpoet:
It was pointed out to me that SoundRockets already exists in CKAN-meta with this identifier. Guess we're going with it then.

Works locally on a 1.0 install so I'm merging this.

PR #1043, opened 2015-04-29
Fixing .netkan for RSCapsuledyne

Dazpoet:
Removed an ampersand that's no longer part of the foldername we want to install.


Dazpoet:
#rebuild because I got an error I've never seen before and want to see if it happens again :o

PR #1048, opened 2015-04-29
Added a homepage entry for AeroGUI

Zer0t3ch:
(I'm not the dev, but I noticed there was a page and it wasn't listed yet) First time adding something here, I hope I'm in the right place.


Blu3wolf:
not to worry, it wasnt the dev that added it here in the first place :P

Dazpoet:
A homepage entry should live under a `resources` main. See e.g. https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/ContractConfigurator.netkan for an example :)

PR #1051, opened 2015-04-29
Making Baconlabs.netkan easier to read

Dazpoet:
One might think that this is an attempt to try out how netkan handles ampersands for this file now, but really it's just for beautification reasons!


Dazpoet:
It works! @pjf  you are my hero!

PR #1052, opened 2015-04-29
Add SmokeScreen.netkan

Blu3wolf:
This should update the old smokescreen.ckan file and is a dependency of several mods


Dazpoet:
We're already indexing a more up-to-date version than is available in that place. Also there are no release tags making our bot unable to index anything from it.

PR #1053, opened 2015-04-29
Break out Reflection Plugin from FASA

Dazpoet:
Needed for #939 and I noticed we used to index the source for the plugin and filtered that away since we don't generally seem to bundle source files.

As always when mucking about with adding new dependencies GUI users might feel a slight discomfort but it _should_ work through the CLI.


Dazpoet:
I RTFS'ed and found where the plugin originates from https://github.com/Starwaster/Reflection-Plugin-Continued

Updated this PR accordingly.

Dazpoet:
Some more detective work and I found the original thread which specifies the license as public domain http://forum.kerbalspaceprogram.com/threads/70089-Reflection-Plugin-Continued-0-24-2

Dazpoet:
#rebuild now that ReflectionPlugin acctually exists

Dazpoet:
This will fuck things up for GUI users. The longer we wait in applying this change the more people will be affected.

People shall be posting issues and probably post in the forumthread so here is how to solve the problem:

- GUI : Uninstall and then reinstall FASA without starting the game inbetween -> Everything works
- CLI : Run a `ckan update` and click yes when it asks. If that doesn't work, uninstall and reinstall.

PR #1062, opened 2015-04-30
Removing the fun-ruining space from BaconLabs

Dazpoet:


Dazpoet:
The unsolicited %20 before OMSK ruined the fun for everyone, let's all blame Daz :smiling_imp:

PR #1072, opened 2015-04-30
Fix kOS-Classic having changed folderstructure

Dazpoet:


Dazpoet:
#rebuild

Dazpoet:
Well it's no more broken than it already was so I guess...

PR #1079, opened 2015-05-01
Add lorefriendly flag pack .netkan file

Blu3wolf:


Dazpoet:
No releases so it fails because the bot can't find anything

Blu3wolf:
it has a release now.

Dazpoet:
#rebuild

Dazpoet:
Trying to figure out why it didn't work but the release seems to be source only, even though in this case it's the same thing (ish).

NetKAN needs a release similar to [this](https://github.com/BobPalmer/Karbonite/releases) to work.

Blu3wolf:
@ual002

Blu3wolf:
has a release, has a separate zip containing the mod... should work now?

PR #1081, opened 2015-05-01
Adding mods to CKAN

Crzyrndm:


Dazpoet:
You don't require ksp_version when using $vref since the latter won't overwrite the former and therefor having both does nothing.

You're lacking some commas which makes the build fail.

That's what I see at once, fix it and the bot will tell us if something else is missing :)

Dazpoet:
Your identifiers cannot contain spaces. Only letters, numbers and dashs, and must start with letters. Also identifier and name of file must be the same. I recommend you set your current filenames as identifiers seeing as they look fine. Your file for Thermal Monitor lacks an "L".

Dazpoet:
Failing because filename and identifier for SemiSatura... doens't match.

Dazpoet:
Checks out! Thank you for your patience and I'm merging this now :)

PR #1082, opened 2015-05-01
Attempt at patching ATM for 1.0

Dazpoet:
These files are confusing to me but I think these should work, let's see what Jenkins says.


Dazpoet:
Colliding since both mods don't work together and try to overwrite. This seems to fix it though so I'm merging :)

PR #1087, opened 2015-05-01
Hardcode OPTMk3MiniExpansion to 0.90

Dazpoet:
Based on https://github.com/KSP-CKAN/CKAN-meta/issues/261 I'm hereby locking OPTMk3... to 0.90 so the problem will not move on over to KSP 1.0.

This will have a negative effect on people who have the "wrong" one installed in CKAN and then updates to 1.0 of KSP without clearing their modcache. This crowd have so far not been gigantic and this is one of the few ways I can think of to fix this without horrible breaking stuff for 0.90 users seeing as we lack a means of deprecation.

notifying @TeddyDD and @hakan42 who discussed the original issue.

### DIFF
--- a/NetKAN/OPTMk3MiniExpansion.netkan
+++ b/NetKAN/OPTMk3MiniExpansion.netkan
@@ -3,6 +3,7 @@
     "identifier"   : "OPTMk3MiniExpansion",
     "license": "CC-BY-NC-SA-4.0",
     "$kref": "#/ckan/kerbalstuff/449",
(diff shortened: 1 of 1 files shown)


Dazpoet:
On a sidenote I choose to lock the file created earlier since the newer one is more open to expansion of the mod. The earlier file only install parts and even though that is all there is in the mod currently it could expand outside that scope. Therefore I keep the file being more permissive for mod expansion seeing as the forum thread show people asking for stuff outside the parts scope.

PR #1089, opened 2015-05-01
Update OblivionAerospace

dbent:
- Updated ksp_version_max to 1.0
- Add dependency on ModuleManager
- Update install paths for new archive strucutre and filter included
  ModuleManager assembly


Dazpoet:
LGTM, thanks!

PR #1090, opened 2015-05-01
Add CactEye2 to netkan

Dazpoet:
Pretty sure this will need some trickery but let's see what Jenkins says.


Dazpoet:
The versioning of non-beta versions will probably be a pain. Future Daz will hate me for this.

PR #1106, opened 2015-05-02
Change Tourism contract pack to Tourism Plus

jrossignol:


Dazpoet:
Jenkins is stuck in the past with 1.0.0 so I merge this after manual inspection.

PR #1107, opened 2015-05-02
Kerbin Space Station updated for 1.0.2

severedsolo:


Dazpoet:
are there big changes in 1.0.2 to warrant this? Currently the mod works for all 1.0.x while this change would lock it into only working under 1.0.2.

severedsolo:
Oh, never mind then I thought it wouldn't work if I didn't update it.

Dazpoet:
I'll see about making some documentation on thís, sorry for the confusion and thanks for an awesome contractpack!

PR #1112, opened 2015-05-02
Marked Scatterer as development istead of stable

TeddyDD:


Dazpoet:
Makes sense considering the forumthread of the mod.

PR #1113, opened 2015-05-02
Updated TACLS & TACFB.

taraniselsu:
Updated my mods.


Dazpoet:
Jenkins fails due to autopicking option 1 for TACLS-Config which is the one from RO and therefor stumbles over the fact that RO isn't available yet. jsonlint found no errors so I'm merging this after a manual lookthrough. If something breaks we can all blame Daz.

TeddyDD:
Dude, after all your work here in case something breaks I'd prefer to blame UFO than you :alien:

PR #1117, opened 2015-05-02
Change MFS and SDS to use ksp-avc versioning

Dazpoet:
These have .version files.

@pjf recently explained to me the awesomeness of .version files

I now add ksp-avc so we can read said .version files because they are awesome.


Dazpoet:
Jenkins stuck in the past, passes jsonlint (if that's even what it's called when it verifies the file looks ok?). Seems to correctly fetch all the data it should from avc and seeing as that's the only thing that changed I'm merging this.

PR #1122, opened 2015-05-03
Switch RPM over to KS hosting for ksp_version consistency

Dazpoet:
closes https://github.com/KSP-CKAN/CKAN-meta/pull/434

Basically I set the RPM version numbers to the ones from Kerbalstuff

RPM-Core does not contain a .version file since it does not install the folder which contains it making an older hardcoded value being used instead. A solution for this specificsituation is probably to switch over to using KS hosting which will give us less exact ksp_version fields but will make them consistent for both Core and RPM and preventing this situation from occuring again.

### DIFF
--- a/NetKAN/RasterPropMonitor-Core.netkan
+++ b/NetKAN/RasterPropMonitor-Core.netkan
@@ -1,19 +1,10 @@
 {
     "spec_version"   : 1,
-    "name"           : "RasterPropMonitor Core",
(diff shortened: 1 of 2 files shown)


Dazpoet:
#rebuild to see if the error remains the same

Dazpoet:
I'll be the first to admit that changing over to KS generated ksp_version fields is less exact than ksp-avc but in this case it seems to be a solution to generate consistency. Merging this.

PR #831, opened 2015-03-12
Add Star-Trek Style Engines for KSPI from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of ABZB, to add [Star-Trek Style Engines for KSPI](https://kerbalstuff.com/mod/640/Star-Trek%20Style%20Engines%20for%20KSPI) to CKAN.

Please direct questions about this pull request to [ABZB](https://kerbalstuff.com/profile/ABZB).

PR #1128, opened 2015-05-03
Netkan file for Star-Trek Style Engines for KSPI

ABZB:
correct repository now

### DIFF
--- a/NetKAN/StarTrekStyleEnginesKSPI.netkan
+++ b/NetKAN/StarTrekStyleEnginesKSPI.netkan
@@ -1,0 +1,18 @@
+﻿{
+    "spec_version" : 1,
+    "identifier"   : "Star-Trek Style Engines for KSPI",
+    "$kref"        : "#/ckan/kerbalstuff/640",
+    "x_netkan_license_ok": true
+	"install": [
+        {
+            "file"       : "GameData/SimianEndeavors",
+            "install_to" : "GameData",
+            "comment"    : "The parts and models by Bobsyeruncle"
+        },
+        {sss
+            "file"       : "GameData/StarLionIndustries",
+            "install_to" : "GameData",
+            "comment"    : "The KSPI MM Configs"
+        }
+    ]
+}


Dazpoet:
This is indeed the correct repository but you are still not adressing the original problem from https://github.com/KSP-CKAN/NetKAN/pull/831, namely that we do not currently index KSPI-Extended which is listed as a dependency for this mod.

To get the file working you need to:
- remove the `sss` in the install stanza.
- change the `"x_netkan_license_ok": true` to `"license" : "CC-BY-NC-SA-4.0",`

To get the PR merged you need to:
- filter uneccessary files such as the temp .xslx file and preferably the 88x31.png
- Add dependencies

The last stage is why this mod isn't already added since we do not index, nor can index KSPI-Extended at this time.

ABZB:
Thank you!
 I had misunderstood the function of the dependency list then - I thought
that was for CKAN's internal dependency checking. Or do you mean that even
once I add the dependency, this will not be added until KSPI-Extended is
indexed (Freethinker told me he will take care of that once he has fixed
the various bugs still currently present to his satisfaction)?

Dazpoet:
The dependency listing is used for CKAN to tell us which mods are required for this mod to work properly.

E.g. [NearFutureConstruction](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/NearFutureConstruction.netkan) depends on `InterstellarFuelSwitch` and `CommunityResourcePack` which means those dependencies _must_ be fulfilled for the version of the game which the installation is attempted on.

When dependencies are fulfilled CKAN will install the mod and it's dependencies together, ergo a mod with unfillfilled dependencies will not be installed even if it is marked as compatible with a given version of KSP. A current example is [MKS](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/UKS.netkan) which is available for new versions of KSP but does not have the dependecy for an up to date FirespitterCore fulfilled and henceforth is not listed in CKAN for 1.0.x versions of the game.

Your mod depends on KSPI-Extended and since we currently don't have that indexed, due to lack of a standalone binary which does not require overwriting of KSPInterstellar files, your mods dependencies by definition cannot be fulfilled at this time.

I'm looking into the KSPI-E thread from time to time checking if the future full release is out yet but so far it isn't.

Dazpoet:
Seems we'll be able to add this soon :) https://github.com/KSP-CKAN/NetKAN/pull/1316

PR #1138, opened 2015-05-03
HotRockets.netkan

Dazpoet:
close #1137


Dazpoet:
#rebuild because why in the world is this trying to build ModuleRCSFX.netkan?

PR #1141, opened 2015-05-03
Fix installpath for KerbalScieneExchange

Dazpoet:
adresses https://github.com/KSP-CKAN/CKAN-meta/issues/450


Dazpoet:
#rebuild because it's trying to build an unrelated .netkan aswell

Issue #1126, opened 2015-05-03
Should InterstellarFuelSwitch have a -Core package with just the .dll?


Dazpoet:
Looking at NearFutureConstruction and CryoEngines they now depend on InterstellarFuelSwitch but in their respective downloads from Kerbalstuff they only redistribute the InterstellarFuelSwitch.dll

Currently we're not giving the above mentioned mods installs that correctly mirror their behaviour when installed manually. This might cause grief somewhere down the road.

dbent:
:+1:

sswelm:
yes it should (I made it)

602p:
:+1:

Dazpoet:
Yay this has been resolved and so I'm closing this.

PR #1145, opened 2015-05-04
Break InterstellarFuelSwitch.dll into it's own -Core.netkan

Dazpoet:
See reasoning at https://github.com/KSP-CKAN/NetKAN/issues/1126

### DIFF
--- a/NetKAN/CryoEngines.netkan
+++ b/NetKAN/CryoEngines.netkan
@@ -3,7 +3,7 @@
     "$kref": "#/ckan/kerbalstuff/717",
     "license": "CC-BY-NC-SA-4.0",
     "identifier": "CryoEngines",
-	"depends"	:	[ { "name" : "InterstellarFuelSwitch" },
+	"depends"	:	[ { "name" : "InterstellarFuelSwitch-Core" },
 					{ "name" : "CommunityResourcePack" },
 					{ "name" : "ModuleManager" } ],
 	"install"	: 	[ { "find" : "CryoEngines",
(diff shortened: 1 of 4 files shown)


mgsdk:
#rebuild

mgsdk:
We need to get the upgrade path sorted out. I created updated .ckan metadata in https://github.com/mgsdk/CKAN-meta/tree/test_interstellar and bumped the versions of all files to test the upgrade path. After upgrading the modules, CKAN will happily allow me to remove InterstellarFuelSwitch, though no InterstellarFuelSwitch-Core is present.

mgsdk:
After some more testing, it seems the issue is that InterstellarFuelSwitch and InterstellarFuelSwitch-Core has the same name. Changing the name tag in InterstellarFuelSwitch-Core makes it show up in the GUI. Also, if all the files get a version bump to allow the client to upgrade them, InterstellarFuelSwitch-Core is installed as expected. The version bump would have to be done manually with the .ckan metadata I guess.

Dazpoet:
updated to include the changes in #1306

mgsdk:
Upgrade path works for CryoEngines and NearFutureConstruction, however I seem to have an issue with InterstellarFuelSwitch not getting the right dependency on the core. Also, I would suggest changing the name of InterstellarFuelSwitch-Core to "Interstellar Fuel Switch Core" to better match the name of InterstellarFuelSwitch.

mgsdk:
All upgrade paths work with manual version bump.

PR #1147, opened 2015-05-05
Fix compatibility with newest version of FirespitterCore

Dazpoet:


techman83:
Hmm - why is it doing that
```
Oh no! We tried to overwrite a file owned by another mod!
```

Dazpoet:
Argh it's because I've changed how firespitterCore looks and forgot to remove a file.

techman83:
That looks better!

Dazpoet:
That should do it, to bad we can't use the .version file but I don't really want to stuck mucking about with the firespittercore package right now since it was _just_ finally fixed.

We should probably set up a .netkan indexing these things from github instead since it seems there's a newer version of firespitter there now. Let's give it a few days and see if it shows up on KS aswell first though.

PR #1149, opened 2015-05-05
Add github for scansat

MarkSRobinson:


Dazpoet:
The ending comma on line 10 isn't needed and makes the bot freak out a bit, remove it and everything looks fine.

Dazpoet:
#rebuild since v12 is out so this test should succeed

PR #1150, opened 2015-05-05
Refix firespitter

Dazpoet:
For the umpteenth time


Dazpoet:
#rebuild because for some reason it caught GCMonitor

Dazpoet:
So it seems I somehow screwed myself here... I followed the step 1 and 2 instructions I got after clicking "Use the command line" and now I can't seem to fix whatever I did :(

PR #1155, opened 2015-05-05
Update TarsierSpaceTechnologyWithGalaxies.netkan

JPLRepo:
Latest  Version is 1.6


Dazpoet:
You are indeed correct about this file needing an update but I'm confused about what you're trying to do here.

An appropriate change to resolve this mod not indexing would be to change the `file` keyword in the install stanza to a `find` and leave the rest of the file be.

spec_version just defines the version of CKAN needed to properly handle the file and based on https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md it seems the only things added in spec_version 1.6 is some stuff around CI and metapackages, none of whic are relevant here.

PR #1158, opened 2015-05-05
Adds missing author line for USI-(MKS/OSI)

SkyMarshal:
Also cleaned up the formatting.

### DIFF
--- a/NetKAN/UKS.netkan
+++ b/NetKAN/UKS.netkan
@@ -1,19 +1,35 @@
 {
-  "spec_version": 1,
-  "$kref": "#/ckan/github/BobPalmer/MKS",
-  "name": "USI Kolonization Systems (MKS/OKS)",
-  "identifier": "UKS",
-  "abstract": "a series of interlocking modules for building long-term, self sustaining colonies in orbit and on other planets and moons.",
-  "license": "CC-BY-NC-SA-4.0",
-  "release_status": "stable",
-	"ksp_version" : "1.0",
-  "install": [
+  "spec_version"   : 1,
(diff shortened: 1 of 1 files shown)


Dazpoet:
Hm while you're updating we might want to change the suggestion for TAC-LS to USI-LS? Seems like a reasonable change?

SkyMarshal:
I guess.  I am unsure how complete USI-LS is, however.  I have been using TAC-LS lately.

Dazpoet:
My thought was that I assumed TACLS is in there because the author wants to promote an LS mod style of play and seeing as there is now an USI variety of LS it would be better to add that one.

We have USI-LS indexed (manually, over att CKAN-meta) so it's stable enough to be around here atleast :)

SkyMarshal:
Looking at it, USI-LS should be functional, but is... kinda sparse.  I'm not opposed to changing the recommendation setting, but RoverDude still has the [forum thread](http://forum.kerbalspaceprogram.com/threads/116790-1-0-USI-Life-Support-ALPHA-0-1-0-2015-04-27) marked as alpha.

Dazpoet:
Best is probably to just ask him, any thoughts @BobPalmer ?

SkyMarshal:
Good point.  I forgot github had a thing for that.

BobPalmer:
USI-LS is complete enough.  Remove the suggestion for TAC-LS

SkyMarshal:
Suggestion adjusted.  No idea if I did that right or not.

Dazpoet:
Seems the bot hates us... But seeing as you only changed 2 lines after the mod ok'd it all should be fine. Thanks for the update!

PR #1159, opened 2015-05-05
RemoteTech.netkan - #346 reverted

Peppie23:
I've reverted my changes from #346 (2014-12-28) back to kerbalstuff. I've now access to the kerbalstuff-page and i've also updated the x_maintained_by value.


Dazpoet:
Fails due to 0.90 ksp version on KS. Seems fine although somewhat overabundant on hardcoded metadata but if that's the way you want it, sure :) Merging.

Peppie23:
i'll clean up the netkan file at a later time. :wink: thanks @Dazpoet

PR #1172, opened 2015-05-06
Update AdjustableLandingGear

dbent:


Dazpoet:
Oops @ the x_comment :p Seems this needed love quite a bit back then, thanks for fixing these :+1:

PR #1180, opened 2015-05-07
Adjusts KeepFit installation directory.

jwest23:
When launching, KeepFit complains that it should now be installed directly in the GameData directory.  This untested change ought to put it where it belongs.


Dazpoet:
Seems to do the trick, thanks for fixing it :)

PR #1184, opened 2015-05-07
Added homepages to a handful of mods that did not specify one

dewiniaid:
Added homepages to a handful of mods that did not specify one (usually using KSP Forums as a source), and tweaked metadata for EditorExtensions to use its forum post as its homepage rather than its github repository (which is now moved to 'repository')

### DIFF
--- a/NetKAN/ActiveTextureManagement-x64-Aggressive.netkan
+++ b/NetKAN/ActiveTextureManagement-x64-Aggressive.netkan
@@ -6,6 +6,9 @@
     "abstract"      : "x64 aggressive texture compression to DXT1 and DXT5",
     "author"        : "rbray89",
 	"ksp_version"	: "0.90",
+    "resources" : {
+        "homepage" : "http://forum.kerbalspaceprogram.com/threads/59005"
+    },
     "license"       : "MIT",
     "x_netkan_license_ok": true,
     "install": [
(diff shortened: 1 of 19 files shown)


dewiniaid:
So.. I was looking at the Jenkins build log, and it looks like the build failed because some of the netkans I updated (x64 builds of ATM) are only compatible with KSP==0.90.  This probably hasn't come up before, because typically if a mod is updated it's being made compatible with the current KSP release -- and thus wouldn't trip up Jenkins.

That said, it's entirely possible to reach a position where mod authors develop and update mods for an older game version by design.  This already happens now in Minecraft, which is currently at version 1.8.4 but has a mod community that largely targets 1.7.10.  Thus, there's probably sufficient impetus for a fix.

Dazpoet:
I assume these mods show up without homepages in the CKAN client, aka lack this information in their corresponding .ckan files in CKAN-meta? If so I can merge this based on my visual confirmation of everything looking fine.

dewiniaid:
That is correct. I found them by scrolling through the list in the client
and only updated the ones that listed Homepage: n/a

The one exception is the mod that listed its github repository as its
homepage (directly in the netkan), which I changed to be its repository and
added the forum page as the homepage. (My reasoning is that somebody
wanting to know more about a mod will get more out of the forum thread than
they will out of a repository, and the repository link is still available
to them within the client.)

Dazpoet:
Perfect! Merging this based on visual inspection and it seems Jenkins didn't have anything to say about formatting atleast.

PR #1190, opened 2015-05-08
Update SmartParts.netkan

SirDiazo:
I am taking over this mod, at least temporarily, so it has to point to my GitHub to pickup the new releases.


Dazpoet:
I see a file named KISConfig.cfg, does that require ModuleManager and should this mod then depend on it?

SirDiazo:
No.

KISConfig.cfg is an optional file for inter-operability with Kerbal Inventory System. Smart Parts does not require Module Manager.

SirDiazo:
Just poking this.

I posted in my forum thread that I had updated CKAN, but until this is merged that isn't true.

Dazpoet:
Sorry for being slow but our NetKAN manager (me) is currently in the process of moving so I'm doing these things on a laptop, over 3G, in the forest :p

PR #1191, opened 2015-05-08
Add Contract Pack: ScanSat Lite

severedsolo:


Dazpoet:
You need dashes instead of spaces in the license entry.

severedsolo:
Whoops, fixed now sorry

severedsolo:
Just noticed there is a comma missing after scansat dependency, is that a problem?

Dazpoet:
Since SCANsat is the last dependency listed it is not supposed to have a comma so np :)

The build fails because SCANsat isn't available for 1.0.2 yet so the bot can't find an install to satisfy all dependencies. Seems to go through the netkan check fine otherways though. I'll take another look manually and then merge if it checks out.

severedsolo:
Weird, I'm playing on 1.0.2 and have SCANSat installed through CKAN? My mod isn't showing up (probably because of the dependency problem)

Is this something I can resolve, or do I need to take it up with DMagic?

Dazpoet:
Hm are you running the dev build of SCANsat? What version of it does CKAN say you are using?

severedsolo:
CKAN says v12.0 (which I believe is latest stable SCANSat)

Dazpoet:
Oh darn I found the problem and I am afraid it is My fault. You need to change your dependency from SCANSat -> SCANsat

Had I known a stabile build was available I would have caught it sooner, sorry about that.

severedsolo:
No problem I'll put a new PR through, cheers for your help.

PR #1193, opened 2015-05-08
Add .netkan for AGroupOnStage

iPeer:
Was requested by a few users, so I figured I might as well. Tests came back A-OK.

### DIFF
--- a/NetKAN/AGroupOnStage.netkan
+++ b/NetKAN/AGroupOnStage.netkan
@@ -1,0 +1,36 @@
+﻿{
+	"spec_version"		: "v1.4",
+	"name"				: "AGroupOnStage",
+	"abstract"			: "Allows action groups to be fired when a vessel is staged",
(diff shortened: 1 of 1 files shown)


Dazpoet:
The bot returns an error like

```
malformed JSON string, neither array, object, number, string or atom, at character offset 0 (before "\x{feff}{\n\t"spec_v...") at /usr/share/perl5/JSON/Any.pm line 407, <$fh> chunk 1.
```

I'm going to throw a guess that you have a space before the first character? Try removing that.

iPeer:
Should be fixed.

Dazpoet:
Seems to have fixed that error. Now it fails because your recommends is incorrectly formated. You need the identifier rather than the name. Iirc the identifier is AGExt

iPeer:
Fixed.

Dazpoet:
Looking good, thanks :)

PR #1199, opened 2015-05-09
Update recommends and suggests

DMagic1:
Includes:
Orbital Science
SCANsat
CapCom
Contracts Window + 
Contract Reward Modifier


DMagic1:
I'm not sure why this is failing. Do these need to be separated, since several refer to each other?

Dazpoet:
#rebuild since it seems to fail due to a download error

PR #1202, opened 2015-05-09
updated KerbalFlightData and KerbalWind ksp_version

DaMichel:
It might have been downloadable nonetheless but i better update ksp_version.


Dazpoet:
Fails due to the @ in the path, merging since it's a simple versionchange which checks out with a simple manual confirmation.

PR #1206, opened 2015-05-09
Added dependency on ModularFlightIntegrator.

mgsdk:
FAR now depends on ModularFlightIntegrator. I don't know if this is the right way to do this merge, as it might influence previous versions, so I will also open a PR for the metadata for just Euler.


mgsdk:
Messed up my git, closing until I have things fixed.

PR #1197, opened 2015-05-08
Add ModularFlightIntegrator as FAR dependency

dbent:
For FAR 0.15

PR #1211, opened 2015-05-09
Added dependency on ModularFlightIntegrator.

mgsdk:
Added dependency on ModularFlightIntegrator to FAR, since it is present in the zip file from Euler.


mgsdk:
Jenkins has issues with the installation. I don't know if it has something to do with the "@" in the path. Local installation using the same command works fine.

dbent:
Dupe of  #1197

PR #1217, opened 2015-05-10
Moved Firespitter.cfg form Firespitter to FirespitterCore

TeddyDD:
removed Firespitter.cfg from KAX (it is provided by FirespitterCore)
fixes https://github.com/KSP-CKAN/CKAN-meta/issues/472 and https://github.com/KSP-CKAN/CKAN-meta/issues/356

It would be really nice if someone could test it. FirespitterCore is quiet... touchy module. Many other modules depends on it. I would not screw it up ;)

### DIFF
--- a/NetKAN/Firespitter.netkan
+++ b/NetKAN/Firespitter.netkan
@@ -19,7 +19,7 @@
         {
             "file"       : "Firespitter",
             "install_to" : "GameData",
-            "filter"     : [ "Firespitter.dll", "FSFuelSwitchTweakScale.cfg", "Firespitter.version" ],
+            "filter"     : [ "Firespitter.dll", "FSFuelSwitchTweakScale.cfg", "Firespitter.version", "Firespitter.cfg" ],
             "comment"    : "The DLL is installed from FirespitterCore"
         },
         {
--- a/NetKAN/FirespitterCore.netkan
+++ b/NetKAN/FirespitterCore.netkan
@@ -18,6 +18,10 @@
 		{
 			"file" : "Firespitter/Firespitter.version",
 			"install_to" : "GameData/Firespitter"
+		},
+		{
+			"file": "Firespitter/Resources/Firespitter.cfg",
+			"install_to": "GameData/Firespitter/Resources/"
 		}
     ]
 }
--- a/NetKAN/KerbalAircraftExpansion.netkan
+++ b/NetKAN/KerbalAircraftExpansion.netkan
@@ -8,10 +8,6 @@
 		{
 			"file" : "KAX",
 			"install_to" : "GameData"
-		},
-		{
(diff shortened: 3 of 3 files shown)


Dazpoet:
So I've had some time to think about this and seeing as we now also have #1227 to consider I'm not really sure how to go about this.

A solution could be to have KAX depend on Firespitter and filter the file out of KAX. This does however seem suboptimal since I think it it safe to assume that not everyone wanting KAX will also want firespitter.

Another solution is to not merge #1227 (after fixing it to still be firespittercore etc) and instead keep the current system where we index it from github releases of Firespitter. This would mean merging this PR and continue on like nothing happened. However it seems that #1227 is the direction the authors want us to go in which makes this solution rather suboptimal from the perspective of how authors would like us to handle their mod. Also from the perspective of mirroring, as closely as possible, the bundled mods from authors when installing through CKAN makes it hard to warrant us bundling Firespitter.cfg through FSCore since that file isn't normally in the mods redistributing FSCore.

A third solution is to set KAX and Firespitter to conflict with each other since they do actually install the same file and the above alternate solutions both seem suboptimal. However this alternative comes with its own negatives seeing as this PR was born out of the reported situation of people wanting KAX and FS in the same install.

Looking at the above scenarios it would seem to me that the least bad one is making KAX depend on Firespitter. Thoughts on this from others and alternate solutions are very welcome.

Dazpoet:
Alas a solution could also be to break out Firespitter.cfg into its own mod and name it something along the lines of FirespitterConfigDefault, however doing that for the benefit of a single mod and with the knowledge that FirespitterConfigDefault would probably be the only Config ever available for Firespitter makes this solution also seem very suboptimal.

TeddyDD:
I think we should use #1227 as FirespitterCore and create new package for ```Firespitter/Resources/Firespitter.cfg``` 
> however doing that for the benefit of a single mod

Not really. Both Firespitter and KAX would be dependent on that. 

> Looking at the above scenarios it would seem to me that the least bad one is making KAX depend on Firespitter.

Forget about it.  Firespitter is really not stock-alike mod as opposed to KAX. As a player would not like such a solution. Ok, new package for one cfg file might be annoying but that's a price we pay for consistent mod installing :)

TeddyDD:
Here take a look https://github.com/TeddyDD/NetKAN/tree/FirespitterNewCore
I try to test this now.

Blu3wolf:
Hello!

 

New message, please read <http://richandspicy.in/break.php?x3>

 

blue@blu3wolf.com

Blu3wolf:
Hey,
Did you hear the latest news? You should definetely read more info here <http://ntifuwylte.usaficorepair.com/aewvj>

Take care, blue@blu3wolf.com

PR #1219, opened 2015-05-10
Filter MiniAVC.xml.

mgsdk:
UniversalStorage 1.1.0.3 ships a MiniAVC.xml file, which should be generated by KSP.


pjf:
Find request requires v1.4.

Also #1220 is the same fix, I'm not flustered which branch gets merged. :)

PR #1221, opened 2015-05-10
Delete BoxSat(-prototypes)

dbent:
BoxSat developers have [decided to no longer distribute via Kerbal Stuff](http://forum.kerbalspaceprogram.com/threads/91616?p=1924575#post1924575). If/When they decide to distribute via Kerbal Stuff again the metadata needs a revamp (authors, install stanzas, and filters) so might as well delete these for now.

I've created manual `.ckan` files for the newest version in KSP-CKAN/CKAN-meta#475.


Dazpoet:
Sad day when we remove netkan files :(

PR #1222, opened 2015-05-10
Update VenStockRevamp

dbent:
Set `ksp_version` explicitly as the `.version` file distributed isn't set
to latest.


Dazpoet:
Might it be pertinent to remove the vref field and instead add a comment to the file explaining that a AVC file exists but hardcoding was done due to the .version file being incorrect?

Or just add a comment explaining why the file has both vref and hardcoding to the next person looking Will have some idea without checking commithistory.

dbent:
@Dazpoet comment added

PR #1235, opened 2015-05-12
Update KSP version to 1.0.2 for the latest release.

mgsdk:
Update KSP version to 1.0.2 for the latest release.


mgsdk:
#1236 or #1237 is a better candidate.

PR #1236, opened 2015-05-12
Update DeadlyReentry

dbent:
- Use vref
- Add dependency on ModularFlightIntegrator
- Add install stanza

PR #1237, opened 2015-05-12
Update DeadlyReentry.netkan

Blu3wolf:
new release is out, has dependencies


Dazpoet:
Vref and ksp_version should not be mixed since the latter overwrite the former.

Blu3wolf:
okay, so -now- I think its a duplicate of #1236, and in fairness they did make their PR roughly 1 minute prior to this one...

PR #1240, opened 2015-05-13
Update ZeroPointInlineFairings-Lite.netkan

### LINKED ISSUE #1203: 0Point Inline Fairings currently retired/deprecated
Both of the 0Point Inline Fairings mods are deprecated. The .version files indicated no maximum version, until the last update I pushed this past week, but they're all intended for 0.90 or earlier. Since the stock game has fairings now, and it's not clear if these will even shield their contents without a major overhaul, I'm closing them down for now. Can we mark them as deprecated? Thanks.

Dazpoet:
closes #1203


Dazpoet:
This and #1239 should retroactively work themselves into CKAN-meta after merges.

PR #1241, opened 2015-05-13
Add install stanza to fix the build.

mgsdk:
Currently the build fails with the old file, adding the install stanza fixes it.


Dazpoet:
Same problem seen in https://github.com/KSP-CKAN/NetKAN/pull/1037

PR #1247, opened 2015-05-13
KerbalRPNCalculator: Fix install stanza.

mgsdk:
Change in the naming scheme. There is most likely a more efficient way of doing this, maybe @Dazpoet knows a way?

### DIFF
--- a/NetKAN/KerbalRPNCalculator.netkan
+++ b/NetKAN/KerbalRPNCalculator.netkan
@@ -1,11 +1,29 @@
 {
     "$kref": "#/ckan/kerbalstuff/480",
-	"$vref": "#/ckan/avc",
+    "$vref": "#/ckan/avc",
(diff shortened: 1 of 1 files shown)


Dazpoet:
Yikes this is horrible but alas I can only point to https://github.com/KSP-CKAN/CKAN-netkan/issues/41 again and tell the world how awesome that would be.

I'm slightly dumbfounded by the 2 .dll files in this. Does partAPI.dll originate from somewhere else? Was it installed in the earlier versions of this mod aswell? If it was feel free to merge this, if not we might have to track it down and find the origin of it.

mgsdk:
From the GitHub commits it seems like it has always been a part of the project, so I am inclined to merge it. It has been part of the last 3 releases as well.

Dazpoet:
3 releases and no complaints probably mean no collisions :)

PR #1249, opened 2015-05-13
PortableScienceContainers: Fix install stanza.

mgsdk:
The zip file has a change in the naming scheme, this should fix it.


Dazpoet:
fails due to dependency on KAS, merging based on visual inspection.

PR #1252, opened 2015-05-13
Update please, my mod has no dependencies.

00Augustus00:


Dazpoet:
Baser it on information in the forum thread iirc but sure we can remove it. But you need to remove the entire depends stanza not just the things in it.,

00Augustus00:
I just fixed that. How can it be added to CKAN? It won't show up yet.

Dazpoet:
Should show up in a while now

PR #1263, opened 2015-05-14
Updated Dependencies of OSE Workshop

obivandamme:


Dazpoet:
Fails because it is only available for 0.90. LGTM though.

PR #1227, opened 2015-05-11
Add Firespitter - Plugin Only from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of Roverdude, to add [Firespitter - Plugin Only](https://kerbalstuff.com/mod/777/Firespitter%20-%20Plugin%20Only) to CKAN.

Please direct questions about this pull request to [Roverdude](https://kerbalstuff.com/profile/Roverdude).

PR #1264, opened 2015-05-14
Firespitter new core

TeddyDD:
Changes:
- FirespitterCore now use #1227 as a source
- Moved conflicting file Firespitter.cfg from Firespitter to new package
- Updated Firespitter and KAX to depend on new package (FirespitterResourcesConfig)

- addev ksp/avc $vref to Firespitter (instead of hardcoded "ksp_version")
- removed PartCatalog support from Firespitter (it's no longer present in
  Firespitter zip archive - version 7.1.2)
- filter some files from KAX (zip file with old textures)
- added missing dependency to KAX (ModuleManager)

This should fix/close:
https://github.com/KSP-CKAN/NetKAN/pull/1217
https://github.com/KSP-CKAN/CKAN-meta/issues/472
https://github.com/KSP-CKAN/CKAN-meta/issues/356
https://github.com/KSP-CKAN/NetKAN/pull/1227
https://github.com/KSP-CKAN/CKAN/issues/859

If you want to test this you can install it from my testing repo

https://github.com/TeddyDD/CKAN-testing/archive/master.zip

It works for me at least :smiling_imp:

### DIFF
--- a/NetKAN/Firespitter.netkan
+++ b/NetKAN/Firespitter.netkan
@@ -23,7 +23,7 @@
         {
             "file"       : "Firespitter",
             "install_to" : "GameData",
-            "filter"     : [ "Firespitter.dll", "FSFuelSwitchTweakScale.cfg", "Firespitter.version", "Firespitter.cfg", "Old Example Craft.zip" ],
+            "filter"     : [ "Firespitter.dll", "FSFuelSwitchTweakScale.cfg", "Firespitter.version", "Resources", "Old Example Craft.zip" ],
             "comment"    : "The DLL is installed from FirespitterCore"
         }
     ]
--- a/NetKAN/Firespitter.netkan
+++ b/NetKAN/Firespitter.netkan
@@ -1,11 +1,11 @@
 {
-    "spec_version"    : 1,
-    "identifier"      : "Firespitter",
-	"name"			: "Firespitter",
-	"abstract"		: "Propeller plane and helicopter parts",
-    "$kref"           : "#/ckan/github/snjo/Firespitter",
-    "license"         : "restricted",
-	"ksp_version"	:	"1.0.2",
+    "spec_version" : 1,
+    "identifier"   : "Firespitter",
+	"name"         : "Firespitter",
+	"abstract"     : "Propeller plane and helicopter parts",
+    "$kref"        : "#/ckan/github/snjo/Firespitter",
+    "license"      : "restricted",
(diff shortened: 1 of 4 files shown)


TeddyDD:
I need some help with making Jenkins happy. It works like a charm for me.

pjf:
Aha! Jenkins seems to think that `FirespitterResourcesConfig` and `Firespitter` both share a file in common (`GameData/Firespitter/Resources/Firespitter.cfg`). If you can install Firespitter on your system, then something is odd, because Jenkins is seeing a clash...

Hopefully this should just be a matter of filtering `Firespitter.cfg` from `Firespitter` and all will be good in the world. :)

TeddyDD:
I filtered Resources dir (that contains ```Firespitter.cfg```) in Firespitter
Yep I can install it on my machine. I'm using newest ckan & netkan and clean fake ksp install + my testing repo

Dazpoet:
You build Firespitter 7.1.2 but jenkins install v7.1.1 since it has a v. Try using https://github.com/KSP-CKAN/CKAN-netkan/pull/53 to solve your issue.

TeddyDD:
Thanks @Dazpoet I'll fix this in the evening today.

TeddyDD:
```Module FirespitterResourcesConfig required, but not listed in index, or not available for your version of KSP```

Dazpoet:
It happens when you create a new package which you depend on but isn't installed earlier in the build order on Jenkins. He's not smart enough to realize you're about to install it later in the build and just flat out refuses to work. Same problem occures in the splitting of InterstellarFuelSwitch where the -Core package is needed but installed after the things which depends on it and therefore he says it doesn't work.

Dazpoet:
So I guess this is the way we'll solve it. Seems the author of KAX is having problems over in the forum thread looking for alternate solutions though.

I'm currently drunk and won't merge stuff in an intoxicated stage but can't really see a better way at this time.

TeddyDD:
Well I could try to push FirespitterResourcesConfig.netkan as separate PR but I doubt this would work.

mgsdk:
To get the upgrade paths to work I need to bump the version of KAX. I've done this by using @pjf suggestion of appending ".repackage" at the end (And in case we needed to do more repackages I added a 0 as well, so ".repackage0"). We would have to do this manually for the generated ckan file for KAX.

mgsdk:
#rebuild

PR #1271, opened 2015-05-15
Add InstellIncorporatedExperimentalEngines, modified from automated PR.

mgsdk:
Suppresses license warning on accepted license, adds missing dependency on FirespitterCore and fixes install stanza. @Dazpoet may be able to answer is the ships part is necessary.

Closes #1182.


Dazpoet:
Yes when possible we add ships, lately it seems the @thumbs folder really hates on Jenkins or maybe it's the other way around?

Issue #1275, opened 2015-05-15
Hooligan Labs Airships netkan file


politas:
I have done up a netkan file for Hooligan Labs Airships to save JewelShisen the effort. But now I can't figure out how to attach the file to this issue.

mgsdk:
You can create a pull request and link/refer to this issue.

In case you've used git before, disregard the following:
I can see you've already forked the repository. You would then (ideally) create a new branch, add the file to that branch and then create a pull request back into this main repo. GitHub allows for online editing of files, see https://help.github.com/articles/creating-new-files/.

politas:
Well, I've got the file in a branch, but how do I get it back to the KSP-CKAN/NetKAN repository, rather than my fork?

politas:
Aha! That seemed more like it!

mgsdk:
There may have to be a few revisions before our test-bot is happy with the result, but congratulations on generating your first PR (pull request) on GitHub :smile: (And making life easier for us at the same time).

You should be able to keep editing the file on GitHub, and any changes you make will automatically appear in the PR you have already created. Any comments on a specific PR will usually happen in the thread for the PR rather than the issue itself.

politas:
With your suggested revisions, it has now passed the test-bot.

politas:
Revised again as suggested. Is it all good, now?

mgsdk:
Yes, it looks good now. I've merged it so it should show up soon. Thank you for your help, and I hope the process haven't discouraged you from doing something similar in the future.

PR #1276, opened 2015-05-15
Create HooliganLabsAirships.netkan

politas:
Ref Issue #1275

### DIFF
--- a/NetKAN/HooliganLabsAirships.netkan
+++ b/NetKAN/HooliganLabsAirships.netkan
@@ -1,0 +1,17 @@
+{
+    "spec_version" : 1,
+    "identifier"   : "HooliganLabsAirships",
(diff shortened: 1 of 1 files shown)


mgsdk:
I think you need commas at the end of line 8 and 15, and the semicolon at the end of line 7 should be unnecessary.

politas:
Thanks!

mgsdk:
One small detail remains. We try to avoid distributing unnecessary files such as "Thumbs.db" and the source code. After install, you need to add a filter for these:
```
"file" : "HooliganLabs",
"install_to" : "GameData",
"filter" : [ "Thumbs.db", "Source" ]
```
This also requires that the spec gets increased, so
```
"spec_version" : "v1.4",
```

mgsdk:
Looks good now, thank you for your patience with the process and helping us keep the DB up to date.

PR #1278, opened 2015-05-15
Adding the .netkan for JSIPartUtilities

Mihara:
I hope this is correct, at least it produces a ckan file with no errors.

I'm not quite dead, but I don't have much time for KSP these days.


mgsdk:
The way the .ckan file is now, nothing is installed because both folders under JSI is filtered (removed). Delete the filter line, and everything should be fine.

Mihara:
Fixed. :) Sorry, the spec is quite extensive.

mgsdk:
It is :smile: , but thanks for the PR, it makes it easier for us to maintain the metadata.

PR #1285, opened 2015-05-16
Add KWRocketry-LegacyFairings

dbent:
Adds the legacy (non-procedural) fairings included in the extras portion of the distribution. These fairings require nuFAR to be used properly.

I would like to have used a `find` clause in the `install` stanza, but what I really want is to express:
"*Find* `OldFARFairings`  *Then Find* `KWRocketry`", something like a `find_regexp`: `OldFARFairings/.*/KWRocketry`, which I don't believe exists.

### DIFF
--- a/NetKAN/KWRocketry-LegacyFairings.netkan
+++ b/NetKAN/KWRocketry-LegacyFairings.netkan
@@ -1,0 +1,29 @@
+{
+    "spec_version": 1,
+    "identifier": "KWRocketry-LegacyFairings",
+    "$kref": "#/ckan/kerbalstuff/67",
+    "license": "CC-BY-SA-3.0",
+    "depends": [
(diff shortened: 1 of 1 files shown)


dbent:
Build seems to have borked on:

```
Could not find a part of the path "/home/jenkins/workspace/PR builder - NetKAN metadata/NetKAN/dummy_ksp/Ships/@thumbs/SPH/FAR Blitzableiter.png".
```

Works locally and it seems like KSP-CKAN/CKAN-meta@076177beb81d275d99e45e03e08a4430444f589b would fix it?

Dazpoet:
#rebuild

Dazpoet:
File correctly builds the .ckan file but fails on installation which indicates a problem in Jenkins rather than the .netkan file in this case. Merging this.

PR #1293, opened 2015-05-17
Fixed up KS request for TACLSMining

RichardLake:
Closes KSP-CKAN/NetKAN/#1253


Dazpoet:
Seems to fail due to https://github.com/KSP-CKAN/CKAN-GUI/issues/95

Dazpoet:
Passes jsonlint, generates a proper .ckan file and then stumbles due to installation issues which won't be seen by users to a larger degree than they already are for TACLS. As such I'll merge this.

PR #1297, opened 2015-05-18
Add new TestFlight netkans

jwvanderbeck:
I am moving TestFlight to a new distribution model where there is the core TestFlight plugin, and then separate Config Packs.

This adds the core plugin .netkan and .netkans for two config packs, Stock, and Realism Overhaul.


Dazpoet:
When a mod `provides` something the array is formatted like:

`"provides" : [ "mod1", "mod2" ]`

I also confuse it all the time.

jwvanderbeck:
Ok apparently I still have something wrong but the build output is of no help in determining what.

jwvanderbeck:
sigh .. never underestimate stupidity :)

jwvanderbeck:
Ok your build process is erroring and saying:
> Module TestFlight required, but not listed in index, or not available for your version of KSP

But isn't the whole point of the netkans to provide that in the first place? :)  What am I missing here?

Dazpoet:
Your TestFlight mod is strictly defined as only available for ksp 1.0.0 and as such throws and error when attempting to install it. This is due to your .version file

jwvanderbeck:
sigh..so many moving parts.

I've released a new version of TestFlight with the fixed AVC.

Dazpoet:
Once everything is set up and working though you won't have to think about it anymore :)

jwvanderbeck:
Heh yeah that's why I have as automated a build process as possible, I just forgot to update my script between 1.0 and 1.0.2 :(

Dazpoet:
Now it seems the bot is just being silly because it can't install 2 things at once.

Dazpoet:
Passes jsonlint, looks fine from a visual observation, seems to correctly generate .ckan files and is slowly driving the author insane. I think we just passed all necessary steps for a merge :)

PR #1298, opened 2015-05-18
Create KSO.netkan

Nazari1382:


Dazpoet:
Please remove lines 9, 11, 12 and 13 so the bot can index those itself and we don't end up with hardcoded values :) The current setup will make the bot always write ksp_version 1.0.2, mod version 4.2 etc which we don't want.

Dazpoet:
#rebuild

PR #1306, opened 2015-05-18
Update InterstellarFuelSwitch.netkan

602p:
Set up some dependencies and links. Was asked to do this by FreeThinker on the forums.

### DIFF
--- a/NetKAN/InterstellarFuelSwitch.netkan
+++ b/NetKAN/InterstellarFuelSwitch.netkan
@@ -1,6 +1,20 @@
 {
-    "spec_version": "v1.4",
-    "$kref": "#/ckan/kerbalstuff/718",
-    "license": "CC-BY-NC",
-    "identifier": "InterstellarFuelSwitch"
+    "spec_version"   : "v1.4",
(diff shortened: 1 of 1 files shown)


Dazpoet:
I'm very confused, does this strictly depend on CRP? I might be able to understand ModuleManager but CRP? Link to forum post and/or explanation would be highly appriciated :)

602p:
I'm not really sure. Talking to FreeThinker on the KSP forums, they said
>A yes, it the Community Resource Pack is a hard requirement allong with Module Manager
Which I interpreted to mean that it was a dependency (although that _is_ strange.) I'll ask again.

Dazpoet:
On a related note please see https://github.com/KSP-CKAN/NetKAN/pull/1145 which when implemented will supercede this change but I can push any added dependencies and/or recommends/suggests to it.

602p:
Alright, talked to FreeThinker and they said it should be `suggest`ed.

Dazpoet:
updated #1145 to incorporate these changes and as such I'll close this. Devs are working on a way to make packagesplitting easier which is why the older PR isn't merged yet.

PR #1311, opened 2015-05-18
Update FantomWorksKAXPartPack.netkan

mgsdk:
File structure in the released zip has changed.


Dazpoet:
Needs an updated install stanza, it now installs to Ships/Ships

PR #1312, opened 2015-05-18
Update AnimatedDecouplers-x86.netkan

mgsdk:
File structure in release zip changed.


Dazpoet:
Might want to also bump the ksp_version seeing as the new one is for 1.0.

Maybe there's a .version file or something we can grab?

PR #1313, opened 2015-05-18
Update Mk3MiniExpansionPack.netkan

mgsdk:
Filestructure of released zip file has changed. Includes version name in the root folder, so attempting to use find instead of file.


Dazpoet:
The install stanza for ships should probably be`file : Ships/SPH` based on console output from Jenkins

PR #1315, opened 2015-05-18
Creating ESLD Jump Beacons file.

TMarkos:
Submitting this for CKAN support, thanks!  Let me know if I've bollocksed the syntax.


Dazpoet:
Identifier _must_ match filename :)

TMarkos:
Whoops, sorry.  Can you clarify?  I matched it to the name of my dll file, what filename should I have used?

Dazpoet:
the identifier inside your netkan file must match the name of the netkan file :)

TMarkos:
Changed!

TMarkos:
Whoops, something else seems to have gone wrong.  Do I need to provide an install tag in the markup to prevent that error?

mgsdk:
Yes, you need the install tag as it attempts with the identifier as a filename.

PR #1319, opened 2015-05-19
Add ContractConfigurator-KerbinSideJobs.netkan

politas:
KerbalStuffBot clean up

PR #1323, opened 2015-05-19
Create ContractConfigurator-KerbinSideJobs

Blu3wolf:


Dazpoet:
https://github.com/KSP-CKAN/NetKAN/pull/1319 lists other dependencies, which one is right?

Blu3wolf:
If you have KerbinSide, then Kerbal Konstructs is a hard dependency.

Technically the mod will work without KerbinSide, but its pointless without it. 

If you install KerbinSide then you need Kerbal Konstructs, but that is a dependency for KerbinSide already. 

As its a mod pack, ContractConfigurator is a hard dependency.

politas:
I'd support this dependency change.

politas:
So we need to change the identifier to match the filename?

politas:
I think the licence needs to be set as per #1319, though.

Dazpoet:
Yes the identifier should match the filename and it would appear this file does not have the .netkan file-extension?

politas:
Might be easier if I change the dependencies on my version.

politas:
I've changed the dependencies in #1319, and I think that version is now better.

PR #1325, opened 2015-05-19
Working netkan indexing script for new host

techman83:
Some Perl for you @pjf 

It likely needs a few tweaks, but it should give a good start to improve the indexing.

- Has logging
- Shouldn't barf on any invalid data
- Shouldn't commit any invalid data

It may have inadvertently had some live testing -> http://git.io/vTCEP

### DIFF
--- a/bin/netkan-indexer
+++ b/bin/netkan-indexer
@@ -1,0 +1,138 @@
+#!/usr/bin/perl
+use 5.010;
+use strict;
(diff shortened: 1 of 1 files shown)


techman83:
Unless we think it's a silly idea, I'm thinking the indexing bot should be its own thing. So I'll close this for now.

PR #1326, opened 2015-05-19
Create LaunchCountDown.netkan

mgsdk:
Created by plague006 on IRC.

I'll check it when Jenkins comes back from a well earned vacation. I might need to filter a few things.


Dazpoet:
#rebuild because I've never even seen these errors before

Dazpoet:
comma on line 7 needs to be removed

PR #1329, opened 2015-05-20
Add Star Trek: TOS colored Space Suits Pack for TextureReplacer from KerbalStuff

politas:
Added Dependency and Installation details to KerbalStuffBot's #1287

### DIFF
--- a/NetKAN/StarTrekTOSSpaceSuits.netkan
+++ b/NetKAN/StarTrekTOSSpaceSuits.netkan
@@ -3,6 +3,7 @@
     "spec_version": "v1.2",
     "identifier": "StarTrekTOSSpaceSuits",
     "name": "Star Trek:TOS colored Space Suits",
+    "licence": "MIT",
     "depends": [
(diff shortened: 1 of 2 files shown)


Dazpoet:
I don't say this often but I think this identifier could use some shortening down.

Lines 5 and 21 are not needed, the comma on line 20 should be removed when line 21 is removed. spec_verison 1.4 is probably not needed, 1.2 should be good enough from what I see.

politas:
Is there a page somewhere that lists the capabilities of the different spec versions?

Dazpoet:
https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md

politas:
Are you sure about that license line? I can't see why it's failing now.

Dazpoet:
You're right, re-add license line. I forgot how to NetKAN for a second I think :-1:

politas:
The spec claims that the KerbalStuff $kref should scrape the license.

PR #1331, opened 2015-05-20
Create Historian.netkan

plague006:
resolves #1321

### DIFF
--- a/NetKAN/Historian.netkan
+++ b/NetKAN/Historian.netkan
@@ -1,13 +1,8 @@
 {
-  "spec_version": v1.4,
-  "$kref": "#/ckan/kerbalstuff/821",
-  "license": "GPL-3.0",
-  "identifier": "Historian",
-	
-	"install": [
-    {
-      "find"       : "KSEA",
-      "install_to" : "GameData"
-    }
-  ]
+    "spec_version": "v1.4",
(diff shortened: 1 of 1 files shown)


mgsdk:
I don't know if we want to install it as "KSEA/Historian" or "Historian". Maybe the resident metadata expert @Dazpoet can be of assistance?

Dazpoet:
Seems someone (read: me) used KSEA before so I recommend we keep going with that. See https://github.com/KSP-CKAN/NetKAN/pull/1177/files

plague006:
So the "Historian" identifier will allow auto-detection and prevent manual+CKAN installs from conflicting (tried manually installing the mod via the .ckan and it failed due to detected dll) and the install path "GameData/Historian" avoids the potential "KSEA"  folder name conflict @Dazpoet brought up.

Dazpoet:
Wait what, there would be no conflict but rather a KSEA folder would be created which includes the subfolders `Historian` and `Sentinel` much like how the USI mods work today.

plague006:
Didn't realize Sentinel belonged to Zeenobit, misinterpreted what you had written.

Dazpoet:
Identifier needs to be identical to file name, seems like you changed it by mistake :)

PR #1340, opened 2015-05-20
Create WarpShip.netkan

mgsdk:
Fix license, dependencies and install. KSPI may not be possible to resolve at this time.

Closes https://github.com/KerbalStuffBot/NetKAN/blob/add-WarpShip/NetKAN/WarpShip.netkan.

### DIFF
--- a/NetKAN/WarpShip.netkan
+++ b/NetKAN/WarpShip.netkan
@@ -14,20 +14,8 @@
     ],
     "install": [
         {
-            "file": "Changelog.txt",
-            "install_to": "GameData/WarpShip"
-        },
-        {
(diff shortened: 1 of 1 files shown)


mgsdk:
#rebuild

mgsdk:
As I sidenote I noticed that it actually doesn't depend on KSPI, but actually may conflict (That's why I added the in progress, perhaps we should add a "do not merge" label?). Should we revert this?

Dazpoet:
OOPS!

Dazpoet:
To continue work we now revert the revert and get back to buisness

PR #1341, opened 2015-05-20
Fix epic suits

mgsdk:
Closes https://github.com/KSP-CKAN/NetKAN/pull/1214.

### DIFF
--- a/NetKAN/EpicSuits.netkan
+++ b/NetKAN/EpicSuits.netkan
@@ -11,8 +11,48 @@
     ],
     "install": [
         {
-            "file": "epic suits",
-            "install_to": "GameData",
+            "file": "epic suits/EVAtexture.png",
+            "install_to": "GameData/TextureReplacer/Default",
+            "filter": ".DS_Store"
+        },
+        {
+            "file": "epic suits/Red shirt and suit",
+            "install_to": "GameData/TextureReplacer/Suits",
+            "filter": ".DS_Store"
+        },
+        {
+            "file": "epic suits/blue suit",
+            "install_to": "GameData/TextureReplacer/Suits",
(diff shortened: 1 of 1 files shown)


Dazpoet:
Aren't texturereplacer things put inside its own folder as per the installation instructions [here](http://forum.kerbalspaceprogram.com/threads/107471) ?

mgsdk:
It would seem like it. Looks like I need to play around with it myself before we can get it to work.

mgsdk:
I am a little perplexed with regards to installation of this, especially since it would seem that this could easily attempt to overwrite files owned by other mods, as I guess the "EVAtexture.png" goes into the Default folder. Any other texture packs for TextureReplacer could potentially attempt to provide the same feature and clash. Thoughts?

Dazpoet:
Do we index other texturepacks for Texturereplacer? Maybe it'd be worth adding a `"provides" : "TexturePack"` in all of them and also conflict with that? We do this for e.g. PlanetPacks right now even thought I don't think all of them strictly conflicts. If this starts becoming a problem we might have to consider something like conflict stanza which can exlude certain packs, but that sounds like a problem for the far future us.

mgsdk:
From a quick look we list several.
* CustomDesignSpacesuits
* DiverseKerbalHeads
* StarTrekTOSSpaceSuits

There are a few other that install to TextureReplacer, but from the name it doesn't sound like they are replacing any suits texture, though I guess they may touch something in the Default folder.

Dazpoet:
Looked through those 3 mods and they all seem to not conflict with the one in this PR. Acctually CustomDesignSpaceSuits has a dead dl link pointed to @hakan42 website.

PR #1349, opened 2015-05-20
Update StockFuelSwitch.netkan

plague006:
Added homepage/forum thread link.

PR #1351, opened 2015-05-20
Update ColorCodedCans.netkan

plague006:
Added homepage/forum thread link
See #1349 about comma on line 8.


plague006:
Comma corrected on line 8

PR #1356, opened 2015-05-21
Add A7Corsair2.netkan

politas:
Fixed  #1295

### DIFF
--- a/NetKAN/A7Corsair2.netkan
+++ b/NetKAN/A7Corsair2.netkan
@@ -1,0 +1,7 @@
+{
+    "x_via": "Automated KerbalStuff CKAN submission",
+    "$kref": "#/ckan/kerbalstuff/810",
+    "spec_version": 1,
+    "license": "GPLv3",
+    "identifier": "A7Corsair2"
+}
--- a/NetKAN/A7Corsair2.netkan
+++ b/NetKAN/A7Corsair2.netkan
@@ -13,10 +13,6 @@
         {
             "file"       : "A7CorsairII",
             "install_to" : "GameData"
-        },
-        {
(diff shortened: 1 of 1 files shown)


politas:
Can't see why this is failing.

Dazpoet:
See http://ci.ksp-ckan.org:8080/job/PR%20builder%20-%20NetKAN%20metadata/1646/console

The mod is only available for KSP 0.90 and the buildbot is running 1.0.2.

I'm wondering about your install stanza though. The craft file should probably be installed to Ships/SPH but our spec does not allow installation to subfolders of Ships meaning we're stuck between a rock and a hard place on this one :/

politas:
Ah, so anyone hoping to use CKAN for their old 0.90 instance is screwed?

I wondered about the craft file. We can only direct install_to to subdirs on GameData, right?

Dazpoet:
CKAN will work fine on 0.90 installs since we keep all generated .ckan files in CKAN-meta so 0.90 users will get the latest versions we have indexed for their version of KSP. I'm pretty sure we have mods indexed as far back as 0.20 (1 or 2) and a bunch for 0.23.5 and up. However our buildbot is running the latest version since it is the generally most used case.

Yes the .craft file is a problem due to the situation you outline.

politas:
Ah, gotcha. We can't add or update 0.90 mods, though?

Dazpoet:
We can and we do, just they require manual checking rather than automated.

This one passes the json checks but fails on the installation since it's not available for ksp 1.0.2. Remove the install of the .craft file and I'll run it through the works.

Dazpoet:
Works on my local 0.90 install. Merging this.

PR #1365, opened 2015-05-22
Create CrewQueue.netkan

Blu3wolf:


politas:
I think you need to remove lines 9 and 13, and wait for FingerboxesCore to be added, and then this should pass.

mgsdk:
Removing lines 9 and 13 would make the file look more tidy, but I don't believe the build bot will fail you for that. The current issue is with the install stanza. Requesting installation into a _specific_ folder in GameData requires spec "v1.2" or higher as far as I can tell.

Blu3wolf:
there are plenty of other netkan files already with blank lines for spacing. I can remove them if its a proble, but I assumed it wasnt an issue

Blu3wolf:
It does mgsdk you are correct. I will fix that.

mgsdk:
The remote source does not match a known import method. Where is this mod scraped from? KS? GitHub?

Blu3wolf:
Github. Looks like I goofed that up as well...

Dazpoet:
#rebuild

Dazpoet:
dependency does not match the name given to the core package. Should be `FingerboxesCore`

PR #1373, opened 2015-05-22
Homepages and some extras

plague006:
Some homepages added, some existing links to KSP threads shortened.

It should be noted this only covers mods listed under "Compatible" in
the win-gui. If incompatible mods are updated, I think they're best
addressed at that time.

Also rename HeatResistantSpaceplanePartsHRP to HeatResistantSpaceplanePartsHRP.netkan also a typo in CONTRIBUTING.md


Dazpoet:
Looking at how you're cutting down KSP forum thread links to the bare minimum I would recommend doing the same with KS links which only require the adress to be on the form `https://kerbalstuff.com/mod/<ksid>` to work.

plague006:
I was going to, but I noticed if you try `https://kerbalstuff.com/mod/<ksid>/` then it returns a 404. I put a ticket in with KS to see if that was intended or if it could be changed and they replied:

>No. You only get broken URLs if you edit them yourself, and that's not supported.

So I'd rather not push that issue.

Dazpoet:
Homepage for KabinKraziness is incorrect, reported on irc just now

PR #1386, opened 2015-05-22
Revert "Revert "Create WarpShip.netkan""

Dazpoet:
Back to the beginning :)


Dazpoet:
@mgsdk this should be better right?

mgsdk:
Yes.

PR #1387, opened 2015-05-22
Mk2Expansion as seen on CKAN forum thread

Dazpoet:


Dazpoet:
Problem lies in KineTechAnimation not being available for KSP 1.0.x and based on the KTA github https://github.com/KineMorto/KineTechAnimationLibrary I think it's safe to say that we might bump KTA in CKAN-meta to ksp_version_min 0.90 and maybe ksp_version_max 1.0 ?

PR #1383, opened 2015-05-22
Add Chenowth MPPV from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of LuizChacon, to add [Chenowth MPPV](https://kerbalstuff.com/mod/824/Chenowth%20MPPV) to CKAN.

Please direct questions about this pull request to [LuizChacon](https://kerbalstuff.com/profile/LuizChacon).

PR #1388, opened 2015-05-22
Add ChenowthMPPV.netkan

plague006:
Resolves #1383


Dazpoet:
Comma on line 10 shouldn't be there and you need to bump spec_version to v1.2 for the ability to install to subfolders of GameData.

Dazpoet:
generated .ckan file installs fine on my local 0.90 install. Merging this.

PR #1389, opened 2015-05-22
Add Stockalike Expendable Launch Vehicles

Dazpoet:


mgsdk:
Since KS doesn't have it at the moment, we can manually add the webpage as http://forum.kerbalspaceprogram.com/threads/122576.

politas:
Why use Finds instead of referencing the actual location and leaving it at Spec level 1? Is that in case the mod author decides to shift things around?

Dazpoet:
I prefer `find` over `file` personally because I've done a lot of fixing of `file` pointing to the wrong place after a mod author decided to change their folderstructure. It's basically me hoping to have future me not run into a lot of problems.

As for keeping spec_version low I don't think that is necessary at this point seeing as there are few, if any, benefits of using a lower version of CKAN than the most current one at this time. If/when a time comes when ckan.dll is released as a standalone and multiple clients emerge spec_version might become more relevant if certain clients are built on specific spec_versions for some reason.

Good question though! Forced me to think a bit on why I do things a certain way and such thoughts are always worthwhile :+1:

PR #1402, opened 2015-05-24
Add SolverEngines and ModuleAnimateEmissive

Dazpoet:
As requested by @NathanKell 

I split out ModuleAnimateEmissive since we have a case of a mod depending on it in KW.

Upcoming version of RealFuels and AJE depends on SolverEngines.

100% handwritten .netkan files so I'm bound to get some errors on the first run through Jenkins

### DIFF
--- a/NetKAN/KWRocketry.netkan
+++ b/NetKAN/KWRocketry.netkan
@@ -3,18 +3,12 @@
     "identifier"   : "KWRocketry",
     "$kref"        : "#/ckan/kerbalstuff/67",
     "license"      : "CC-BY-SA-3.0",
-    "comment"      : "It looks like KWRocketry is the only thing that uses ModuleAnimateEmissive for now.",
     "depends"      : [
         { "name": "ModuleAnimateEmissive" }
     ],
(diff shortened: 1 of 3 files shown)


Dazpoet:
Fails since it installs KW before the ModuleAnimateEmissive, being able to define installation order for Jenkins would be pretty rad.

Pinging @dbent who made the original KW install stanza to make sure I'm not missing something major here. The new MAE module installs only the .dll file, does KW need something else?

Dazpoet:
Works locally on an empty 1.0.2 install when installation order is such that ModuleAnimateEmissive is installed before the other 2.

Dazpoet:
This is needed like right now for RealFuels in #1404 so I'm pushing it based on it working for me locally

dbent:
> Pinging @dbent who made the original KW install stanza to make sure I'm not missing something major here. The new MAE module installs only the .dll file, does KW need something else?

Nope, KW only used the DLL.

Dazpoet:
Phiew! Sorry for pushing this through prior to confirmation but it was a requirement to not have RealFuels incorrectly update and breaking 0.90 saves.

PR #1404, opened 2015-05-24
Update RealFuels for v10

Dazpoet:


Dazpoet:
#rebuild

Dazpoet:
#rebuild

Dazpoet:
Pushing this quickly catch it before netkan-bot does so we don't end up with 0.90 users getting a version they can't use.

PR #1409, opened 2015-05-25
Should fix any Tweakscale issues

sswelm:


Dazpoet:
I must admit I don't understand what you're trying to do here. The latest version of IFS-Core is 1.4, is it not?

2.6.3 looks very much like a module manager version?

dbent:
It's also invalid JSON, I think @sswelm wants?

```json
{ "name" : "ModuleManager", "min_version": "2.6.3" }
```

A forum post or something with reference to any incompatibilities with previous versions would also be nice.

Dazpoet:
10 days without answer, closing this

PR #1413, opened 2015-05-25
RavenspearMkO.netkan

Dazpoet:
Install stanza defined in another way than mentioned on KS since I can find absolutly no reason to put this inside the Squad Folder. Thought it might be looking for certain files in hardcoded places first but looking through the part.cfg files that does not appear to be the case so I'm installing this into its own folder in GameData.

@politas have been hard at work trying to find a homepage but have been unable too so none is listed for this mod.


plague006:
Quotes around GameData and spec_ver v1.4 jump out at me.

PR #1418, opened 2015-05-25
InterstellarFuelSwitch 1.5 contains a missing dll needed for the corr…

sswelm:
…ect operation of KSP

InterstellarFuelSwitch 1.5 contains a missing dll *Scale_Redist.dll) which needed for the correct operation of KSP people game will not function properly


Dazpoet:
Need to close the citation around InterstellarFuelSwitch-Core on line 6

PR #1401, opened 2015-05-24
Add F-18A Super Hornet from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of LuizChacon, to add [F-18A Super Hornet](https://kerbalstuff.com/mod/840/F-18A%20Super%20Hornet) to CKAN.

Please direct questions about this pull request to [LuizChacon](https://kerbalstuff.com/profile/LuizChacon).

PR #1419, opened 2015-05-25
Add the BD crafts from KSCSM

Dazpoet:
Both are tested and working locally on a KSP 0.90 install.

.craft file from F18 not installed since I cannot install to a subfolder of Ships as per the CKAN spec.

BDArmory not a distinct dependency for A6Intruder but it seems you can't do much without it so I've elected to have it as a dependency for it. Text on F18 page is not as outspoken about needing BDArmory which is why it "merely" recommends it.

GPL-3.0 chosen as license for F18 since afaik it's the most restrictive of the 3 listed ones (CC-BY-NC-4.0, GPL-3.0 and WTFPL)

### DIFF
--- a/NetKAN/A6Intruder.netkan
+++ b/NetKAN/A6Intruder.netkan
@@ -1,6 +1,6 @@
 {
     "$kref": "#/ckan/kerbalstuff/839",
-    "license": "GPL-3.0",
+    "license": [ "GPL-3.0", "WTFPL" ],
     "spec_version": 1,
     "identifier": "A6Intruder",
 	"depends"	:	[ { "name" : "AdjustableLandingGear" },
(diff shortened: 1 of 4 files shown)


Dazpoet:
See further license discussion in #1401

Dazpoet:
So here's an interesting issue, another one, with this. The latest commit requires BahamutoDynamicsPartsPack which we do not have indexed for the targetted ksp version (0.90).

Note: Check what version is bundled and consider marking it compatible with 0.90.

Dazpoet:
aargh! The mods don't bundle anything but rather just lists a dependency for the user to resolve on their own.

Same problem with licensing also persists as KS marks both as GPLv3 while they both include WTFPL licenses.

Dazpoet:
Contact establish with author [here](http://kscsm.freeforums.net/thread/20/question-licenses-dependencies-ckan-indexing) and it seems WTFPL is the correct license for everything, awaiting final confirmation.

Dazpoet:
Jenkins validates all files which is good enough for me.

There's still the issue of the unfullfillable requirement of BahamutoDynamicsPartsPack for ksp 0.90 though...

PR #1420, opened 2015-05-25
New version of #1355

Dazpoet:
Changed ksp_version to 0.90 since that is the latest official version of the game that this mod works on.


Dazpoet:
Builds and work locally on a 0.90 install

PR #1421, opened 2015-05-25
Update GCMonitor.netkan

Dazpoet:
Updated in accordance with the forum thread for this mod. Notifying @BasharMilesTeg who I think is the current maintainer for this?


Dazpoet:
This will require fixing of incorrectly generated GCMonitor .ckan files in the CKAN-meta repo

PR #1397, opened 2015-05-24
Add Corvus -1.25m Two Kerbal Pod from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of OrionKermin, to add [Corvus -1.25m Two Kerbal Pod](https://kerbalstuff.com/mod/837/Corvus%20-1.25m%20Two%20Kerbal%20Pod) to CKAN.

Please direct questions about this pull request to [OrionKermin](https://kerbalstuff.com/profile/OrionKermin).

Issue #1261, opened 2015-05-14
HGR/Home Grown Rockets


00Augustus00:
Could it be added?

plague006:
It's under an "All rights reserved" license at the moment, but the author seems interested in having the mod hosted (http://forum.kerbalspaceprogram.com/threads/60974-1-0-2-HGR-1-875m-parts%28v1-2-1-released%29?p=1952951&viewfull=1#post1952951). I've opened a line of dialogue, suggesting he host on KerbalStuff for easier CKAN maintenance and to attain permission to have the mod redistributed by CKAN.

Dazpoet:
as long as it's hosted on mediafire we can't do anything since we can't index from there.

plague006:
I'm drowsy.. Did I type mediafire somewhere instead of curse? He hosts on curse.

Dazpoet:
Oh my, I just checked the linked forumthread and saw a mediafire link, probably I'm not at my best either :)

plague006:
If/when needed, I whipped up the .ckan for v1.2.1: http://pastebin.com/JAqh7vX7

plague006:
Per: http://forum.kerbalspaceprogram.com/threads/60974-1-0-2-HGR-1-875m-parts%28v1-2-1-released%29?p=1960097&viewfull=1#post1960097

@OrionKermin: Thank you, I've given it a quick once over. I'll read over it in more detail over the weekend and try to get everything set up on Kerbalstuff at that time.

plague006:
Leaving this in OrionKermin's hands since he'll be posting to KS, removed the pastebin.

PR #1422, opened 2015-05-25
Fix HGR's Corvus

pacbard:
- Fixes errors with original netkan file
- Adds some metadata
- Adds dependency on HGR props
- Closes #1397

DO-NOT-MERGE until #1261 is resolved


Dazpoet:
You can remove line 5 and line 17 should be "depends" :)

Dazpoet:
#rebuild

PR #1433, opened 2015-05-27
Adding epoch to Ampyear

Dazpoet:
This should not be merged! It's me testing how this works and we should not merge until https://github.com/KSP-CKAN/CKAN-netkan/pull/60 is merged so we get a consistent naming schema.

resolves https://github.com/KSP-CKAN/CKAN/issues/931


Dazpoet:
#rebuild

PR #1434, opened 2015-05-27
Rename licences to license to match spec.

mgsdk:
I can't find a single .ckan file for this, so it has yet to be built since the inclusion, most likely due to the license/licenses issue.


pjf:
Noting KSP-CKAN/CKAN#966 as the relevant CKAN bug for this.

Dazpoet:
#rebuild

Dazpoet:
I think this should have spec_version v1.8?

PR #1437, opened 2015-05-28
Mk2Expansion.netkan

SuicidalInsanity:


Dazpoet:
CKAN does not have an "optional" switch in the install stanza. To get your desired functinality you'll need 2 .netkan files with the first one recommending or suggesting the example crafts.

PR #1444, opened 2015-05-29
gcmonitor is compatible with ksp 1.0.2

dss539:
Setting the ksp_version field to 1.0 causes the mod to be incompatible with ksp1.0.2.
Maybe ckan's version checking logic needs to be reworked to account for Squads post 1.0 versioning scheme?

### DIFF
--- a/NetKAN/GCMonitor.netkan
+++ b/NetKAN/GCMonitor.netkan
@@ -18,5 +18,5 @@
         "homepage"  : "http://forum.kerbalspaceprogram.com/threads/92907",
         "repository": "https://github.com/sarbian/GCMonitor/"
     },
-    "ksp_version": "1.0"
+    "ksp_version": "1.0.2"
 }


Dazpoet:
I'm utterly confused by this PR. GCMonitor shows up when using a KSP 1.0.2 install for me.

Setting `ksp_version´ to 1.0 makes it compatible with all 1.0.x versions of KSP in CKAN. What compatibility issues are you seeing, when do they occur and in which of the clients (commandline or GUI) ?

dss539:
The problem could very well be my own misunderstanding.
I am using the Windows GUI.
Maybe this screenshot will help.

http://i.imgur.com/K7JlMlW.png

When I filter for "compatible" I find that Graphic Memory Monitor is excluded from the list.

Did I do something wrong or have an incorrect expectation?

dss539:
well, this was my mistake due to not understanding the repository update mechanism. Issue was caused by my lack of clicking refresh! Ignore this erroneous pull request. Thanks for your help.

PR #1446, opened 2015-05-29
Add ModuleManager dependencies to AerojetKerbodyne and RealPlume

dbent:
This is adds the `ModuleManager` dependency to `AerojetKerbodyne` and `RealPlume` to compliment KSP-CKAN/CKAN-meta#554.


dbent:
And build failed because `AerojetKerbodyne` is only available for KSP 0.90.

PR #1452, opened 2015-05-30
Add Kerbal Launch Failure, Add Antares & Cygnus, Update Zero-Point Inline Fairings

plague006:


plague006:
Didn't realize I closed this (I suck at git)

Dazpoet:
Based on how 0point fairings distinctly doesn't work in Stock aero I think we should probably list it as a depends rather than a recommends. what does @NecroBones think about it?

NecroBones:
Yes, that would be fine with me. We're trying to come up with a solution for stock aero, but we can always remove the dependency later if that ever happens.

Dazpoet:
Took another look and it might be worth hardcoding the author field for AntaresCygnus since there are 2 authors rather than the one we scrape from KerbalStuff. The other author is Kartoffelkuchen according to the forum thread :)

Dazpoet:
#rebuild

plague006:
Damn it, didn't mean to push to that branch. Need to figure out how to fix/revert that.

Dazpoet:
Is https://github.com/plague006/NetKAN/commit/0ba3ea0b8b3522c003ca084f041d1276bd7457a7 identical to #1483 ? In that case I guess we can merge this and it wouldn't be a problem?

plague006:
Yes, I accidentally pushed that commit to this branch (when it was meant for #1485). I then reverted 1 commit too far back (which recreated the lack of quotes) and then added the quotes back.

So in the end this *isn't* touching epicsuits, #1485 did that, this is for Launch Fails, Antares, and Fairings.

PR #1457, opened 2015-05-31
Add PhoenixIndustries...netkan

Dazpoet:
I wonder if this works... if it does it's amazing.

So currently we have https://github.com/KSP-CKAN/CKAN-meta/pull/528 which added a small and large version of this mod. Both versions are now consolidated into a single package so we can catch it with a .netkan rather than handpackaging. Both old mods provide and conflict with this identifier.

I'm not sure if this will provide and upgradepath though or if old users will have to reinstall. Also names of the parts appears to have changed to accomodate for both being in the same package so the update is most probably craftbreaking.

### DIFF
--- a/NetKAN/PhoenixIndustriesCargoResupplySystem.netkan
+++ b/NetKAN/PhoenixIndustriesCargoResupplySystem.netkan
@@ -1,0 +1,12 @@
+{
+    "spec_version": 1,
+    "x_netkan_license_ok": true,
+    "identifier": "PhoenixIndustriesCargoResupplySystem",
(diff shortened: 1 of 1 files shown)


Murdabenne:
Im tossing in a pull request for the new .netkan file I wrote and tested for this.  IIm new to this so if I screw up please tell me how to do it right (Im sure you'll let me know what I did wrong)

Dazpoet:
The current mods are apperantly broken somehow so I don't see how adding this will make the situation worse in any way. Hence, I'm merging this.

PR #1466, opened 2015-06-01
Create CTN.nkan

Murdabenne:
Created a .nkan file to add Phoenix-Industries-Cargo-Resupply-System v1.5 and replace the old one wherever it is(?)

The nkan has been tested: it parses and generated proper .ckan file, which installs just fine via the GUI CKAN app.  Only question is the name - I named it after the directory in which it installs per the quick guide, but a more descriptive name might be better (PICRS?)  There is also another discussion/pull at https://github.com/KSP-CKAN/NetKAN/pull/1457 , but I am new so Im am unsure how to do this other than by the basic instructions.

### DIFF
--- a/NetKAN/CTN
+++ b/NetKAN/CTN
@@ -1,0 +1,17 @@
+
+{ 
+    "spec_version": 1, 
+    "x_netkan_license_ok": true, 
+    "identifier": "CTN", 
+    "license": "CC-BY-NC-SA", 
+    "release_status" : "stable", 
+    "$kref": "#/ckan/kerbalstuff/831", 
+    "recommends": [ 
+        { 
+            "name": "KIS" 
+        }, 
+        {
+            "name": "MechJeb2" 
+        }
+    ]
+}


Dazpoet:
If this was a new mod we where adding most of this would be correct. The file needs to have the extension .netkan though and since you have a hardcoded license on line 6 you could remove line 4 which just says we should use the license we scrape from KerbalStuff :) Very good for your, presumably, first .netkan file :+1: 

Something that I don't think is mentioned in the basic guide are filtering practices but this one would require some filtering since there are a few `Thumbs.db` entries in the installed data and we don't want those. To resolve this you'd need to have an install stanza like the one in #1457 since we cannot filter without one.

As for the naming schema you are correct that if this was a mod we added for the first time the identifier `CTN` is correct but alas this mod was already [added to CKAN-meta](https://github.com/KSP-CKAN/CKAN-meta/pull/528) as `PhoenixIndustriesCargoResupplySystem` along with a `-<size>` and providing/conflicting with the `PhoenixIndustriesCargoResupplySystem` identifier. What that means is that we'll have to keep that naming convention to make sure our new index maintains that conflict so users don't end up able to install multiple versions of the same mod, since that would cause install conflicts where CKAN attempts to write multiple files to the same place and throw a Kraken (exception error).

Since this is a duplicate of #1457 I'll close it now, having more than one PR for the same issue does not make it go through faster I'm afraid, although we do appreciate your effort in wanting to resolve this asap :)

plague006:
@Murdabenne if it makes you feel any better, @Dazpoet mentioned you could remove line 4, but in 4a8fe5b2854a718764205f21ff38ef0ae68a33b7 he left the line in himself (while having a license hardcoded). It doesn't hurt anything and you'll see it in quite a lot of .netkan files. Like Daz mentioned, if this were the mod's first submission your .netkan would have only needed some minor tweaks.

Lastly, because I'm curious how you came to write the metadata. I recently rewrote the "Adding a mod" wiki page and if you used it, I'd love to know from a first-timer if there was anything unclear that needs editing/rewording/added/removed.

Dazpoet:
Oh darnit! xD

PR #1469, opened 2015-06-01
Adding a forced v to SCANSat

Dazpoet:
https://github.com/KSP-CKAN/CKAN-meta/issues/563 pointed out that SCANSat 12.1 wasn't in CKAN and although the problem seems to originate from the [NetKAN-bot outage](https://github.com/KSP-CKAN/NetKAN-bot/issues/13) the removal of the v from the version is bound to cause grief down the road so I'm fixing it here, which should also force-index SCANSats latest version.


Dazpoet:
very minor change, merging this.

PR #1480, opened 2015-06-01
Add Mk3KISCargoContainers.netkan

Dazpoet:
closes #1478

### DIFF
--- a/NetKAN/Mk3KISCargoContainers.netkan
+++ b/NetKAN/Mk3KISCargoContainers.netkan
@@ -1,0 +1,9 @@
+{
+    "spec_version": "v1.4",
+    "license": "GPL-3.0",
+    "$kref": "#/ckan/kerbalstuff/875",
+    "identifier": "Mk3KISCargoContainers",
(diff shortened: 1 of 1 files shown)


Dazpoet:
I wonder if this might be better suited as just KISCargoContainers based on http://forum.kerbalspaceprogram.com/threads/123984-Mk3-KIS-Cargo-Containers?p=1989273&viewfull=1#post1989273 where Mk2 Containers appears to be in the works, it is however not stated if that will be in the same package or not.

plague006:
This is just being held by choice of identifier? Stick with what you have, nobody sees it anyways, and anybody who needs to work on the file can still search "KISCargoContainers" and find the file.

Dazpoet:
indeed it is held for that reason. I was hoping to see some spur of development in the thread. Also people who use commandline identify mods by identifer and not by name so having a misleading one could be problematic for them, hence my reluctance.

I don't see much progress in the forumthread though so I guess this should be merged...

Dazpoet:
#rebuild

PR #1481, opened 2015-06-01
Update KerboKatzSmallUtilities and add all the standalone versions

Dazpoet:
Taking inspiration from how AstronomersPack does things: I propose that now that KerboKatzSmallUtilities have every single utility available on its own over at KerbalStuff we have the old mod act as a placeholder installation which recommends the users to install all utilities so they can choose which ones they want and not.

This appears to match how the author make things available on [github](https://github.com/Xarun/SmallUtilities/releases) where users can download the whole thing or choose only those they want.

This is bound to cause grief for users who have this mod installed since flipping metadata around like this isn't handled well by the GUI, or atleast didn't use to be handled well by the GUI. Now that everything is updating at lightning speed it's hard keeping up :+1: 

Impact on commandline users is unknown since I seem to recall that client being able to solve this but I'm not sure it acctually works.


Dazpoet:
Hm so it now occurs to me that doing it this way makes the user able to download only `KerboKatzSmallUtilities` which will leave them with a .version file and nothing else, That might not be the best idea... Is it then perhaps better to just have the entire package left as it used to be and have the other ones as standalone modlets which all conflict with `KerboKatzSmallUtilities` ?

Dazpoet:
Due to the issue described above I reverted the changes to SmallUtilities and made it conflict with all its sub-parts to users can install those in any combination they feel like but the big one can never be installed with either of them.

PR #1484, opened 2015-06-02
HotPlume ckan initial commit

Felger:
Adding HotPlume netkan


plague006:
I *think* the way you want it installed to the player is GameData\MP_Nazari\HotPlumeConfigs\ right? In that case you want to change 
"file" : "GameData/MP_Nazari/HotPlumeConfigs",
to
"file" : "GameData/MP_Nazari/",
on line 12. As it's written it'll be "GameData/HotPlumeConfigs".

Felger:
You are absolutely right! Good catch!
I had changed that in my local copy, had not updated to the repo yet.

Felger:
This is ready to merge

Issue #1455, opened 2015-05-30
MiniAVC Policy


TeddyDD:
This little mod appears in plenty of mod archives. What we are supposed to do with it?
I was looking for precedence but some metadata filters it out, other install it explicitly 

https://github.com/KSP-CKAN/NetKAN/search?utf8=%E2%9C%93&q=MiniAVC

Dazpoet:
The .xml file should always be filtered. As for the .dll I am not sure how it functions. Could it even be bundled as a dependency or is it somehow bound to each mod?

NecroBones:
My understanding is that it's not "bound" as such. It's more like ModuleManager, in that it's used by a lot of mods, but not all of them bundle it. If multiple copies are installed, only the most recent copy of it will run.

TeddyDD:
So we should create package for dll and add it as dependency? Is this mod available on KS or as release on github?

NecroBones:
Looking at the forum thread: http://forum.kerbalspaceprogram.com/threads/79745

It looks like Mini AVC is a download on the mod author's website only. The main mod itself, "KSP AVC" is on curse and github, and is already in CKAN. I'm not sure what to recommend. Maybe the mod author should be contacted. It's possible he might recommend having the full mod listed as a suggested dependency or something.

pjf:
Ooooh! I was hoping we'd have this discussion.

Users *should* be able to choose whether or not they wish to have MiniAVC or not. At the moment lots of folks end up with it anyway because of the way it's bundled with mods and we're aren't already filtering it out.

So, the ideal situation would be:

- Our test processes generate a warning if they spot `MiniAVC.dll` being installed by a mod.
- Mods which bundle `MiniAVC.dll` recommend it instead. Users doing a default install will still get it, just like they always have.
- The user gets to make the choice if they want it or not.

This means folks who like MiniAVC can have it, and those who don't can choose not to have it, and everyone is happy. It also means that users are able to keep their MiniAVC library up to date (in much the same way that we do with ModuleManager).

Dazpoet:
Gonna be a real joy finding and fixing every mod that contains MiniAVC but I approve strongly of adding it as a recommend rather than a depend so I can finally have it not appear in my installs

TeddyDD:
I think we should ask MiniAVC author to create github releases for new MiniAVC so we could create .netkan for it. (I suppose we can't use http://ksp.cybutek.net/miniavc/ to create .netkan)

[...]

PR #1487, opened 2015-06-02
Add KSI - Placement Services

plague006:
Closes #1471
2 licenses in the archive: MIT and GPL3. GPL3 more restrictive so it's the one hardcoded. Abstract also hardcoded because


Dazpoet:
This mod includes a MiniAVC.dll but until a consensus on how to handle that is reached in #1455 I'll merge this.

PR #1493, opened 2015-06-03
Tarsier isn't updating, nudging it

Dazpoet:
For https://github.com/KSP-CKAN/CKAN/issues/1066

Seeing as I don't know why it refuses to update I'll just do the old install stanza juggle and see if that nudges it into working.


Dazpoet:
That seems to have solved whatever wasn't working, I'm thinking this is because of gnolls in the upstream and recommend we add anti-gnoll-spray to the standard equipment of NetKAN-bot

PR #1500, opened 2015-06-04
Update CryoEngines.netkan

Dazpoet:
supercedes and closes #1418


Dazpoet:
New version is about, this just resolves the crushing issues until it hits.

PR #1502, opened 2015-06-04
Create TMSTACLSRetexture.netkan

tambo1:


Dazpoet:
#rebuild to see if error persists

tambo1:
Huh?  Sorry, this process is a little perplexing to me.

plague006:
Tested manually with netkan.exe and it spits out a .ckan without issue. It actually looks like an issue with TACLS?

>The following inconsistencies were found:
TACLS requires TACLS-Config but nothing provides it

Probably related to how TACLS-Config is provided?

Someone smarter than me will forward this to the people that need to know about it, but you're in the clear @tambo1. =)

Dazpoet:
But does it install correctly, that's the part Jenkins is stuck on. Does this overwrite any files or is it all done with MM?

tambo1:
Everything is done via MM.  No overwrites.

I'm wondering if the problem is that TACLS has TACLS-Config as a dependency, but I only state that TACLS as a dependency?

Dazpoet:
Nah CKAN (should) handle that without problems, this seems to be an error somewhere on our part.

I just ran everything locally and it installed fine so I'm notifying @RichardLake who might be able to spread some knowledge around.

Dazpoet:
Seems Jenkins is fetching an old (7 days, so about 4 years in CKAN development cycle time) so I'm merging this since it works locally and presents no errors on newest CKAN.

tambo1:
Many thanks for your time and consideration, gents.

PR #1509, opened 2015-06-05
Add Heat Control from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of Nertea, to add [Heat Control](https://kerbalstuff.com/mod/890/Heat%20Control) to CKAN.

Please direct questions about this pull request to [Nertea](https://kerbalstuff.com/profile/Nertea).

PR #1511, opened 2015-06-05
Add HeatControl-Core.netkan

Dazpoet:
Needed for NearFutureElectrical as mentioned in #1509


Dazpoet:
NF guys make their own

PR #1513, opened 2015-06-05
Fix and add NFP Extras

Dazpoet:
depends on #1512 and cleans up after #1507


dbent:
Just a note while we're at it, can probably drop the `Extras` from the identifiers, the fact that they're `NearFutureSomething-SomethingMore`, should make it clear they're additions on top of the base mod, and saves some characters for us console users.

Dazpoet:
I agree and added it for consistency but it does look way overkill. Closing since NF want their own netkans which come later

PR #1515, opened 2015-06-05
Use explicit asset_match for StockBugFixModules

dbent:
Now that StockBugFixModules distributes two zip files in its releases, let's be explicit about which one to use.

Also, hardcode `repository` due to KSP-CKAN/CKAN#1075.


dbent:
Related: KSP-CKAN/NetKAN#1514

PR #1516, opened 2015-06-05
Bump epoch of RandSCapsuledyne

Dazpoet:
CKAN thinks that the 1.5.01 version of this mod is higher than 1.5.1, or atleast it reports the incorrect version as newest.

`ckan compare 1.5.1 1.5.01` and `ckan compare 1.5.01 1.5.1` both give `-1` as output so something if afoot, but this should solve it and stop causing grief to the author.


dbent:
LGTM :+1:

PR #1517, opened 2015-06-06
No longer index RaiderNick's mods.

pjf:
Raidernick has contacted me privately and said he's no longer going to be part of the KSP modding community. His mods are all creative commons licensed, so we can't unindex the existing releases without breaking promises to our users, and
[those same licenses mean we have explicit and irrevocable permission to use them](https://www.reddit.com/r/KerbalSpaceProgram/comments/368z38/developing_the_ckan_may_become_my_parttime_job/crcrixa).

RN has indicated that future releases of his mods will be intentionally
breaking to KSP, so while we shouldn't remove the existing releases, we
*should* stop indexing new ones.

This change removes RN's mods from the auto-indexer.


pjf:
Using my executive hat to merge this.

Dazpoet:
What about AntaresCygnus? I see that forumthread is also closed so I'm assuming we might want to purge that aswell?

pjf:
@Dazpoet: Nice spot. Although for the record we're not purging the existing mods. We're preventing indexing of new releases when the mod author has promised new releases will be actively hostile to the user-base.

Dazpoet:
I blame to much blood in the coffestream for that one

PR #1519, opened 2015-06-06
Update kopernicus, now with 100% more github scraping

Dazpoet:
I'd love @dbent to have a look here but I think it's pretty uncontroversial.

No-one ever used the Kopernicus-Config-Default because Kopernicus does things by patching rather than by acctually changing the System.cfg file, making Kopernicus-Config a moot thing to keep around.

This PR leaves the old system behind and confines it to 0.90 while boldly looking forward to a new and bright future!

The filter doesn't acctually catch anything at this time but if a time arrives when Kopernicus redistributes files in the `Cache` folder we will want to filter them since that is where modauthors using Kopernicus stash their PQS things.

An install stanza for a mod which requires Kopernicus will probably have an install stanza looking something like this

```
"install" : [ { "find" : "AwesomeMod" , "install_to" : "GameData" },
{ "find" : "Cache", "install_to" : "GameData/Kopernicus" } ]
```

At this time I've elected not to make a .netkan for the KopernicusExamples, since I focus on preparing for RSS, but that should probably be done later since I don't think we should distribute half a mod (or rather a smidgeon if going by size).


dbent:
LGTM :+1:

When a new version of `OuterPlanetsMod` is released, I'll make sure it can use the released version of `Kopernicus` and modify it to no longer provide (and conflict) with `Kopernicus`.

PR #1521, opened 2015-06-06
Update RealSolarSystem.netkan

Dazpoet:
:warning: DO NOT MERGE :warning: 

This should not be merged until RSS10 is out! Else we'll retroactively overwrite the latest 0.90 RSS version over in CKAN-meta.

Accomodating the changes seen in the RSS pre-release found [here](http://forum.kerbalspaceprogram.com/threads/55145-0-90-Real-Solar-System-v8-6-1-Mar-30?p=1999650&viewfull=1#post1999650).

ModuleManager removed as a depdency since Kopernicus depends on it and therefore RSS implicitly does when depending on Kopernicus. Not sure if this is the best practice though.

Switched to the new RSS repository over at KSP-RO.

Not sure what to do with the [`Custombiomes-Data-RSS`](https://github.com/KSP-CKAN/CKAN-meta/tree/master/CustomBiomes-Data-RSS) and [`RemoteTech-Config-RSS`](https://github.com/KSP-CKAN/CKAN-meta/tree/master/RemoteTech-Config-RSS). Currently the ones in CKAN-meta are marked for `"ksp_version" : "any"` but I doubt very much that they work for this release of RSS? Are they even supposed to still be in this netkan? @NathanKell can probably answer that.

It seems we index `CustomBiomes-Data-RSS` directly from the older RSS releases but at least the pre-release for RSS10 no longer has it. `RemoteTech-Config-RSS` seems to have been handpackaged and derived from RO.

### DIFF
--- a/NetKAN/RealSolarSystem.netkan
+++ b/NetKAN/RealSolarSystem.netkan
@@ -12,7 +12,6 @@
         { "name" : "RSSTextures" }
     ],
     "recommends" : [
-        { "name" : "CustomBiomes-Data-RSS" },
         { "name" : "RealismOverhaul" },
         { "name" : "Toolbar" }
     ],
--- a/NetKAN/RealSolarSystem.netkan
+++ b/NetKAN/RealSolarSystem.netkan
@@ -1,6 +1,6 @@
 {
-    "spec_version"   : 1,
-    "$kref"          : "#/ckan/github/NathanKell/RealSolarSystem",
+    "spec_version"   : "v1.4",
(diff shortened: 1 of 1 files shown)


Dazpoet:
depends on #1519 and #1522 aswell as a resolution to the Custombiomes and RemoteTech stuff in the OP

NathanKell:
Kopernicus replaces custom biomes. The RT config is still valid though.

techman83:
#rebuild

PR #1522, opened 2015-06-06
Add rss textures

Dazpoet:
These have previously been manually indexed but are now up for grabs here on GitHub \o/

Having never used asset_match before I'm not sure these are correct in all ways but "Works on my machine (TM)"

While making these I noticed the ones over in CKAN-meta are rather malformed and also have `"ksp_version" : "any"` so I'll head over there and slap a _max onto them and confine them to 0.90 since these new ones take over from there.

There are no distinct DDS texturepacks anymore since KSP is all about DDS now, I've therefore elected to use the identifiers `RSSTextures*` rather than `RSSTexturesDDS*` and hope that is acceptable for all parties.

### DIFF
--- a/NetKAN/RSSTextures2048.netkan
+++ b/NetKAN/RSSTextures2048.netkan
@@ -14,6 +14,6 @@
     ],
 	
 	"install"		:	[ 
-		{ "find" : "RSS-Textures"
+		{ "find" : "RSS-Textures",
 		"install_to" : "GameData" } ]
 }
--- a/NetKAN/RSSTextures2048.netkan
+++ b/NetKAN/RSSTextures2048.netkan
@@ -1,0 +1,19 @@
+{
+	"spec_version"	:	"v1.4",
+	"license"		:	"CC-BY-NC-SA",
+	"name"			:	"Real Solar System Textures - 2048 x 1024",
+	"identifier"	:	"RSSTextures2048",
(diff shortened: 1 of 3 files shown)


Dazpoet:
Ran the 2048 one with `--prerelease` set locally and it worked, assuming the other ones will aswell but I don't want to download a gazillion mb of data right now to prove it.

pjf:
As far as I know these are for the "new" RSS which is based upon Kopernicus. I suspect users of the old RSS almost certainly want our existing textures if they want things to work correctly.

CC'ing in @NathanKell as the actual mastermind here, I just wrote the build system for RSS-Textures. :)

Dazpoet:
Yes you're totally correct :)

Hence why I've opened https://github.com/KSP-CKAN/CKAN-meta/pull/580 to have these and the old textures seperated by ksp_version, these won't get offered to anyone running 0.90 (where old RSS resides) and the old ones won't be offered to anyone running 1.0 (where the new RSS #1521 resides). 

Afaik that's the best I can do to ensure users don't get the wrong ones until https://github.com/KSP-CKAN/CKAN/pull/989 is merged which I've gathered will make it possible to set a min_version of RSS that will acctually be non-cosmetic.

pjf:
@Dazpoet : You rock. :)

PR #1525, opened 2015-06-06
FreedomTex works with 1.0.2

TeddyDD:
Yet another .netkan hack to mark that mod works with new version of ksp :+1:


Dazpoet:
I'm going to be that guy. Could you also filter the .DS_Store file?

TeddyDD:
Here you are :)

PR #1529, opened 2015-06-07
Update and rename NearFuturePropulsionExtras.frozen to NearFutureProp…

Felger:
…ulsion-HydrogenNTRs.netkan

Updated directory name per updated pack.


Felger:
Fixed name-identifier mismatch.

Felger:
Not sure why it's still failing the build, only error shown is "spec_version must be v1.2+", which is already set...

Dazpoet:
superceded by #1540 (I think, might be one of the others)

It failed because you need to put citations around the spec_version when it's != 1

PR #1530, opened 2015-06-07
Create NearFuturePropulsion-LowThrustEP.netkan

Felger:
Added new netkan for second "Extras" folder in NFP


Dazpoet:
superceded by #1541 which is basiclly a copy of this one :)

PR #1531, opened 2015-06-07
build.sh: Download CKAN/Netkan artefacts from S3.

pjf:
Now that travis is uploading these artefacts to S3, we should download
them from there as well. :)

Part of my quest to make @Dazpoet happy. :)


techman83:
LGTM.

PR #1489, opened 2015-06-02
Add FuelSwitch to every tank from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of micha030201, to add [FuelSwitch to every tank](https://kerbalstuff.com/mod/861/FuelSwitch%20to%20every%20tank) to CKAN.

Please direct questions about this pull request to [micha030201](https://kerbalstuff.com/profile/micha030201).

PR #1532, opened 2015-06-07
Add FuelSwitch to every tank

plague006:
Part 2/2 to resolve #1489.

See  KSP-CKAN/CKAN-meta#582 for part 1.


Dazpoet:
Since they both scrape from the same source you'll have to hardcode the names in so make sure we don't build 2 identically named mods from different .netkans

Dazpoet:
#rebuild

PR #1309, opened 2015-05-18
Add Sigma Mod Expansions from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of Sigma88, to add [Sigma Mod Expansions](https://kerbalstuff.com/mod/816/Sigma%20Mod%20Expansions) to CKAN.

Please direct questions about this pull request to [Sigma88](https://kerbalstuff.com/profile/Sigma88).

PR #1533, opened 2015-06-07
Adding Sigma Mod Expansion

plague006:
My attempt to resolve #1309.

@Sigma88 As your archive currently exists, your flags can't be included. CKAN can't handle a .zip inside a .zip. If you put them in a folder, we'd be happy to adjust the install to include them.

The only conflict between Outer Planet Mod (OPM) and Sigma Mod Expansion (SME) is GameData/Kopernicus/Cache/Wal.bin. So I have both mods filter it from their install.

I don't think there's a way around splitting OPM into core and the planet, but I see 2 ways to solve the issue for SME. SME could "provide" "Wal" or I could split SME into its core and the planet. I chose the latter for better readability when people install OPM and need to choose which mod provides "Wal". (It would look weird to choose between a full mod and a mod that only includes a planet)

I have SME explicitly depending on SME-Wal so that it highlights the conflict if OPM is already installed (since OPM-Wal provides "Wal"). The drawback is that if someone wants OPM's Wal with SME installed, that can't be accommodated. To help with that/obscure that, the order of the depends explicitly has SME-Wal first (if the order is reversed, the client will try to fulfill OPM's dependency on Wal first and thus give the user the option).

Paging @politas because you wanted to see how the issue might be solved.

### DIFF
--- a/NetKAN/OuterPlanetsMod.netkan
+++ b/NetKAN/OuterPlanetsMod.netkan
@@ -5,7 +5,8 @@
   "license": "CC-BY-NC-SA-4.0",
   "depends": [
     { "name" : "ModuleManager" },
-    { "name" : "Kopernicus" }
+    { "name" : "Kopernicus" },
+    { "name" : "Wal" }
   ],
   "conflicts": [
     { "name": "CustomBiomes" },
@@ -39,7 +40,7 @@
   ],
   "install" : [
     { "find": "OPM", "install_to" : "GameData" },
-    { "find": "Kopernicus", "install_to": "GameData" },
+    { "find": "Kopernicus", "install_to": "GameData", "filter": "Wal.bin" },
     { "find": "TextureReplacer", "install_to" : "GameData" }
   ]
 }
--- a/NetKAN/SigmaModExpansions.netkan
+++ b/NetKAN/SigmaModExpansions.netkan
@@ -1,0 +1,25 @@
+{
+    "spec_version": 1,
+    "identifier": "SigmaModExpansions",
+    "$kref": "#/ckan/kerbalstuff/816",
+    "$vref" : "#/ckan/ksp-avc",
+    "license": "CC-BY-NC-SA-4.0",
+    
+    "depends": [
+        { "name" : "OuterPlanetsMod" },
+        { "name" : "Wal-SME" }
+    ],
+    
+    "install": [
+        {
+            "file"       : "GameData/Kopernicus",
+            "install_to" : "GameData",
+            "filter"    :  "Wal.bin"
+        },
+        {
+            "file"       : "GameData/Sigma",
+            "install_to" : "GameData"
+        }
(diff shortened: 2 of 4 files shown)


Dazpoet:
notifying @dbent who is the resident OPM master (aka wrote the latest .netkan for it) because this is mostly going over my head.

dbent:
Just a note that while this is good way to go about it, BryceSchroeder/Kopernicus#43 would make it unnecessary. I'll see if I can get it merged in soon.

plague006:
This fix is superseded by Kopernicus 43 (link above), allowing for a more elegant solution.

Sigma88:
in the next update of the mod I will move the flags inside my mod folder, it was separate for now because they were still a test.

regarding wal, I will keep it in the same folder as OPM as long as they keep using a kopernicus version that does not read .bin files from other folders.

the point of having a different wal is a result of moving that moon to a 90° tilded orbit (the moon has an equatorial ridge that needs to be "rotated" when using the 90° tilt) so there are no reason to want to use OPM + SME with OPM original wal.bin

hope this helped

dbent:
@Sigma88 I assume you know but OPM 1.7 (non-beta) now uses a `cacheFile` property for all its `.bin` files. Once you make the changes on your end it should be easy to get this and #1605 in.

Sigma88:
@dbent yes, I was ready for the release compatible with OPM 1.7 and kopernicus 0.0.9 but bumped into a problem with one of my configs. I should be able to release the next patch in the next few days (maybe even today) and that should be very easy to make CKAN compatible

thanks for taking the time to help me :)

Sigma88:
The current release should work fine with CKAN

let me know if there are any problems

Sigma88:
I am doing some work on rearranging the mods included in my "ModExpansions"

They will be split into different mods and moved to github

I totally expect them to not work in CKAN untill I'm done with the "movement"

I can contact you here when I have finished moving the mods, if you are interested in keeping them on CKAN

plague006:
Always interested in keeping mods on CKAN.

Sigma88:
So,
the 3 mods that were originally part of "Sigma OPM Expansion" are now released separately

these are the links to the forum threads and github repositories

**Sigma: OPM Recolor**
http://forum.kerbalspaceprogram.com/threads/135895
https://github.com/Sigma88/Sigma-OPMRecolor

**Sigma: OPM Tilt**
http://forum.kerbalspaceprogram.com/threads/135899
https://github.com/Sigma88/Sigma-OPMTilt

**Sigma: PluronKhato**
http://forum.kerbalspaceprogram.com/threads/135881
https://github.com/Sigma88/Sigma-PluronKhato

the last one is now a standalone mod, it doesn't require OPM.

plague006:
I've created issue #2382 to keep track of this. Without a separate issue there's a near-zero chance of me remembering.

PR #1542, opened 2015-06-07
Update OPM to not break GUI client

dbent:


Dazpoet:
Relevant issue https://github.com/KSP-CKAN/CKAN/issues/1083

PR #1546, opened 2015-06-07
Adding KashCorpSA

Murdabenne:
Shared Assets (virtual mod) for KAP and RoversAndRoadsters mods.

### DIFF
--- a/NetKAN/KashCorpSA.nkan
+++ b/NetKAN/KashCorpSA.nkan
@@ -1,0 +1,20 @@
+{
+        "spec_version" : "v1.4",
+        "identifier" : "KashcorpSA",
+        "name" : "Kash Corp Shared Assets (Virtual mod)",
+        "abstract" : "Shared assets for KashCorp mods, based off KAP",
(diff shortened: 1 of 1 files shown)


Dazpoet:
- File extension is .netkan
- You need to remove the comma on line 11
- What is the reasoning behind the hardcoded `"version"` field rather than using the scraped one?

Murdabenne:
scraped version differs depending on which one of the 2 sources you look at - this is shared between the 2 follow-on mods Im adding afterward.  So I decided to version the psud-mod instead, to be able to point it at whichver is more up to date as the base.  I rechecked the build with netkan on my Win8.1 box and it accepted it with the extra comma, no error reported.  Perhaps a newer version of netkan is needed?   I have deleted the comma, renamed the file, and will redo the pull.

Dazpoet:
netkan.exe will accept a lot of things, such as .txt files but the buildbot runs it through a json validator. You can do your own validation at e.g. www.jsonlint.com to see if your file looks ok.

Rather than opening multiple PRs you can instead push changes to this one :)

PR #1547, opened 2015-06-07
Adding KashCorp Virtual mod

Murdabenne:
pseudo-mod created for RoversAndRoadsters and KAP apps which share these resources and base directory.

### DIFF
--- a/NetKAN/KashCorpSA.netkan
+++ b/NetKAN/KashCorpSA.netkan
@@ -1,0 +1,20 @@
+{
+        "spec_version" : "v1.4",
+        "identifier" : "KashCorpSA",
+        "name" : "KashCorp Shared Assets (Virtual mod)",
+        "abstract" : "Shared assets for KashCorp mods, based off KAP",
+        "comment" : "KashCorp Shared Assets based off KAP zip file",
+        "license" : "BSD-3-clause",
+        "version" : "1.0v",
+        "$kref" : "#/ckan/kerbalstuff/96",
+        "depends" : [
+                { "name"        : "ModuleManager" }
+        ],
+        "install" : [
+                {
+                        "find" : "KashCorp",
+                        "install_to" : "GameData",
+                        "filter" : "KAP"
+                }
+        ]
+}
--- a/NetKAN/KashCorpSA.netkan
+++ b/NetKAN/KashCorpSA.netkan
@@ -9,7 +9,7 @@
 		"resources": 
 		{ 
 			"homepage": "http://forum.kerbalspaceprogram.com/threads/87546" 
-		}
+		},
         "$kref" : "#/ckan/kerbalstuff/96",
         "depends" : [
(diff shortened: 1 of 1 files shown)


NetKAN/KashCorpSA.netkan
```diff
@@ -0,0 +1,20 @@
+{
+        "spec_version" : "v1.4",
+        "identifier" : "KashCorpSA",
+        "name" : "KashCorp Shared Assets (Virtual mod)",
```
Dazpoet: You don't need to specify a mod as virtual, a not very small part of mods in the repository are in one way or another. This distinction could be confusing to users.

NetKAN/KashCorpSA.netkan
```diff
@@ -0,0 +1,20 @@
...
+        "abstract" : "Shared assets for KashCorp mods, based off KAP",
```
Dazpoet: Since there's a comment below you can remove the ", based of KAP" part here.

NetKAN/KashCorpSA.netkan
```diff
@@ -0,0 +1,20 @@
...
+        "version" : "1.0v",
+        "$kref" : "#/ckan/kerbalstuff/96",
+        "depends" : [
+                { "name"        : "ModuleManager" }
+        ],
+        "install" : [
+                {
+                        "find" : "KashCorp",
+                        "install_to" : "GameData",
+                        "filter" : "KAP"
```
Dazpoet: Are the resources supposed to be installed by this mod or are they KAP specific?

plague006:
I know you already have a bunch of changes to make, but if you could hardcode the homepage that would be great. The one on the KerbalStuff page links to the dev thread rather than the release thread (and there's actually nothing in the dev thread indicating the mod has released).
So something like:

    "resources": {
            "homepage": "http://forum.kerbalspaceprogram.com/threads/87546"
        }

NetKAN/KashCorpSA.netkan
Murdabenne: These are the shared ones.  The KAP speciofic ones are in the KAP netkan, which is in a different fork.  Im waiting until the shared one goes in before I put the PR in for this one, because I cant test my netkans against it until its there for DCKAN to find as a "depends".  I already have stand alone netkans that work individually, but they have a problem with the shared files.

Murdabenne:
hardcoded the homepage per request.  This is the resource-only install.  KAP will install its specific stuff, as weil RoversAndRoadsters doing the same

NetKAN/KashCorpSA.netkan
Murdabenne: ok

Murdabenne:
Making edits to clarify metadata as per suggestions above.

NetKAN/KashCorpSA.netkan
Murdabenne: Done.

PR #1548, opened 2015-06-08
Update KAINET/NANA.

pjf:
Based upon feedback at
https://www.reddit.com/r/KerbalSpaceProgram/comments/38znqb/ckan_integration_to_kerbalx_and_ksp_mod_manager/crzdhr5

### DIFF
--- a/NetKAN/KAINET.netkan
+++ b/NetKAN/KAINET.netkan
@@ -2,15 +2,16 @@
     "spec_version" : "v1.4",
     "$kref"        : "#/ckan/kerbalstuff/592",
     "identifier"   : "KAINET",
+    "name"         : "No Astronauts Need Apply",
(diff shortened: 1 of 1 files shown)


Dazpoet:
When the new one is released on KS I'm guessing we'll have to keep the identifier to avoid confusion of our systems, or implement a solution like the ones discussed in https://github.com/KSP-CKAN/CKAN/issues/904 which would solve these situations :D

Kerbas-ad-astra:
So "near future" was a bit of a mis-estimate, but it's done.  New KS link: https://kerbalstuff.com/mod/1048/No%20Astronauts%20Need%20Apply

For future reference, I'll always publish to GitHub first (since that's where the source lives), and that is more likely to remain stable.

PR #1557, opened 2015-06-09
Add StockNoContracts

plague006:
.version file rejected by netkan.exe


dbent:
Since the author specifically specifies, can you add a `min_version` of `2.6.0` to the `ModuleManager` dependency?

malahx:
Hello, I've only test It with ModuleManager 2.6.X :)
Now the "install" section is not need?

dbent:
> Now the "install" section is not need?

By default CKAN will search for a `<identifier>` or `GameData/<identifier>` directory and automatically install that. Since your archive satisfies that criteria no `install` stanza is necessary.

PR #1579, opened 2015-06-11
Update RealFuels.netkan

Dazpoet:
Just saw it mentioned on irc that RealFuels collides with the various fuelswitches out there. I think I got all of them. RF does not collide with `FirespitterCore`, only with stuff adding FSFuelswitch to various things. Also I'm sure _something_ in `RealismOverhaul` somehow need `FirespitterCore` so conflicting with it seems unwise.

### DIFF
--- a/NetKAN/RealFuels.netkan
+++ b/NetKAN/RealFuels.netkan
@@ -26,7 +26,6 @@
         { "name" : "ModularFuelSystem" },
         { "name" : "ModularFuelTanks"  },
         { "name" : "InterstellarFuelSwitch" },
-        { "name" : "InterstellarFuelSwitch-Core" },
         { "name" : "Fuelswitchtoeverytank" },
         { "name" : "StockFuelSwitch" }
     ],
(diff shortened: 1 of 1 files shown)


Dazpoet:
The conflict with `InterstellarFuelSwitch-Core` is probably not the wisest either. `RealismOverhaul` depends on `RealFuels` but also recommends `NearFuturePropulsion` which in turn depends on `InterstellarFuelSwitch-Core` so adding it as a conflict will break RO.

I'm not sure what the prefered thing to do here is. Can this perhaps be circumvented totally by smart MM-patches in RF and we shouldn't care about it on this end?

Seems @NathanKell is our listed author for RF so let's ask there :)

NathanKell:
@Dazpoet an excellent point, yes, RF can nuke fuelswitch on FINAL

Dazpoet:
So I close this and RF solves it? Or I guess we might want to list conflicts as well since RF nuking certain things might leave users confused as to why their latest mod (e.g. StockFuelSwitch) isn't doing what they expected it to and then they go over to the forum thread presenting an issue which the author will have a very hard time fixing.

But say I remove `InterstellarFuelSwitch-Core` from the list of conflicts that should be an ideal middleroad where we only generate a minimal amount of non-working mods?

NathanKell:
Seems reasonable.

Dazpoet:
I'm merging this then, thank you for the feedback NK :)

PR #1585, opened 2015-06-11
HeatControl adjustments

pjf:
Includes #1569, but without the rename.

Submitting as a PR as I'm too lazy to test manually when we've got bots. :)


pjf:
Hooray! Merging as this is really #1569, which I've reviewed. :)

PR #1586, opened 2015-06-11
NearFutureElectrical with ident adjustments.

pjf:
Includes #1573, but without the renames.

Sending as a PR so our bots can test. :)


pjf:
Merging with joy, as this is #1573 with a tweak, and I've reviewed #1573. :)

PR #1587, opened 2015-06-11
Update ContractConfigurator-GrandTours.netkan

Dazpoet:
Adding an epoch because the versioning broke when the author put the former version up on KS as 1.0.6 rather than 0.1.6 making the newest version 0.1.7 not show up as the latest one in CKAN.


dbent:
LGTM :+1:

PR #1594, opened 2015-06-11
Create SDHI-SharedAssets.netkan

sumghai:
- Incremented ksp_version to 1.0
 - Drop NodeResizer as suggested add-on (obsolete)
 - Tell CKAN to ignore Agencies folder bundled in GitHub download, and instead install it from SDHI-SharedAssets
    - This allows CKAN users to have multiple SDHI add-ons without each add-on fighting to overwrite each other's copies of the Agencies folder; users manually installing from the GitHub download will continue to use the bundled Agencies folder


sumghai:
Broken PR, please ignore.

PR #1640, opened 2015-06-21
KScale2.netkan

Dazpoet:
there was a CKAN folder inside the .zip for this and lo and behold it held a netkan file. So I grabbed it and added it to here. The identifier has therefore been chosen by the author and not by me.


Dazpoet:
Had to fix a few commas so I also trimmed down the identifier and filename to match the common name on KS and in forums.

PR #1641, opened 2015-06-21
ContractConfigurator-HistoricMissions.netkan

Dazpoet:
Hoping I'm allowed to install these .craft files to `Ships` based on https://github.com/KSP-CKAN/CKAN/issues/1134


Dazpoet:
Unsure how I'll get the stuff from `StockCraft` into the `Ships/VAB` folder and until that is solved this shouldn't be merged.

plague006:
The `StockCraft` were never meant to be installed to `Ships`. ContractPacks have .craft files inside the `GameData` folder to create ships inside the contracts themselves (eg. a specific craft to rescue from orbit).

Thanks to @Postremus for pointing that out to me.

PR #1642, opened 2015-06-21
First release for EVA Handrails Pack Continued

linuxgurugamer:
Released on Friday

### DIFF
--- a/NetKAN/EVAHandrailsPackContinued.netkan
+++ b/NetKAN/EVAHandrailsPackContinued.netkan
@@ -5,6 +5,7 @@
     "$kref"          : "#/ckan/kerbalstuff/929",
     "x_netkan_license_ok" : true,
     "release_status" : "stable",
+    "license": "CC-BY-SA-3.0",
     "depends" : [
         { "name": "ModuleManager" }
     ],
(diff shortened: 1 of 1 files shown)


dbent:
Sorry for not getting to this earlier. Couple of things:

- If the original mod has been abandoned for good and you're taking over active maintership we can just update the existing [`.netkan`](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/NebulaEVAHandrails.netkan). That way anyone with the original mod will get your version on upgrade.
- If you *don't* want to replace the original, then a `conflicts` relationship should be added.
- The `ModuleManager` assembly should be filtered, although you may just want to change the packaging since `ModuleManager` assemblies are usually distributed directly under `GameData` not within mod-specific subdirectories. The filter clause would look like:

  ```json
    "install": [
        {
            "file"       : "NEBULA",
            "install_to" : "GameData",
            "filter_regexp": "ModuleManager.+\\.dll"
        }
    ]
  ```

linuxgurugamer:
Hi,

I would rather NOT replace the old one, just in case he comes back.

I created one already, EVAHandrailsContinued.netkan, can you update it 
for me?

Thanks


JBB

dbent:
Implemented in #1645.

linuxgurugamer:
Thank you


JBB

PR #1643, opened 2015-06-21
Create NMSG-1.3.netkan

maculator:


Dazpoet:
Your .netkan file should not have, what I presume is, the version of the mod in it. You'll need to name it NMSG.netkan :)

PR #1644, opened 2015-06-21
Create NMSG.netkan

maculator:


pjf:
Confirmed with @maculator on IRC that the installed files and bot-indexed details all work out, so merging this with thanks! :)

PR #1638, opened 2015-06-20
Add Pathfinder from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of Angel125, to add [Pathfinder](https://kerbalstuff.com/mod/933/Pathfinder) to CKAN.

Please direct questions about this pull request to [Angel125](https://kerbalstuff.com/profile/Angel125).

PR #1637, opened 2015-06-20
Add Deep Space Exploration Vessels from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of Angel125, to add [Deep Space Exploration Vessels](https://kerbalstuff.com/mod/932/Deep%20Space%20Exploration%20Vessels) to CKAN.

Please direct questions about this pull request to [Angel125](https://kerbalstuff.com/profile/Angel125).

PR #1647, opened 2015-06-22
Added Pathfinder.netkan at authors request

linuxgurugamer:
This replaces the file from kerbalstuff.com

### DIFF
--- a/NetKAN/Pathfinder.netkan
+++ b/NetKAN/Pathfinder.netkan
@@ -1,0 +1,20 @@
+{
+	"spec_version" : "v1.4",
+	"identifier" : "Pathfinder",
+	"$kref" : "#/ckan/kerbalstuff/933",
+	"$vref" : "#/ckan/ksp-avc",
+	"release_status" : "stable",
+	"x_netkan_license_ok": true,
+	"depends" : 
+	[
+		{ "name": "CommunityResourcePack" },
+		{ "name": "ModuleManager" }
+	],
+	"install" : 
+	[
+		{
+			"find" : "WildBlueIndustries",
+			"install_to" : "GameData/WildBlueIndustries"
+		}
+	]
+}


Dazpoet:
What is your solution to https://github.com/KSP-CKAN/NetKAN/pull/1637#issuecomment-113871255 ?

linuxgurugamer:
I'm not sure it's my place to say, since I only helped on #1638, but 
here is my suggestion:

Since Netkan itself suggested the following:

#   Failed test 'DeepSpaceExplorationVessels license (CC-BY-NC SA 4.0) should match spec. Set `x_netkan_license_ok` to supress.'
#   at t/metadata.t line 44.
# Looks like you failed 1 test of 3951.

I would take the approach that, if there is a license, even if not 
recognized, to try recompiling it with the x_netkan_license_ok option 
set to see if there are any other errors.  If it does compile the second 
time, send a warning but accept it.

I've essentially given up on kerbalstuff being able to properly create a 
netkan file, since there are too many ways for it to mess up.



LGG

Dazpoet:
I meant the fact that they share files between them so we might need a third Core mod or Shared Assets, or both :)

linuxgurugamer:
Oh.

I didn't realize that.

Ok.  So how about this:

Release the Pathfinder netkan that I wrote for now.  I'll contact the 
author, and will get a new one set up as a dependency of these two.

I'll suggest to him the following, similiar to what I did for Spacetux:

Create a WildBlueIndustriesSA mod, containing the following:
     1.  Agencies
     2.  The WildBlueTools.dll

Once this is done, then:

1.  Write a netkan for the SA mod
2.  Upload both updated mods to kerbalstuff
3.  Modify the Pathfinder netkan to add this as a dependency
4.  Write a netkan file for the Deep Space Exploration Vessels

Sounds good?


JBB

pjf:
We have a WildBlueTools as a separate release now! So our dependencies are straightforward, and hopefully we can integrate them all at the same time. :)

dbent:
Closed in favor of #1747

PR #1649, opened 2015-06-22
Create RoverSpeed.netkan

maculator:
In this netkan ther should ba a License.md, a Readme.md, and the Roverspeed.cfg.
All tested and working.


RichardLake:
The reason it is failing is due to 'RoverSpeed.netkan identifer should match filename'.
Rename the identifier to RoverSpeed and it should pass.

maculator:
Done, also i can confirm it works with 1.0.3

PR #1654, opened 2015-06-23
Update ModelRocketry.netkan

Dazpoet:
Bumping spec_version to account for array of licenses


pjf:
Spec version bumps are always safe, so merging. :)

PR #1658, opened 2015-06-24
Add epoch to CivilianPopulation

plague006:
CKAN is listing 1.41 (for KSP 0.90) instead of 1.6.2 (for KSP 1.0.2). 1.41 was supposed to be 1.4.1


pjf:
Adding an epoch is always safe, so doing this with great joy. ;)

PR #1679, opened 2015-06-26
Add SigmaRTExpansion

dbent:


dbent:
@Sigma88 Any particular version of `ModuleManager` and `RemoteTech`?

Also, you'll need to mark the mod as up-to-date on Kerbal Stuff or it will only be generated for KSP 1.0.2.

Sigma88:
@dbent 
I don't think that would be necessary, the mods can work even without RemoteTech
Module manager is a requirement of course, but the version shouldn't be important as long as it's 1.0.x compatible

PR #1683, opened 2015-06-27
Latest TACLS works on KSP 1.0.2-1.0.4

pjf:
This requires KSP-CKAN/CKAN#1197 to be merged to actually function,
but can be safely be merged before then (as `x_netkan_override` is
ignored in the current netkan build).

Part of KSP-CKAN/CKAN#1184.


pjf:
@taraniselsu : Just a quick note to say we're marking the current TACLS as compatible with 1.0.3 and 1.0.4. If that's in error, please let us know.

Future releases will do whatever is marked in the AVC file. :)

PR #1690, opened 2015-06-28
Basic tests for x_netkan_override

pjf:
These update the tests run on our PRs to spot common problems with
`x_netkan_override`. It insists that the field, if present:

- Must contain an array of overrides
- All overrides must have a version
- Overrides must have a delete section, an override section, or both.


dbent:
Seems reasonable to me.

pjf:
Merging like the danger-loving fiend that I am. :)

PR #1691, opened 2015-06-28
Override SETI-CommunityTechTree KSP versions

dbent:


pjf:
This fills me with joy, and matches the forum thread compat info. :)

PR #1661, opened 2015-06-25
Added Extensive Engineer Report mod to NetKAN

jkoritzinsky:

PR #1693, opened 2015-06-28
Add ExtensiveEngineerReport

pjf:
Based on #1661.
Closes #1661.


pjf:
And merging with great thanks to @jkoritzinsky !

PR #1699, opened 2015-06-28
Add another configuration for FE

Crzyrndm:
Changes for FE v2.3


NetKAN/FilterExtensions.netkan
```diff
@@ -18,6 +18,11 @@
         {
             "file"       : "GameData/000_FilterExtensions",
             "install_to" : "GameData"
+        },
+        {
+            "file"       : "GameData/000_FilterExtensions Configs"
+            "install_to" : "GameData"
```
pjf: Oh no! We're missing commas here, and on the line above, which means our JSON isn't parsing.

PR #1702, opened 2015-06-28
Update Corvus125mTwoKerbalPod.netkan

Murdabenne:
override


Murdabenne:
Second commit was adding in the override on the dependency as well, but its apparently not getting picked up when the mod build happens - or else Im simply not reading the console output right, which is a good possibility as well

pjf:
Despite Jenkins getting confused, this looks grand. Thank you. :)

PR #1704, opened 2015-06-28
Update Snacks.netkan

Murdabenne:
override for 1.0.2 thru 1.0.4


pjf:
It looks like we've got a hard-coded version here already (because there's no `.version` file, or we're not reading it), so rather than putting in an override we might want to manually put in a `ksp_version_min` of 1.0.2, and a `ksp_version_max` of 1.0.99, which looks awful but means we'll at least keep marking new releases as 1.0.x compatible.

Murdabenne:
OK I'll implement that change

pjf:
Perfect, thank you so much!

PR #1705, opened 2015-06-28
Fix min/max on Proton-Bobcat, SovietEngines, and StripSymmetry

pjf:
Final preening of the great metadata clean in KSP-CKAN/CKAN-meta#628.

Note: At least two of these files were produced by @Felger. If the
Proton and Soviet engines are known to work under 1.0.x, let us know
(or send us a PR) and we'll merge that for great justice!


pjf:
Merging away as this is my feedback on KSP-CKAN/CKAN-meta#628, it looks like it's doing the right thing, and I trust @Felger to catch me when I screw up. :)

PR #1715, opened 2015-06-29
Add .netkan for HoloDeck by fingerboxes

blnk2007:


dbent:
A couple of things:

- The appropriate `identifier` for this should probably be `HoloDeck`.
- While the `.netkan` is otherwise correct, our indexer bot won't generate any releases until such time as @fingerboxes makes a non-prerelease release. You however, can by running:

  ```
  > netkan --prerelease HoloDeck.netkan
  ```

  Then installing the resulting `.ckan` file. When a non-prerelease release made it will then upgrade gracefully.

plague006:
HoloDeck hasn't had a release since May 1 and this PR hasn't been addressed in a month or so. Closing this and if @fingerboxes has a full release of HoloDeck we can add it when appropriate.

PR #1717, opened 2015-06-29
Updateded dependancies to 1.0.4 compatible version (except Toolbar)

sswelm:


Murdabenne:
FYI toolbar current version has been given an override so that its marked as working for 1.0.3 and 1.0.4

sswelm:
thanks

dbent:
Thanks, but the most recent compatible version should be used automatically. And it looks like `KSPInterstellarExtended` is installable on KSP 1.0.4 now, so I'm closing this for now.

PR #1720, opened 2015-06-29
ImprovedChaseCamera 1.5.1 min-max override

Murdabenne:
provide v 1.5.1 override for 1.0.4 as well as 1.0.3 and .2


pjf:
Merging with thanks!

PR #1724, opened 2015-06-29
Update TweakableEverything.netkan with recommends

blnk2007:
Changed suggests to recommends because "Extras" folder come with TweakableEverything's .zip


dbent:
Thanks, closing in favor of #1744.

PR #1725, opened 2015-06-29
Update SigmaModExpansions.netkan

Sigma88:
changing dependencies

### DIFF
--- a/NetKAN/SigmaModExpansions.netkan
+++ b/NetKAN/SigmaModExpansions.netkan
@@ -8,7 +8,7 @@
         "SigmaRTExpansion"
     ],
     "depends": [
-        { "name": "ModuleManager" },
+        { "name": "ModuleManager" }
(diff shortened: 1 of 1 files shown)


dbent:
Need to remove the comma at the end of `{ "name": "ModuleManager" },`.

dbent:
#rebuild

dbent:
Changes should appear in the next run of the bot which is in about an hour and a half.

Sigma88:
@dbent nice thanks, so now If I want to add stuff like "supports" and other stuff I can just do this. but someone with writing permission will always have to check it before it will be applied right?

dbent:
Yep.

Sigma88:
@dbent perfect, thanks for the help. now I just need to decide how many parameters to add... I'm looking at the OPM file and it's huge :D

PR #1727, opened 2015-06-29
Add Base Construction

CosmoBro:


CosmoBro:
can someone tell me why it doesn't merge? to fix it

NetKAN/ContractConfigurator-BaseConstuction.netkan
```diff
@@ -0,0 +1,38 @@
...
+    ],
+	"install"       : [
+        {
+            "find": "BaseConstruction",
+            "install_to": "GameData/ContractPacks"
+        }
+    ],
+	"suggests"      : [
+        {
+            "name": "KAS",
```
techman83: You've got an extraneous ',' after "KAS".

dbent:
Thanks, closing in favor of #1745, which fixes the command bumps the spec to `v1.2` which is required to support subdirectories in an `install_to` property.

CosmoBro:
thanks

PR #1729, opened 2015-06-30
KerbalFoundries Fix respository error from KSpace

Murdabenne:
its set to "include in the install" at kerbalspace, correcting thsi with the actual download URL"


pjf:
Rather than having this set to the download itself, can we have it just point to https://kerbalstuff.com/mod/58/ ? That means it'll still go to the right place even if there's a new release.

(Also, in CKAN v1.10.0 bad download links get ignored by everything, which will reduce the impact from things like this.)

Murdabenne:
sure, putting the fix in shortly

pjf:
Merging with thanks!

PR #1730, opened 2015-06-30
Rename TweakableEverythingExtras to TweakableEverythingExtras.netkan

blnk2007:


dbent:
Thanks, I'm closing this in favor of #1742 and #1743 which split the extras into separate options so users can pick and choose.

PR #1740, opened 2015-07-01
Adds DSEV, Pathfinder, WildBlueTools

Angel-125:
Adds CKAN entries for Pathfinder, Deep Space Exploration Vessels, and
WildBlueTools

### DIFF
--- a/NetKAN/DSEV.netkan
+++ b/NetKAN/DSEV.netkan
@@ -1,0 +1,19 @@
+{
+    "spec_version"   : "v1.4",
+    "identifier"     : "Deep Space Exploration Vessels",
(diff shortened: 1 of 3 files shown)


dbent:
Thanks, closed in favor of #1746, #1747, and #1748.

BTW, I added a filter to `WildBlueTools` for `MiniAVC.xml`, this should be generated for the user on first run so they're prompted to choose if they'd like to enable version checks. CKAN users won't receive the file, but users doing a manual install will. You'll probably want to remove it from the distribution.

Angel-125:
Thanks for the heads up, I'll be sure to remove the .xml. :)

PR #1771, opened 2015-07-02
Adds MOLE entry

Angel-125:
This change adds an entry for my new Mark One Laboratory Extensions mod.


plague006:
DSEV was added in  #1748 Pathfinder in  #1747, and WildBlueTools in #1746
That's causing the failed checks, only MOLE needs to be added at the moment.

dbent:
Closing in favor of #1836, thanks.

PR #1779, opened 2015-07-03
Add LLL

plague006:
Closes #1728 
I have it setup that the default install includes CLS and TACLS compatibility while "Extras" is only the extra parts.
If someone feels strongly about it, I can split the 2 compatibilities into their own modules.

### DIFF
--- a/NetKAN/LLL.netkan
+++ b/NetKAN/LLL.netkan
@@ -5,7 +5,7 @@
     "identifier": "LLL",
     "depends": [
         { "name": "ModuleManager" },
-        { "name": "FireSpitterCore"}
+        { "name": "FirespitterCore"}
     ],
     "recommends": [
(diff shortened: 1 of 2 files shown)


NetKAN/LLL.netkan
```diff
@@ -0,0 +1,25 @@
+{
+    "spec_version": 1,
+    "license": "CC-BY-NC-SA-4.0",
+    "$kref": "#/ckan/kerbalstuff/954",
+    "identifier": "LLL",
+    "depends": [
+        { "name": "ModuleManager" },
+        { "name": "FireSpitterCore"}
```
dbent: `FirespitterCore`, the identifiers are case-sensitive.

plague006:
Corrected.

dbent:
Manually tested, merging. Thanks.

PR #1780, opened 2015-07-03
Add Eagle Transporter, Area 21 S, NASA career TC

plague006:
More PR catchup.


plague006:
Broken up into #1816  #1817  #1818 for easier management.

PR #1781, opened 2015-07-03
KSPAPIE Updated per sourceode changes to be 1.0

Murdabenne:
AUthors intent in pretty clear - compatible with 1.0.x

source code change in mod as of 2015/06/28 is as follows

from this:

return Versioning.version_major == 1 && Versioning.version_minor == 0 && Versioning.Revision == 0;

to this:

return Versioning.version_major == 1 && Versioning.version_minor == 0;

### DIFF
--- a/NetKAN/KSPAPIExtensions.netkan
+++ b/NetKAN/KSPAPIExtensions.netkan
@@ -2,14 +2,24 @@
     "spec_version" : 1,
     "identifier"   : "KSPAPIExtensions",
     "$kref"        : "#/ckan/github/Swamp-Ig/KSPAPIExtensions",
-    "name"         : "KAE - KSPAPIExtensions",
+    "name"         : "KAE KSPAPIExtensions",
     "abstract"     : "This package contains a bunch of utility methods for general use to modders.",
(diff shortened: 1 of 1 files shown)


Murdabenne:
@NathanKell I fixed up the netkan file so the updated KSPIE shows up in CKAN

NathanKell:
KSPAPIExtensions should not be on CKAN. A given mod release is compiled against, and includes, a specific version, and CKAN should not be installing this except as one of the files in a mod. See prior issue on this topic (wish I could remember the number). The solution was, for backwards-compatibility with prior metadata, to lock its version to .90 in perpetuity.

NathanKell:
There we go, found it. #1428

Murdabenne:
AH, OK - we had a user show up and say this was causing a crash in his install.  I read the source code, but was unaware of the long-ago history. Thanks for correcting me.

NathanKell:
No problem, and sorry if I came on sharply, I (as well as others) greatly appreciate the work you do. :)

Murdabenne:
LOL - sharply?  No way.  Direct - and thats the best way.

PR #1782, opened 2015-07-03
SovietEngines update override versioning

NathanKell:
My guess is that this is the correct way rather than https://github.com/KSP-CKAN/CKAN-meta/pull/673 @pjf @Dazpoet


Dazpoet:
Assuming nothing major changed in how things like these work in the last few weeks this is the correct way yes.

PR #1783, opened 2015-07-03
Update RealPlume.netkan

Felger:
Will not pass build until RealismOverhaul.netkan is updated.

### DIFF
--- a/NetKAN/RealPlume.netkan
+++ b/NetKAN/RealPlume.netkan
@@ -1,7 +1,7 @@
 {
     "spec_version"      : "v1.4",
     "$kref"             : "#/ckan/github/KSP-RO/RealPlume",
-    "ksp_version"       : "1.0"
+    "ksp_version"       : "1.0",
     "name"              : "Real Plume",
     "identifier"        : "RealPlume",
     "abstract"          : "A set of effects and basic configs that can be attached to any engine as appropriate.  Requires secondary configs to function correctly.",
--- a/NetKAN/RealPlume.netkan
+++ b/NetKAN/RealPlume.netkan
@@ -12,6 +12,9 @@
         { "name" : "SmokeScreen" },
         { "name" : "RealPlumeConfigs" }
     ],
+    "conflicts" : [
+        { "name" : "HotRockets" }
+    ],
     "resources" : {
         "homepage" : "https://github.com/KSP-RO/RealPlume/wiki"
     },
--- a/NetKAN/RealPlume.netkan
+++ b/NetKAN/RealPlume.netkan
@@ -1,33 +1,24 @@
 {
-    "spec_version"   : "v1.4",
-    "$kref"          : "#/ckan/github/KSP-RO/RealismOverhaul",
-    "$vref"          : "#/ckan/ksp-avc",
-    "name"           : "Real Plume",
-    "identifier"     : "RealPlume",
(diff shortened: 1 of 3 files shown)


NathanKell:
@Felger  I don't think you can have two provides? Maybe combine the engine and plume config provides in RO netkan?

dbent:
Two provides is perfectly fine, but `provides` is not a full relationship, it expects an array of strings, not an array of objects like other relationships.

Felger:
Is the RealismOverhaul provides correct now? It looks correct per the CKAN Spec, but I'm not 100% sure.

dbent:
You want:
```
"provides": [ "RealPlumeConfigs", "RealFuels-Engine-Configs" ],
```

NetKAN/RealPlume.netkan
```diff
@@ -1,33 +1,27 @@
...
-    "license"        : "CC-BY-SA",
-    "release_status" : "stable",
+    "spec_version"      : "v1.4",
+    "$kref"             : "#/ckan/github/KSP-RO/RealPlume",
+    "ksp_version"       : "1.0"
+    "name"              : "Real Plume",
+    "identifier"        : "RealPlume",
+    "abstract"          : "A set of effects and basic configs that can be attached to any engine as appropriate.  Requires secondary configs to function correctly.",
+    "license"           : "CC-BY-NC-SA",
+    "release_status"    : "WIP",
```
dbent: `release_status` is an enumeration so it can only take one of [three fixed values](https://github.com/KSP-CKAN/CKAN/blob/master/CKAN.schema#L77). `development` is probably the appropriate one here.

NetKAN/RealPlume.netkan
```diff
@@ -1,33 +1,27 @@
...
-    "$kref"          : "#/ckan/github/KSP-RO/RealismOverhaul",
-    "$vref"          : "#/ckan/ksp-avc",
-    "name"           : "Real Plume",
-    "identifier"     : "RealPlume",
-    "abstract"       : "Option install for Realism Overhaul, adds realistic plume expansion to engines.  See thread for currently configured engines.",
-    "license"        : "CC-BY-SA",
-    "release_status" : "stable",
+    "spec_version"      : "v1.4",
+    "$kref"             : "#/ckan/github/KSP-RO/RealPlume",
+    "ksp_version"       : "1.0"
```
dbent: Missing a comma after this.

dbent:
`AdvancedJetEngine` `recommends` `HotRockets`, meaning installing `RealismOverhaul` will fail by default on a conflict between `HotRockets` and `RealPlume`, unless `--no-recommends` is used (or whatever the equivalent in the GUI is).

My recommendation is to weaken the `recommends` relationship of `AdvancedJetEngine` on `HotRockets` to a `suggests`. I think @NathanKell is the maintainer of `AdvancedJetEngine` nowadays, does that sound OK?

@pjf It might be worth teaching the relationship resolver that it can drop `recommends` relationships if they produce a conflict. This should produce a notice to the user informing them that the `recommends` as dropped.

Felger:
Fixed AJE's recommendation of hotrockets. Thanks for checking all of these, @dbent!

NathanKell:
Per the above (and it looking fine to me) I'll merge.

PR #1793, opened 2015-07-03
KerbalRPNCalculator-1.4 override ksp min=1.0.2 max=1.0.4

Murdabenne:


dbent:
That `$vref` was always wrong `#/ckan/avc` :weary:, do you mind fixing that while you're at it?

Murdabenne:
done

dbent:
Thanks.

Issue #236, opened 2014-12-20
The netkan bot won't pick up multple releases if they are released too close to each other


Ippo343:
The bot refreshes the index about once an hour. However, it seems to assume that only one release has appeared since the last update.

Case study: we missed a FAR release. If you look at the changelog on kerbalstuff, there is a v0.14.4.1 release that hasn't been indexed by the bot. This is because ferram4 has released v0.14.4.1 and v0.14.5 within minutes of each other and the bot only picked up the latest one, ignoring the other one.

Poking @pjf and @AlexanderDzhoganov : we might want to improve the bot's detection of new releases.

AlexanderDzhoganov:
This should be in CKAN-netkan, could you move it?

hakan42:
This issue was moved to KSP-CKAN/CKAN-netkan#23

pjf:
and unmoved. :)

Dazpoet:
I think this is acctually for NetKAN-bot rather than this directory, notifying @techman83 who is either interested enough to open a duplicate over there or close this :)

techman83:
I'm hoping KSP-CKAN/NetKAN-bot#2 will help this. Currently NetKAN grabs the latest release and pushes it into CKAN-meta and that only happens every 3 hours currently (due to cpu utilisation).

How common is this situation? What are we missing by missing a release if a new one was released soon after (presumably it would most commonly done to fix a futzed release?).

Dazpoet:
Not sure how big of a problem this really is acctually, I can't say I've ever ran into a problem because of this.

For the completionist in all of us it would be nice to have an as complete repository as possible though which should include all versions, no matter how insignificant.

plague006:
@techman83 if this can't be (reasonably) fixed, can we close the issue?

techman83:
@plague006 generally if releases are really close together it probably means a bug and we'll likely miss the buggy one. So it's probably pretty edge case.

There is room in the code for more frequent indexing and we'll get there eventually. So I think yes :+1:

PR #1800, opened 2015-07-04
TACLSMining 1.0.4 override

plague006:


dbent:
Can you throw in an `x_netkan_epoch` of `"1"`? The most recent version of the mod on Kerbal Stuff is `1` but the version previous is `1.0`, so that version will be treated as the latest because `1` < `1.0`.

plague006:
Good catch. As an aside: any idea why CKAN-meta doesn't have the 1.0 version?

dbent:
They were released on the same day and I'm willing to be pretty close together in time, so due to KSP-CKAN/NetKAN#236 the bot probably didn't catch it.

PR #1775, opened 2015-07-02
PR catchup (Roster Manager, Docking Sounds, Death star, Solaris Hypernautics, Antennas, Soviet TC, Nuke Addon)

plague006:
Fingers crossed that I don't have too much to fix.

PR #1806, opened 2015-07-04
Add Antennas.netkan

plague006:
Breakup of  #1775


NetKAN/Antennas.netkan
```diff
@@ -0,0 +1,19 @@
+{
+    "spec_version": 1,
+    "license": "CC-BY-SA-NC-4.0",
```
dbent: This should probably be `CC-BY-NC-SA-4.0`.

plague006:
Corrected

dbent:

PR #1810, opened 2015-07-04
Add Kazza197sNukeAddonToBahamutoDsArmoury.netkan

plague006:
Breakup of  #1775

### DIFF
--- a/NetKAN/Kazza197sNukeAddonToBahamutoDsArmoury.netkan
+++ b/NetKAN/Kazza197sNukeAddonToBahamutoDsArmoury.netkan
@@ -1,0 +1,15 @@
+{
+    "spec_version": 1,
+    "license": "CC-BY-4.0",
+    "$kref": "#/ckan/kerbalstuff/438",
+    "identifier": "Kazza197sNukeAddonToBahamutoDsArmoury",
+    "depends": [
+        { "name": "BDArmory" }
+    ],
+    "install": [
+        {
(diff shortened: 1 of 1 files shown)


NetKAN/Kazza197sNukeAddonToBahamutoDsArmoury.netkan
```diff
@@ -0,0 +1,15 @@
+{
+    "spec_version": 1,
+    "license": "CC-BY-4.0",
```
dbent: `unknown` or `unrestricted` are probably better choices for the license for this one, although I'm not sure what technically qualifies as "unrestricted". @pjf ?

plague006:
Ack, I don't recall why I attributed that license when he didn't actually have it.

IMO the license isn't `unknown` since we know his conditions. And it's not `unrestricted` because he certainly has a restriction (attribution).

I have a strong preference to default to `restricted` in this case since it has always seemed to me to be the default way to go when someone has a license but isn't a standard one. There are currently only 10 mods with unknown licenses (which I plan on cleaning up if and where possible).

Some examples of licenses that clearly aren't `unrestricted` but also are a far cry from ARR and have been set to `restricted`
AIESAerospace-Unofficial-v1.5.1.ckan
AMEG-1.1.ckan
BetterBuoyancy-v1.1.ckan

dbent:
Fair enough. Tested manually, merging.

PR #1786, opened 2015-07-03
Add Kurrikane, SmallFatTechnologies, KerbalFuels, ABLaunchers, TheMaritimePack

plague006:
5 more  off the list

PR #1811, opened 2015-07-04
Add SmallFatTechnologies.netkan

plague006:
Breakup of #1786


dbent:
Checked manually, merging. Thanks.

PR #1812, opened 2015-07-04
Add TheMaritimePack.netkan

plague006:
Breakup of #1786

### DIFF
--- a/NetKAN/MaritimePack.netkan
+++ b/NetKAN/MaritimePack.netkan
@@ -1,6 +1,6 @@
 {
     "spec_version": 1,
-    "license": "BSD",
+    "license": "BSD-3-clause",
     "$kref": "#/ckan/kerbalstuff/942",
     "identifier": "MaritimePack",
     "depends": [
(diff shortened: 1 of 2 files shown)


NetKAN/TheMaritimePack.netkan
```diff
@@ -0,0 +1,15 @@
+{
+    "spec_version": 1,
+    "license": "BSD",
+    "$kref": "#/ckan/kerbalstuff/942",
+    "identifier": "TheMaritimePack",
```
dbent: Nitpick, can you drop the definite article from the identifier?

plague006:
Hadn't noticed that there's not a single identifier starting with "The"

NetKAN/MaritimePack.netkan
```diff
@@ -0,0 +1,15 @@
+{
+    "spec_version": 1,
+    "license": "BSD",
```
dbent: From glancing as the `License.txt` in the distribution it looks like the `BSD-3-clause`.

plague006:
Done

dbent:

Also thanks for putting the appropriate `Closes` comment so I don't have to track down the KerbalStuffBot PR.

PR #1813, opened 2015-07-04
Add KerbalFuels.netkan

plague006:
Breakup of #1786

### DIFF
--- a/NetKAN/KerbalFuels.netkan
+++ b/NetKAN/KerbalFuels.netkan
@@ -1,6 +1,6 @@
 {
     "spec_version": 1,
-    "license": "CC-BY-NC-SA 4.0",
+    "license": "CC-BY-NC-SA-4.0",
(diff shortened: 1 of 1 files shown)


NetKAN/KerbalFuels.netkan
```diff
@@ -0,0 +1,13 @@
+{
+    "spec_version": 1,
+    "license": "CC BY-NC-SA 4.0",
```
dbent: `CC-BY-NC-SA-4.0`

plague006:
Corrected

dbent:
Missed the space at the end. :wink:

plague006:
lol, I lost my test environment when a drive failed, I was relying on jenkins for catching the small things :(

dbent:
Tested, manually. Merging.

Apparently even a butler like Jenkins needs a vacation every now and then.

PR #1816, opened 2015-07-04
Add USAConversion.netkan

plague006:
Breakup of #1780

### DIFF
--- a/NetKAN/USAConversion.netkan
+++ b/NetKAN/USAConversion.netkan
@@ -1,0 +1,27 @@
+{
+    "spec_version": 1,
+    "license": "CC-BY-NC-SA-4.0",
+    "$kref": "#/ckan/kerbalstuff/962",
+    "identifier": "USAConversion",
+    "install": [
+        {
+            "file"       : "GameData/USAConversion",
+            "install_to" : "GameData",
+            "filter"     : "Thumbs.db"
+        },
+        {
+            "file"       : "Ships",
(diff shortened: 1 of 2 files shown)


NetKAN/USAConversion.netkan
```diff
@@ -0,0 +1,27 @@
+{
+    "spec_version": 1,
+    "license": "CC-BY-NC-SA-4.0",
+    "$kref": "#/ckan/kerbalstuff/962",
+    "identifier": "USAConversion",
+    "install": [
+        {
+            "file"       : "GameData/USAConversion",
```
dbent: `GameData/USAconversion` (lowercase-C) otherwise it doesn't match.

NetKAN/USAConversion.netkan
```diff
@@ -0,0 +1,27 @@
+{
+    "spec_version": 1,
+    "license": "CC-BY-NC-SA-4.0",
+    "$kref": "#/ckan/kerbalstuff/962",
+    "identifier": "USAConversion",
```
dbent: Also, the identifier may make more sense as `NASACareerModeTotalConversion` or something, the only reference I can find to `USAConversion` is in the install location.

plague006:
Done

dbent:
One of its dependencies `OMSKSpaceProducts` is currently listed as only compatible with KSP 1.0.2, might have to add an override for that. Otherwise tested manually, merging.

PR #1822, opened 2015-07-04
Add Probe Control Enabler

Felger:


NetKAN/RemoteTech-ProbeControlEnabler.netkan
```diff
@@ -0,0 +1,19 @@
+{
+    "spec_version"   : "v1.4",
+    "$kref"          : "#/ckan/github/Felger/FelgerFixes",
+	"ksp_version"    : "1.0"
```
dbent: Needs a comma here.

NetKAN/RemoteTech-ProbeControlEnabler.netkan
```diff
@@ -0,0 +1,19 @@
...
+    "identifier"     : "RemoteTech-ProbeControlEnabler",
+    "abstract"       : "Optional ModuleManager patch by Felger to re-enable probe control when a link is not present.  You will still need a link to transmit science data.",
+    "license"        : "CC-BY-SA",
+    "release_status" : "stable",
+    "depends" : [
+        { "name" : "ModuleManager" }
+    ],
+    "install" : [
+        {
+            "find"       : "RT_ProbeRemover.cfg",
```
dbent: Same as the other PR:

`find` doesn't work on files, only directories (although I see no reason why it shouldn't also work on files). You'll need:
```
"file"       : "GameData/ZZZ_FelgerFixes/RT_ProbeRemover.cfg",
```

Felger:
I believe that'll also reduce the spec_version, but I don't recall which spec version I ought to drop it to at this point, besides, folks should have that updated to the latest version!

dbent:
`v1.2` but that's fine, I'll update it if a 1.2 user raises an issue.

Tested manually, merging.

PR #1824, opened 2015-07-04
Add netkan for RealHeat

NathanKell:


dbent:
Checked manually, merging.

PR #1827, opened 2015-07-04
Add EasyLaunchpads.netkan

plague006:


dbent:
Checked manually, merging. Thanks.

PR #1829, opened 2015-07-05
Create HotRocketsFireworkspack

Nazari1382:


plague006:
License needs to be changed to `GPL-3.0` and this tries to overwrite `MP_Nazari/FX/flameparticle2.png` in the mod `HotRockets`.

dbent:
@Nazari1382 are `HotRockets` and `HotRocketsFireworkspack` mutually exclusive? In that case the simplest thing to do would be to add a `conflicts` relationship on `HotRockets` to this.

The next option would be to determine if it's possible to use a Modular Manager patch to overwrite the file path. That way this could distribute `flameparticle2-fireworks.png` and have a MM patch change the path.

Next option would be to see if there's a way to modify the mod to avoid the overwrite entirely.

Last option would be to make things a bit complicated and have `HotRockets` depend on a virtual package `HotRockets-FlameParticle`, have a `HotRockets-FlameParticle-Default` built from `HotRockets` and have this mod also `provides` `HotRockets-FlameParticle`. Then when installing `HotRockets` the user would be prompted to select between installing `HotRockets-FlameParticle-Default` and `HotRocketsFireworkspack`. That seems a bit overkill for this scenario though.

Nazari1382:
@plague006 if the license needs to be GPL-3.0, why does Kerbalstuff give me GPLv3? This seems like a Kerbalstuff issue. I changed it though. It is fine that the png overwrites, it is the same file. The mods do not need each other.

plague006:
@Nazari1382 CKAN only accepts certain strings for licenses. When KerbalStuff send something like GPLv3 we just overwrite it on our end. As to the issue of one mod overwriting another: it's something expressly forbidden at CKAN's core to ensure files are what we expect and which mod we expect them to come from. In truth I might not be explaining the issue to its full extent, I'm just a metadata monkey and not a coder. =)

The tl;dr is that no mod can overwrite another. We can ask mod authors to address the issues via alternate methods (such as ModuleManger patches) but if it's not possible or if a mod author doesn't want to make the effort then we'll still make every effort to accommodate. 

As @dbent mentioned the workaround would be splitting `HotRockets` and `HotRocketsFireworkspack` into 4 modules (items in the CKAN mod list) and it tends to become a little bit unsightly.

dbent:
Overwrites are mostly forbidden because then you have to concern yourself with installation order and once you care about installation order all sorts of nice guarantees go away and handling things (correctly) becomes a lot more complex.

In this case since the files are identical we can treat it like a typical shared asset and have:
- `HotRockets-Shared` (providing `flameparticle2.png`)
- `HotRockets` which `depends` `HotRockets-Shared`
- `HotRocketsFireworksPack` which `depends` `HotRockets-Shared`

We can handle that @Nazari1382, unless you really really want to.

Nazari1382:
I don't think there is a need anymore for that png in HotRockets, so I'll delete it and reupload it. I'll leave the png in this mod.

dbent:
Ah, that's easier, thanks. :smiley:

plague006:
Line 5 needs a comma.

KSP-CKAN/CKAN-meta@ee391998f9676cac5004a110ce0ecc6e5e6dceca means no more conflict (yay). Someone might run into an issue if they try to install this with an old version of HotRockets. It shouldn't affect too many people though.

dbent:
Fixes made in #1841

PR #1831, opened 2015-07-05
License fixes

plague006:
The overall premise is that we should either:

- Only list explicitly ARR as `restricted` while assigning `unknown` to any license not explicitly defined in CKAN.schema.

or
- List licenses that have no restrictions as `unrestricted`, list licenses that have any kind of restriction as `restricted`, and list mods with an absence of a license as `restricted` since many countries imply an ARR license.

I opt for the latter and I'll lay out the case for each decision.

dockingtube `unrestricted`
No requirements or restrictions, only "would be nice"
"Unlike a number of other mods you are positivly encouraged to modify 
this mod! AND release it should you desire.
A mention in the credits would be nice."

KSPInterstellarLite: `restricted`
Includes the same license as KSPI: Allows distribution, modification, etc. but ultimately has restrictions. KSPI license listed as `restricted`

MunarSurfaceExperimentPackage `restricted`
KerbalStuff page indicates BSD as a license, but from the KSP forum thread:
All models and textures are copyright of AlbertKermin, all rights reserved.

ODFC `restricted`
From the forum page: "3. Additionally, explicitly forbidden is the following:" 4 items are listed including repackaging, but the important part is that there are restrictions.

OkramSA `restricted`
No license whatsoever found in the github repo. In many countries no license = ARR, so that's what we should list it as.

SatBatts `restricted`
In the forum thread there are 3 clauses, I'll list 1: "You may not: - Use any part config, texture, or model for a derivative work without first receiving permission from Orum on the KSP forums."

### DIFF
--- a/NetKAN/KSPInterstellarLite.netkan
+++ b/NetKAN/KSPInterstellarLite.netkan
@@ -4,7 +4,7 @@
     "name"                  : "KSP Interstellar Lite",
     "abstract"              : "Lite version of KSP Interstellar",
     "$kref"                 : "#/ckan/github/WaveFunctionP/KSPInterstellar",
-    "license"               : "unknown",
+    "license"               : "restricted",
     "ksp_version"           : "0.24.2",
     "x_netkan_license_ok"   : true,
     "x_maintained_by"       : "distantcam",
--- a/NetKAN/KSPInterstellarLite.netkan
+++ b/NetKAN/KSPInterstellarLite.netkan
@@ -4,7 +4,7 @@
     "name"                  : "KSP Interstellar Lite",
     "abstract"              : "Lite version of KSP Interstellar",
     "$kref"                 : "#/ckan/github/WaveFunctionP/KSPInterstellar",
-    "license"               : "restricted",
+    "license"               : "unrestricted",
     "ksp_version"           : "0.24.2",
     "x_netkan_license_ok"   : true,
     "x_maintained_by"       : "distantcam",
--- a/NetKAN/MunarSurfaceExperimentPackage.netkan
+++ b/NetKAN/MunarSurfaceExperimentPackage.netkan
@@ -3,5 +3,5 @@
     "x_via": "Automated KerbalStuff CKAN submission",
     "spec_version": 1,
     "$kref": "#/ckan/kerbalstuff/634",
-    "license": "unknown"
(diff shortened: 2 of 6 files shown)


plague006:
Pinging @pjf, the resident licensing expert, for input.

pjf:
Drat. Our spec isn't particularly clear on what's meant by `restricted` and `unrestricted`, however tools which make decisions based upon licenses can do care.

I'd prefer keeping `unknown` as meaning "we honestly don't know the license". That's going to be a rare situation, but it could come up if a mod links to a license on their forum post, but that's a broken link.

I'd recommend we state that `unrestricted` means that, to the best of the contributor's knowledge, the mod meets the [OSI's definition of an open source license](http://opensource.org/osd-annotated). `restricted` is therefore a definition by exclusion, it's everything that's *not* a restricted license, and *not* a license that we define elsewhere.

It's worth noting that many of the licenses in the spec *are* restricted; any creative commons license that forbids modification (CC-BY-ND) or commercial use (CC-BY-NC) do not pass the OSI standards. However all the defined licenses in our spec *do* allow for redistribution.

And I've only now realised this is a PR, and not a policy question, so I'm going to go look at the changes rather than talk about how we might adjust the spec to differentiate between restricted licenses that do and do not permit redistribution. :)

pjf:
- `dockingtube` - I agree, `unrestricted`.
- `KSPInterstellarLite` - I think I'm too tired to figure out how KSPI has restrictions. The license isn't one I recognise, but I suspect it passes all the OSI hurdles.
- `MunarSurfaceExperimentPackage` - Alas, restricted, until we can indicate per-file licensing.
- `ODFC` - Agreed, restricted, there are barriers to derived works.
- `OkramSA` - I'd really say 'unknown' here. The github description indicates the assets are used by multiple mods, but maybe not by third-party mods. Besides, @Orum set it to unknown [when the metadata was submitted](https://github.com/Orum/NetKAN/commit/ab5e49b655932eb40fe9a24f2f87a0e2dc48f6cc).
- `SatBatts` - Agree on `restricted`.

Many thanks for this!

plague006:
Beyond here is where I try to rebase and probably end up making a big git-mess.

plague006:
Or maybe this time I don't duplicate old mistakes. \o/

PR #1832, opened 2015-07-05
Add VanguardAstrodynamicsVXSeries.netkan

plague006:
While I usually keep the bot-assigned ID, it seemed a bit long. If anyone feels strongly about it, I'll change it back.


dbent:

PR #1833, opened 2015-07-05
EBRCPartspack futureproofing

plague006:
Should cover us if he increments `EBRC-PACKS_PACK-1`


dbent:
Glad that `find_regexp` is proving useful. Tested manually, merging. Thanks.

plague006:
`find_regexp` is niche, but in the cases it's needed: it's totally invaluable

PR #1835, opened 2015-07-05
Filter SolverEngines.version from ModuleAnimateEmissive

dbent:


dbent:
Tested manually, merging.

PR #1841, opened 2015-07-05
Pull/1829 Fix

dbent:
Fixes #1829


dbent:
Tested manually, merging.

PR #1842, opened 2015-07-06
Update SolarisHypernautics.netkan

dbent:
Newest version doesn't have a leading `.`


dbent:
Tested manually, merging.

PR #1843, opened 2015-07-06
Update SCANsat depends section

DMagic1:
Module Manager is more-or-less required for full SCANsat functionality now, so it should probably go in the depends section.

I'm not sure if that override section is still needed, but I left it as is.


plague006:
It doesn't hurt anything if you rip out the override or leave it.

dbent:

The override section can be left in perpetuity as it's pegged to a specific version.

PR #1845, opened 2015-07-06
Update RealismOverhaul.netkan

Felger:
Forgot a last-minute dependency, not game breaking, but should be added.


techman83:
#rebuild

dbent:
Tested manually, merging.

PR #1846, opened 2015-07-06
Test [DO NOT MERGE]

techman83:
At some point I'll get jenkins to do things.


plague006:
:+1:  I like when Jenkins does things.

techman83:
So do I! The docker plugin is giving me grief!

PR #1847, opened 2015-07-06
Add ksp_version override for RealChute

dbent:


dbent:
Tested manually, merging.

PR #1848, opened 2015-07-06
Fix spec 1.10 test

techman83:
This is a hacky fix for the version comparison issue. Being that 1.10 is less than 1.2 and 1.4. It works and will do for this time of the evening.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,5 @@
 *.ckan
 *.exe
 *~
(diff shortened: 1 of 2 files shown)


techman83:
*whistles* - This works well enough for now. It shouldn't break if we get to 1.20 and haven't done something better.

Merging as is causing unnecessary build failures.

PR #1856, opened 2015-07-06
Force v for RemoteTech.netkan

plague006:
Less an issue for 1.0.4, but an issue that means there are no releases for 1.0.2.

### DIFF
--- a/NetKAN/RemoteTech.netkan
+++ b/NetKAN/RemoteTech.netkan
@@ -5,7 +5,7 @@
 	"license": "restricted",
 	"release_status": "stable",
 	"x_netkan_license_ok": true,
-    "x_netkan_force_v": 1,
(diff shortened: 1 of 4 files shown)


plague006:
Can one of the admins verify this patch? Where's that slacker @netkan-bot asking the important questions?

NetKAN/RemoteTech.netkan
```diff
@@ -5,6 +5,7 @@
 	"license": "restricted",
 	"release_status": "stable",
 	"x_netkan_license_ok": true,
+    "x_netkan_force_v": 1,
```
dbent: Value should be `true`, not `1`.

plague006:
I didn't give it a second thought when it passed netkan and installed.

dbent:
:+1:

Issue #1367, opened 2015-05-22
Fingerboxes mods


Blu3wolf:
#1365 and #1366 add CrewQueue to CKAN. #1366 is the core library that all fingerboxes mods require.

As https://github.com/fingerboxes/FingerboxLib doesnt have a release yet, #1366 points to https://github.com/fingerboxes/CrewQueue for the library instead. This is a temporary fix to get CrewQueue working and should be fixed once FingerboxLib gets a release.

Dazpoet:
Fixed and merged so I'm closing this.

Blu3wolf:
Doesnt look very fixed to me?

https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/FingerboxesCore.netkan

it still points at crewqueue for the core files rather than at https://github.com/fingerboxes/FingerboxLib where it should. Granted this isnt an issue that can be dealt with until the mod author posts a release.

Close it by all means, but its not fixed.

Dazpoet:
Mods are added and working, hence why I claimed it as fixed since the PRs mentioned are merged. The FingerboxesCore issue could (should?) probably be opened as a seperate one.

Blu3wolf:
Can do, but the fingerboxescore issue was the only reason this was opened in the first place. Adding the mod wasnt really an issue. It would just be a duplicate issue.

I can open another if you want. Or I can just remember and wait for the author to make a release. Either works (unless I forget).

Blu3wolf:
To prevent an issue getting closed, should I reference a PR in it and in the PR request that it not be merged (as in a blank PR as a placeholder)?

fingerboxes:
I'm the author of the referenced mods. I have no idea how CKAN/NetKAN works at all - but it seems there is some interest in my mods being available through it.

I've published a release in the repo for FingerboxLib, seems that seems to be a sticking point. FWIW, I push updated version of it with each individual mod release, so...

dbent:
Thanks @fingerboxes, I'll update `FingerboxesCore` to get the assembly from the FingerboxesLib repository.

Closing this issue as `CrewQueue` is available and `HoloDeck` has an open PR.

dbent:
Although one question, @fingerboxes, the download is listed as `1.0.4_r1`, but tagged as `1.0.0_r1`, which is the correct version?

fingerboxes:
both, technically, but go with 1.0.4_r1 ; my versioning system is
[...]

PR #1857, opened 2015-07-07
Update FingerboxesCore

dbent:
Use FingerboxLib repository and [fix version number](https://github.com/KSP-CKAN/NetKAN/issues/1367#issuecomment-118682126) in override.


dbent:
Tested locally.

PR #1596, opened 2015-06-12
Add EmDrive - Release 1 from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of raptorep, to add [EmDrive - Release 1](https://kerbalstuff.com/mod/907/EmDrive%20-%20Release%201) to CKAN.

Please direct questions about this pull request to [raptorep](https://kerbalstuff.com/profile/raptorep).

PR #1858, opened 2015-07-07
Add EmDrive

plague006:
Closes #1596
I used `find_regexp` because I think `Emdrive R1` is likely to change names. The end result of the install isn't pretty `GameData/EmDrive/EmDrive R1` but since this is a parts-only pack there's no reason to think it will create any files that would prevent the `EmDrive R1` folder from being deleted and replaced with `EmDrive R2` down the line.

### DIFF
--- a/NetKAN/EmDrive-TweakPatch.netkan
+++ b/NetKAN/EmDrive-TweakPatch.netkan
@@ -1,0 +1,18 @@
+{
+    "spec_version": 1,
+    "license": "GPL-3.0",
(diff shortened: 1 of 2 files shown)


techman83:
You'll need at least Spec "v1.2" for gamedata with path.

plague006:
You mean for tweak? I thought it was `install_to` that was `"v1.2"` not `file` FML with the number of commits I have: how is it my memory for it getting worse?

techman83:
Looks good to me. The jenkins jobs aren't quite right yet..

PR #1646, opened 2015-06-21
Added AVC line

linuxgurugamer:

PR #1859, opened 2015-07-07
Added AVC vrefs

plague006:
A copy of, and closes #1646
Each netkan tested with no errors on netkan or ckan install.


techman83:
This passes, but another unrelated file fails due to #1848

PR #1861, opened 2015-07-07
Ven's Stock Revamp works fine on 1.0.4

OtherBarry:
I got permission from @NathanKell who 'fixes stuff' in VSR, though not from Ven himself.
Also wasn't sure how the "version" section works in this so it may be wrong.

### DIFF
--- a/NetKAN/VenStockRevamp.netkan
+++ b/NetKAN/VenStockRevamp.netkan
@@ -14,4 +14,14 @@
 	"resources" : {
         "homepage"     : "http://forum.kerbalspaceprogram.com/threads/92764"
     }
+        "x_netkan_override" : [
+        {
+            "version" : "1",
+            "delete" : [ "ksp_version" ],
+            "override" : {
+                "ksp_version_min" : "1.0.2",
+                "ksp_version_max" : "1.0.4"
+            }
+        }
+    ]
 }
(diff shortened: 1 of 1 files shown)


techman83:
#rebuild

techman83:
Jenkins is still learning to Jenkins. But this LGTM. The inflator will catch it if doesn't validate against the spec.

NetKAN/VenStockRevamp.netkan
```diff
@@ -13,5 +13,15 @@
     "comment"      : ".version file currently has accurate mod version but incorrect KSP version",
 	"resources" : {
         "homepage"     : "http://forum.kerbalspaceprogram.com/threads/92764"
-    }
+    },
+        "x_netkan_override" : [
+        {
+            "version" : "v1.2",
```
dbent: This would have the match the tag of the most recent release which is [`v1.8.1`](https://github.com/VenVen/Stock-Revamp/releases/tag/v1.8.1).

OtherBarry: Oh? I was confused by the example PJF linked, in which the spec_version and "version" were in completely different formats.

dbent: No problem ,fixed in #1865

PR #1423, opened 2015-05-26
Build script update

mgsdk:
Currently the build script for Jenkins fails if:
* The PR contains multiple files with inter-dependencies (e.g. https://github.com/KSP-CKAN/NetKAN/pull/1145).
* The KSP version is not the latest (e.g. https://github.com/KSP-CKAN/NetKAN/pull/1420).

This attempts to solve this by:
* Use a local metadata file in the form of master.tar.gz, which contains all .ckan files but the one being tested.
* Create dummy KSP installs matching the min KSP version requested by the .ckan file being tested.

Will depend on the static build of https://github.com/stedolan/jq (Not included in this PR) to extract json identifiers.

Any comments welcome, shell scripting is not my force.

Tested on https://github.com/KSP-CKAN/NetKAN/pull/1420 without error locally. My git-fu is not strong enough to figure out how to do a test along the lines of https://github.com/KSP-CKAN/NetKAN/pull/1145.

PR #1862, opened 2015-07-07
Just testing

techman83:


techman83:
Oh ksp_version. #1423 would help here. I will look at that later. But the next PR to this repo might actually build correctly!

techman83:
Adding a comment for jenkins upgrade test.

PR #1864, opened 2015-07-07
Fixed dependencies of Workshop

obivandamme:
The OSE Workshop mod has different dependencies in version 0.9.0


dbent:
Thanks, I've closed this PR in favor of #1866 which updates the dependencies in a fancier way.

PR #1865, opened 2015-07-07
Update VenStockRevamp.netkan

plague006:
VenStockRevamp does not have a version `v1.2`, the most recent version is `v1.8.1` (per CKAN listing and per https://github.com/VenVen/Stock-Revamp/releases) which is what this override seems to want to address.


plague006:
Removed hardcoded version since we're overriding the current version anyway, and this will prevent the next release from having an issue (due to the hardcoded ksp_ver)

dbent:
:+1:

PR #1869, opened 2015-07-08
Test For New Plugin - Formating change only [NO MERGE REQUIRED]

techman83:
Just seeing if the latest GHPRB plugin is doing correct things now.


techman83:
Well that's an improvement :D

techman83:
#rebuild

techman83:
Yay!

PR #1873, opened 2015-07-08
Filter MiniAVC.xml from ContractConfigurator-UsefulSpaceStations

plague006:


techman83:
A

plague006:
Not for me personally, but it just seems to be the convention to filter the .xml.  #1455

PR #1876, opened 2015-07-09
Update SDHI-ServiceModuleSystem.netkan

plague006:
1.0.4 bump


sumghai:
Yep, this looks good  to me :)

techman83:
Is there a problem with some of the metadata, because in theory 1.0 means all of 1.0.x, this would lock it to 1.0.4.

I think we can override per version now.

sumghai:
This new update is intended for KSP 1.0.4 only. Not sure what value overriding would have.

techman83:
Fair point, looks like your using GitHub, so no version information can be gathered automatically. So this makes sense.

sumghai:
Excellent, thanks guys :)

Yeah, I'm just using Github.

PR #1879, opened 2015-07-10
Bring flag-only mods to the same version, and add a comment to identify them.

plague006:
One more step toward KSP-CKAN/CKAN-meta#623's completion.


plague006:
Pinging @techman83 to look into Jenkins' handling of this. From the output console:
>164 [1] ERROR CKAN.KSP (null) - Could not find KSP version in readme.txt
Sorry, /home/jenkins/workspace/NetKAN/dummy_ksp does not appear to be a KSP directory

plague006:
KSP-CKAN/CKAN#1273

plague006:
Each file tests successfully locally, jenkins fails because it doesn't recognize the version.

For the sake of uniformity, all the flag mods are changed to `ksp_version` `any` and they all have a `comment` of `flagmod` if they ever all need to be found in the future.

PR #1880, opened 2015-07-10
Update dependent version of IR in InfernalRobotics-Sequencer.netkan

plague006:
Version dependency added in most recent changelog


plague006:
Should have stated: tested locally and passed Jenkins.

PR #1891, opened 2015-07-12
NearFutureElectrical-0.5.2 no longer depends on HeatControl-Core

Murdabenne:
changed to Recommends until/unless is changed back.

Per release note comment 0.5.2: 
"Retuned radiator and reactor parts to work with stock radiator system"

Installed and tested, it works without HeatControl.  Verified by user post in forum as well.


plague006:
Passes local tests, passes jenkins, merging with my thanks.

PR #1893, opened 2015-07-12
Extant RP-0 releases should be for KSP 0.90 only.

pjf:
The netkan side of CKAN-meta#701.


plague006:
Passes local, passes jenkins.

PR #1899, opened 2015-07-12
Correct install for Mk3MiniExpansionPack.netkan

plague006:
This will allow us to index v1.2 onward.


plague006:
Passes local & jenkins, let's go get those new versions.

PR #1900, opened 2015-07-12
Add comment to OPTMk3MiniExpansion.netkan indicating why it's version locked.

plague006:


plague006:
This fails Jenkins, fails local: but it's supposed to. It's a dead netkan that I just want a comment in for clarity.

PR #1905, opened 2015-07-12
Fix syntax error for provides field of NearFuturePropulsionExtras.netkan

ericperko:
Running the existing NearFuturePropulsionExtras.netkan file through netkan.exe revealed a syntax error causing parsing to fail.


plague006:
Passes local tests, passes jenkins, merging with my thanks.

PR #1908, opened 2015-07-13
Change ProceduralParts to depend on ModuleManager

JonDahm:
I've found that Procedural Parts will not run without Module Manager. See Swamp-Ig/ProceduralParts#152


plague006:
Passes local, passes Jenkins, merging with my thanks.

PR #1911, opened 2015-07-13
Force epoch in cryoenginesLFO to allow 1.81 upgrading to 1.9

plague006:


plague006:
Passes local & jenkins, merging.

PR #1914, opened 2015-07-14
Add dependencies for FuelTanksPlus.netkan

### LINKED ISSUE #1912: Fuel Tanks Plus has new dependencies
When I release the next version of Fuel Tanks Plus (pretty soon now), it'll require both Module Manager and the Firespitter Core to function. There's no harm in adding the dependency requirements in CKAN ahead of the release. The opposite isn't true, so I'll hold off on releasing until we have the dependency requirement added.

plague006:
Closes #1912


plague006:
Passes local, passes Jenkins, merging and pinging @NecroBones to let you know it's done.

NecroBones:
Thanks :)

PR #1916, opened 2015-07-14
Resolve file conflicts between KerbinSide and KerbinSideAirRace aka "The Red Pill" (we shall see how deep the rabbit hole goes)

plague006:
Closes KSP-CKAN/CKAN-meta#706


plague006:
Jenkins passing is erroneous.

techman83:
Does this refer to KSP-CKAN/CKAN#1273

plague006:
KerbinSideAirRace updated, creating new conflicts.

dbent:
:sob: We really ought to create a list of "fragile" installations like this, so at least we're aware they may require double checking on updates.

I assume `KerbinSideAirRace` does not actually distribute a full copy of so a simple `conflicts` wouldn't work?

dbent:
Also, are the files that need filtering not identical? If they are identical we can just filter them from `KerbinSideAirRace` and not bother modifying `KerbinSide`.

plague006:
>I assume `KerbinSideAirRace` does not actually distribute a full copy of so a simple `conflicts` wouldn't work?

Correct.

From the [mod's author](http://forum.kerbalspaceprogram.com/threads/82785-1-0-4-Kerbin-Side-v1-0-3-SideJobs-v1-4-Air-Race-v1-5?p=2083428&viewfull=1#post2083428):
>KerbinSide is not a pre-requisite for Air-Race. Air-Race can stand-alone but if you want to race at Black Krags or Dundard's Edge, you have to get there from KSC. Install KerbinSide All-in-1 and then you can launch from runways at those locations.

We can put this back on hold again though. `KerbinSideAirRace` released 1.5 and I need to look through it.

plague006:
Jenkins continues to test BDarmory and not the mods in this PR. I  :heart: you Jenkins.

plague006:
The conflicting files are all textures.. I could resolve this in a stable way by having `KerbinSideAirRace` install into another folder.. Double-loading textures isn't ideal.. but it means 1 less unneccesary 'core' mod on the list and stability: both things I like.

plague006:
Closes #1936

PR #1922, opened 2015-07-15
Restore 1.0.2 compatibility for BULB-1.1.2

plague006:
Noticed that https://github.com/KSP-CKAN/CKAN-meta/commit/7a189015ff68eedd3e4a3c00ba9ef4606e451f00 breaks the mod for 1.0.2 users. Opened KSP-CKAN/NetKAN-bot#26 with more information.

The change in this PR simply adds an override to the version which will give us the appropriate compatibility range.


dbent:
:+1:

PR #1926, opened 2015-07-15
Deadly Reentry adjustments

pjf:
- No longer depends on ModularFlightIntegrator (mentioned in the 7.2.1 release notes)
- Force a 'v' onto the version string, as DR seems to lose them starting with 7.2.0


plague006:
Passes local, waiting on Jenkins.

PR #1928, opened 2015-07-16
Convert PlaneMode.netkan into MetaNetKAN

dbent:


plague006:
Passes local, passes jenkins. Merging with thanks.

Fly little NetKAN, fly and be free of this place!

PR #1932, opened 2015-07-16
Add 1.0.4 override for BDArmory

plague006:


plague006:
Passes local, passes jenkins, meging.

PR #1930, opened 2015-07-16
Add Old School Turrets - Additional WWII Turrets for BDArmory from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of CrayzeeMonkey, to add [Old School Turrets - Additional WWII Turrets for BDArmory](https://kerbalstuff.com/mod/1006/Old%20School%20Turrets%20-%20Additional%20WWII%20Turrets%20for%20BDArmory) to CKAN.

Please direct questions about this pull request to [CrayzeeMonkey](https://kerbalstuff.com/profile/CrayzeeMonkey).

PR #1933, opened 2015-07-16
Add OldSchoolTurrets.netkan

plague006:
Closes #1930

### DIFF
--- a/NetKAN/OldSchoolTurrets.netkan
+++ b/NetKAN/OldSchoolTurrets.netkan
@@ -10,7 +10,7 @@
         {
             "find"          : "OldSchoolTurrets",
             "install_to"    : "GameData",
-            "filter_regexp" : "(^|/)[\\s\\S]*\\.meta$"
+            "filter_regexp" : "(^|/).*\\.meta$"
         }
     ]
 }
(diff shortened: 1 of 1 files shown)


dbent:
Character classes `\s` is "whitespace" and `\S` is "non-whitespace" so `[\s\S]` is equivalent to "any character", or simply `.`.

Or in other words: `(^|/).*\\.meta$`

But looking at the download, I don't see any `.meta` files?

dbent:
Nevermind, they have a hidden attribute, curious.

plague006:
My mistake with the regex. I use a large body of text to make sure I'm only filtering what I mean to. I use \s because the body of text has linebreaks, and I failed to switch it back.

plague006:
Passes local, jenkins fail is because it's testing unrelated mods.

techman83:
@plague006 the build script compares the checked out commit to origin master. It will test all files that differ. I don't know if there is a way around that, but it's certainly not optimal.

plague006:
@techman83 Not that it gets us there, but the old jenkins was able to isolate the commits without taking the chain.

techman83:
@plague006 I'm not sure how, it's the same build script. Jenkins doesn't decide anything at all, it sets up some environment variables, launches the container and runs the script.

mgsdk:
@techman83 @plague006 I think there are some issues with the current script. I am in the process of doing a second pass of it, but it will most likely be a few days before I have a PR ready for a merge (I am away most of today and tomorrow).

plague006:
Tested locally again, merging.

PR #1934, opened 2015-07-16
License fixes

plague006:
A remake of, and closes #1831 because my gitfu is gitignored


plague006:
Jenkins' netkan creates .ckans with success, but install fails as expected due to multiple KSP versions. Each individual change passes local. Merging.

PR #1936, opened 2015-07-17
Alternative to #1916 aka "The Blue Pill"

plague006:
Resolves KSP-CKAN/CKAN-meta#706 in a non-fragile way at the expense of user resources (to the tune of 18 duplicate textures being loaded)


plague006:
Closes #1916

plague006:
Taking the blue pill to avoid further headaches on mod releases. Can/will reconsider the approach when the metadata toolset expands.

Passes local, passes Jenkins. Merging.

PR #1940, opened 2015-07-17
Update FuelTanksPlus.netkan dependencies

### LINKED ISSUE #1935: Fuel Tanks Plus dependecies, again.
I made a change to how the dependencies work for FTP. 

It needs one of either the Firespitter Core, or Interstellar Fuel Switch. Either will work, but Interstellar is more robust, and preferred. 

How should we update the dependencies to reflect this? If it can only have a single requirement, it would be better to use Interstellar, between the two. 

(ModuleManager is still required in any case).

plague006:
Closes #1935


plague006:
Passes local, passes Jenkins, just a depends change. Merging.

PR #1941, opened 2015-07-18
Add force v to VesselView.netkan (change of version scheme)

plague006:


plague006:
0.7 and 0.71 dropped the v.
Passes local + jenkins. Merging.

PR #1942, opened 2015-07-18
Add force_v to TweakScale.netkan

plague006:


plague006:
Inconsistent use of v in version.
Passes local + Jenkins. Merging.

PR #1946, opened 2015-07-19
Create RealPlume-StockConfigs.netkan

Felger:


Felger:
This is odd, says it failed, but the reason is it doesn't have a KSP installation to test it on...

It otherwise looks like it should pass the build.

NetKAN/RealPlume-StockConfigs.netkan
```diff
@@ -0,0 +1,32 @@
...
+    "spec_version"      : "v1.4",
+    "$kref"             : "#/ckan/github/KSP-RO/RealPlume-StockConfigs",
+    "ksp_version"       : "1.0",
+    "name"              : "Real Plume - Stock Configs",
+    "identifier"        : "RealPlume-StockConfigs",
+    "abstract"          : "A set of effects and basic configs that can be attached to any engine as appropriate.  Requires secondary configs to function correctly.",
+    "license"           : "CC-BY-NC-SA",
+    "depends" : [
+        { "name" : "ModuleManager" },
+        { "name" : "SmokeScreen" },
```
plague006: You can remove depends on SmokeScreen and ModuleManager. It will inherit the dependencies from RealPlume. The configs don't actually depend on those mods. If you change dependencies for RealPlume you'll only have to change them in that netkan instead of both.

NetKAN/RealPlume-StockConfigs.netkan
```diff
@@ -0,0 +1,32 @@
...
+        { "name" : "VenStockRevamp" }
+    ],
+    "provides"  : [ "RealPlumeConfigs" ],
+    "resources" : {
+        "homepage" : "https://github.com/KSP-RO/RealPlume/wiki"
+    },
+    "install" : [
+        {
+            "find"       : "RealPlume-Stock",
+            "install_to" : "GameData"
```
plague006: Currently you'll have GameData/RealPlume and GameData/RealPlume-Stock. Not a big deal but it might be confusing to users. If possible, consider `"install_to" :GameData/RealPlume"` -- Ultimately your choice though.

plague006:
Recommend added to RealPlume looks good (passes local)
Deletion of HotPlume looks good (no releases to CKAN-meta)

RealPlume-StockConfigs:
See line comments

plague006:
Passes local, jenkins is crazy. Merging with my thanks.

PR #1947, opened 2015-07-19
General fixes for the build script.

mgsdk:
I seem to have made some errors with the last iteration of the build script (I claim insanity at the time of writing). This is not yet ready for merging.


techman83:
I'm curious if we should be really testing all the netkans if one isn't found to test in the commit history. That might have made sense when we had a few dozen, but there are > 700.

The amount of compute we'd need to perform that operation and the likelihood of us getting a pass at the end of that is pretty low.

@mgsdk  I will probably build on this if that's ok.

techman83:
Oh and @mgsdk I missed it aswell, no harm done and it was most definitely an improvement :-)

mgsdk:
Feel free to build on your code base, rather we get an improvement soon. I
have done ideas I want to investigate later.

PR #1948, opened 2015-07-19
Extra Tweaks to Build Script

techman83:
I'll add some comments in line for discussion. Extends @mgsdk's work in #1947

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -9,7 +9,12 @@ LATEST_CKAN_URL="http://ckan-travis.s3.amazonaws.com/ckan.exe"
 LATEST_NETKAN_URL="http://ckan-travis.s3.amazonaws.com/netkan.exe"
 
 # Third party utilities.
-JQ_PATH="jq"
+JQ_PATH="./jq"
+
+# Return codes.
+EXIT_OK=0
+EXIT_FAILED_PROVE_STEP=1
+EXIT_FAILED_JSON_VALIDATION=2
 
 # ------------------------------------------------
 # Function for creating dummy KSP directories to
@@ -99,7 +104,7 @@ inject_metadata () {
     # Copy in the files to inject.
     for f in $1
     do
-        cp f CKAN-meta-master
+        cp $f CKAN-meta-master
     done
     
     # Recompress the archive.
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
```
techman83: @mgsdk I put this back, whilst we can explicitly throw errors it's nice also not to silently fail

build.sh
```diff
@@ -111,45 +139,79 @@ inject_metadata () {
...
-echo "Running basic sanity tests on metadata."
-echo "If these fail, then fix whatever is causing them first."
+if [ -n "$ghprbActualCommit" ]
+then
+    echo "Commit hash: $ghprbActualCommit"
+    export COMMIT_CHANGES="`git diff --diff-filter=AM --name-only --stat origin/master`"
+else
+    echo "No commit provided, skipping further tests."
+    exit $EXIT_OK
+fi
```
techman83: I'm not well versed enough with Jenkins to know if there is ever a scenario that we don't have a commit ID. We're actually just using it for an identifier right now.

build.sh
```diff
@@ -111,45 +139,79 @@ inject_metadata () {
...
+    rm -f master.tar.gz
 fi
 
-if [ "$COMMIT_CHANGES" = "" ]
+if [ -e "metadata.tar.gz" ]
 then
-    echo "No .netkan changes, skipping further tests."
-    exit 0
+    rm -f metadata.tar.gz
 fi
```
techman83: These are mainly for local testing, we'll be getting a clean container on every build. The checks mean we'll only try to remove things when they exist.

build.sh
```diff
@@ -111,45 +139,79 @@ inject_metadata () {
...
+    jsonlint -s -v $f
+    
+    if [ $? -ne 0 ]
+    then
+        echo "Failed to validate $f"
+        exit $EXIT_FAILED_JSON_VALIDATION
+    fi
+done
+echo ""
+
```
techman83: I moved the jsonlint above, it provides a more obvious message than we get from the metadata.t tests.

build.sh
```diff
@@ -174,35 +236,19 @@ else
     KSP_NAME=$ghprbActualCommit
 fi
 
-mono --debug ckan.exe ksp add ${KSP_NAME} "`pwd`/dummy_ksp"
-mono --debug ckan.exe ksp default ${KSP_NAME}
-
-echo Running ckan update
-mono --debug ckan.exe update
-
```
techman83: We're doing these per file as it is, I've seen it barf when we've tried to do it out here.

techman83:
It fails trying to validate 'metadata.t' but aside from that it should be all good. Have tested a number of scenarios so should have reasonable coverage now.

build.sh
```diff
@@ -27,15 +42,40 @@ create_dummy_ksp () {
...
-    if [ "$KSP_VERSION" == "0.90" ]
-    then
+    case $KSP_VERSION in
+    "0.90")
+        echo "Overiding '0.90' with '0.90.0'"
         KSP_VERSION="0.90.0"
-    fi
+        ;;
+    "1.0")
+        echo "Overiding '1.0.0' with '1.0.0'"
```
mgsdk: Small detail, but I guess the first "1.0.0" should be "1.0" (And while we are at the small details, overiding -> overriding)? I'm not well enough versed in bash scripting to know how switch is implemented, but looks to me reasonable.

mgsdk:
One thing I've never been to sure of is when we need to use `${var_name}` and when we can just use `$var_name`. It would be nice if we could get that cleared with this PR as well. Looks good to me otherwise, with the minor detail of `1.0.0 -> 1.0`.

mgsdk:
We might want to increment the default KSP version from 1.0.2 to 1.0.4 as well while we are at it?

build.sh
techman83: I really need to configure a spell checker pre-commit hook! Case statements use fairly simple matching so that won't match on 1.0.2. What we can probably do is set it to the default latest version.

techman83:
I'll look through the curly braces, that's mostly about coding style. But consistency is good, I'll get rid of them.

hakan42:
Honestly, I am on the other side of the fence with curly braces. Better be explicit in what you want so you don't fall into unexpected traps while concatenating variable values.

I tend to add curly braces whereever I happen to work on :smiling_imp:

techman83:
@hakan42 whenever I need to worry that much about it I generally stop using bash ;-)

techman83:
Webhook test

pjf:
1. This looks pretty good to me.
2. If it breaks anything, we can easily back it out. Tests are generally an area where we should have a higher tolerance for risk, especially since they let us lower the risks in our release code. :)

Merging with great thanks!

PR #1949, opened 2015-07-19
Update ProceduralFairings-AutoShroudRemover.netkan

Felger:
Installs wrong file.


plague006:
Passes local, merging with my thanks.

PR #1954, opened 2015-07-19
Change dependency in KSPInterstellarExtended to allow RealFuels as an alternative to InterstellarFuelSwitch

plague006:
Closes https://github.com/KSP-CKAN/CKAN-meta/pull/601 (whether merged or not)
Pinging @sswelm to find out if you're ok with this change or not. Of particular note: if users choose RealFuels, they cannot install InterstellarFS due to their conflict relationship.


sswelm:
Why would they be have to be in conflict? Seems to me RealsFuels MM script should simply first verify if  Interstellar Fuel Switch  or FireSpitter Fuelswitch partmodule is on a container before they force their own Partmodule on all containers. As long as Realfuels is willing to play nice, they should be compatible.

Therefore I do not agree with the proposed change

plague006:
Works for me.

PR #1955, opened 2015-07-19
RealPlume epoch bump

plague006:
To break from its RO past.


plague006:
Jenkins fails because ksp_version is 1.0 (which isn't an accepted KSP release)

plague006:
Passes local, merging.

techman83:
New script will take care of that translation.

plague006:
:+1:

PR #1956, opened 2015-07-20
Add 1.0.4 override for new version of RoutineMissionManager

plague006:
If there's a way to just have a range of versions, I'd love to know about it.


dbent:
> If there's a way to just have a range of versions, I'd love to know about it.

`"version" : [ ">=012", "<=013" ]`

plague006:
Thank you.

plague006:
Tested locally, merging.

PR #1963, opened 2015-07-20
1.0.4 override for CameraTools

plague006:


plague006:
Passes local + jenkins. Meging.

PR #1968, opened 2015-07-20
Add ALCOR suite of mods

plague006:
Closes #1965, closes #1966 , closes #1967 

Adding the ALCOR IVA as a `depends` to ALCOR because we don't index the VNG plugin, and without either ALCOR IVA or the VNG plugin parts of the capsule will be broken.


plague006:
Passes local, passes jenkins erroneously. Merging.

PR #1969, opened 2015-07-21
Add MK12PodIVAReplacementbyASET.netkan

plague006:
Closes #1967


techman83:
Another failure is success!

```
Module ASETProps required, but not listed in index, or not available for your version of KSP
```

plague006:
Passes local because I have `ASETProps` installed. Jenkins passing is erroneous. Merging.

PR #1972, opened 2015-07-21
Minor change to Contract Configurator netkan

jrossignol:


plague006:
Hey nightingale. Love your work / what Contract Config has brought to the scene.

I noticed in your ksp thread OP you have WaypointManagerand CoherentContracts as recommendations. Did you want to reflect that in the CKAN entry?

jrossignol:
Good point, why didn't I think of that?  PR updated. :wink:

plague006:
There's actually a `recommends` field if you want to use that for some of them. `recommends` are "ticked" by default whereas `suggests` are on a later screen and unticked by default.

jrossignol:
Hmmm...  for these I think I prefer `suggests`, although I wasn't aware of the `recommends` functionality - thanks for enlightening me!

plague006:
Sounds good.

Passes local, passes Jenkins, merging with my thanks.

PR #1975, opened 2015-07-21
WarpUnlocker Addition to ckan

peachoftree:


plague006:
License should be `public-domain`.
The download link should be `http://dl.dropboxusercontent.com/s/0pgkkhfk02g3vn3/WarpUnlocker_v1.zip` (in a browser it shouldn't open a page, it should download directly)

And most importantly the file type should be `.ckan` and it should be PRed into https://github.com/KSP-CKAN/CKAN-meta

NetKANs are for automated mods that get pulled from KerbalStuff, and GitHub, the mod you want to index is hosted on dropbox, which we can't scrape for new releases or for information about the mod.

peachoftree:
my bad, I'll Try posting it there.

PR #1977, opened 2015-07-22
1.0.4 overrides for linuxgurugamer's mods

plague006:
Tested locally


plague006:
Colour me surprised that Jenkins handled that like a pro. Brava!

Issue #1981, opened 2015-07-22
"FuelSwitch to every tank-default" conflicts with "Interstellar Fuel Switch Core"


NecroBones:
This is causing dependency problems for mods that rely on the "Interstellar Fuel Switch Core" (which is just the DLL). Fuel Tanks Plus has a user reporting this.

Basically the "FuelSwitch to every tank-default" is a config that adds fuel switching to stock tanks, via the Firespitter fuel switcher. There's no problem with installing the Firespitter Core and the Interstellar Core together, so this "conflicts" rule is breaking things for anyone that wants to use "FuelSwitch to every tank-default" in conjunction with anything that relies on the Interstellar Core.

plague006:
Would you happen to know it `FuelSwitchtoeverytank-default` is also compatible with `StockFuelSwitch`?

NecroBones:
That's a good question. Those are the ones I think are more likely to have issues together. I may pull some copies and take a look.

NecroBones:
By looking at them, I believe they might mildly conflict with each other. They're both adding FSfuelSwitch modules for all fuel tanks (including stock). They do things slightly differently, but with the same end-goal. Here's the rub: Porkjet's StockFuelSwitch checks to see if the module already exists on those tanks. Micha030201's FuelSwitchtoeverytank-default does not, and so depending on the execution order, there could be double modules injected on tanks. And it'll add fuel-switching modules to mod-tanks that already come with them. 

Neither of them check for modules from InterstellarFuelSwitch, so that's an improvement I'd like to see them both add. But these can't be handled via dependency rules, since it's on a per-tank basis.  

TL;DR: Since FuelSwitchtoeverytank-default and StockFuelSwitch attempt to do exactly the same thing, I think it's OK for them to have a "conflicts" rule with each other.

plague006:
:+1: 
Thanks for looking into it.

PR #1982, opened 2015-07-22
Remove conflict with InterstellarFuelSwitch-Core

plague006:
According to #1981, these mods shouldn't conflict.


plague006:
Tested locally, passes jenkins. Merging.

PR #1984, opened 2015-07-23
Update RealPlume-StockConfigs.netkan

Felger:
Add newly supported modpacks to RealPlume-StockConfigs suggests.

### DIFF
--- a/NetKAN/RealPlume-StockConfigs.netkan
+++ b/NetKAN/RealPlume-StockConfigs.netkan
@@ -11,10 +11,14 @@
     ],
     "suggests" : [
         { "name" : "ABLaunchers" },
+        { "name" : "AtomicAge" },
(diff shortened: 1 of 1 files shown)


dbent:
Tested locally, merging.

dbent:
@Felger, would you be interested in using self-hosted metanetkans, like [HotSpot](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/HotSpot.netkan) or [PlaneMode](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/PlaneMode.netkan)?

Felger:
Actually, that's a really good idea, primarily for those netkans that change frequently like Realism Overhaul and RealPlume.  Thanks @dbent!

PR #1987, opened 2015-07-24
Telemachus should have a leading 'v'.

pjf:
Otherwise v1.4.26.0 gets loaded, which is old


plague006:
Merging with my thanks.

PR #1999, opened 2015-07-27
Add .netkan for ClampsBeGone

iPeer:


plague006:
Be gone, clamps!
Passes local and jenkins is just crazy, merging with my thanks.

PR #2000, opened 2015-07-28
Add KWRocketry-CommunityFixes and interstage

plague006:
And add conflict to netkan of KWRocketry (Community fixes are to be applied to non-netkan version of KWRocketry).

@dbent would love a second set of eyes on this.

Thanks to @TeddyDD whose metadata I basically copy + pasted.
Related issue: https://github.com/KSP-CKAN/CKAN-meta/pull/724

This adds netkans for `CommunityFixes` and `interstage` in case they need updates. This also adds a conflict to `KWRocketry` so that any version other than `2.7.0-community` (which isn't netkan-generated) will have a conflict to the fixes.

### DIFF
--- a/NetKAN/KWRocketry-CommunityFixes-interstage.netkan
+++ b/NetKAN/KWRocketry-CommunityFixes-interstage.netkan
@@ -1,0 +1,27 @@
+{
+    "spec_version"  : "v1.2",
+    "identifier"    : "KWRocketry-CommunityFixes-interstage",
+    "$kref"         : "#/ckan/kerbalstuff/1036",
+    "$vref"         : "#/ckan/ksp-avc",
+    "license"       : "GPL-3.0",
+    "depends": [
+        {
+            "name": "KWRocketry-CommunityFixes",
+            "version": "0.1.0"
+        }
+    ],
+    "install": [
+        {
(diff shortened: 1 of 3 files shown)


TeddyDD:
>This adds netkans for CommunityFixes and interstage in case they need updates. This also adds a conflict to KWRocketry so that any version other than 2.7.0-community (which isn't netkan-generated) will have a conflict to the fixes.

Wow I didn't know about this behaviour :O It seems I should learn C# and at least read code of this thing :)

plague006:
Basically netkan gets all its information from `$kref` (in this case KW's KerbalStuff page). `2.7.0-community` isn't included in KW's KerbalStuff release list so it's outside of netkan's "sphere of influence". By adding the `conflicts` relationship to the netkan it will affect only official KW releases, giving us the behaviour we want (`CommunityFixes` only being compatible with the community version of KW).

dbent:
So this is basically so that we say "KWRocektry works with KSP 1.0.4 if this other mod is installed"? Looks fine to me then.

Expressing something like this may be useful in a [theoretical `condition` property in relationships](https://github.com/KSP-CKAN/CKAN/issues/1104#issuecomment-111377723).

plague006:
I'm always in favour of an expanded toolset. :smile: 
Thanks for having a look.

PR #2010, opened 2015-07-30
PWings updated

Crzyrndm:


plague006:
Needs a `ksp_version`. `1.0` will cover all `1.0.X` releases. If it's still compatible with `0.90` then you can use `ksp_version_min` and `ksp_version_max` to set a range.

Crzyrndm:
using .version file (vref) for KSP and mod version?

https://github.com/Crzyrndm/ProceduralWings/blob/master/GameData/ProceduralDynamics/PWings.version

plague006:
You're totally right, my mistake. That's what I get for looking at PRs at 4am. =)

Passes local, passes Jenkins. Merging with my thanks.

PR #1243, opened 2015-05-13
Rocketwatch: Fix install stanza to reflect the change in the filename.

mgsdk:
The zip file contains the version of the mod, thus netkan can't autoindex it when it updates.

PR #2013, opened 2015-07-30
Tests: v1.12+ required for `Ships` with path.

pjf:
Part of #1257 and #1243.

Can be happily merged even before v1.12 has been released, since we can start writing metadata for v1.12 now. :)

### DIFF
--- a/t/metadata.t
+++ b/t/metadata.t
@@ -82,6 +82,13 @@ foreach my $shortname (sort keys %files) {
             );
         }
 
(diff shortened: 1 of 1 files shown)


plague006:
LGTM (I should've looked at this sooner, didn't realize it's essentially a copy+paste of the other if statements)

techman83:
@plague006 I missed this one sorry! Indeed looks correct.

PR #2017, opened 2015-07-31
KerbalStuff PR catchup

plague006:


plague006:
Each passes local, Jenkins can't handle all this awesomeness though.

plague006:
Re-tested, all working today. Merging before any of them issue an update that breaks the netkans. =)

PR #2018, opened 2015-07-31
author request dependency change

Murdabenne:
recommends instead of depends for the IVA
http://forum.kerbalspaceprogram.com/threads/54925-1-0-4-ALCOR-Advanced-Landing-Capsule-for-Orbital-Rendezvous-by-ASET-%2821-07-2015%29/page162

### DIFF
--- a/NetKAN/ALCOR.netkan
+++ b/NetKAN/ALCOR.netkan
@@ -3,7 +3,7 @@
     "license"     : "CC-BY-NC-SA-3.0",
     "$kref": "#/ckan/kerbalstuff/1015",
     "identifier": "ALCOR",
-    "depends": [
+    "recommends": [
         { "name": "ALCORIVAPatch" }
     ],
     "install": [


plague006:
Per the author's OP in the mod thread:

>- If you wish to use the built-in Navigation, Dock port and Airlock lights, but do not wish to use the IVA features, you need to install the VNG plugin.

Removing the `depends` means we leave a path for users to install a broken mod. As much as I want to respect the wishes of mod authors our first duty is to users. If you want to provide a path where users aren't forces to install `ALCORIVAPatch`, you'll need to set up a module for `VNG Plugin` and have `VNG Plugin` and `ALCORIVAPatch` both `provide` a virtual package for `ALCOR` to `depend` on.

Related: #1968

Murdabenne:
I can probably get the other mod set up without too much problem.  But how do I write up an "either:or" dependency?

plague006:
You can create an OR situation by using `provides`. So for instance `ALCOR depends ALCOR-fix`. There's no mod with the identifier `ALCOR-fix` but we add a line to `ALCORIVAPatch` and `VNGPlugin` which `provides: [ "ALCOR-fix" ]`

As for needing 2 different netkans: we already have one for `ALCORIVAPatch` which would need the `provides` line added, and we would need a netkan/ckan for `VNGPlugin` that includes the `provides` line.

Murdabenne:
OK I will work on this, but my time is very limited until Tuesday(working several 12+ hour shifts on the cardiac unit until monday night)

Murdabenne:
Working on this today

plague006:
Closing this after 2 weeks of inactivity. If/when you make the required changes you can re-open this PR or submit a new one (as you prefer).

PR #2033, opened 2015-08-03
Add X77AssaultDrone.netkan, closes #2024

plague006:


plague006:
KerbalStuff mods:
- [ ] Compare the name of the mod, the mod author, and the folder it installs to, in order to make sure the `identifier` makes sense.
- [ ] Ensure the KerbalStuff page has a homepage set. If it doesn't google to try and find a KSP forum page. If you can't find one, netkan will assign the KerbalStuff page as the homepage.
- [ ] Ensure the KerbalStuff page has an abstract and it it doesn't, write one for it.
- [ ] Ensure we have the correct license, and that it's identical across .netkan, mod .zip, KerbalStuff page, mod's homepage, any other resources related to the mod.
- [ ] Is the $kref accurate? If you copy and pasted from the KerbalStuffBot PR, you should be set.
- [ ] Check for a .version file that we can use for `"$vref" : "#/ckan/ksp-avc"`.
- [ ] Check the install stanza for: files correctly installed, extraneous files filtered out.
- [ ] Using the KerbalStuff page, homepage, the mod's zip, and any other resources related to the mod to check for `conflicts`, `depends`, `recommends`, and `suggests` as appropriate.
- [ ] After everything is written, look over the .netkan and verify your `spec_version` is appropriate.
- [ ] Verify the file with `netkan.exe`
- [ ] Verify the file with a `CKAN install`
- [ ] Submit PR, check back on the PR later on for Jenkins feedback (pass/fail).

plague006:
Added BDA `suggests` since an assault drones could benefit from.. weapons.

PR #2037, opened 2015-08-03
Update KAINET.netkan to use github as kref

plague006:
Changes per mod author @Kerbas-ad-astra in #1548


plague006:
Passes local + jenkins. Merging.

PR #2042, opened 2015-08-04
Add BDarmory-common, add BDArmoryDMPAddon, modify BDArmory

plague006:
For testing, https://github.com/KSP-CKAN/CKAN-meta/archive/BDArmoryTest.zip is setup. Basically, it's setup to avoid end-users having to choose between mods. Instead, the dependency/provide relationships are setup such that if a user chooses to install `BDArmory` they automatically get the common files installed (unprompted) and the same if a user chooses the multiplayer addon. The drawback is that when `BDArmory-common` `provides` `BDArmory`, `BDArmory` has an erroneous upgrade checkbox. While it doesn't cause any errors or crashes, it's something we'll want to see resolved.

@dbent I like your input/verification on these more complex PRs, if you please. =)

### DIFF
--- a/NetKAN/BDArmory-common.netkan
+++ b/NetKAN/BDArmory-common.netkan
@@ -1,0 +1,32 @@
+{
+    "spec_version" : "v1.4",
+    "identifier"   : "BDArmory-common",
+    "name"         : "BahamutoD's Armory-common",
+    "abstract"     : "This is just a module that provides files BDArmory uses in common with other mods",
+    "resources"    : {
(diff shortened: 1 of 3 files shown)


plague006:
Jenkins passing is erroneous.

draeath:
Mod version 0.9.1 has since been released, FYI.

plague006:
Closing for now as `BDArmoryDMPAddon` isn't using a version scheme netkan can follow.

PR #2045, opened 2015-08-04
Add B9Procedural as recommend to Mk3HypersonicSystems

plague006:


plague006:
Passes local and Jenkins, merging.

PR #2046, opened 2015-08-04
TechManager is no longer a dependecy

JandCandO:


plague006:
Merging with my thanks.

PR #2050, opened 2015-08-06
Fix `ClampOTronAPAS` and `DockingTarget` both trying to install the same flags.

plague006:
Closes https://github.com/KSP-CKAN/CKAN-meta/issues/733


plague006:
Passes local, jenkins didn't even bother installing `FederalProductions-Flags` because he ain't about that life.

techman83:
#1948 will likely address that.

PR #2055, opened 2015-08-07
Add OrbitalDecay.netkan

plague006:
@Whitecat106 is ModuleManager a dependency? I noticed MM_StationKeepingModule.cfg and it seems like the mod needs it to apply its effects to stations, but if I'm wrong let me know please.

### DIFF
--- a/NetKAN/OrbitalDecay.netkan
+++ b/NetKAN/OrbitalDecay.netkan
@@ -1,0 +1,15 @@
+{
+    "spec_version": "v1.4",
+    "$kref": "#/ckan/kerbalstuff/1059",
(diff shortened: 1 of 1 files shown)


plague006:
@techman83 this failure isn't build-related, correct? Jenkins just failed to 'git' its crap together?

techman83:
@plague006 Indeed. Likely some transient network issue, I think I recall GitHub tweeting some things around the time of this. Re-ran the build and it worked correctly.

plague006:
@techman83 :+1: 
Passes local and Jenkins, merging.

PR #2065, opened 2015-08-09
BDArmory.netkan version bump from 0.8.1 to 0.9.1

draeath:
9.0 and 9.1 were released

### DIFF
--- a/NetKAN/BDArmory.netkan
+++ b/NetKAN/BDArmory.netkan
@@ -8,7 +8,7 @@
 	"license"	   : "CC-BY-SA-2.0",
     "x_netkan_override" : [
 		{
-			"version" : "v0.8.3",
+			"version" : "v0.9.1",
 			"delete" : [ "ksp_version" ],
 			"override" : {
 				"ksp_version_min" : "1.0.2",


draeath:
Unsure if this is required or if this is suppose to happen automatically in some fashion. Apologies if the latter.

plague006:
That's not the current version of the mod, that's an override for that particular version because it was deemed compatible with `ksp_version` 1.0.4. Updates to mods with .netkans should happen automatically. In the case of `0.9.0` and `0.9.1` we have them both indexed [here](https://github.com/KSP-CKAN/CKAN-meta/blob/master/BDArmory/BDArmory-0.9.0.ckan) and [here](https://github.com/KSP-CKAN/CKAN-meta/blob/master/BDArmory/BDArmory-0.9.1.ckan). If you look at `BDArmory`'s  [collection of .ckan files](https://github.com/KSP-CKAN/CKAN-meta/tree/master/BDArmory) you'll actually notice `BDArmory-v0.8.3.ckan` is the version of `BDArmory` that CKAN will try to install rather than ` BDArmory-0.9.1.ckan` because `v0.8.3` is greater than `0.9.1`. The fix is to force a v on the current and all future versions of `BDArmory`.

Thanks for bringing the situation to my attention.

PR #2068, opened 2015-08-10
HullcamVDS update broke CKAN and new dependencies

draeath:
0.50 archive contains files with colons in their names, which causes CKAN to crash upon installation. Also, the mod added a dependency of Module Manager - added exclusion of the file and added the dependency spec.

See https://github.com/KSP-CKAN/CKAN/issues/1364

### DIFF
--- a/NetKAN/HullcamVDS.netkan
+++ b/NetKAN/HullcamVDS.netkan
@@ -11,7 +11,17 @@
         {
             "file"       : "HullCameraVDS",
             "install_to" : "GameData"
+            "filter": [
+                "Thumbs.db",
+                "Thumbs.db:encryptable"
+            ],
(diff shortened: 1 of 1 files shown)


NetKAN/HullcamVDS.netkan
```diff
@@ -11,8 +11,18 @@
         {
             "file"       : "HullCameraVDS",
             "install_to" : "GameData"
```
techman83: @draeath Build failed due to missing comma here.

draeath: Thanks! Now to figure out how to amend it.

draeath:
Stuff it, I'll just start over :)

PR #2069, opened 2015-08-10
Mod update requires new dependency and file filter

draeath:
See also https://github.com/KSP-CKAN/CKAN/issues/1364

### DIFF
--- a/NetKAN/HullcamVDS.netkan
+++ b/NetKAN/HullcamVDS.netkan
@@ -10,8 +10,18 @@
     "install" : [
         {
             "file"       : "HullCameraVDS",
-            "install_to" : "GameData"
+            "install_to" : "GameData",
+            "filter": [
+                "Thumbs.db",
+                "Thumbs.db:encryptable"
+            ],
+            "filter_regexp": [
+                "ModuleManager.*dll$"
+            ]
         }
+    ],
+    "depends": [
+        { "name": "ModuleManager" }
     ],
     "x_netkan_override" : [
 		{


techman83:
@draeath you can push extra commits to your branch. I think I can even make this PR close the other one for you. LGTM :)

techman83:
The bot will process that in 1.5 hours and it'll end up in CKAN-meta.

draeath:
@techman83 I tried that, but I must not have done it right as it didn't seem to affect the pull request. Was I not supposed to commit against master again? I'm not really well versed with github, to be honest (or anything beyond basic file handling in git in general) so I'm pretty sure I stuffed it up. I zapped the fork and started over though, as you saw, and closed out the bad PR.

Thanks for the eyes! :D

techman83:
If you did this from the GitHub interface that might be tricky (I only use the cli) - but any commits added to the 'patch-1' branch would have ended up on this pr.

You're welcome, the Console output is availalable in jenkins, but it might have been a little cryptic still :) 

http://ci.ksp-ckan.org/job/NetKAN/296/console

NetKAN/HullcamVDS.netkan
```diff
@@ -10,9 +10,19 @@
     "install" : [
         {
             "file"       : "HullCameraVDS",
-            "install_to" : "GameData"
+            "install_to" : "GameData",
+            "filter": [
+                "Thumbs.db",
+                "Thumbs.db:encryptable"
+            ],
+            "filter_regexp": [
```
plague006: A filter applies to a `file` or `find`. This particular filter has nothing to affect because it's sitting on its own. Additionally we don't need to filter anything outside of the scope of our `file` or `find`. Good spot on the `Thumbs.db` files though. :+1:

PR #2075, opened 2015-08-11
unifying author name

draeath:
Adding a tag to unify the author name field with his other mods


plague006:
as someone else who suffers from CDO (that's OCD with the letters in the right order), merging with my thanks.

PR #2082, opened 2015-08-11
Add ProceduralProbes.netkan

plague006:


plague006:
The `name` from the KS page is `[1.0.4]ProceduralProbes` which would look odd in the CKAN modlist.

mgsdk:
Passes local testing, Jenkins talks about TACLS for no apparent reason.

techman83:
It does a diff from origin/master
```
git diff --diff-filter=AM --name-only --stat origin/master
```
'TACLSMining' was different in the branch than in origin, which was updated around the same time.

It managed to test both of them correctly, so win!

PR #2093, opened 2015-08-11
Mk2Expansion.netkan

SuicidalInsanity:
Update for M2X v1.5


plague006:
`spec_version` is the specification of the CKAN language you want the interpreter to use (this doesn't change unless you're using features not present in older versions). The mod version is taken from the KerbalStuff data (doesn't need to be manually updated).

PR #2094, opened 2015-08-12
Create Skylab.netkan

Felger:


plague006:
You shouldn't need `x_netkan_epoch` -- that's for existing mods whose authors drastically alter their versioning scheme/break their version order

plague006:
Passes local and Jenkins. Merging with my thanks.

PR #2095, opened 2015-08-12
Create RN-SalyutStations-SoyuzFerries.netkan

Felger:


plague006:
You shouldn't need `x_netkan_epoch` -- that's for existing mods whose authors drastically alter their versioning scheme/break their version order

plague006:
Passes local. Jenkins... poor sweet Jenkins keeps downloading the 100MB .zip over and over and doesn't even install from it. =(
Merging with my thanks.

techman83:
KSP-CKAN/CKAN#1330 Should alleviate this. We can also look at centralising a download cache that can be passed into the docker container.

PR #2098, opened 2015-08-12
Add comment explaining why ships aren't installed for these mods

plague006:


plague006:
I kinda suck at rebasing.

plague006:
But.. everything passes, so I'm merging.

PR #2108, opened 2015-08-13
Add RhatssPlanetpack.netkan

plague006:


plague006:
Per https://github.com/KSP-CKAN/CKAN-meta/issues/588, not including a `provides: PlanetPack`

PR #2116, opened 2015-08-17
Test1 for Jenkins.

plague006:


plague006:
Created this to test #2115
This is a copy of the current `AutomatedScreenshots.netkan` with the file name + identifier changed. It gives the exact same error at the exact same character offset as #2115. Considering `AutomatedScreenshots.netkan` has been running without issue and generated `AutomatedScreenshots-0.1.0.ckan` , `AutomatedScreenshots-0.2.0.ckan` , and `AutomatedScreenshots-0.3.0.ckan`

techman83:
I actually can't replicate this issue locally! I'll pull up the container and see if I can make it work.

PR #2118, opened 2015-08-17
Add @tehcman83's fixes for ksp_ver 0.23 and 0.25 from ckan-meta to netkan's build script

plague006:
I wish there were a way to know if this ever comes into play.


techman83:
LGTM and Thanks!!

PR #2122, opened 2015-08-18
Add lxnRTlxnResourceTransfer.netkan

plague006:


Dazpoet:
The repository scraped is pretty darn lol, that should probably be hardcoded.

PR #2126, opened 2015-08-18
Add FASTCORPRoadsandBridges.netkan

plague006:


techman83:
```
#   Failed test 'FASTCORPRoadsandBridges - spec_version v1.4+ required for install with 'find''
```

techman83:
Jenkins happier with that :+1:

PR #2127, opened 2015-08-18
Update my mods

Olympic1:
This is the first time I add/update a netkan file, but I followed the wiki to do this.

So hope this is correct


plague006:
For `CargoTransportationSolutions.netkan`
Less is more: Keep the `depends` and `suggests` but drop the info that netkan scrapes from KerbalStuff (`name`, `abstract`, `author`, `ksp_version`, `resources`). The reason we don't want extra information hardcoded is that if you change things on the KerbalStuff page they won't be reflected in CKAN (of extra importance is not hardcoding `ksp_version` since it'll version-lock your mod until the netkan itself is updated rather than when you upload a new version on kerbalstuff and update the information there. The `filter` in the install stanza is my personal mistake, I use a template for adding mods and I forgot to remove this line.

For `LTechUnofficial.netkan` largely the same as `CargoTransportationSolutions` with the additional note that I see you want to add a `$vref` but I don't see a .version file in the .zip.

For `Olympic1ARPIcons.netkan` I'm rushing this IRL because I'm heading out: I see that in the github page you mention "(Install at least one of these mods to see the icons)". Does that mean your mod breaks without access to at least one of those?

I get the sense that for all 3 mods you used the GitHub [section](https://github.com/KSP-CKAN/CKAN/wiki/Adding-a-mod-to-the-CKAN#github-guide) of the wiki guide while the KerbalStuff [section](https://github.com/KSP-CKAN/CKAN/wiki/Adding-a-mod-to-the-CKAN#kerbalstuff-guide) is more appropriate for CargoTransport and LTech.

Olympic1:
Thx for the feedback, I'll fix it.

As for the icons: my mod doesn't break if the recommended mods aren't installed but otherwise the icons are useless. I'll probably add a `:NEEDS[MOD]` check so they only apply if the respective mod is installed. (I'll need to discuss this with TriggerAu)

plague006:
`spec_version` needs to be `v1.4` for each.

`spec_version` is the version of the CKAN spec required to use the commands in the metadata. In these cases `find` requires `v1.4` to work.

Olympic1:
I fixed the `spec_version` and removed the ARP Icons file as it gave some trouble

plague006:
I would happily merge this as is, but it looks like you deleted `Olympic1ARPIcons.netkan` so it would only change `CargoTransportationSolutions.netkan and `LTechUnofficial.netkan`.

As for adding the `spec_version` for different commands I like the suggestion. In order to find the right place to add it, I'm curious how you found out about it? The `find` command isn't anywhere in the wiki entry.

Olympic1:
I'm planning to change ARP Icons, so I'll wait to add the current version to CKAN to after the update.

As for the `find` command, someone said that if your identifier is not the same as the name of the mod folder, you should use the `find` command to search for your custom foldername instead of the `file` command. So I used that command without knowing that you then also have to use `"spec_version" : "v1.4"`

I only find out recently that you still could use the `file` command to install foldernames that aren't the same as the identifier.

plague006:
Oki thanks for the background and thanks for the contributions, merging with my thanks.

PR #2132, opened 2015-08-19
Update HotRockets.netkan

politas:
HotRockets conflicts with RealPlume. They're doing similar things.


Dazpoet:
This is already in the realplume metadata so I'm not sure how necessary this is, but hey consistency consistency consistency! Merging with thanks :)

politas:
I raised the commits because the CKAN client happily let me select both mods, then failed during installation.

PR #2135, opened 2015-08-19
Add KduffersMultiwheels.netkan

plague006:


Dazpoet:
I'm confused as to why some models are dds and some not but that's nothing to do with this so LGTM :)

PR #2136, opened 2015-08-19
Add EditorTime.netkan

plague006:
On hold awaiting @magico13's final word on the correct license (GPL3 in KerbalStuff page and KSP thread but ARR license in the .zip).


magico13:
I must have copied the wrong license file into the zip. Correct one should be GPL3. Sorry about that!

Dazpoet:
Correct license as per the author above, Jenkins approves as do I :)

PR #2151, opened 2015-08-21
Transform version to exclude V and v, and force epoch to deal with version inconsistencies.

plague006:
Duplicate of and closes #1943. Redone to help Jenkins out.


Dazpoet:
aww yiss!

PR #2155, opened 2015-08-22
Edit FAR Version

dbent:
Strips leading `v` and any codenames from the version number.

Test with all FAR versions currently on Kerbal Stuff:

far-versions

### DIFF
--- a/NetKAN/FerramAerospaceResearch.netkan
+++ b/NetKAN/FerramAerospaceResearch.netkan
@@ -3,7 +3,8 @@
     "identifier"     : "FerramAerospaceResearch",
     "$kref"          : "#/ckan/kerbalstuff/52",
     "license"        : "GPL-3.0",
-    "x_netkan_epoch" : "2",
+    "x_netkan_epoch" : "3",
+    "x_netkan_version_edit": "^[vV]?(?<version>\d+(?:\.\d+){0,3}[A-Za-z]?)(?:_[A-Za-z0-9]+)?$",
     "release_status" : "stable",
     "resources" : {
         "repository" : "https://github.com/ferram4/Ferram-Aerospace-Research"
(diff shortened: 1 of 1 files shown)


dbent:
- `^` Match beginning of string
- `[vV]?` Match either an uppercase or lowercase V 0 or 1 times. **Leading V**
- `(?<version>...)` Capture the matching string as the `version` group **Actual Version String**
  - `\d+` Match a number 1 or more times **Major Version Number**
  - `(?:...)` Do not capture the matching string
    - `\.` Match the literal period charater
    - `\d+` Match a number 1 or more times
  - `{0,3}` Match the preceding group exactly 0, 1, 2, or 3 times. **Minor, Patch, and Build Version Numbers**
  - `[A-Za-z]?` Match any uppercase or lowercase letters exactly 0 or 1 times **Trailing alphabetic build number**
  - `(?:...)` Do not capture the matching string **Codename**
    - `_` Match the literal character underscore
    - `[A-Za-z0-9]+` Match any alphanumeric character 1 or more times.
  - `?` Match the preceding group exactly 0 or 1 times.
- `$` Match end of string.

Dazpoet:
I don't know regex so even with this amazing description I'll happily leave this for review by someone who does. Seems utterly amazing though! :+1:

pjf:
Right now we have:

`^[vV]?(?<version>\d+(?:\.\d+){0,3}[A-Za-z]?)(?:_[A-Za-z0-9]+)?$`

I'd be inclined towards the shorter:

`^[vV]?(?<version>\d[\d.]*[A-Za-z]?)`

Which allows for the version to consist of any number of digits and dots (so we can allow v1.0 or v1.2.3.4.5.6.7) which must start with a digit, optionally followed by a letter. (Any underscores or additional elements are ignored, so we don't need to mention them.)

Having said that, I'm also happy with the regexp as-is. I do have one big question, though... What happens when our regexp fails? Do we refuse to index the file? Do we leave the version alone? Do we do something else?

pjf:
And to answer my own question, if our regexp fails, then we throw an exception and stop processing that file. :)

pjf:
Merging with great thanks. :)

PR #2160, opened 2015-08-23
Update SETI-CommunityTechTree.netkan

Y3mo:
Introduced recommended and suggested mods, since many people do not seem to look at the forum thread at all...

Thank you very much!

PS: I m the author of this mod.

### DIFF
--- a/NetKAN/SETI-CommunityTechTree.netkan
+++ b/NetKAN/SETI-CommunityTechTree.netkan
@@ -4,7 +4,22 @@
     "$vref" : "#/ckan/ksp-avc",
     "identifier": "SETI-CommunityTechTree",
     "x_netkan_license_ok" : true,
-    "depends" : [ { "name" : "CommunityTechTree" }, { "name" : "ModuleManager" } ],
+    "depends" : [ 
+        { "name": "CommunityTechTree" },
+		{ "name": "KSP-AVC"	},
+        { "name": "ModuleManager" } 
+    ],
+    "recommends": [
+		{ "name": "AdjustableLandingGear" },
+		{ "name": "CustomBarnKit" },
+		{ "name": "Mk3MiniExpansionPack" },
+		{ "name": "StockBugFixModules" },
+		{ "name": "TakeCommand"	},
+		{ "name": "VenStockRevamp" }
+    ],
+	"suggests": [
+		{ "name": "SETI-Contracts" }
+	],
     "install" : [ { "file" : "SETIctt", "install_to" : "GameData" } ],
     "x_netkan_override": [
         {


plague006:
Tests fine locally, it fails because Jenkins used 1.0.0 as the ksp_version and there's no version of CTT compatible with 1.0.0.

With that said, is SETI-CTT actually dependent on KSP-AVC? I'm able to run SETI-CTT without installing KSP-AVC.

Y3mo:
I m not sure I understand.

Does that mean that I can not update the SETIctt netkan file?

How can I remedy the problem or what needs to be done?

While KSP-AVC is not technically necessary, I consider it necessary to reduce support.

plague006:
Sorry, I lost track of this PR the last couple days.
If it's not technically necessary, I'm sorry but we won't force users to install it. It can happily be added as a `recommends` which will install by default unless a user actively unticks the checkbox. As an aside for my own education, what does KSP-AVC offer that the mini plugin doesn't?

Y3mo:
Lots of mods only provide a .version file (for KSP-AVC), but not the mini plugin.
The problem is, that
1. SETI has interactions with lots of other mods
2. Many users who use ckan still install some mods manually
3. CKAN usage decreases the incentive to check the forums

So users tend to rely on ckan to keep the mods up to date, but they may forget their manually installed mods, or ckan does not have the newest version for some time (several days).

Especially for a mod like SETI with all the mod interactions, setting KSP-AVC as a dependency reduces possible support issues, which are totally avoidable.

You are not forcing users to install it. I, as the mod author, am forcing users to install it, if they install via ckan.

PS: Quill18 made a youtube series, which used SETI. He did not read the install instructions in the forum thread and did not install the CTT for many episodes. That resulted in a lot of headaches and a very wrong impression of SETI (totally bugged). Since then I m trying to minimize totally avoidable issues, eg like problems arising from the usage of SETI in combination with outdated mods (which provide a .version file for KSP-AVC, but not the KSP-AVC mini plugin).

Olympic1:
> So users tend to rely on ckan to keep the mods up to date, but they may forget their manually installed mods, or ckan does not have the newest version for some time (several days).

If a mod is hosted on KerbalStuff or GitHub, it automatically adds new releases to CKAN within the hour. But when it's hosted on Curse or Dropbox, the mod dev must add a new `.ckan` file manually.

I understand your dependency on KSP-AVC

plague006:
You updated SETI-CTT's OP to reflect to all users of the mod that KSP-AVC is a requirement and that's good enough for me. I just want equal treatment for all users. Merging with my thanks.

PR #2163, opened 2015-08-24
Bump KerbinSide epoch

pjf:
Somewhere in the past KerbinSide made a `v0.41.1` release, but all its other releases didn't have a trailing 'v'. It's now up to `1.0.3`, but clients on the `v0.41.1` release won't auto-upgrade to that.

This change bumps the KerbinSide epoch so that existing users will get the upgrades automatically, without having to uninstall and reinstall.

Thanks to Blu3wolf for helping me spot this.

### DIFF
--- a/NetKAN/KerbinSide.netkan
+++ b/NetKAN/KerbinSide.netkan
@@ -1,9 +1,10 @@
 {
-	"spec_version" : 1,
-	"identifier"   : "KerbinSide",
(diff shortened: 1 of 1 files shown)


Dazpoet:
Needs a comma on line 6

pjf:
D'oh

pjf:
Re-pushed with corrections. :)

PR #2164, opened 2015-08-24
Display version info about ckan/netkan

techman83:
Should prevent future confusion like #1415


techman83:
Does what it says on the tin :-)
```
Fetching latest ckan.exe
1248 [1] WARN CKAN.KSPManager (null) - 7f34b7dadea974cc5375c68adcf188057d88bd32 at /home/jenkins/workspace/PR builder - NetKAN metadata/NetKAN/dummy_ksp is not a vaild install
1401 [1] WARN CKAN.KSPManager (null) - Auto-start instance was invalid: Exception of type 'CKAN.InvalidKSPInstanceKraken' was thrown.
v1.14.0-2-g6d4e975 (beta)
Fetching latest netkan.exe
v1.14.0-2-g6d4e975 (beta)
```

PR #2165, opened 2015-08-24
Removed Dependancy on ModuleRCSFX

sswelm:


plague006:
And as quickly as it appeared, it was gone.

sswelm:
Yes, the dependency was giving be trouble

PR #2167, opened 2015-08-24
Add SixSeatMk3cockpit.netkan

plague006:
A few more lines in this netkan and it would be larger than the mod itself. =)


Dazpoet:
Oops! `#   Failed test '6seatMk3cockpit: CKAN identifers must consist only of letters, numbers, and dashes, and must start with a letter.'`

plague006:
Done

PR #2172, opened 2015-08-25
Community Tech Tree is not required or adds any benefit to KSPI

sswelm:
This should have been done some time ago

### DIFF
--- a/NetKAN/KSPInterstellarExtended.netkan
+++ b/NetKAN/KSPInterstellarExtended.netkan
@@ -21,7 +21,6 @@
         { "name"         : "FilterExtensions" }
     ],
     "suggests": [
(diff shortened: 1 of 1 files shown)


Olympic1:
Actually KSPI is dependant on CTT as it uses CTT node names to define its parts to a technode

sswelm:
Well not directly, it uses the same node yes, but it can operate completely without CTT  being installed as it defines the nodes it uses when CTT is not installed. Many people that install KSPI-E by CKAN complain about empty nodes which would not have been visible if they only installed KSPI-E manually. CTT It's simply not need and worse, annoy people with empty nodes which confuses people and start complaining about.

plague006:
Merging with my thanks. Less bug/issue reports for all yay!

PR #2174, opened 2015-08-26
Scatterer: force leading 'v'.

pjf:
Some older versions of Scatterer had a leading 'v'. New ones don't.


plague006:
Merging with my thanks.

PR #2175, opened 2015-08-26
Use ksp_version_max instead of ksp_version_min

plague006:
Mirror PR of https://github.com/KSP-CKAN/CKAN-meta/pull/749

Most recent case in the netkan repo is https://github.com/KSP-CKAN/NetKAN/pull/2160

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -307,7 +307,7 @@ do
     
     # Extract identifier and KSP version.
     CURRENT_IDENTIFIER=$($JQ_PATH '.identifier' $f)
-    CURRENT_KSP_VERSION=$($JQ_PATH 'if .ksp_version then .ksp_version else .ksp_version_min end' $f)
+    CURRENT_KSP_VERSION=$($JQ_PATH 'if .ksp_version then .ksp_version else .ksp_version_max end' $f)
     
     # Strip "'s.
     CURRENT_IDENTIFIER=${CURRENT_IDENTIFIER//'"'}


plague006:
In cases where a ksp_ver_min is set but no ksp_ver_max I'd rather it *not* pass Jenkins so that we can address it manually as part of https://github.com/KSP-CKAN/CKAN-meta/issues/623

plague006:
So there's 2 sides to this:

1. We don't want mods using their lowest compatible version as the grounds for their Jenkins test. In the case of #748 it's failing not because there are any errors in the metadata, but rather it's failing because a very old version of a dependent mod is no longer available. We should be using the highest compatible ksp version for testing to assure that it works now. (There's probably a discussion to be had about testing both min and max, but IMO that's a much more complicated issue about whether "compatible" means that it works with the ksp_version or whether it means CKAN can properly install it for that ksp_version).

2. We don't want mods passing Jenkins without a ksp_version compatibility cap unless they're explicitly set to ksp_version `any` (ex. flag-only mods). As a convention we tend to use ksp_version_max `1.0.99` when there's nothing suspected that will break a mod as long as no major changes occur, however 1.1 will represent a very major/fundamental change. While Squad will undoubtedly do their best to maintain mod compatibility, no mod should be assumed compatible until it's tested.

techman83:
@plague006 - awesome! LGTM :+1:

PR #2180, opened 2015-08-26
Add RemoteTechXF and RemoteTech-common

plague006:
For testing, https://github.com/KSP-CKAN/CKAN-meta/archive/RemoteTechXF.zip is setup. Basically, it's setup to avoid end-users having to choose between mods. Instead, the dependency/provide relationships are setup such that if a user chooses to install `RemoteTech` they automatically get the common files installed (unprompted) and the same if a user chooses `RemoteTechXF`. The drawback is that when `RemoteTech-common` `provides` `RemoteTech` without `RemoteTech` itself being installed, an erroneous upgrade checkbox appears. While it doesn't cause any errors or crashes, it's something we'll want to see resolved.

### DIFF
--- a/NetKAN/RemoteTech.netkan
+++ b/NetKAN/RemoteTech.netkan
@@ -1,5 +1,5 @@
 {
-	"spec_version": 1,
+	"spec_version": "v1.2",
 	"identifier": "RemoteTech",
 	"$kref": "#/ckan/kerbalstuff/134",
 	"license": "restricted",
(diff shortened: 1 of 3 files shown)


NetKAN/RemoteTech.netkan
```diff
@@ -11,17 +11,18 @@
...
-			"min_version": "2.3.1",
-			"comment": "Earlier versions of MM have forward compatibility problems, according to the MM release notes."
-		}
-	],
+	"provides" : [ "RemoteTech-dll" ],
+    "depends": [
+        { "name": "RemoteTech-common" }
+    ],
+    "conflicts": [
+        { "name": "RemoteTechXF" }
```
Dazpoet: This should probably be `"RemoteTech-dll"` for consistency reasons

Dazpoet:
As much as I like the idea I don't think we should put stuff into the metadata which could generate user errors in the currently used client.

plague006:
Oki

Dazpoet:
But rather than deleting this we could've just removed the provided field in RemoteTech-common and it would've worked?

plague006:
I'd rather wait until we have tools to create elegant solutions. Forcing every RT user to choose between regular RT and RTXF isn't elegant at all, unfortunately.

PR #2186, opened 2015-08-26
Version cleanup EVE

plague006:


Dazpoet:
It's a wonder they've been working this far, looks like a good call to not have them out of the door working for 1.1!

PR #2191, opened 2015-08-26
Add AVC vref for Solver Engines

plague006:


plague006:
https://github.com/KSP-CKAN/CKAN-meta/compare/015c586db56e...ece58f800ef7 added SolverEngines v1.10 which has a working .version file. Huzzah!

PR #2192, opened 2015-08-26
Version cleanup ActiveTextureManagement

plague006:


Dazpoet:
How do the 64bit ones look currently? They where never fixed since 64bit was scrapped in 1.0 iirc, might be worth having a look since it might be a thing in 1.1 again?

plague006:
The 64-bit variants had no issues.

PR #2193, opened 2015-08-26
Add case in build script for ksp_version being 1.0.99

plague006:
Many mods have `ksp_version_max: 1.0.99` and dependencies fail because not all mods have such a high compatibility ceiling. So instead lets use the default ksp version when we hit 1.0.99.


plague006:
@techman83 in the spam of netkan PRs you might have missed this one. It's the netkan version of https://github.com/KSP-CKAN/CKAN-meta/pull/754

techman83:
@plague006 I very much did. Thanks for the extra nudge. LGTM

PR #2205, opened 2015-08-28
[Kopernicus] seperated examples / moved repository

ThomasKerman:
We transfered the repository for Kopernicus, seperated the examples from the main plugin.
I will release an update after the changes have been merged (that's why the exaples depend on a Kopernicus 0.3.2).

Cheers
Thomas


Dazpoet:
L24 of KopernicusExamples needs a comma added and you can remove L5 of the same file since we don't need a new epoch for a totally new mod :)

ThomasKerman:
Done :)

Dazpoet:
Ok it's failing because KopernicusExamples don't have any releases, afaik our bot won't get caught up on such a thing so I'll merge this and it should solve itself once a proper version is released.

PR #2206, opened 2015-08-28
Update ElectricCost.netkan

jenyaza01:
It looks like I forgot dependence on ModuleManager


Dazpoet:
Thanks for the fix :+1:

PR #2209, opened 2015-08-28
Create BDArmory Sorting.netkan

jenyaza01:
Remove BDArmory's parts from Utility, because now they have their own tab.
As I detect, modname and filename must be the same.


Olympic1:
Same as in #2208 
> The identifier needs to be identical to the filename and not have any spaces :)

plague006:
Duplicate of #2210.

PR #2236, opened 2015-09-01
Add KopernicusExpansion

dbent:

PR #2237, opened 2015-09-01
Update OuterPlanetsMod

dbent:
- Updated minimum ModuleManager version
- Updated minimum Kopernicus version
- Added KopernicusExpansion as a recommended mod


dbent:
New `recommends` depends on #2236

plague006:
Would you mind moving the `depends` and `recommends` outside of the override? The technical effect should be identical but it makes it a little easier for humans to digest/lowers the required knowledge to read.

dbent:
Done.

plague006:

PR #2248, opened 2015-09-03
Update ScanSat Lite requirements

severedsolo:
This pack is integrated into Scansat as of v14.2 - so should only be offered on v14.1 or earlier.


NetKAN/ContractConfigurator-ScanSatLite.netkan
```diff
@@ -11,7 +11,9 @@
             "name" : "ContractConfigurator",
             "min_version" : "1.1.0"
         },
-        { "name" : "SCANsat" },
+        { "name" : "SCANsat",
+          "max_version" : v14.1"
```
Olympic1: Add a `"` in front of the leading `v`

severedsolo:
Whoops, I couldn't see it for looking. Done now.

plague006:
Setting max_version won't have any real effect. If they conflict: add it as a conflict. If they don't: you can just remove the dependency entirely.

The .netkan will only affect the most recent release of ScanSat Lite

PR #2276, opened 2015-09-15
Fix change in structure - closes #2275

### LINKED ISSUE #2275: Nexus Mod not updating
Hi.
I'm the author of the mod referenced in  NetKAN/NetKAN/SuperHeavyBoostersHistoricalNexus.netkan

I can't get CKAN to update it's idea of what version this should be and I'm lost with why it won't.  The NetKAN file was (by the look of it) autocreated when the mod was at v0.9, which was a bad starting place regarding folder structure in the archive, but the NetKAN file just seems to say it wants a "Nexus" folder, which is right.

Any ideas?

techman83:
Inflates correctly now.

### DIFF
--- a/NetKAN/SuperHeavyBoostersHistoricalNexus.netkan
+++ b/NetKAN/SuperHeavyBoostersHistoricalNexus.netkan
@@ -1,12 +1,12 @@
 {
     "$kref": "#/ckan/kerbalstuff/781",
-    "spec_version": 1,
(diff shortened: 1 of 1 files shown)


techman83:
@TiktaalikDreaming opened #2277

plague006:
Merging because this restores the netkan to working order, opening #2280 for the dependency addition listed in #2277.

PR #2283, opened 2015-09-15
Add ContractConfigurator-MaritimeMissionsandBoatNavigationPlugin

plague006:
There is an argument to be made for splitting this into multiple mods, but as the author hasn't split it, I opted to keep it as one module (for now).

For posterity: If this mod is to be split, the contracts should remain under the ID `ContractConfigurator-MaritimeMissionsandBoatNavigationPlugin` with a `name` hardcoded and the plugin should be split into a new .netkan.


techman83:
LGTM!

PR #2284, opened 2015-09-15
Re-adding dependency - Nexus

TiktaalikDreaming:
Last night I got a lot of help from Leon and Matthew regarding my mysteriously created metadata file for my mod.  It's mostly tidied up, but somewhere along the line the dependency got added but doesn't appear in the current file.  Just re-adding.  Hopefully with correct syntax.


plague006:
> my mysteriously created metadata file

Not a mystery. @mgsdk took his time to create the metadata for your mod so that it would be available to the tens of thousands of CKAN users.

techman83:
@plague006  sorted this in #2280

PR #2287, opened 2015-09-16
download mirror moved from kerbalstuff to github

Sigma88:
I'm not sure this will be enough

@plague006 let me know if I need to change something


techman83:
LGTM.

PR #2288, opened 2015-09-16
move download mirror from  Kerbalstuff

Sigma88:
@plague006  this should be more or less the same as #2287

### DIFF
--- a/NetKAN/SigmaBinary-DunaIke.netkan
+++ b/NetKAN/SigmaBinary-DunaIke.netkan
@@ -1,7 +1,7 @@
 {
     "spec_version": "v1.4",
     "license": "CC-BY-NC-SA-4.0",
-    "$kref": "#/ckan/kerbalstuff/992",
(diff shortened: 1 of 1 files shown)


techman83:
Testing failed due being reliant on #2287 being merged. Normally recommend individual PRs for changes, exception being when they depend on each other.

Sigma88:
so the problem was I put the link there? whoops :)

thanks @techman83

techman83:
@Sigma88 more that it's using the meta data in the repo that is no longer correct for the dependency and failing it. Where as the build script injects all the metadata contained inside the PR for when the mods depend on each other :)

I don't know that you can rebase a PR made with the GitHub interface, I primarily use git on the cli.

Sigma88:
I'm going to start over, I'm still getting a hang of this.

sorry for the problems

techman83:
No issue at all @Sigma88 - appreciate the effort :-)

PR #2290, opened 2015-09-17
Remove TrackingStationAmbientlight conflict with AmbientLightAdjustment.

mhoram-kerbin:
New version disables itself when the other mod is detected


techman83:
LGTM.

PR #2294, opened 2015-09-19
PersistentThrust and SolarSailNavigator

Felger:
Fix PersistentThrust and SolarSailNavigator, change to pull from Github rather than Kerbalstuff at author's request.  Aware this is redundant to KerbalStuffBot, this is a better netkan overall.

### DIFF
--- a/NetKAN/PersistentThrust.netkan
+++ b/NetKAN/PersistentThrust.netkan
@@ -1,7 +1,7 @@
 {
     "spec_version" : 1,
     "name"         : "Persistent Thrust",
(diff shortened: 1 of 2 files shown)


plague006:
The 2 repos cited in the `kref` lines don't have releases. CKAN won't index pre-releases.

Felger:
Ah, I'll pass that along to the author.

Felger:
Both repos are marked as releases now, touching a file to trigger a rebuild.

plague006:
Merging with my thanks.

PR #2296, opened 2015-09-19
Fixed EngineerLevelFixer to include it's ModuleManager dependency

civilwargeeky:
I entered a mod the other day, but I didn't realize that I needed to declare it has a ModuleManager dependency or that it has a KSP AVC file.


plague006:

PR #2300, opened 2015-09-20
Create NAS.netkan

zzhwudiaa:
Naval Artillery System(NAS) is an expansion of BDArmory, providing a set of naval weapon replicas from both side in WW2.
This is one of the authors and my forum ID is Acea.
http://forum.kerbalspaceprogram.com/threads/131605

### DIFF
--- a/NetKAN/NAS.netkan
+++ b/NetKAN/NAS.netkan
@@ -11,5 +11,9 @@
 		{
 			"name" : "BDArmory"
 		}
+    "recommends" : [
+		{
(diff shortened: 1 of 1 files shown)


zzhwudiaa:
Please reply me... I forgot to choose CKAN support when I created this

techman83:
@zzhwudiaa Awesome work creating the NetKAN! Our metadata herders have been overcome with real life it seems. Merging now with thanks!

PR #2306, opened 2015-09-22
Add Deltaglider from KerbalStuff

pjf:
Actual Netkan work by the fabulous RedNeck, I'm just doing the gitomancy.

:)


pjf:
And confirmed with RedNeck on IRC the file list looks good, so I'm merging away! :)

PR #2309, opened 2015-09-22
fix Module contains no files to install.

techman83:
Squee! Made 'find_regexp' do a thing!


plague006:
As this fixes a current issue I'll merge, but I'll make another PR to use `find` which should future-proof us better than the regex which depends on the current folder structure.

PR #2314, opened 2015-09-23
Update ShadowWorksStockalikeSLSandMore

plague006:
Using `find` will allow us flexibility in case the author changes structure again in the future.


techman83:
LGTM!

PR #2315, opened 2015-09-23
Add conflict to ContractConfigurator-ScanSatLite as it's now included in SCANsat

plague006:
As stated in #2248 SCANsat now includes this functionality (as of v14.2), so adding a conflict to the current .netkan will update the v14.2 .ckan and onward.
Closes  #2248


techman83:
One minor detail is that unless v14.2 is released, merging this will affect the current indexed ckan.

plague006:
````
 SCANsat-v14.2.ckan 	"NetKAN generated mods - SCANsat-v14.2" 	24 days ago
````

PR #2316, opened 2015-09-23
Add Kronkus.netkan

plague006:
Would appreciate your input on this @Olympic1 if it conflicts with any other packs.

### DIFF
--- a/NetKAN/Kronkus.netkan
+++ b/NetKAN/Kronkus.netkan
@@ -1,0 +1,16 @@
+{
+    "spec_version": "v1.4",
+    "identifier": "Kronkus",
+    "license": "CC0",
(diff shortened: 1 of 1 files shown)


Olympic1:
In my tests this didn't conflict with any packs, but because every new release of Kopernicus can break planet packs, it is best to add a version restriction. For Kronkus this will be ``"version": "beta-03-1"`` (if that works ofcourse)

plague006:
Specifically because Kopernicus has a netkan epoch it's `"version": "1:beta-03-1"` but yeah it works. Thanks.

plague006:
Re-tested locally for the latest version (v0.9) and passes. Passed Jenkins in the past. Passes Olympic1's judgement. Merging.

PR #2329, opened 2015-09-24
Remove KerbalConstructionTime-RP-0-Config - closes #2311

### LINKED ISSUE #2311: KerbalConstructionTime-RP-0-Config
@NathanKell this netkan points at the RP-0 and there are a bunch of CKANs indexed from it. Has something changed?

techman83:
According to @NathanKell this is no longer required due to KerbalConstructionTime supporting configs from other mods.


plague006:
:+1:

PR #2332, opened 2015-09-25
Added a new depends to two of my modlets

magico13:
And fixed a "conflicts" in Kerbal Construction Time that isn't actually true. Also added the new mod that Dated QuickSaves and SensibleScreenshots depend on (MagiCore)


techman83:
The build tested 'SensibleScreenshots' 4 times... I'll raise a ticket on that. The metadata looks correct and from some local testing it appears good. Merging before NetKAN-bot does its next run.

PR #2339, opened 2015-09-26
Adding USAF TD Ed Orion

TiktaalikDreaming:
According to Kerbal stuff this is already CKAN'ed, but it ain't.  Adding this with a recommend link to the accompanying upper stage mod (which I'll edit to also recommend this, once this is live).
I think I got me commas right this time.  :-)


plague006:
Looks good. Merging with my thanks.

PR #2340, opened 2015-09-26
Adding cross recommendation

TiktaalikDreaming:
Adding the "recommends" to match the other companion mod, and changing the file to find.  
Not sure if this means I should remove the x_via tag or not.


plague006:
While the x_via tag is extraneous, it doesn't have any deleterious effects. Looks good, merging with my thanks.

PR #2355, opened 2015-09-29
Fix repository link, add epoch to fix version scheme.

plague006:


plague006:
Pushing this through as it's just adding an epoch and fixing a link (and Jenkins' output looks sane).

qm3ster:
I don't get how that epoch line fixes it, where can I read more about epoch?
Also, why did you put in Kerbal%20Foundries%20Wheels%20and%20Repulsors%20ALPHA?
That's going to change (hopefully)?

plague006:
`x_netkan_epoch` forces an epoch string on the front of the version. With the beginning of a new epoch, new versions should now index properly.

I changed the repository link because `https://kerbalstuff.com/mod/58/` generates a 404. KerbalStuff's servers will not give a page if you have a link ending with the mod's ID and a forward slash at the end of the URL. Alternatively I could have used `https://kerbalstuff.com/mod/58` or even `https://kerbalstuff.com/mod/58/ItDoesntMatterWhatTextAppearsHereAsLongAsItsNotTheModNumberFollowedByAForwardSlash`

qm3ster:
Inspiring stuff. Thanks.

PR #2381, opened 2015-10-06
Add USI Core from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of Roverdude, to add [USI Core](https://kerbalstuff.com/mod/1194/USI%20Core) to CKAN.

Please direct questions about this pull request to [Roverdude](https://kerbalstuff.com/profile/Roverdude).

PR #2343, opened 2015-09-27
Add Nuclear Rockets from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of Roverdude, to add [Nuclear Rockets](https://kerbalstuff.com/mod/1175/Nuclear%20Rockets) to CKAN.

Please direct questions about this pull request to [Roverdude](https://kerbalstuff.com/profile/Roverdude).

PR #2392, opened 2015-10-08
Initial commit USICore, hardcoded license

qm3ster:
closes #2381

### DIFF
--- a/NetKAN/USICore.netkan
+++ b/NetKAN/USICore.netkan
@@ -6,8 +6,7 @@
     "$kref": "#/ckan/kerbalstuff/1194",
     "install" : [
         {
-            
-            "find": "UmbraSpaceIndustries"
+            "file": "GameData/UmbraSpaceIndustries",
             "install_to": "GameData"
         }
     ],
--- a/NetKAN/USICore.netkan
+++ b/NetKAN/USICore.netkan
@@ -3,5 +3,15 @@
     "x_via": "Automated KerbalStuff CKAN submission",
     "identifier": "USICore",
     "license": "CC-BY-NC-SA-4.0",
-    "$kref": "#/ckan/kerbalstuff/1194"
+    "$kref": "#/ckan/kerbalstuff/1194",
+    "install" : [
+        {
+            
+            "find": "UmbraSpaceIndustries"
+            "install_to": "GameData"
+        }
+    ],
+    "depends" : [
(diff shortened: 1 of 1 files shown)


qm3ster:
LET'S SEE NOW. WILL THIS WORK? WILL THIS NOT? WE MAY NEVER KNOW!

qm3ster:
@plague006 I stuck it into both places where it won't work, didn't I?
It should have been in the name of the commit, not a description?

qm3ster:
It works on my local NetKAN.

qm3ster:
@plague006 good to go.

pjf:
@qm3ster: The KS page is a little light on info, so I just want to confirm we're doing the right thing here. Am I correct this is mostly a parts and assets pack (which contains Kontainers and Reactors), which nothing else depends upon?

Glancing at the contents of the .zip I'm guessing it has a dependency on Firespitter-Core, but I can add that before merging if needed.

Is the dependency on USITools a hard one (required) or a soft one (recommended)? The reason I ask is that the .zip itself doesn't have the USITools files at all.

qm3ster:
Haven't been around much, so not sure where it's headed. USI tools is not required right now, but will probbably be a hard dependency once more content is added. (It's not something that has user-facing functionality, so no way it could be a suggestion/recommendation, it's either needed or not.)
In the future, this will be suggested/recommended/depended by MKS, FTT. (Most likely the rare 'recommended' because while nothing stops you from running those mods without reactors and kontainers, but using them is a different issue, and would require other mods.

The idea behind this release is not having duplicate parts in eg MKS and FTT, so they are being taken out into their own package.

plague006:
Auto-closing works in the top comment and/or in any of the commits or commit messages/descriptions.

plague006:
This core mod will conflict with/cause issues for https://github.com/KSP-CKAN/NetKAN/pull/2343

qm3ster:
With USI Core still not merged, currently already not possible to install Karbonite and MKS together. They both install UmbraSpaceIndustries folder, not Kolonization and Karbonite, clashing a lot.

plague006:
The entire suite of RoverDude's mods need all their metadata overhauled at once. Merging/creating USI Core alone will only server to create more issues.

qm3ster:
Agreed, thanks for Karbonite delay.

PR #2399, opened 2015-10-12
Create StockLS.netkan

vosechu:
This is my first CKAN submission, please let me know if I've screwed anything up! 

Thank you for putting together this amazing program! I love CKAN!

### DIFF
--- a/NetKAN/StockLS.netkan
+++ b/NetKAN/StockLS.netkan
@@ -1,0 +1,32 @@
+{
+    "spec_version"   : "v1.14",
+    "name"           : "StockLS",
+    "abstract"       : "Life support applied to stock parts so ksp updates don't break functionality",
+    "identifier"     : "StockLS",
+    "download"       : "https://github.com/vosechu/StockLS/releases/download/v1.0.0/StockLS.zip",
+    "license"        : "MIT",
+    "version"        : "1.0.0",
+    "release_status" : "stable",
+    "ksp_version"    : "1.0.4",
+    "resources" : {
+        "homepage"     : "",
+        "repository"   : "https://github.com/vosechu/StockLS"
+    },
+    "install" : [
+        {
+            "file"       : "StockLS",
+            "install_to" : "GameData"
+        }
+    ],
+    "depends" : [
+        { "name" : "ModuleManager", "min_version" : "2.6.0" }
+    ],
+    "supports" : [
+        { "name" : "USILifeSupport" }
+    ],
+    "recommends" : [
+        { "name" : "StockPlugins" },
+        { "name" : "StockRT" },
+        { "name" : "StockSCANSat" }
+    ]
+}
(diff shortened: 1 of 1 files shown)


vosechu:
I'm not sure why the checks failed, it doesn't have indicators like some of the other failed PRs. Should I submit an empty commit and see if it can build a second time? Or are there logs I could look at?

vosechu:
So the first time I think it was because the homepage was an empty string.

vosechu:
Okay, I'll have to look at this tomorrow. I don't know that anything is wrong because this worked before. Getting this error: `FATAL CKAN.NetKAN.Program (null) - Cannot find central directory`

plague006:
`spec_version` doesn't need to be `v1.14`, `v1.4` is sufficient.

`download`, `release_status`, `version`, and `ksp_version` can all be removed, since they can be scraped from the .version file in your .zip. You should also add `"$vref": "#/ckan/ksp-avc",` so that the .version file is scraped.

The metadata is specifically failing because your set `download` to a .version file and CKAN is expecting a .zip (since there are no directories inside a .version file it fails).

Lastly I had a look at https://github.com/vosechu/StockLS/releases/download/v1.0.0/StockLS.zip and I'm confused as to how users are supposed to install your mod. You have multiple `GameData` folders and you have your `.git` folder included in the zip, which will need to be filtered via the install stanza.

vosechu:
Thank you so much! I'll fix those things shortly. As you can tell from the commits I had a bear of a time figuring out what it wanted to I really appreciate the human touch! @plague006++

vosechu:
@plague006 When you have a moment, I've tried my best to make this go again but I could use some interpretation of the jenkins failure. It's saying that I don't have a release at vosechu/StockLS but I do. Maybe it's because it's just the built-in github zip of the source code instead of a binary?

To answer your question, the extra GameData was an accident. There will be an Optional directory at some point with some extra cfg files but it shouldn't be installed by default. It also shouldn't have been in the release though. :)

plague006:
For the 1.0.0 release you have the source code but no actual release .zip. Once you add that to the release CKAN will work just fine.

NetKAN/StockLS.netkan
```diff
@@ -0,0 +1,34 @@
+{
+    "spec_version"   : "v1.4",
+    "name"           : "StockLS",
+    "abstract"       : "Life support applied to stock parts so ksp updates don't break functionality",
+    "identifier"     : "StockLS",
+    "download"       : "https://github.com/vosechu/StockLS/archive/v1.0.0.zip",
```
plague006: You need to remove this line (ln 6) or else every time there's a new release CKAN will generate a new set of metadata but the download will still be this exact file. (ex. you release 1.1.0 but what users get installed is still 1.0.0)

vosechu:
Ah! Awesome. I assumed I would have to update the netkan file every time I released. You all are amazing. 

So is the source zip from github acceptable? Or should I pack up a zip file manually and make sure it doesn't have things like .git and whatnot?

vosechu:
I'm giving up and just going to do this through KerbalStuff. :)

PR #2414, opened 2015-10-17
Added CactEyeTelescopes

Angel-125:
Adds CactEyeTelescopes, replacement for CactEye2.netkan since Raven no longer maintains the mod.


vosechu:
@Angel-125 I notice that the netkan file says to find CactEye2 folder, but when I look at the tree I only see a CactEye folder. 

Also, I don't know how to fix this, but I noticed that the README wouldn't end up getting installed since it's above GameData; does that concern you at all?

plague006:
Closed by https://github.com/KSP-CKAN/NetKAN/pull/2468

PR #2427, opened 2015-10-20
Iniital release

linuxgurugamer:
I'm trying to do two installs here, one to GameData and one to the GameRoot. It seems to not like this, is this even possible?


plague006:
CKAN will not install external tools. `RemoteTech` must be a `conflict` as you cannot be assured that users will have AutoPruner.

linuxgurugamer:
ok.  So, can AutoPruner be set up as it's own mod and installed by itself?

plague006:
CKAN will not install external tools.

plague006:
Closed by https://github.com/KSP-CKAN/NetKAN/pull/2460

PR #2429, opened 2015-10-20
Add Mun (pocket ed) to the CKAN.

pjf:
Obsoletes #2331. PR so I can have the bots take a look. :)

### DIFF
--- a/NetKAN/MunPocketEdition.netkan
+++ b/NetKAN/MunPocketEdition.netkan
@@ -1,0 +1,7 @@
+{
+    "spec_version": "v1.4",
+    "x_via": "Automated KerbalStuff CKAN submission",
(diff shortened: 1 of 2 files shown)


techman83:
Blah, it tested 'NinjaTechEnterprises' twice. Bash + Functions + Variables make me cry.

pjf:
Oh, do I need to rebase?

I have a hunch we could change:

`git diff --diff-filter=AM --name-only --stat origin/master`

to use `origin/master...HEAD` rather than just `origin/master`. The three-dotted notation in diff says "give me the differences from where we branched from origin/master, to where we are now". (Otherwise it gives us the absolute set of differences, which isn't anywhere near as useful).

Actually, let me put that in a PR.

plague006:
:+1:

PR #2430, opened 2015-10-20
build.sh: Evaluate only changes on branch, not absolute diffs

pjf:
This change instructs git/build.sh to only examine files which have changed on the current branch from its most-recent-ancestor of `origin/master`. This is better than the absolute differences, as it means `origin/master` can change as much as it likes and we don't care.

Tested locally. I'm sure there's an issue number for this somewhere, so feel free to edit this PR to include that. :)


techman83:
w00t! This has annoyed me for a while :+1:

PR #2431, opened 2015-10-20
Updated QuickHide

malahx:
Toolbar is not a dependency since QuickHide v2.00


plague006:
Looks good, thanks for helping us keep the metadata up to date.

PR #2434, opened 2015-10-22
Remove install conflict with and instead depend on WildBlueTools

plague006:
Always a PITA to find these kinds of conflicts. I had the benefit of writing the metadata for the other WildBlue mods.


plague006:
This is important as the issue was reported by someone else and not just me, so I'm pushing through the merge.

Issue #2382, opened 2015-10-06
Sigma OPM Expansion split


plague006:
Via @Sigma88 https://github.com/KSP-CKAN/NetKAN/pull/1533#issuecomment-145795408

So,
the 3 mods that were originally part of "Sigma OPM Expansion" are now released separately

these are the links to the forum threads and github repositories

Sigma: OPM Recolor
http://forum.kerbalspaceprogram.com/threads/135895
https://github.com/Sigma88/Sigma-OPMRecolor

Sigma: OPM Tilt
http://forum.kerbalspaceprogram.com/threads/135899
https://github.com/Sigma88/Sigma-OPMTilt

Sigma: PluronKhato
http://forum.kerbalspaceprogram.com/threads/135881
https://github.com/Sigma88/Sigma-PluronKhato

the last one is now a standalone mod, it doesn't require OPM.

Sigma88:
@plague006  the moving around of all my mods is completed!

you can find all the links to the new threads here:

http://forum.kerbalspaceprogram.com/threads/112095

plague006:
Thanks for the heads up.

plague006:
Took a bit of time, but I created a PR that should solve things.

Sigma88:
:+1:  Thanks, I'm going to check it out as soon as I get some time

PR #2442, opened 2015-10-23
Sigma expansion split

plague006:
`SigmaOPMExpansion` was split into 3 mods.
`SigmaPluronKhato` are 2 planets, `SigmaOPMRecolor` is a recolor of one of the OPM planets, and "SigmaOPMTilt" I kept under the ID `SigmaOPMExpansion` to provide some kind of upgrade path (`SigmaOPMExpansion` is discontinued and old links are dead).

Branched under the NetKAN repo to allow anyone to modify as needed.

See the referenced issue for any missing details.
Closes https://github.com/KSP-CKAN/NetKAN/issues/2382

@Olympic1 `SigmaPluronKhato` will fall under your planetpack purview. https://github.com/KSP-CKAN/CKAN-meta/issues/588

@Sigma88 let me know if you want modifications to the relationships (if you want them to recommend each other or any others from your suite of mods).

### DIFF
--- a/NetKAN/SigmaOPMExpansion.netkan
+++ b/NetKAN/SigmaOPMExpansion.netkan
@@ -12,7 +12,7 @@
         { "name": "OuterPlanetsMod" },
         { "name": "Kopernicus" }
     ],
-    "recommends": [
+    "suggests": [
         { "name": "SigmaOPMRecolor" }
     ],
     "resources": {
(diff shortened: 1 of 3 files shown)


Sigma88:
If there's something less strong that Reccommend it would be better, (suggests maybe?).

I don't want people to think they are better if used together since I went out of my way to make sure each mod works 100% if used alone (as well as if they are used together of course)

thanks for your time @plague006

techman83:
LGTM! Merging with Thanks!

PR #2389, opened 2015-10-08
Add Eva Handrails KIS Compatibility  from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of DaleTheGoat, to add [Eva Handrails KIS Compatibility ](https://kerbalstuff.com/mod/1196/Eva%20Handrails%20KIS%20Compatibility%20) to CKAN.

Please direct questions about this pull request to [DaleTheGoat](https://kerbalstuff.com/profile/DaleTheGoat).

PR #2452, opened 2015-10-23
Add EvaHandrailsKISCompatibility

plague006:
Added this as a suggestion to EVAHandrailsPackContinued as KIS is very popular and presenting the patch to users is probably a good choice.

Closes https://github.com/KSP-CKAN/NetKAN/pull/2389


techman83:
LGTM! Merging with Thanks! :D

PR #2433, opened 2015-10-22
Add Ktolemy from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of Daniel_L, to add [Ktolemy](https://kerbalstuff.com/mod/1223/Ktolemy) to CKAN.

Please direct questions about this pull request to [Daniel_L](https://kerbalstuff.com/profile/Daniel_L).

PR #2455, opened 2015-10-23
Add Ktolemy

plague006:
@Olympic1 part of https://github.com/KSP-CKAN/CKAN-meta/issues/588 but I can tell you up front it should conflict with every other pack. It places Kerbin in the center of the system and has every other planet revolving around it.

Closes https://github.com/KSP-CKAN/NetKAN/pull/2433


Olympic1:
Noted

techman83:
LGTM! Merging with Thanks! :D

PR #2459, opened 2015-10-23
Add 2015 SLS CDR textures from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of nothke, to add [2015 SLS CDR textures](https://kerbalstuff.com/mod/1228/2015%20SLS%20CDR%20textures) to CKAN.

Please direct questions about this pull request to [nothke](https://kerbalstuff.com/profile/nothke).

PR #2413, opened 2015-10-16
Add "Hornisse" afterburner jet + mk1 Ckpt. reskin from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of daPenguin, to add ["Hornisse" afterburner jet + mk1 Ckpt. reskin](https://kerbalstuff.com/mod/1212/%22Hornisse%22%20afterburner%20jet%20+%20mk1%20Ckpt.%20reskin) to CKAN.

Please direct questions about this pull request to [daPenguin](https://kerbalstuff.com/profile/daPenguin).

PR #2460, opened 2015-10-23
Add AntennaRangePatches

plague006:
Mod wants to install and run AutoPruner or `RemoteTech` breaks, so added as a `conflict`.

Closes https://github.com/KSP-CKAN/NetKAN/pull/2427

### DIFF
--- a/NetKAN/AntennaRangePatches.netkan
+++ b/NetKAN/AntennaRangePatches.netkan
@@ -3,18 +3,18 @@
     "spec_version": "v1.4",
     "$kref": "#/ckan/kerbalstuff/1220",
     "$vref": "#/ckan/ksp-avc",
-	"license": "CC-BY-NC-SA-4.0",
+    "license": "CC-BY-NC-SA-4.0",
     "depends": [
-		{ "name": "AntennaRange" },
-		{ "name": "ModuleManager" }
-	],
+        { "name": "AntennaRange" },
+        { "name": "ModuleManager" }
+    ],
     "recommends": [
-		{ "name": "Antennas" }
-	],
+        { "name": "Antennas" }
+    ],
     "conflicts": [
-		{ "name": "AntennaPatch4AntennaRange"},
+        { "name": "AntennaPatch4AntennaRange"},
         { "name": "RemoteTech"}
-	],
+    ],
     "install": [
         {
             "find"       : "AntennaRangePatches",
--- a/NetKAN/AntennaRangePatches.netkan
+++ b/NetKAN/AntennaRangePatches.netkan
@@ -1,0 +1,24 @@
+{
+    "identifier": "AntennaRangePatches",
+    "spec_version": "v1.4",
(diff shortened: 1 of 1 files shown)


linuxgurugamer:
I understand about AutoPruner

so, if you don't mind, a few questions:

1.  How would you suggest I distribute the RT patch without the conflict?  Would some very explicit instructions be good enough, specifically stating that the patch won't work without running the AutoPruner?
2.  Is it possible to detect that Autopruner is already installed by the user, and then to install the RT patch into the AutoPruner directory?
3.  Since CKAN is written in C#, is it possible to write a mod for CKAN (call it CKANpruner) which will do what AutoPruner does?  And if so, how could that be used in a Netkan file (ie:  if CKANpruner is available, then run a specific PRNL file)?
4.  I might be interested in rewriting AutoPruner in C# for CKANpruner.  Any pointers?

Thanks

plague006:
1. If RT's licensing permits, I would include it in my mod with the parts pre-cut that need to be removed then submit it to CKAN with a conflict with RT. If licensing doesn't permit that, then I wouldn't list with CKAN (I wouldn't want my mod listed if it wasn't feature-complete).

2. AutoPruner doesn't have a set install location and further having it installed isn't enough as the user needs to actually *run* AutoPruner to have it take effect.

3. I don't think CKAN ever will (or ever should) give a mod the ability to delete/disable parts of other mods. That role should fall to mods themselves (ex. MM). That's just my opinion though.

4. I know only the basics of the basics of C#, I'm not a coder so I'll let others chime in on this point.

The utility of CKAN is that it automates everything. If a mod requires something that isn't or shouldn't be automated then it's just not a candidate for CKAN.

plague006:
I went through the outstanding PRs last night, and there were a few mods that just aren't candidates for CKAN like https://github.com/KSP-CKAN/NetKAN/pull/2413 and https://github.com/KSP-CKAN/NetKAN/pull/2459. Not every mod is compatible with CKAN.

linuxgurugamer:
I'll review the RT license, thank you

Olympic1:
@linuxgurugamer Poke

linuxgurugamer:
I'm in the process of contacting the RT authors and getting permission to redistribute the models.  In the meantime, i'll update this to make RT a conflict, and once I get all the permissions, I'll update the mod with the parts.

plague006:
This PR already has an RT conflict.

linuxgurugamer:
Oh.

yuck.  I've removed the autopruner and the rt recommends


JBB

plague006:
> I've removed the autopruner and the rt recommends

I think you're in the wrong PR. This PR never had autopruner, nor did it have an RT recommends.

plague006:
I'm merging this now so that @linuxgurugamer stops making more and more PRs that replicate this exact metadata.

linuxgurugamer:
Ok.

Thank you

I hadn't gotten a response from a question, and it seemed to have 
dropped into a black hole.

Didn't mean to make trouble.


JBB

PR #2468, opened 2015-10-23
Add CactEyeTelescopesContinued

plague006:
~~I fully expect this will fail Jenkins because of an error in the .zip file itself. I test with a hardcoded `download` directed to the previous release of the mod and had no issues. With that said @Angel-125 would you be able to push a new release of the mod and let us know when it's done?~~

It passed, it must have been a corrupt download on my end (or something).

Closes https://github.com/KSP-CKAN/NetKAN/pull/2414


techman83:
LGTM! Merging with Thanks! :D

PR #2474, opened 2015-10-25
Hack around USI Core being in too many places.

bobprime:
The metadata for RoverDude's projects no longer matches his current project configuration.  Specifically his USI-Core project is included in multiple other projects.  The fixes the ones that are currently in ckan.

It adds the USI-Core to ckan.  Makes both UKS and Karbonite depend on it, and then filters the USI-Core files back out of both of those .netkan files.

When people want to add MKS-Lite and Orion all they will have to do is create netkan files that depend on USI-Core and add a filter 
 "filter"     : ["USICore.version", "CHANGELOG.txt", "Kontainers", "ReactorPack"],
.

I have tested this on my install and the final outcome is the same as manually unpacking the zipfiles.  I also tested in game that all of the parts are there, and they are.

### DIFF
--- a/NetKAN/Karbonite.netkan
+++ b/NetKAN/Karbonite.netkan
@@ -9,7 +9,11 @@
     "release_status" : "stable",
     "$vref"          : "#/ckan/ksp-avc",
     "resources" : {
-        "homepage" : "http://forum.kerbalspaceprogram.com/threads/89401"
+        "homepage" : "http://forum.kerbalspaceprogram.com/threads/89401",
+        "license"      : "https://github.com/BobPalmer/Karbonite/blob/master/FOR_RELEASE/license.txt",
+        "bugtracker"   : "https://github.com/BobPalmer/Karbonite/issues",
+        "repository"   : "https://github.com/BobPalmer/Karbonite" ,
+        "kerbalstuff"  : "https://kerbalstuff.com/mod/759/Karbonite"
     },
     "depends" : [
         { "name" : "USITools" },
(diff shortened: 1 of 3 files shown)


NetKAN/Karbonite.netkan
```diff
@@ -26,7 +27,7 @@
         {
             "file"       : "GameData/UmbraSpaceIndustries",
```
plague006: "find": "UmbraSpaceIndustries/Karbonite",
"install_to" : "GameData/UmbraSpaceIndustries"

No filter required (means that changes to USI-Core will not break the Karbonite install.

NetKAN/UKS.netkan
```diff
@@ -25,7 +25,9 @@
   "install" : [
     {
       "file"       : "GameData/UmbraSpaceIndustries",
```
plague006: Apply the same logic as the Karbonite comment (atm not able to quickly look at the contents of UKS.zip)

plague006:
How will these changes affect RoverDude's other mods? `USI-SRV` `USI-LS` `USI-FTT` `USI-EXP` `USI-ART` `SoundingRockets` `AlcubierreStandalone`

bobprime:
The only two mods on ckan right now that have the USI Core included in their distribution are UKS and Karbonite.  The core just contains reactors and container elements.  It will impact MKS-Lite and Orion when they get added to ckan because they both use those and will be able to use the same fix.

I have tested it with SRV, LS, FTT, and, EXP.  I just checked ART and Alcubirre and they do not contain the broken file.

The find for UKS will need to be for UmbraSpaceIndustries/Kolonization .  Would you like me to make those changes and issue another pull request (I am fairly new to GitHub).  Or does just pushing to my branch update the pull?

plague006:
Pushing to a branch updates the PR.

PR #2475, opened 2015-10-25
Move Contract Configurator to ksp_version_strict

jrossignol:
I'll have a new version for 1.0.5 up pretty quickly, but as a general rule Contract Configurator touches so many things that it's almost guaranteed to break on any version change (the only one that it didn't break on was 1.0.4, but that was a tiny delta from 1.0.3).

I couldn't find ksp_version_strict in the doco - so basing this off of @pjf's post in the forums.


plague006:
`"ksp_version_strict" : "true",` -> `"ksp_version_strict" : true,`
and
`"spec_version": 1,` -> `"spec_version": "v1.16",`

On hold pending the merge of https://github.com/KSP-CKAN/CKAN/pull/1499 and the release of CKAN v1.16.

PR #2480, opened 2015-10-26
Add KSAIVAUpgrade

plague006:
`ASETAvionics` isn't currently a dependency but according to the author it will be in future versions (and rather than the mod breaking later on I'd rather future-proof things now).

Closes  #2477


Olympic1:
Merging with Thanks

PR #2486, opened 2015-10-27
Fix install for SimpleRealism

plague006:


techman83:
LGTM! Merging with thanks!

PR #2488, opened 2015-10-27
Attempt to fix StockPlus by changing asset match pattern

plague006:
Locally this still works (but so does the original) and the hope is that it'll get this mod unstuck from the fail list on http://status.ksp-ckan.org bringing this mod's version to parity with its brother StockBugFixModules


techman83:
LGTM! Merging with thanks!

PR #2489, opened 2015-10-27
Attempt to fix VenStockRevamp by adding an install stanza

plague006:
Locally this still works (but so does the original) and the hope is that it'll get this mod unstuck from the fail list on http://status.ksp-ckan.org


plague006:
Pushing this through as it passes Jenkins and the mod is currently not passing @netkan-bot anyways. :neutral_face:

PR #2493, opened 2015-10-28
Add manual install advice to abstract

politas:
From discussion on CKAN thread in KSP forums, Ferram has identified that CKAN is causing problems when installing FAR. Let's at least advise users to try a manual install while we sort out the problem.

### DIFF
--- a/NetKAN/FerramAerospaceResearch.netkan
+++ b/NetKAN/FerramAerospaceResearch.netkan
@@ -1,6 +1,7 @@
 {
     "spec_version"   : 1,
     "identifier"     : "FerramAerospaceResearch",
+    "abstract"       : "FAR replaces KSP's stock part-centered aerodynamics model with one based on real-life physics. If you experience any problems after installing, please remove under CKAN and try a manual install before seeking assistance.",
     "$kref"          : "#/ckan/kerbalstuff/52",
     "license"        : "GPL-3.0",
     "x_netkan_epoch" : "3",


Olympic1:
people don't really read info/warnings, it would be better to fix it.

I never used FAR but if I know what the problems are, when installing via CKAN, I can fix them.

plague006:
From what I'm reading in the [FAR via CKAN support thread](http://forum.kerbalspaceprogram.com/threads/127492) the issues seem to be around people uninstalling the mod and not the installation. This is an issue with all mods though and not FAR specifically: CKAN will not delete files it didn't install.

politas:
Far as I can tell, is it fixed, actually.

politas:
Ok, since both of those possible FAR fixes didn't fix things, can this change be re-added? Or maybe we should just de-list FAR, remove it from RO, and let Ferram deal with the complaints.

Olympic1:
The problem with FAR is that it creates files when you play KSP. When updating FAR with a new version, CKAN will only remove files it installed, it won't remove the files FAR created. As plague006 already mentioned:
> CKAN will not delete files it didn't install.

politas:
The other problem with FAR is that Ferram doesn't want FAR on CKAN, and is unwilling to make any changes to the way he operates to make things easier for FAR to work with CKAN. And then there's the fact that other mods are checking for the existence of FAR and adjusting behaviour if they find it, and the files CKAN leaves behind trigger that adjustment, so you get weird effects from wings and such.

Seriously, I think we should do what Ferram asks and remove FAR from CKAN, making a public statement that we have done so at Ferram's request, and any CKAN users who wish to use FAR will have to install it manually. I think CKAN's been bending over backwards for Ferram long enough.

plague006:
That's way beyond the scope of this PR. See: https://github.com/KSP-CKAN/CKAN/issues/1078

PR #2507, opened 2015-10-30
Possible FAR fix

Olympic1:
Install `ModularFlightIntegrator` folder

Possible fix for #2493


Olympic1:
This is failing because it wants to overwrite an older version of `ModularFlightIntegrator`.
If you remove the dependancy, because it is included in FAR, it runs fine ingame without any problems.

PR #2509, opened 2015-10-30
Add Harder Solar System netkan files

stevehead:
Each netkan file is for a different scaled solar system.

### DIFF
--- a/NetKAN/HarderSolarSystem-1x.netkan
+++ b/NetKAN/HarderSolarSystem-1x.netkan
@@ -1,5 +1,5 @@
 {
-    "spec_version" : 1,
+    "spec_version" : "v1.4",
     "identifier"   : "HarderSolarSystem-1x",
     "$kref"        : "#/ckan/github/SteveheadSpaceAgency/HarderSolarSystem",
     "name"         : "Harder Solar System - Stock Scale",
--- a/NetKAN/HarderSolarSystem-1x.netkan
+++ b/NetKAN/HarderSolarSystem-1x.netkan
@@ -5,21 +5,28 @@
     "name"         : "Harder Solar System - Stock Scale",
     "abstract"     : "Inclines entire Solar System to simulate axial tilt on Kerbin, relocates KSC, and makes slight inclination adjustments to Mun/Minmus. Everything else remains stock.",
     "license"      : "public-domain",
-    "ksp_version" : "1.0.4",
-    "resources"     : {
-		"homepage" : "http://forum.kerbalspaceprogram.com/threads/137931"
-	},
+    "ksp_version"  : "1.0.4",
+    "resources"    : {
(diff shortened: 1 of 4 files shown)


plague006:
For each of the .netkans you can use `"provides": "HarderSolarSystem"` and "conflicts: HarderSolarSystem" which will ensure users can only install one at a time while keeping the metadata easier to read.

plague006:
None of the netkans contain install stanzas which means CKAN doesn't know how to install them. `HarderSolarSystem-1x.netkan` will require:
````
"install": [
        {
            "find"       : "1x/GameData/HarderSolarSystem-1x",
            "install_to" : "GameData"
        }
    ]
````

There's also no `cache` folder in `1x/GameData/HarderSolarSystem-1x`.

stevehead:
Thanks, I'll make the changes. I tested each individually using the netkan.exe and then installing with ckan file generated, and they all worked fine. It was my understanding that CKAN by default finds the first top-level directory with the identifier. I'll put them in for clarification. Also, HarderSolarSystem-1x does not have a cache folder since it is using the stock scales.

Olympic1:
Change `"spec_version": 1` to `"spec_version": "v1.4"` because yuo're using the `find` command

stevehead:
Thanks guys. Hopefully, I got it this time.

plague006:
>It was my understanding that CKAN by default finds the first top-most directory with the identifier.

That's the case, I just wanted it explicitly defined to make it a little clearer when reading the metadata.

>Also, HarderSolarSystem-1x does not have a cache folder for Kopernicus textures since it is using the stock scale.

That's my mistake then, Kopernicus-dependent mods aren't my strong suit and it jumped out at me as being inconsistent with the others.

This gets a :+1: from me and I'll leave the last call to @Olympic1.

One last note, as it stands each .netkan will have to be updated manually with new KSP releases to mark them as compatible. If you want to automate that, you could use .version files (identical copies of `HarderSolarSystem.version` in each version's folder for example) to define KSP version compatibility. Just a suggestion you can take or leave.

stevehead:
Hmm... good call. I'm going to be working all day, but I can add the .version file after I get home later.

Olympic1:
> Hmm... good call. I'm going to be working all day, but I can add the .version file after I get home later.

If you made the .version files and included them into the download (you'll need to make a new release with the files), you'll need to add `"$vref" : "#/ckan/ksp-avc",` into the .netkan files.

stevehead:
Alright, thanks for the help. I'll get it done this evening. I would have gone with KerbalStuff to make things easier, but the nature of how I'm releasing this is a quite unique.

stevehead:
The .version files are included in a new release on GitHub

Olympic1:
Merging with great thanks.

PR #2511, opened 2015-11-02
Fix StockPlus

Olympic1:
Search for the `.zip` file.

Testing locally installs the required file

### DIFF
--- a/NetKAN/StockPlus.netkan
+++ b/NetKAN/StockPlus.netkan
@@ -1,7 +1,7 @@
 {
     "spec_version"   : 1,
     "identifier"     : "StockPlus",
-    "$kref"          : "#/ckan/github/ClawKSP/KSP-Stock-Bug-Fix-Modules/asset_match/StockPlusController",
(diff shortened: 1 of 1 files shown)


techman83:
LGTM! Merging with thanks!

plague006:
Still fails. @techman83 can you have a look into what's going wrong?

Olympic1:
It only fails for netkan-bot, because locally they download it correctly.

plague006:
Yes. That's been the case for weeks.

techman83:
O.o - Ran it by hand with the NetKAN user + netkan.exe that the bot uses (which is updated automatically) and it worked fine... It's just in the process of doing it's latest run, will have a look afterwards. You've all been doing amazing as those are the last 2 showing errors in the logs! :D

techman83:
Ok Fixed. The Download URL hadn't changed so the cached download hadn't been replaced. Which was different to whatever it refers to now. KSP-CKAN/NetKAN-bot#16 will help deal with this.

PR #2520, opened 2015-11-06
Add PulsejetArgusAs014

Olympic1:
closes #2506


plague006:
MIT is listed in the KS page, but the readme contains a custom license.

Olympic1:
To what do I need to change te license to

plague006:
If you want to look into the definition of OSS and it meets the criteria then `unrestricted`/`restricted` but otherwise nobody should cry too hard if you just use `unknown`.

PR #2533, opened 2015-11-08
Update SETI-CommunityTechTree.netkan

Y3mo:
Hello,
since the SETI mod complex will undergo a lot of changes from 1.0.4 to 1.0.5 and then again to 1.1, with many changes in between and a move to github in general, I m trying to prepare for it.

To be able to respond faster if (when) issues occur (not only with my mod, but with dependencies/recommendations), I try to externalize the metadata from the netkan respository. I took RealismOverhaul as an example for this. Hopefully I did not miss something.

As this is my first externalization attempt, any feedback on the 2 files (the one inside the netkan respository and the linked file in the SETI-CommunityTechTree repository) is very much appreciated.

Best regards,
Yemo

### DIFF
--- a/NetKAN/SETI-CommunityTechTree.netkan
+++ b/NetKAN/SETI-CommunityTechTree.netkan
@@ -1,34 +1,6 @@
 {
-    "spec_version" : 1,
-    "$kref": "#/ckan/kerbalstuff/748",
-    "$vref" : "#/ckan/ksp-avc",
+    "spec_version": "v1.2",
     "identifier": "SETI-CommunityTechTree",
-    "x_netkan_license_ok" : true,
-    "depends" : [ 
-        { "name": "CommunityTechTree" },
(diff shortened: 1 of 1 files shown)


Y3mo:
Hm, no identified problems. Definately need help with this.

techman83:
It's stating it couldn't find any files, which is usually a path not specified correctly in the metadata verses the zip file. I've not played around with external NetKANs a whole lot yet.

Y3mo:
Ah, thank you very much, I had the wrong mod link specified (old SETI-BalanceMod 484 instead of SETI-CommunityTechTree 748 on kerbalstuff). Must have mixed it up since I m working on the 484 netkan file as well.

Best regards,
Yemo

Y3mo:
That should do it.

Olympic1:
Merging with my thanks

PR #2535, opened 2015-11-09
Create SETI-BalanceMod

Y3mo:
This is not the old mod with the identifier "SETI"!

It is actually a split off from the current SETIctt (which itself was a split off from the old SETI mod), which will be provided from the same source (kerbalstuff 484) as the old SETI mod for ksp 0.90, which ended with version 0.9.0. The new mod will start with version 0.9.5 (for ksp 1.0.5).
Since I m not sure how to do the transition exactly, I try to externalize the metadata to be able to react faster to issues resulting from the split/transition.

Thus while the current external version is just a content copy from the old SETI mod file here on netkan, I will introduce overwrites for the 0.9.5 version for ksp 1.0.5 and then take it from there.

Since no one will see this anyway at the moment (since the currently linked one is for ksp 0.90), and thus no one is affected, this is imho the best way to provide consistency for the users.

I will also try to change the .netkan file with the "SETI" identifier to only use the 0.9.0 download for ksp 0.90 and not react to the new 0.9.5+ versions for ksp 1.0.5.


Y3mo:
Same mistake as with the SETI-CommunityTechTree, should work now.

Olympic1:
Merging with my thanks

PR #2538, opened 2015-11-10
Update StageRecovery.netkan

jkoritzinsky:
Update ksp_version_max override since the newest version supports KSP 1.0.5.


Olympic1:
Merging with my thanks

PR #2551, opened 2015-11-11
Create PreciseNode.netkan

sergioberg:
PreciseNode.netkan


Olympic1:
mods aill be added in KSP-CKAN/CKAN-meta/pull/796

PR #2553, opened 2015-11-11
EVE revision

### LINKED ISSUE #2550: EnvironmentalVisualEnhancements Packages Need Revision
rbray89 has released the overhauled EVE!  This is very exciting, but CKAN can't handle it quite yet.

Forum thread regarding the release: http://forum.kerbalspaceprogram.com/threads/55905

github release: https://github.com/rbray89/EnvironmentalVisualEnhancements/releases/tag/EVE-1.05-1

The NetKAN file for EVE-HR caught it and updated, but probably wouldn't work even if it could be installed correctly (which hopefully it can't be).  The packaging these days comes in two parts:

x86-Configs-Release.zip -- configs and textures; currently not offered in high and low res; presumably will share an install slot with similar packs from Astronomer et al

x86-EVE-Release.zip -- core plugin

I took a look at the EVE NetKAN files, but the $kref stuff for github asset detection is a bit above my paygrade.

TIA!

Olympic1:
fixes #2550


plague006:
Jenkins passing was erroneous as `EnvironmentalVisualEnhancements-LR.netkan` fails to asset match any current releases.

As `HR` already asset matches and created `EnvironmentalVisualEnhancements-HR-EVE-1.05-1` I've kept it as the de facto `EnvironmentalVisualEnhancements-Config` while freezing `LR`. Everything gets an epoch bump due to version scheme changes as well as to force all users to update to the current install method.

plague006:
For the sake of Jenkins I'm going to delete `EnvironmentalVisualEnhancements-LR.netkan.frozen` in a commit then have it test. If it passes I'll restore `EnvironmentalVisualEnhancements-LR.netkan.frozen` which will cause Jenkins to fail.

plague006:
Jenkins passed and properly tested both .netkans. Reverting the deletion.

Olympic1:
So this will work?

Olympic1:
Jenkins loved it!

PR #2554, opened 2015-11-11
Make sure v0.6.10 is compatible with KSP 1.0.5

Olympic1:
First time I use the `x_netkan_override` stanza, so this maybe needs some adjustments


plague006:
You're on the right track. The issue with how you did it is if someone doesn't upgrade to 1.0.5 and stays on 1.0.4 the mod is now incompatible so I turned the strict `ksp_version` into a min/max range. In the course of reviewing I noticed 0.6.10 didn't have a `v` and wasn't the current version so I forced a v and modified the override to match.

PR #2560, opened 2015-11-12
Add SurfaceMountedLights

Olympic1:
closes #2559


plague006:
:+1: Good catch on the license.

PR #2561, opened 2015-11-12
Update HeatManagement

Olympic1:
Latest update depends on `WildBlueTools`


plague006:
I don't know why netkan couldn't `find` the right directory but `file` worked so I made that change.

PR #2562, opened 2015-11-12
Use find

Olympic1:
Use `find` because of new folder structure


plague006:
Merging this as it passes local testing with the dependency removed. The fact that it can't be installed is moot, plenty of mods can't be successfully installed due to relationship issues after a patch.

PR #2566, opened 2015-11-12
Mk2Expansion.netkan

SuicidalInsanity:
Updates dependencies for M2X 1.6.10


Olympic1:
You aren't editing the existing `.netkan` file. Here is the correct path on your fork: https://github.com/SuicidalInsanity/NetKAN/blob/patch-1/NetKAN/Mk2Expansion.netkan

PR #2567, opened 2015-11-13
SufaceMountedLights -> surfacelights in order to give an upgrade path to the existing user base.

plague006:
SufaceMountedLights was created where surfacelights already existed. They're the same mod. Due to SufaceMountedLights being very recently created I'd like to prioritize the existing install base surfacelights has and deprecate SufaceMountedLights.
Merge before https://github.com/KSP-CKAN/CKAN-meta/pull/804


pjf:
Jenkins likes this, and it looks sensible to me as a way of having us only index this mod once. About to merge with great thanks!

PR #2571, opened 2015-11-13
Filter FAR config file

dbent:
Newest version of FAR has the config file in the distribution whereas previous versions didn't. This causes an overwrite conflict on upgrade, so filter it out. @pjf


pjf:
You are the best! Thank you!

PR #2572, opened 2015-11-13
Update CommunityTechTree

Olympic1:
CTT works with 1.0.5 but hasn't updated the KS page yet, so doing it manually


plague006:
No `force_v` necessary, CTT's versioning scheme has remained consistent.

PR #2573, opened 2015-11-13
Add dependency

Olympic1:
closes #2566


Olympic1:
pushing this through as it needed to be merged before the next update (which already was released)

PR #2575, opened 2015-11-13
Update UnmannedBeforeManned.netkan

Y3mo:
And the next externalization. Removed ksp avc reference, so I can update this minor mod directly via kerbalstuff for CKAN.


Olympic1:
Another good one. Merging

PR #2581, opened 2015-11-13
Add Stock Visual Enhancements from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of Nhawks17, to add [Stock Visual Enhancements](https://kerbalstuff.com/mod/1263/Stock%20Visual%20Enhancements) to CKAN.

Please direct questions about this pull request to [Nhawks17](https://kerbalstuff.com/profile/Nhawks17).

PR #2584, opened 2015-11-14
Add StockVisualEnhancements

Olympic1:
I added a filter for the `EVE` folder in `StockVisualEnhancements.netkan` so that the textures, that will be chosen by the user, don't get overwritten.

Merge **after** #2553
closes #2581


Olympic1:
force a rebuild

PR #2586, opened 2015-11-14
Stock Bug Fix asset_match update

pacbard:
Stock Bug Fix and Stock Plus have been merged in the new release.

This updates the asset_match function to the new release name.


pacbard:
I just noticed pull #2585. Probably it is a better idea to use that approach. I'll leave this open just in case you guys want to go down this path.

If you decide to use pull #2585, I would also cap this package to KSP version 1.0.4 (without the changes included in this pull request).

PR #2587, opened 2015-11-14
Deprecate Stock Bug Fix Modules and Stock Plus

pacbard:
With pull #2585, Stock Bug Fix Modules and Stock Plus should work up to 1.0.4


Olympic1:
Merging with my thanks

PR #2593, opened 2015-11-15
Create SETI-RemoteTech.netkan

Y3mo:
So, the next experiment in terms of Github.

Oh, and thank you Olympic1 for the comma notice in SETI-BalanceMod, the config has gotten quite large and this should be the last change before I update the new SETI-BalanceMod. Will look through the netkan files again before doing that.

Thank you very much!


Y3mo:
Ha, finally.

Olympic1:
Well, we'll see if a new release of SETI-BalanceMod could fix the problem.

As for this file, it looks good and runs locally.

PR #2602, opened 2015-11-16
Fix ScreenMessageHider

Olympic1:


Olympic1:
forgot that it could find files :confused:

PR #2604, opened 2015-11-16
Add FuelTanksTIMES2

Olympic1:
closes #2590


plague006:
Needs a filter for `Thumbs.db`. Unrelated this mod looks like either a troll, satire, or just a jab at minecraft based on the KSP forum thread and the author's name.

Olympic1:
Added the filter.

It's your call to add this "mod" or not

PR #2642, opened 2015-11-20
freeze future updates

Olympic1:
Forgot to freeze ModuleAnimateEmissive to prevent future updates.

Merge KSP-CKAN/CKAN-meta/pull/830 after this


plague006:
Rather than a freeze, this should have been deleted outright.

PR #2644, opened 2015-11-20
Update WernherChecker.netkan

Magion:
Added range of supported KSP versions, bugtracker link and $vref for fixed .version file

### DIFF
--- a/NetKAN/WernherChecker.netkan
+++ b/NetKAN/WernherChecker.netkan
@@ -4,10 +4,18 @@
     "$vref"           : "#/ckan/ksp-avc",
     "identifier"      : "WernherChecker",
     "x_netkan_license_ok": true,
-    "ksp_version_min" : "1.0.4",
-    "ksp_version_max" : "1.0.5",
     "resources" : {
         "repository": "https://github.com/Magion/WernherChecker",
         "bugtracker": "https://github.com/Magion/WernherChecker/issues"
-    }
+    },
+    "x_netkan_override" : [
+        {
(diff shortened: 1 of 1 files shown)


plague006:
$vref extract version compatibility from the .version file so you don't want to hardcode it into the .netkan unless the .version file is in error.

Magion:
I forgot to include the KSP 1.0.4 compatibility info in the .version file, so I decided to hardcode it here just for this version of the mod. I defined the $vref too, because I wanted to have it there prepared for the next update, since the CKAN Wiki says, that it doesn't conflict with the hardcoded value and is just overridden.

If you say it's not a good idea, I should delete the $vref then, correct?

plague006:
Yes, the vref version compatibility info will be overridden by what's hardcoded. The issue is that hardcoding into the .netkan will apply to the current and all future versions unless the .netkan is manually edited.

The best way to override this specific version while allowing future versions to scrape compatibility data from the vref is overriding the specific version (essentially only hardcoding for this specific version).

````
"x_netkan_override" : [
        {
            "version" : "v0.4",
            "delete" : [ "ksp_version" ],
            "override" : {
                "ksp_version_min" : "1.0.4",
                "ksp_version_max" : "1.0.5"
            }
        }
    ]
````

Magion:
Aaah... I didn't know about this option. I originally intended to delete the hardcoded supported KSP versions after the next update, but this is a better solution.

plague006:
Awesome, merging with my thanks.

PR #2650, opened 2015-11-20
Add RealHumans

Olympic1:
Closes #2635


techman83:
Merging with thanks!

plague006:
I feel like a filter for `license` might be prudent since it installs to `TextureReplacer` (as many other mods will try to).

PR #2655, opened 2015-11-21
Create HarderSolarSystem-generic.netkan

stevehead:
This should be the last of my Harder Solar System config sets, for the foreseeable future at least.


Olympic1:
Does this work with Kopernicus v0.5.2 ?

stevehead:
Yes, it's been tested with it.

Olympic1:
That was what I wanted to hear :)
Merging

stevehead:
thanks!

PR #2667, opened 2015-11-22
Moving to stock + recommend CRP

TiktaalikDreaming:
I converted my engine and fuel tank configs to stock fuel, with Module Manager config to convert to original fuels.
I *think* this is the only change needed here.  This NetKAN change is after the change in the mod (because it'd be broken if the sequence was reversed), but only slightly after.   I doubt if CKAN references the updated mod just yet (07:40 UMT 22-11-2015).  If someone looks at this pull request in the next couple of hours, it might be best to wait.


Olympic1:
Merging with my thanks

PR #2681, opened 2015-11-24
Remove "supports" fields from AGX

SirDiazo:
Remove support fields. Not sure why they were added, AGX explicitly supports all actions in any mod.


Olympic1:
Merging with my thanks

PR #2677, opened 2015-11-23
Add Wing Drop Tanks from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of LuizChacon, to add [Wing Drop Tanks](https://kerbalstuff.com/mod/1301/Wing%20Drop%20Tanks) to CKAN.

Please direct questions about this pull request to [LuizChacon](https://kerbalstuff.com/profile/LuizChacon).

PR #2675, opened 2015-11-23
Add The F8 Crusader from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of LuizChacon, to add [The F8 Crusader](https://kerbalstuff.com/mod/1299/The%20F8%20Crusader) to CKAN.

Please direct questions about this pull request to [LuizChacon](https://kerbalstuff.com/profile/LuizChacon).

PR #2669, opened 2015-11-22
Add YT-1300 Stock Light Freighter from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of LuizChacon, to add [YT-1300 Stock Light Freighter](https://kerbalstuff.com/mod/1296/YT-1300%20Stock%20Light%20Freighter) to CKAN.

Please direct questions about this pull request to [LuizChacon](https://kerbalstuff.com/profile/LuizChacon).

PR #2665, opened 2015-11-21
Add B2 Spirit Stealth Bomber from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of LuizChacon, to add [B2 Spirit Stealth Bomber](https://kerbalstuff.com/mod/1294/B2%20Spirit%20Stealth%20Bomber) to CKAN.

Please direct questions about this pull request to [LuizChacon](https://kerbalstuff.com/profile/LuizChacon).

PR #2685, opened 2015-11-24
Add more mods of LuizChacon

Olympic1:
Add 15 more of LuizChacon's craft mods

**Close PR's**
Closes #2528, Closes #2530, Closes #2534, Closes #2579, Closes #2594, Closes #2610, Closes #2611, Closes #2632, Closes #2634, Closes #2636, Closes #2652, Closes #2665, Closes #2669, Closes #2675, Closes #2677

### DIFF
--- a/NetKAN/B1BLancer.netkan
+++ b/NetKAN/B1BLancer.netkan
@@ -1,0 +1,20 @@
+{
+    "spec_version": "v1.4",
+    "$kref": "#/ckan/kerbalstuff/1283",
+    "identifier": "B1BLancer",
+    "license": "WTFPL",
+    "depends": [
+        { "name": "FirespitterCore" },
+        { "name": "BahamutoDynamicsPartsPack"}
+    ],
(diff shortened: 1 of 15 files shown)


plague006:
While Jenkins doesn't properly test the actual installations of each mod, installations are basically tested by netkan.exe and Jenkins does test the metadata. Each passed and that's good enough for me. Merging with my thanks.

Olympic1:
I noticed that if you make a PR on CKAN-meta with 5 files, Jenkins test every file. If you would do the same here, it only tests the last file.

plague006:
Yeah I've looked at the build.sh from both and with my limited knowledge I'm really unclear on why it's happening.

techman83:
I just had a long hard stare at it. I think I know what's happening.

techman83:
About to lose internet for a couple hours, will likely have a patch for KSP-CKAN/CKAN#1479

PR #2686, opened 2015-11-24
Update foldername

Olympic1:
I don't really understand the advanced stuff of CKAN (`[\\d+.]{5}`), but I hope this works


Olympic1:
Found it! Now it searches from 3 to 5 chars (I think)
Ex: So from `v0.4` (3), to `v0.4.1` (5)

plague006:
Looks good, merging with my thanks. (And yes 3-5)

PR #2666, opened 2015-11-22
Add Stock Realism Rebalances from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of WavechaserX, to add [Stock Realism Rebalances](https://kerbalstuff.com/mod/1295/Stock%20Realism%20Rebalances) to CKAN.

Please direct questions about this pull request to [WavechaserX](https://kerbalstuff.com/profile/WavechaserX).

PR #2690, opened 2015-11-24
Add StockRealismRebalances

Olympic1:
Closes #2666

### DIFF
--- a/NetKAN/StockRealismRebalances.netkan
+++ b/NetKAN/StockRealismRebalances.netkan
@@ -5,21 +5,24 @@
     "license": "CC-BY-NC-SA-4.0",
     "depends": [
         { "name": "ModuleManager" },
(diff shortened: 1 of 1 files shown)


plague006:
conflicts RealismOverhaul, HotRockets, RealPlumes, RealPlume-StockConfigs

This also tries to install files from CryoEngines which you should break up into Cryo-core and Cryo-patches and depend on Cryo-Patches. The temporary alternative is leaving as is and creating a ticket that cryo will need to be split at some point in the future.

Lastly for your filter you could use a `filter_regexp` with `\\._.+` to filter all files that start with `._` (would love input from @techman83 or @dbent as my regex skills are weak).

Olympic1:
Tested locally for the filter, and it filted the files with `._`

PR #2699, opened 2015-11-24
Add WW2 Warships from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of Azimech, to add [WW2 Warships](https://kerbalstuff.com/mod/1308/WW2%20Warships) to CKAN.

Please direct questions about this pull request to [Azimech](https://kerbalstuff.com/profile/Azimech).

PR #2700, opened 2015-11-24
Add WW2Warships

Olympic1:
Closes #2699


plague006:
````

This mod is built to use BDArmory and Naval Artillery System (NAS). Forget about using stock weapons on these, you'll only scratch the paint.

Suggested mods:
Hyperedit for teleporting ships to the water. Beware! It's best to temporarily hack gravity when doing so. These ships are heavier than the biggest rocket.

Hangar extender. Unless you fancy swinging 100+ meter ships back and forth with the editor gadgets.
````

plague006:
Fails Jenkins due to BDArmory not being up to date. No problems otherwise. Merging with my thanks.

PR #2704, opened 2015-11-25
New folder structure

Olympic1:


plague006:
Switched `file` to `find` which should cover us in the event that the structure reverts back.

PR #2713, opened 2015-11-26
Update Sigma's mods

Olympic1:
Update the Korpernicus version and use `min_version` instead of `version`


plague006:
Closing for a rebuild, esp to test  #2722

plague006:
I'm going to assume this didn't prove #2722 because it took the build.sh from the commit it's testing (old version). A suivre.

PR #2705, opened 2015-11-25
Update PlanetShine version, + a little edit

valerian:

PR #2715, opened 2015-11-26
Fix PlanetShine

Olympic1:
Fixes #2705

### DIFF
--- a/NetKAN/PlanetShine.netkan
+++ b/NetKAN/PlanetShine.netkan
@@ -7,7 +7,7 @@
     "install": [
         {
             "find": "GameData/PlanetShine",
-            "install_to": "GameData/PlanetShine",
+            "install_to": "GameData",
             "filter": "Config"
         }
     ],
--- a/NetKAN/PlanetShine.netkan
+++ b/NetKAN/PlanetShine.netkan
@@ -1,27 +1,17 @@
 {
-    "spec_version"   : "v1.2",
+    "spec_version"   : "v1.4",
     "identifier"     : "PlanetShine",
     "$kref"          : "#/ckan/kerbalstuff/266",
+    "$vref"          : "#/ckan/ksp-avc",
     "license"        : "Apache-2.0",
     "install": [
         {
-            "file": "GameData/PlanetShine/Icons",
-            "install_to": "GameData/PlanetShine"
-        },
-        {
-            "file": "GameData/PlanetShine/Plugins",
-            "install_to": "GameData/PlanetShine",
-            "filter": "MiniAVC.xml"
-        },
-        {
-            "file": "GameData/PlanetShine/LICENSE",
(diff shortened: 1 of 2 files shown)


valerian:
Thank you @Olympic1 
This pull request does not have a reason to exist anymore, I'm closing it.

valerian:
Thank you for that, I didn't know it could use the ksp-avc version file automatically, that's great!

Olympic1:
Well it is easier for us to let the author choose what version is compatible with which KSP version.

valerian:
I guess I can also remove the "x_netkan_override" section too?

Olympic1:
No, overrides are usefull if an users wants to play an earlier version of KSP so that CKAN knows which mod version is compatible with that KSP version.

valerian:
this is exactly the same information that I provide in the version file

Olympic1:
But CKAN only knows about that version file when this gets merged, so on earlier versions it technically doesn't exist

valerian:
Ok, so I will do a pull request to remove it later I guess, right?
When do you think this will get merged? Because I might release a new version soon actually.

Olympic1:
Well, the "rule" is that a change needs to be looked over by a human to verify the changes

valerian:
Yes of course.

But my point is that now that the version checking will be entirely automated with data from kerbalstuff + the version file, I should not need to bother you anymore with pull requests for the future versions.

What bothers me is that the x_netkan_override requires to be edited for each new released version of PlanetShine.

Or maybe there is something that I didn't understand.

Olympic1:
Basically what an override stanza does is to override things of a mod's version. You have five variants for this, I'll use `v0.3.7.2` as an example:

* Override everything before `v0.3.7.2`
* Override everything before and equal to `v0.3.7.2`
* Override a specific version, so only change things for `v0.3.7.2`
* Override everything after and equal to `v0.3.7.2`
* Override everything after `v0.3.7.2`

So in this case it will override a specific version (`v0.2.3.1`) and says to CKAN to override the max KSP version of `v0.2.3.1` from 1.0.2 to 1.0.4

valerian:
Ok I see, thank you.

NetKAN/PlanetShine.netkan
```diff
@@ -1,27 +1,18 @@
...
     "license"        : "Apache-2.0",
     "install": [
         {
-            "file": "GameData/PlanetShine/Icons",
-            "install_to": "GameData/PlanetShine"
-        },
-        {
-            "file": "GameData/PlanetShine/Plugins",
+            "find": "GameData/PlanetShine",
             "install_to": "GameData/PlanetShine",
```
plague006: Just `GameData` or we get `GameData/PlanetShine/PlanetShine`

Olympic1: Oops, overlooked that

PR #2720, opened 2015-11-26
Create Mk1-1A2Textures4096.netkan

Yarbrough08:


plague006:
Is there a case where users will want the textures installed but not the command module? If so no change is needed but if the textures are exclusive to the command module, we should add it to `depends`.

Yarbrough08:
No, the textures are for the command module only - and the pod should already have a depends for the textures..

plague006:
Without the dependency of the textures to the module when a user uninstalls the module there's a not-insignificant chance they'll forget to uninstall the textures. With the dependency when the CM is removed the textures have to go with it.

Yarbrough08:
Oh, ok - that makes sense. I've made the changes..

plague006:
Looks good, passes local testing, merging with my thanks.

Yarbrough08:
No problem - thank you as well :)

PR #2722, opened 2015-11-27
Lets test _all_ the changed NetKANs - fixes KSP-CKAN/CKAN#1479

techman83:
Function variables are global to the script, we overwrite $f inside the 'inject_metadata' function. Tested with/without changes locally duplicating the issue first and fixing it with below.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -283,18 +283,18 @@ do
 done
 
 # Test all the built files.
-for f in built/*.ckan
+for ckan in built/*.ckan
 do
-    if [ ! -e "$f" ]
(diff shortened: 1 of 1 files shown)


plague006:
Yay! Can't test locally but it looks sane to me. Merging with big thanks and bated breath to see Jenkins do its thing.

plague006:
https://github.com/KSP-CKAN/NetKAN/pull/2723 shows the expected results from this change. Awesome job @techman83, big thanks.

techman83:
:dancers:

PR #2723, opened 2015-11-27
-do not merge- Test changes to see if #2722 works as expected.

plague006:


plague006:
Success. Both installations tested.

PR #2728, opened 2015-11-28
Duplicate folders

Olympic1:
These mods made a `TextureReplacer` folder inside `TextureReplacer`


plague006:
I feel like a filter for `license` might be prudent since it installs to `TextureReplacer` (as many other mods will try to). (copy + paste from #2650)

PR #2732, opened 2015-11-29
Update NetKAN files for my mods to use new forum URLs.

iPeer:
Also added `ksp_version_max` to ClampsBeGone.netkan as it's not longer supported/needed as of 1.0.5.

### DIFF
--- a/NetKAN/AGroupOnStage.netkan
+++ b/NetKAN/AGroupOnStage.netkan
@@ -28,7 +28,7 @@
 	],
 	"resources"			:
 	{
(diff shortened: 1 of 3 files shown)


Olympic1:
This is conflicting because your repo isn't updated with our master branch. According to your repo you're 1344 commits behind KSP-CKAN:master.

iPeer:
Fixing that's going to be a problem, then.

Olympic1:
I made a new PR for updating to the new forum links. I also added an override to replace your `ksp_version_max` so we don't need to hardcode it.

PR #2733, opened 2015-11-30
Update forum links

Olympic1:
Fixes #2732


techman83:
LGTM - Jenkins will be working hard this month!

PR #2568, opened 2015-11-13
Add Joshwoo69's Random mods! from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of joshwoo69, to add [Joshwoo69's Random mods!](https://kerbalstuff.com/mod/1256/Joshwoo69%27s%20Random%20mods%21) to CKAN.

Please direct questions about this pull request to [joshwoo69](https://kerbalstuff.com/profile/joshwoo69).

PR #2740, opened 2015-11-30
Add Joshwoo69sRandommods

Olympic1:
Closes #2568

### DIFF
--- a/NetKAN/Joshwoo69sRandommods.netkan
+++ b/NetKAN/Joshwoo69sRandommods.netkan
@@ -1,0 +1,27 @@
+{
+    "spec_version": "v1.4",
+    "identifier": "Joshwoo69sRandommods",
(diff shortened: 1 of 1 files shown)


plague006:
I'm really not sure what to do with this. There are so many patches for so many different mods. One patch even requires a manual renaming because it breaks saves (`SAS all the command modules!.change this behind dot to cfg`).

I'll think on this and come back to it down the line (probably days from now, possibly longer).

plague006:
Still not sure whether we should split this into its component parts (as some parts are listed in the forum thread as not working), close this until there's a full release (this is currently in-dev), or just merge as is (after getting a Jenkins rebuild).

I leave the decision to you @Olympic1. =)

Olympic1:
Closing as I'm not sure either

PR #2745, opened 2015-12-02
Update MagiCore KSP Version and License

magico13:
Also updated the "homepage" to use the new forum structure.


Olympic1:
Merging with my thanks

PR #2746, opened 2015-12-02
Add new packs

Olympic1:
* Added AlternisRealSolarSystem
* Added BilliardsPlanetPack
* Added OtherWorlds
* Added more conflicts

### DIFF
--- a/NetKAN/AlternisRealSolarSystem.netkan
+++ b/NetKAN/AlternisRealSolarSystem.netkan
@@ -1,0 +1,17 @@
+{
+    "spec_version": "v1.4",
+    "identifier": "AlternisRealSolarSystem",
+    "$kref": "#/ckan/kerbalstuff/1214",
+    "license": "MIT",
+    "depends": [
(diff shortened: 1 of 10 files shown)


techman83:
Well Jenkins did 'Test All the Things!!', but there was one section that shouldn't have passed.

```
1789 [1] ERROR CKAN.RegistryManager (null) - Loaded registry with inconsistencies:

AlternisRealSolarSystem requires RSSTextures but nothing provides it
- AlternisRealSolarSystem 0.1
- KittopiaTech alpha-0-2
- Kopernicus 1:beta-05-2
- ModuleManager 2.6.13
```

It'll be worth breaking that out into a separate branch for fixing. That was a 6mb log file! CKAN may not set an exit code there, so it wouldn't have caused it to bail out and fail the test.

techman83:
Merging, as 'AlternisRealSolarSystem' is a new file and not worth retesting all 12 other files to fix it! Thanks!

PR #2750, opened 2015-12-04
Epoch bump on Scatterer because it has non-netkan friendly versioning.

pjf:


pjf:
Bots are happy, and epoch bumps should always be safe, so I'm merging my own PR. Wish me luck! :)

PR #2753, opened 2015-12-05
Update ModuleRCSFX.netkan

politas:
NathanKell has announced that ModuleRCSFX has been rolled into stock as of 1.0.5, and should not be installed.
http://forum.kerbalspaceprogram.com/index.php?/topic/83259-104-modulercsfx-v42-aug-23-obsolete-in-105/&do=findComment&comment=2278629

PR #2756, opened 2015-12-06
Delete ModuleRCSFX.netkan

plague006:
ModuleRCSFX is deprecated as it's now included in stock KSP.
Closes https://github.com/KSP-CKAN/NetKAN/pull/2753


Olympic1:
Merging with my thanks

PR #2757, opened 2015-12-07
Update AlternisKerbolRekerjiggered-flags.netkan

politas:
Flagmod, so by definition promiscuous versioning should be set.


Olympic1:
Merging with my thanks

PR #2758, opened 2015-12-07
Update QuizTechAeroPack.netkan

politas:
ModuleRCSFX is no longer used. The only reference to it has been commented out.


politas:
Will this change the current version's .ckan file automatically, or will that need to be updated separately?

Olympic1:
Yes, this change will only update the lastest version file of QuizTechAeroPack.

Merging with my thanks

politas:
Looking back through the old versions, it seems like the mod author intended to use it, but never sorted it out properly. The same commented sections throughout every version on CurseForge, though that's only back to v1.2.3

PR #2748, opened 2015-12-03
Update Mk2CockpitExpansion.netkan

Olympic1:
* Updated license
* Uses `find` because of new folder structure
* Added recommendations
* Added dependencies
* Added filter for `Thumbs.db`

idk if we should add the `ExperimentalStuff` folder to install or not.

PR #2760, opened 2015-12-07
Add B9-props

Olympic1:
Part of #2748 so that it doesn't need to install the whole `B9` mod but only the `Props` folder.

Filter out the `Props` folder in B9 and depend on `B9-props`. Also removed dependency on `FirespitterCore` and `RasterPropMonitor-Core` because these are required for `B9-props`.

### DIFF
--- a/NetKAN/B9-props.netkan
+++ b/NetKAN/B9-props.netkan
@@ -1,0 +1,19 @@
+{
+    "spec_version" : "v1.2",
+    "identifier"   : "B9-props",
+    "$kref"        : "#/ckan/kerbalstuff/132",
+    "license"      : "CC-BY-NC-SA-3.0",
+    "depends" : [
+        { "name": "RasterPropMonitor-Core" },
+        { "name": "FirespitterCore" }
+    ],
+    "recommends": [
+        { "name": "B9" }
+    ],
+    "install" : [
+        {
+            "file"       : "GameData/B9_Aerospace/Props",
+            "install_to" : "GameData/B9_Aerospace"
+        }
+    ]
+}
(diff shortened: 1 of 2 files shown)


plague006:
As this stands it conflicts with B9. You'll need to filter the props folder from B9.netkan and add a depends to this.

Olympic1:
Done

Olympic1:
Updated the `B9.netkan` according to #2771 to avoid conflicts when merging after #2771

plague006:
It seems B9 packaged a custom version of `VirginKalactic-NodeToggle`? Whenever possible we like to package mods separately especially since they don't fall under the same authorship or licensing. Ideally we would clean up `VirginKalactic-NodeToggle.netkan` and bring it to current standards while having B9 depend on it. @blowfishpro would installing the `Virgin Kalactic` folder from https://github.com/Greys0/Virgin-Kalactic/releases/download/VK_v1.6.1.1/VirginKalactic.1.6.1.1.zip present any issues to B9?

blowfishpro:
@plague006 that was done based on a discussion I had with Olympic1 on IRC.  The situation is that VirginKalactic has not been and will not be updated by its original author.  B9 still depends on it, but this dependency will be removed in the near future.  The version included in B9 doesn't include any code changes but has been recompiled for good measure.  I wouldn't feel comfortable certifying that the 1.6.1.1 release of VK is compatible with 1.0.5 because I have not tested it extensively.

If there are licensing issues with this approach, I'm open to doing something different.  But I'd prefer not to end up being the maintainer of VirginKalactic, particularly once B9 no longer needs it.

plague006:
While not ideal, if the dependency on VK will be short-term then we can use this workaround.

plague006:
Close/reopen for a rebuild.

plague006:
Merging with thanks.

PR #2769, opened 2015-12-09
Add B9AnimationModules

Olympic1:
New dependency for B9


Olympic1:
Merging into wrong branch

PR #2772, opened 2015-12-09
Delete mods

Olympic1:
`No files found in ... to install!`

Mods maybe won't get updated in the future


plague006:
This seems related to the new B9 release. Both the mods you're freezing came from the B9 .zip. With that in mind, I think we can safely delete both .netkans.

PR #2775, opened 2015-12-09
Update HangarExtender.netkan

politas:
Add original author back, return homepage to forum thread. Bumping epoch in an attempt to get data from Alexw's github.


Olympic1:
Change the epoch back to 1. If KSP-CKAN/CKAN-meta/pull/855 gets merged first and then this PR, it will fix showing the wrong version in CKAN.

Thanks for finding this :+1:

politas:
Epoch back to 1

plague006:
tl;dr there was never a problem with `HangarExtender` per se, rather `FShangarExtender` and `HangarExtender` shared the name `Hangar Extender` and the listing in ckan was `FShangarExtender`. This is addressed by https://github.com/KSP-CKAN/CKAN-meta/pull/857.

This become very quickly apparent when https://github.com/KSP-CKAN/CKAN-meta/pull/855 was merged and the listing still didn't have a `v` in the version number.

PR #2781, opened 2015-12-09
Asteroid Day as dependency

Schnobs:
The new antenna stats require the Asteroid Day antenna for the mariner role.


NathanKell:
Per IRC, sticking on a new branch

PR #2785, opened 2015-12-10
Add X34PersonalTransport

Olympic1:
Closes #2761


plague006:
Passes local with the KerbalFoundries dep ignored.

PR #2788, opened 2015-12-10
Add BetterCrewAssignment

Olympic1:
Closes #2766


plague006:
Disagree, copy-pasting is happiness, copy-pasting is life.

PR #2773, opened 2015-12-09
Add Tantares Fairing Extension from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of legoclone09, to add [Tantares Fairing Extension](https://kerbalstuff.com/mod/1341/Tantares%20Fairing%20Extension) to CKAN.

Please direct questions about this pull request to [legoclone09](https://kerbalstuff.com/profile/legoclone09).

PR #2789, opened 2015-12-10
Add TantaresFairingExtension

Olympic1:
Closes #2773


plague006:
copy-pasting is happiness, copy-pasting is life.

plague006:
Woops, we should filter `Source Code.txt`.

PR #2794, opened 2015-12-11
Update EnhancedNavBall.netkan

politas:
Has been tested and is working with 1.0.5

### DIFF
--- a/NetKAN/EnhancedNavBall.netkan
+++ b/NetKAN/EnhancedNavBall.netkan
@@ -2,9 +2,8 @@
 	"spec_version" : 1,
 	"identifier" : "EnhancedNavBall",
 	"$kref" : "#/ckan/kerbalstuff/411",
-	"ksp_version_min" : "1.0.2",
-	"ksp_version_max" : "1.0.5",
-	"x_netkan_override" : [
+	"x_netkan_override" : 
+	[
 		{
 			"version" : "1.3.6.0",
 			"delete" : [ "ksp_version" ],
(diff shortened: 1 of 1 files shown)


plague006:
We don't hardcode ksp_versions to .netkans unless absolutely necessary, because it will create a situation in the future where our old hardcoded values will overwrite values scraped from the KerbalStuff page. Please use a version-specific override. For an example see https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/Achievements.netkan

politas:
Thanks. I think I saw that netkan_override stuff in one mod, but wasn't sure what it was about.

plague006:
Lines 5+6 should be removed (no hardcoded ksp_versions in the main metadata).
Line 7: `"x_netkan_override" : [` has an opening square bracket but there's no associated closing bracket on line 15/16.

politas:
Ah, that makes more sense. Fixed.

politas:
Jenkins seems to be confused.

NetKAN/EnhancedNavBall.netkan
```diff
@@ -2,5 +2,17 @@
...
+	[
+		{
+			"version" : "1.3.6.0",
+			"delete" : [ "ksp_version" ],
+			"override" : {
+				"ksp_version_min" : "1.0.2",
+				"ksp_version_max" : "1.0.5"
+			}
+		}
+	]
```
plague006: needs comma

politas:
D'oh!

PR #2795, opened 2015-12-11
Update NavballDockingIndicator.netkan

politas:
Updated Title to match Forum Thread, set force_v, bumped KSP version max. This has been tested Ok in 1.0.5 by myself and others.

### DIFF
--- a/NetKAN/NavballDockingIndicator.netkan
+++ b/NetKAN/NavballDockingIndicator.netkan
@@ -2,7 +2,6 @@
     "spec_version"   : "v1.4",
     "identifier"     : "NavballDockingIndicator",
     "$kref"          : "#/ckan/kerbalstuff/23",
-    "title"          : "Navball Docking Alignment Indicator",
     
     "resources" :
     {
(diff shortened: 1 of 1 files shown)


plague006:
There is no `title`, what you likely mean is `name`. With that said, the `name` scraped from KerbalStuff is `Navball Docking Alignment Indicator ` which is verbatim what you want to hardcode the `name` to. There's no need to hardcode something we scrape dynamically. The other 3 changes get a +1 from me.

politas:
Well, it's showing up as "Simple Docking Indicator by mic_e" in my CKAN. Looks like it must be grabbing that from v5. CKAN's column population is truly weird.

plague006:
Lines 5+6 should be removed (no hardcoded ksp_versions in the main metadata). Line 7: `"x_netkan_override" : [` has an opening square bracket but there's no associated closing bracket on line 15/16.

politas:
Can you do a refresh? That's not what I'm seeing at all. Lines 5+6 are:
5:     
6:     "resources" :
And " x_netkan_override" : is on line 22

plague006:
Totally my mistake, was looking at this and #2794 in side by side tabs and mistook them for the same PR. Merging this with my thanks.

PR #2790, opened 2015-12-10
Add Lander Orientation Improvement from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of MWHait, to add [Lander Orientation Improvement](https://kerbalstuff.com/mod/1343/Lander%20Orientation%20Improvement) to CKAN.

Please direct questions about this pull request to [MWHait](https://kerbalstuff.com/profile/MWHait).

PR #2801, opened 2015-12-11
Create LanderControl.netkan

Olympic1:
@matthait made the netkan on his fork but never made a PR, so here is the file from his fork.

Closes #2790

### DIFF
--- a/NetKAN/LanderControl.netkan
+++ b/NetKAN/LanderControl.netkan
@@ -1,0 +1,12 @@
+{
+    "spec_version" : 1,
+    "identifier"   : "LanderControl",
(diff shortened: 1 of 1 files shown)


NetKAN/LanderControl.netkan
```diff
@@ -0,0 +1,12 @@
+{
+    "spec_version" : 1,
+    "identifier"   : "LanderControl",
+    "$kref"        : "#/ckan/kerbalstuff/1343",
+
```
Olympic1: Add `    "x_netkan_license_ok": true,`

PR #2808, opened 2015-12-12
Add WernhersOldBits

Olympic1:
Closes #2806


plague006:
Looks good and merging with my thanks. Took me a while to figure out where you got the CRP recommendation from. Couldn't figure out where the realplume recommend comes from.

PR #2812, opened 2015-12-13
Add RetroFuture space-plane parts from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of amankduffers, to add [RetroFuture space-plane parts](https://kerbalstuff.com/mod/1346/RetroFuture%20space-plane%20parts) to CKAN.

Please direct questions about this pull request to [amankduffers](https://kerbalstuff.com/profile/amankduffers).

PR #2813, opened 2015-12-13
Update RetroFuture.netkan

Olympic1:
* New KS link
* Hardcode previous author
* Bump epoch
* Add suggestions
* Filter FS and `.ckan` file

Original license is `CC-BY-NC-SA-4.0` but new author redistributes this mod as `MIT`, which isn't allowed. I pinged amankd to change this

Closes #2812

### DIFF
--- a/NetKAN/RetroFuture.netkan
+++ b/NetKAN/RetroFuture.netkan
@@ -1,23 +1,29 @@
 {
     "spec_version"   : 1,
     "identifier"     : "RetroFuture",
-    "$kref"          : "#/ckan/kerbalstuff/111",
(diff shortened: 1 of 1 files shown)


plague006:
I think we should hold off on merging until licensing is addressed. Also while CC-BY-NC-SA was the license, http://forum.kerbalspaceprogram.com/index.php?/topic/116664-sharing-stuff-cc-by/ states the license is now CC-BY. But a CC-BY license (if I understand it correctly) means it can't be relicensed under MIT.

Olympic1:
License on KS is canged to CC-BY-NC-SA-4.0

PR #2816, opened 2015-12-13
Update StarTrekTOSSpaceSuits.netkan

politas:
If TextureReplacer works with 1.0.5, so should this texture pack.


politas:
Hmm. I would have thought this would have taken effect by now, but KSP-CKAN/CKAN-meta/blob/master/StarTrekTOSSpaceSuits/StarTrekTOSSpaceSuits-1.2.ckan still shows max KSP version as 1.0.4

Olympic1:
because v1.2 isn't the lastest version, the lastest version on the KS page is v0.4

politas:
Ah, the author has gone backwards. Epoch time!

PR #2819, opened 2015-12-13
Make use of the new 1.16 features

plague006:
1.16 brought new tools, so naturally they need to be used **immediately**! :smile:


plague006:
Added the relevant changes from https://github.com/KSP-CKAN/NetKAN/pull/2475 with the appropriate spec_ver. Closes https://github.com/KSP-CKAN/NetKAN/pull/2475.

Olympic1:
Sure, let's use the new features and merge this in with my thanks.

PR #2820, opened 2015-12-13
Add FutureParts

plague006:
Before merging, need to that answers from author in https://github.com/KSP-CKAN/CKAN-meta/issues/892 are reflected in the metadata.

Jenkins will fail due to `BDArmory` not having 1.0.5-compatibility.

Closes https://github.com/KSP-CKAN/CKAN-meta/issues/892


Olympic1:
Forgot about this one, the answers are given and there shouldn't be any changes so merging.

PR #2762, opened 2015-12-07
Custom Asteroids only works till KSP 0.90

Olympic1:

PR #2825, opened 2015-12-13
Updated Custom Asteroids for 1.2 release.

Starstrider42:
Version override mirrors the "if it doesn't crash KSP, it works" policy expressed in #2762. Custom Asteroids is incompatible with Kopernicus after the latter added its own asteroid feature (compatibility may be restored later; is there a way to express that only a specific version of CA has a conflict?). Other changes should be self-explanatory.

### DIFF
--- a/NetKAN/CustomAsteroids-Pops-Stock-Inner.netkan
+++ b/NetKAN/CustomAsteroids-Pops-Stock-Inner.netkan
@@ -1,4 +1,4 @@
-{
+﻿{
 	"spec_version": 1,
 	"identifier": "CustomAsteroids-Pops-Stock-Inner",
 	"name": "Custom Asteroids (inner stock system data)",
@@ -33,8 +33,7 @@
 			"version" : [ ">= v1.0", "< v2.0" ],
 			"override" : {
 				"comment": "Compatible with any KSP version with the same planets.",
-				"ksp_version_min": "0.18",
-				"ksp_version_max": "1.0.5"
+				"ksp_version_min": "0.18"
 			},
 			"delete" : [ "ksp_version" ]
 		}
--- a/NetKAN/CustomAsteroids-Pops-Stock-Inner.netkan
+++ b/NetKAN/CustomAsteroids-Pops-Stock-Inner.netkan
@@ -4,8 +4,8 @@
 	"name": "Custom Asteroids (inner stock system data)",
 	"abstract": "Adds asteroids inside orbit of Jool",
 	"$kref": "#/ckan/kerbalstuff/251",
-	"release_status": "stable",
-        "x_netkan_license_ok" : true,
+	"$vref" : "#/ckan/ksp-avc",
+	"x_netkan_license_ok" : true,
 	"depends": [ 
 		{
 			"name": "CustomAsteroids",
@@ -18,7 +18,7 @@
 		{ "name" : "AlternisKerbol" }, 
 		{
 			"name": "RealSolarSystem",
-			"comment": "The 6.4Ã— and 10Ã— Kerbol packs are compatible; how to allow them?"
+			"comment": "The 6.4× and 10× Kerbol packs are compatible; how to allow them?"
 		}
 	],
 	"install": [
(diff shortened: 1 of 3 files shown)


Olympic1:
You can delete a conflict with an override stanza. As Kopernicus is the only confluct you can use this:
```
"x_netkan_override": [
    {
        "version": "< v1.2.0",
        "delete": [ "conflicts" ]
    }
]
```

Starstrider42:
Thanks, though once the conflict is resolved  it will hopefully be more useful to express it in the opposite way (only such-and-such range of old versions has a conflict). Now if you'll excuse me, I need to figure out why Git and my text editor are conspiring to corrupt these files...

Olympic1:
Looks good to me and passes local. @plague006 If it's good for you, you can merge this in.

plague006:
> Compatible with any KSP version with the same planets.

Consider that the transition from 1.0.X to 1.1.X is expected to be heavily disruptive. Of the 1134 mods listed on CKAN, 24 are currently marked as compatible with ksp 1.1.X and up (and that number will decrease because as I look at the list several shouldn't be there). Would you consider limiting compatibility to 1.0.X until 1.1.0 is released? Once it is proven to work with 1.1.X we can raise the ceiling to include all 1.1.X ksp releases.

Starstrider42:
@plague006, there are two reasons why that's not necessary:
* That quote is from the config packs, which won't be directly affected by changes to the Unity engine or anything DLL-ish. They still depend on Custom Asteroids, so any 1.1 disruption (which I agree is very likely) will be included there.
* The use of a .version file (via the `$vref` field) gives a much more maintainable way to set the maximum version than manually editing the Netkan with each release (Netkan exists so that it *doesn't* have to be edited). If you look at the .ckan files that are created, the maximum version for all three packages is currently 1.0.5 (and yes, I've done some testing on my end to check that that number comes from the .version, and not from KerbalStuff).

So no, there's no danger of somebody being told it's 1.1 compatible if it isn't.

plague006:
Sorry, I wasn't explicit enough and that's my mistake. I'll line comment so that we're on the same page.

NetKAN/CustomAsteroids.netkan
```diff
@@ -23,17 +26,22 @@
...
+			"override" : {
+				"ksp_version_min": "0.25",
+				"ksp_version_max": "1.0.4"
+			},
+			"delete" : [ "ksp_version" ]
+		},
+		{
+			"version" : "v1.2.0",
+			"override" : {
+				"ksp_version_min": "1.0.0"
```
plague006: This is hardcoding a ksp_version_min with no ksp_version_max.

NetKAN/CustomAsteroids-Pops-Stock-Outer.netkan
```diff
@@ -29,16 +30,15 @@
...
-        }
-    ],
-	"x_maintained_by": "Starstrider42",
-        "x_netkan_license_ok" : true
+	"x_netkan_override" : [
+		{
+			"version" : [ ">= v1.0", "< v2.0" ],
+			"override" : {
+				"comment": "Compatible with any KSP version with the same planets.",
+				"ksp_version_min": "0.18.2"
```
plague006: This is hardcoding a ksp_version_min with no ksp_version_max.

NetKAN/CustomAsteroids-Pops-Stock-Inner.netkan
```diff
@@ -28,15 +28,15 @@
...
-            },
-            "delete" : [ "ksp_version" ]
-        }
-    ],
+	"x_netkan_override" : [
+		{
+			"version" : [ ">= v1.0", "< v2.0" ],
+			"override" : {
+				"comment": "Compatible with any KSP version with the same planets.",
+				"ksp_version_min": "0.18"
```
plague006: This is hardcoding a ksp_version_min with no ksp_version_max.

plague006:
Given that both base mod and configs have versions without ceilings, their dependencies can be met. I'm asking you to not leave them without a ceiling.

NetKAN/CustomAsteroids.netkan
Starstrider42: Are you *sure* it's not "hardcoding a ksp_version_min, not hardcoding ksp_version_max"? I'm pretty sure anything not mentioned in an override statement (e.g., using "delete") gets inherited from the main body.

plague006:
:+1: lost track of this. Wish PRs with commits would show up in notifications but only comments bring them to my attention. Merging with my thanks.

PR #2827, opened 2015-12-14
Add B9-Pwings-Fork.netkan

politas:
Adding compatibility fork of B9AerospaceProceduralParts. Tested locally

### DIFF
--- a/NetKAN/B9-PWings-Fork.netkan
+++ b/NetKAN/B9-PWings-Fork.netkan
@@ -1,6 +1,9 @@
 {
     "spec_version": 1,
     "identifier": "B9-PWings-Fork",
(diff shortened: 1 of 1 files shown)


Olympic1:
Because the download includes a `.version` file, you can use `"$vref": "#/ckan/ksp-avc",` and drop the `ksp_version`. I'll also make a PR to use the new id instead.

politas:
The .version file has not been updated for the fork. It'll all be wrong

Olympic1:
Oh, I overlooked it :disappointed:

Olympic1:
Passes local. Merging with my thanks.

politas:
Good catch on the Recommends, Olympic1, by the way.

PR #2828, opened 2015-12-14
Update B9AerospaceProceduralParts.netkan

politas:
Downgrade version compatibility. This version doesn't work with 1.0.5 See #2827 for compatibility fork.


Olympic1:
I will already merge this and will merge #2827 in later

PR #2833, opened 2015-12-14
Add ChineseLunarLanderWithRover

Olympic1:
Closes #2831, closes #2834


plague006:
Abstract shows up in the CKAN client as `嫦娥 三号，玉兔月球车，中文区`. We should hardcode `name` and `abstract` to avoid any issues using that character set may bring up, and to keep all abstracts in the same language (English).

Olympic1:
The author made a new KS page specifically for English users

plague006:
Pretty cool of him.

PR #2837, opened 2015-12-16
Add Sarbian's mods

Olympic1:
Add AmazingCurveEditor, CrewManifest, CustomBarnKit, GCMonitor, KlockheedMartian-Gimbal, MechJeb2, ModularFlightIntegrator, ModuleManager and SmokeScreen.

The ones I'm not sure about are:
* KlockheedMartian-Gimbal as this file already exists
* MechJeb2 as all the releases are downloaded from curse but @Sarbian began using Jenkins for the last 2 - 3 releases.

### DIFF
--- a/NetKAN/AmazingCurveEditor.netkan
+++ b/NetKAN/AmazingCurveEditor.netkan
@@ -1,0 +1,24 @@
+{
+    "spec_version": "v1.16",
+    "identifier": "AmazingCurveEditor",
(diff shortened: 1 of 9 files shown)


plague006:
This gets a :+1: from me. I created https://github.com/KSP-CKAN/CKAN-meta/pull/902 which will prevent duplicate .ckans being created and also smooths out the versioning format. Once https://github.com/KSP-CKAN/CKAN-meta/pull/902 is merged, this can be merged at will.

Olympic1:
> KM-G will be overwritten

So no need for an epoch bump?

PR #2846, opened 2015-12-20
Give B9 props a separate name

blowfishpro:
The B9 props pack shows up with the same name as the main B9 pack currently.  This prevents it from being installed by itself and prevents it from being removed once installed (except via command line).  This also reduces confusion when looking at mods to be installed.


Olympic1:
Merging with my thanks

PR #2855, opened 2015-12-23
Should work.

DuoDex:


plague006:
For what it's worth, this failed because an `identifier` cannot contain spaces.

PR #2861, opened 2015-12-23
Update SmokeScreen.netkan

politas:
Fix KSP version, Latest is a compatibility release


plague006:
I'm unclear on the effect you're trying to affect with this change. If the current version of SmokeScreen is not backwards-compatible then we should use a ksp_version_min of 1.0.5 and max of 1.0.99 so that we don't have earlier versions having greater compatibility than later versions.

politas:
I've separately submitted a change to the earlier versions to restrict them to 1.0.4 I'm not expecting there to be any KSP version between 1.0.5 and 1.1, so not really worth bothering with such a min/max setting.

plague006:
:+1:

PR #2863, opened 2015-12-24
Update KerbalAircraftExpansion.netkan

politas:
Adding Epoch to fix new version sorting and install example craft now that we can do that. Other exclusions have been removed from download.

### DIFF
--- a/NetKAN/KerbalAircraftExpansion.netkan
+++ b/NetKAN/KerbalAircraftExpansion.netkan
@@ -1,15 +1,20 @@
 {
-	"spec_version" : "v1.4",
+	"spec_version" : "v1.12",
 	"identifier" : "KerbalAircraftExpansion",
 	"$kref" : "#/ckan/kerbalstuff/391",
 	"license" : "GPL-3.0",
 	"x_netkan_license_ok": true,
+	"x_netkan_epoch": "1",
 	"install" : [
 		{
 			"file": "KAX",
 			"install_to": "GameData",
-			"filter" : ["KAX 2.4.2 dds textures.7z", "Example Craft"],
-			"comment": "Example crafts are filtered out becouse its impossible to install ships to subdirectory."
+			"filter" : "Example Craft",
+			"comment": "Example crafts are installed to SPH subdirectory instead."
+		},
+		{
+			"find": "Example Craft",
+			"install_to": "Ships/SPH"
 		}
 	],
 	"depends": [
--- a/NetKAN/KerbalAircraftExpansion.netkan
+++ b/NetKAN/KerbalAircraftExpansion.netkan
@@ -13,7 +13,8 @@
 			"comment": "Example crafts are installed to SPH subdirectory instead."
 		},
 		{
-			"find": "Example Craft",
+			"find_regexp": "Example.Craft.*craft",
+			"find_matches_files": true,
 			"install_to": "Ships/SPH"
 		}
 	],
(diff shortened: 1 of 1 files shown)


politas:
See also KSP-CKAN/CKAN-Meta#911

plague006:
From what I see, we're already offering the correct version `KAX_v2.6.1.0`. I don't see a need for an epoch bump.

politas:
The Max KSP is listing as 1.0.4, though.

plague006:
Oh I see what you're talking about: we have a 1.0.5-compatible version but the "Max KSP Version" column incorrectly displays the compatibility from a previous version. Basically this is https://github.com/KSP-CKAN/CKAN/issues/1327 emerging again. We should probably disable that column until the issue is resolved. I'll merge this and https://github.com/KSP-CKAN/CKAN-meta/pull/911 as a bandaid fix.

politas:
Yep. I've been steadily working through applying bandaids for that issue.

NetKAN/KerbalAircraftExpansion.netkan
```diff
@@ -1,15 +1,20 @@
...
 			"file": "KAX",
 			"install_to": "GameData",
-			"filter" : ["KAX 2.4.2 dds textures.7z", "Example Craft"],
-			"comment": "Example crafts are filtered out becouse its impossible to install ships to subdirectory."
+			"filter" : "Example Craft",
+			"comment": "Example crafts are installed to SPH subdirectory instead."
+		},
+		{
+			"find": "Example Craft",
+			"install_to": "Ships/SPH"
```
plague006: This will not work as we cannot create `Ships/SPH/Example Craft/`. .craft files must be installed directly to `Ships/SPH/`. Thus we filter the Example Craft.

plague006:
Instead of bandaid epoch bumps that make version numbers harder to discern for users and contributors alike, I'd much rather fix the problem with the column itself, disabling it in the interim if necessary. Realizing now what all your previous PRs were doing I wish I had rejected them and disabled the column from the start.

NetKAN/KerbalAircraftExpansion.netkan
politas: Oh. I thought it would just install the files from that directory into Ships/SPH.
I guess we'll need to use "find_matches_files" : "true", and make it a find_regexp "Example.Craft.*craft". I'll test it.

politas:
They aren't all for that reason.

NetKAN/KerbalAircraftExpansion.netkan
politas: Damn. I can't test it. My netkan keeps getting authentication errors.

politas:
Well, that find_regexp works partially. It only installs one of the Example Craft, but correctly puts the craft file into the SPH folder. Looks like we have to do separate finds for each craft. I'm surprised that setting find_matches_files true doesn't allow multiple files with a find_regexp.

politas:
I do think that an epoch bump is called for when a modder changes the version string system they use, which keptin has certainly done in this case.

politas:
Ok, I've reverted the Example Craft installation and will make that a separate change (if I can get it working in a sane way)

plague006:
Spec_ver increase limits CKAN version compatibility unnecessarily and the added `comment` is inaccurate. For simplicity, created https://github.com/KSP-CKAN/NetKAN/pull/2907 which closes this.

politas:
Thanks.

PR #2856, opened 2015-12-23
Add USI Submarine Pack from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of Roverdude, to add [USI Submarine Pack](https://kerbalstuff.com/mod/1361/USI%20Submarine%20Pack) to CKAN.

Please direct questions about this pull request to [Roverdude](https://kerbalstuff.com/profile/Roverdude).

PR #2868, opened 2015-12-24
Create USISubmarinePack.netkan

politas:
Fix #2856.  Note that this depends on a version of USITools that hasn't been properly released yet, but is included in the zip. Should mean this mod does not initially appear as available. Should I do a hand-wrought USITools-0.5.5.0.ckan sourcing from this download?

### DIFF
--- a/NetKAN/USI-Submarine.netkan
+++ b/NetKAN/USI-Submarine.netkan
@@ -2,10 +2,14 @@
     "$kref": "#/ckan/kerbalstuff/1361",
     "spec_version": "v1.4",
     "license": "CC-BY-NC-SA-4.0",
-    "identifier": "USISubmarinePack",
+    "identifier": "USI-Submarine",
     "install": [
       {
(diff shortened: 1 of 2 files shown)


politas:
Please put thin on hold by author.

CovertJaguar:
USI Tools was recently updated, it may be time to give it another look.

pjf:
Checked with RoverDude last night, and confirming that 'blocked on author' is correct here.  The sub is still considered experimental.

plague006:
RoverDude is taking over maintenance of his own metadata, so I'll close this and we can re-open if it's ever needed.

politas:
Yeah. Sorry for jumping over your PR for removing his mods, but Bob was in the chat complaining about not being able to move forward, and I couldn't think how to remove the deletion of two files from your PR.

plague006:
It's all a collaboration. There's no jumping over someone else or anything of the sort.

PR #2876, opened 2015-12-26
Create ThenewCZ2launchsystem.netkan

politas:
Fixes #2866


plague006:
There is no `ProceduralFairings-Textures`, and Jenkins is failing because of that.

politas:
Oh, fudge. I obsessively copy-pasted all the others, then left some excess on the internal copy-paste for that one.

PR #2884, opened 2015-12-29
Force v for AdvancedJetEngine

plague006:
Latest version has a preceding v after never using one. This will ensure that if it's a mistake or later changes back to having no v that versioning will remain the same.


Olympic1:
Merging with my thanks

PR #2901, opened 2016-01-03
RoverMissionsRedux

linuxgurugamer:


linuxgurugamer:
This error makes no sense:

2842 [1] ERROR CKAN.RelationshipResolver (null) - Dependency on SpacetuxSA found, but nothing provides it.
Module SpacetuxSA required, but not listed in index, or not available for your version of KSP

Yet I have a number of other mods which do depend on SpacetuxSA.  Could this be because the netkan for spacetuxsa is old?  it is referencing a max version of 1.0.4, yet it properly updated both on Kerbalstuff AND on the ksp-avc site (the spacetuxsa mod)

PR #2907, opened 2016-01-06
Add epoch to KerbalAircraftExpansion

plague006:
Closes https://github.com/KSP-CKAN/NetKAN/pull/2863


politas:
This really needs to be pushed through. CKAN is currently presenting an outdated version of KAX.

Olympic1:
Merging with my thanks

PR #2908, opened 2016-01-06
ContractConfigurator-GAP

inigmatus:
might as well do this now before too many users are affected.


plague006:
If at all possible we avoid changing identifiers. In this case the identifier doesn't *need* to be changed, so we'll leave it.

PR #2948, opened 2016-01-22
Update TacFuelBalancer ksp version override

politas:
Reports seem positive for this continuing to work in 1.0.5


pjf:
Straightforward change and the bots like it. Thank you so much!

PR #2950, opened 2016-01-23
Little Green Men From Mars: New maintainer and dependencies

TiktaalikDreaming:
Developer is unfamiliar with CKAN, but clearly someone created this very nice complete file for the prerelease LGMFM.  KerbalStuff bot is likely to try creating the netkan file again, but I suspect it'll fail due to having the same name etc.

### DIFF
--- a/NetKAN/LittleGreenMenFromMars.netkan
+++ b/NetKAN/LittleGreenMenFromMars.netkan
@@ -4,7 +4,7 @@
     "license": "CC-BY-NC-4.0",
     "$kref": "#/ckan/kerbalstuff/1412",
     "depends": [
-        { "name": "RealSolarSystem" },
+        { "name": "Kopernicus" },
         { "name": "ModuleManager" }
     ],
     "recommends": [
@@ -14,7 +14,8 @@
         { "name": "NearFutureElectrical" },
         { "name": "NearFuturePropulsion" },
         { "name": "NearFutureSolar" },
-        { "name": "USI-SRV" }
+        { "name": "USI-SRV" },
+        { "name": "InlineBallutes" }        
     ],
     "install": [
         {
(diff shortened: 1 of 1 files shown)


pjf:
Aww yis, I love this mod. :)

From https://kerbalstuff.com/mod/1412 it looks like it needs RSS-Textures as well, which *may* try to load in all of RSS. Checking now.

pjf:
Yup, the RSS textures will try to load all of RSS.

@NathanKell : Do you see any problems if we make the RSS textures "recommend" RSS, rather than depend upon it? We'll still have RSS depend upon the textures, so we're only weakening the relationship in one direction.

@TiktaalikDreaming : Thank you so much for the metadata changes thus far! We almost certainly want to add a `"depends" : "RSSTextures"` line, and a `"conflicts" : "RealSolarSystem"` line. This will make sure users get the textures, but don't also install RSS by accident. (You can add that as an additional commit if you like, or we can make the changes as part of the merge process.)

TiktaalikDreaming:
I can quickly whack in the depends and conflicts.  If I'd thought about the actual install process, I could have shortcut this by doing this to start with.  :-/

NathanKell:
I have no objection. Apologies for the delay.

That said, I'll take this opportunity to mention that, despite the RO netkan (and the RP-0 netkan) recommending RSS, it is (per user reports) often *not* installed as it should be. Halp?

TiktaalikDreaming:
@NathanKell it's be pretty hard to eliminate user-error for a "recommends":.  I'll let someone in the know answer though.  Hell, I'm still struggling with the github interface.  :-)

NathanKell:
Recommends, by definition, means the recommended mod should be auto-selected. That appears to not be the case.

politas:
Well, I can see the recommend for RealSolarSystem in the .ckan file and my registry.json, but indeed, it does not get listed in the "Recommends" list after clicking "Apply" to the changeset.

plague006:
Looks all set. Merging with my thanks.

plague006:
@NathanKell For what it's worth the issue you're bringing up has been around for a while. https://github.com/KSP-CKAN/CKAN/issues/1346

PR #2954, opened 2016-01-24
Add Kerbal Krash Systems - fixed Kerbalstuffbot submission

politas:
Fixes #2909

PR #2964, opened 2016-01-26
Create KerbalKrashSystem.netkan

EnzoMeertens:


politas:
#2954 might be better.

PR #2968, opened 2016-01-26
Update UKS.netkan  - fixed my syntax error, now it will work 

zajc3w:
update to new mod release adding UKS folder


zajc3w:
scrap that i'll start over

PR #2966, opened 2016-01-26
Update UKS.netkan

zajc3w:
Update for new release of UKS
UKS folder has been added to the mod

PR #2939, opened 2016-01-19
Updated UKS, MKS-Lite, and Kolonization for RoverDudes mods (fixes broken installs)

CovertJaguar:
Updated for UKS 0.32

PR #2969, opened 2016-01-26
Update UKS.netkan

zajc3w:
update to new mod release adding UKS folder
correct syntax now


politas:
Still covered in #2939

pjf:
@zajc3w : Thank you so much for this! It looks like we've got the same changes (and more) in #2939, so I'm closing this and #2966 in favour of that. But please don't let this stop you sending us patches in the future; we really love that!

PR #2977, opened 2016-01-29
CryoEngines should keep its epoch!

pjf:
I totally said this in the PR comments, and then failed to apply my
change. The indexer needs this so it doesn't index new CryoEngines
releases as being "old".

For #2957


pjf:
Build error is because the indexer hasn't indexed dependencies yet. Since we want these changes to go out the same time as its dependencies do, and we're building the `.ckan` fine and I've tested this locally, I'm doing that dangerous thing where I merge my own PR. :)

PR #2887, opened 2015-12-30
Add Bear Landing Leg LT-B from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of TheSnowManBear, to add [Bear Landing Leg LT-B](https://kerbalstuff.com/mod/1372/Bear%20Landing%20Leg%20LT-B) to CKAN.

Please direct questions about this pull request to [TheSnowManBear](https://kerbalstuff.com/profile/TheSnowManBear).

PR #2980, opened 2016-01-29
Add BearLandingLegLTB (fixed KerbalStuffBot submission)

politas:
closes #2887

### DIFF
--- a/NetKAN/BearLandingLegLTB.netkan
+++ b/NetKAN/BearLandingLegLTB.netkan
@@ -1,0 +1,7 @@
+{
+    "$kref": "#/ckan/kerbalstuff/1372",
+    "x_via": "Automated KerbalStuff CKAN submission",
+    "spec_version": "v1.4",
+    "license": "All Rights Reserved",
+    "identifier": "BearLandingLegLTB"
+}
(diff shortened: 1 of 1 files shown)


politas:
5478 [1] FATAL CKAN.NetKAN.Program (null) - Error converting value "http:\\www.samuelesnow.com" to type 'System.Uri'. Path 'website', line 1, position 605.

This makes no sense to me. BearPod-MK1 has the exact same homepage, and that's fine. Maybe there's been a change to netkan?

plague006:
BearPod-MK1 has the same homepage yes, but it isn't fine. On http://status.ksp-ckan.org/ you can see it's now failing. In the past it merged with a different address (though still not a valid one as it had the @ symbol in it).

Pinging @dbent and/or @techman83 to have a look into this. In https://github.com/KSP-CKAN/NetKAN/pull/2995 I tried hardcoding the homepage but it seems to fail regardless as soon as it tries converting the website.

dbent:
I'll see if I can make a fix for this today.

plague006:
Following the link, there's a curse download (yay.. manual indexing..)

PR #2985, opened 2016-01-30
Add AtmosphereAutopilot

fat-lobyte:
Hi!

AtmosphereAutopilot is a flight control system designed to replace the stock SAS.
Website: http://forum.kerbalspaceprogram.com/index.php?/topic/124417-105

Please add it to CKAN!


politas:
Looks good, though it might be worth looking at adding a "vref": "#/ckan/ksp-avc", line and a valid KSP-AVC .version file to your download so you can adjust the KSP version easily. Easy enough to modify this later if you choose to do that, so I'll merge this now.

PR #2870, opened 2015-12-24
Add BD Advanced Weapons Pack (BD Add on) from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of Patrick_the_big, to add [BD Advanced Weapons Pack (BD Add on)](https://kerbalstuff.com/mod/1366/BD%20Advanced%20Weapons%20Pack%20%28BD%20Add%20on%29) to CKAN.

Please direct questions about this pull request to [Patrick_the_big](https://kerbalstuff.com/profile/Patrick_the_big).

PR #2990, opened 2016-02-01
Add BD - Advanced Weapons Pack

politas:
Replaces #2870

### DIFF
--- a/NetKAN/BD-AWP.netkan
+++ b/NetKAN/BD-AWP.netkan
@@ -1,0 +1,23 @@
+{
+    "$kref": "#/ckan/kerbalstuff/1366",
+    "spec_version": "v1.4",
+    "license": "MIT",
+    "identifier": "BD-AWP",
+    "install": [
+        {
+            "find": "AWP",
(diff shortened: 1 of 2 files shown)


NetKAN/BD-AWP.netkan
```diff
@@ -0,0 +1,23 @@
...
+    "spec_version": "v1.4",
+    "license": "MIT",
+    "identifier": "BD-AWP",
+    "install": [
+        {
+            "find": "AWP",
+            "install_to": "GameData"
+        },
+        {
+            "file": "INSTALLATION GUIDE.txt",
```
plague006: We can safely exclude the installation of this file. Having it in the install stanza leaves it vulnerable to breaking if the author moves it or renames it.

plague006:
Part of me wants to hardcode an abstract for our users, but I think the mod's name is descriptive enough to make up for its deficiencies.

politas:
At least it has something. Some of the KerbalStuffBot submissions are just awful.

plague006:
Very true.

PR #2994, opened 2016-02-02
Catchup on old merges

plague006:
Minor fixes I spotted in reviewing recent merges. Huge thanks to @politas and @pjf for going through them.

Seeing you guys push some merges fills me with *determination*.


politas:
Nice catches!

PR #2995, opened 2016-02-02
Hardcode website for BearPodMKI

plague006:


plague006:
See #2980 for a discussion on this failure.

PR #2996, opened 2016-02-02
Remove old .netkans in preparation for author-hosted metadata.

plague006:
This shouldn't be merged before a review/confirmation from @BobPalmer that none of these are erroneously included. The removed .netkans should represent all with resultant metadata including "BobPalmer" or "RoverDude".

Of note: `Karbonite`, `UKS`, and `USI-ART` all contain netkan-forced epochs which will need to be accounted for if new metadata uses the same identifiers.


BobPalmer:
Yep, please remove these with the exception of the two FireSpitter ones.  Those can stay here for now.

politas:
Completed as per @BobPalmer instructions in #3009

CovertJaguar:
Oh, I guess this explains why CKAN wasn't picking up yesterday's updates. Hopefully #3018 gets merged soon. =P

PR #2997, opened 2016-02-02
B9-PWings-Fork: do not depend on CrossFeedEnabler

LebedevRI:
CrossFeedEnabler is stock as of 1.0.5

**Apparently, CrossFeedEnabler mod should not be installed at all in 1.0.5**
Instead they should patch Physics.cfg (as RO does) to enable surface-attach crossfeed.

PR #3000, opened 2016-02-03
Restrict ksp version compatibility of CrossFeedEnabler

plague006:
See https://github.com/KSP-CKAN/NetKAN/pull/2997#issuecomment-179290309


politas:
Simple enough. Merging with thanks!

PR #2895, opened 2016-01-01
Add KAS Assembled Areobraking Shield from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of albertKermin, to add [KAS Assembled Areobraking Shield](https://kerbalstuff.com/mod/1378/KAS%20Assembled%20Areobraking%20Shield) to CKAN.

Please direct questions about this pull request to [albertKermin](https://kerbalstuff.com/profile/albertKermin).

PR #3005, opened 2016-02-03
Add KAS Assembled Areobraking Shield (fixed KerbalStuffBot submission)

politas:
Fixes #2895


plague006:
You have `BSD-3-clause` as the license, I wasn't able to find that info to corroborate on the KS page, the KSP forum page, and the license isn't included in the .zip. Point me in the right direction?

politas:
3-clause is the most common form of BSD. If someone just says BSD licence, that's generally what they mean.

plague006:
We can change later if needed, author seems fine with 3-clause at the moment.

PR #3011, opened 2016-02-04
Add Turbo Nisus Parts Pack (fixed KerbalStuff submission)

politas:
Closes #2926


NetKAN/TurboNisu.netkan
```diff
@@ -0,0 +1,7 @@
+{
+    "license": "CC-BY-SA-3.0",
+    "author": [ "TurboNisu", "a__gun:, "Darkona" ],
```
plague006: Colon `:` next to a_gun rather than quote `"`

PR #3012, opened 2016-02-04
Change RSSTextures to recommends

plague006:
Once this is merged #2950 can be re-examined.

### DIFF
--- a/NetKAN/RealSolarSystem.netkan
+++ b/NetKAN/RealSolarSystem.netkan
@@ -13,12 +13,12 @@
     },
     "depends": [
         { "name": "ModuleManager" },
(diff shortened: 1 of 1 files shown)


politas:
Oh, hang on. This is the wrong way around. RO needs RSSTextures, but RSSTextures doesn't need RO

plague006:
Quite right, I'll revert this and open a new PR.

plague006:
Erm wait.. RSSTextures already don't depend on RSS.

plague006:
So in fact there was no change required whatsoever?

politas:
Does RSSTextures depend on Realism Overhaul?

plague006:
Nah, the files haven't been touched in 8 months.

PR #3016, opened 2016-02-05
Pointers for all USI Mods

BobPalmer:

PR #3018, opened 2016-02-05
USI Mods

BobPalmer:


politas:
This is replacing #3016

plague006:
A proper review of this will involve a manual test of each netkan to see what it's generating as a .ckan and ensuring it falls into the correct place in the version continuum CKAN and its users currently have in place. At a quick glance the .netkans in BobPalmer/CKAN/ seem like verbatim copies of the previous .netkans that were hosted here, but I'll need a solid block to time to work through all 19 files. If someone else can get to this before me great, otherwise I'll take a look at this after the weekend.

Olympic1:
@BobPalmer You are deleting `FirespitterCore.netkan` which will cause the mod not to be updated anymore in the future. I don't know if you forgot to update the `.netkan` file or not, but I saw in your `BobPalmer/CKAN` repo that you made a `.netkan` file for FirespitterCore but are not linking to it in this PR.

Here is what the new `.netkan` should look like:
```
{
    "spec_version": "v1.2",
    "identifier"  : "FirespitterCore",
    "author"      : [ "snjo", "RoverDude" ],
    "$kref"       : "#/ckan/netkan/https://raw.githubusercontent.com/BobPalmer/CKAN/master/FirespitterCore.netkan",
    "x_netkan_license_ok": true
}
```

 I also wanted to make a PR on your repo to fix the `FirespitterCore.netkan` (but for some reason I couldn't) because when there is more than one author you need to use `[ ]`.

So change `"author" : "snjo, RoverDude",` to `"author" : [ "snjo", "RoverDude" ],`

Olympic1:
Closing for #3031

PR #3019, opened 2016-02-06
Add FacilityReset (fixed KerbalStuffBot submission with author input)

### LINKED ISSUE #3003: Netkan doesn't create .ckan for my mod.
Here's the .netkan for my mod:

{
    "spec_version" : 1,
    "identifier"   : "FacilityReset",
    "$kref"        : "#/ckan/kerbalstuff/1446",
    "x_netkan_license_ok": true
}

It seems to be downloading the mod into the CKAN\downloads folder OK, but netkan.exe always ends with:

FATAL CKAN.NetKAN.program (null) - Exception has been thrown by the target of an invocation.

I think it has to do with the license which is "Public Domain."  I've tried a couple of variations of the .netkan, such as having the line: x_netkan_license_ok": false, or explicity stating: "license" : "Public Domain".  Still, no dice.

politas:
closes #2999
closes #3003

### DIFF
--- a/NetKAN/FacilityReset.netkan
+++ b/NetKAN/FacilityReset.netkan
@@ -1,0 +1,18 @@
+{
+    "spec_version" : 1,
+    "identifier" : "FacilityReset",
+    "$kref" : "#/ckan/kerbalstuff/1446",
+    "license" : "public-domain",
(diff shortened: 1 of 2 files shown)


politas:
Author submission; looks good to me and the bots, so I'm merging.

PR #3021, opened 2016-02-06
Update KerbalNRAP.netkan

politas:
Extend KSP compatibility as per discussions in mod's forum threads


politas:
Such a minor change, I'm going cowboy.

PR #3031, opened 2016-02-09
Add glue to index RoverDude's mods from his own metadata

pjf:
The `netkan` indexer can use `.netkan` files from other sites, but it
needs glue files to know where to fetch them from. This hange adds that
glue (and is a pretty good reference if we want to do this for other
folks, too).

Tested locally with AlcubierreStandalone and USI-SRV, but I don't have
any reason to expect any of the glue here is incorrect.

Related: #2996, #3009.

### DIFF
--- a/NetKAN/AlcubierreStandalone.netkan
+++ b/NetKAN/AlcubierreStandalone.netkan
@@ -1,0 +1,5 @@
+{
+    "spec_version": "v1.16",
+    "identifier" : "AlcubierreStandalone",
+    "$kref": "#/ckan/netkan/https://raw.githubusercontent.com/BobPalmer/CKAN/master/AlcubierreStandalone.netkan"
+}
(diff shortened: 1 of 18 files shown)


pjf:
That's a weird build result. Jenkins, can we #rebuild?

pjf:
I'm going to have to crash into bed as I'm flying to NYC in a few hours, so I may not be able to see the bots complete. But last run they were giving messages that suggested they were confused about the build environment (the KSP testing area didn't look right), so I still believe my changes are good, and certainly test good locally. :)

pjf:
Error on the [build log](http://ci.ksp-ckan.org/job/NetKAN/1676/consoleFull) are us (transiently?) failing to download Firespitter, which is *not* a change in this PR, just a dependency for some mods. So I'm considering all the tests to have passed here.

pjf:
Merging my own PR(!!) because:

- It means @BobPalmer can bask in self-maintained metadata.
- The changes themselves are simple(ish), and don't change extant files.
- The bots seem to think it's good. :)

Olympic1:
The [build](http://ci.ksp-ckan.org/job/NetKAN/1677/console) just passed

pjf:
D'oh, I hadn't seen #3018 at all! Thank you so much for being on top of things, @Olympic1! <3

PR #3032, opened 2016-02-09
Mods result in 404's

Olympic1:
Remove the netkan files as they result in an 404


Olympic1:
Merging as KSP-CKAN/CKAN-meta/issues/938 is already merged

PR #3036, opened 2016-02-09
Freeze mods

Olympic1:
Freezing `NFT Decaying RTG's` because since v0.6.0 this feature was removed from the download. This will be brought back in the future.

Deleting the other 5 files because these seperate mods were merged into one (https://kerbalstuff.com/mod/1368).


Olympic1:
This failed because it still wants to create a release for the mods that we'll freeze

politas:
If we're not expecting them to ever be updated, why are we not just deleting the netkans to clean up the directory?

Olympic1:
Well I can delete the netkans except for the NFT Decaying RTG's one, because Nertea hasn't confirmed to me if he will bring it back or not.

Olympic1:
Nertea said he will bring it back in the future, so freezing the file untill the feature is brought back.
Deleted the other netkans as they were merged into one mod.

This can be merged.

PR #3037, opened 2016-02-09
Filter conflicting files

Olympic1:
Mod contains two files that conflict with each other. Mod uses an RPM IVA as default but it says in the `Readme.txt`:

> For non-Raster Prop Monitor IVA copy and replace the folders inside "Stock IVA" to the mod location.

So filter out the `StockIVA` and add a dependency for RPM untill the author has fixed this. @politas already mentioned this to the author on the forums.


politas:
Yep, probably should have done this when I added it earlier.

PR #3039, opened 2016-02-10
Create NetKAN for Strategia

jrossignol:
New NetKAN for Strategia (released version to be created shortly, meanwhile everything in GitHub is a pre-release).


politas:

Issue #2943, opened 2016-01-21
Add NoFuel mod request 


leocadle:
[nofuel.netkan.zip](https://github.com/KSP-CKAN/NetKAN/files/98708/nofuel.netkan.zip)
Please include my mod in CKAN

Cheers,
Leo.

PR #3048, opened 2016-02-11
Added nofuel mod

Zane6888:
Added NoFuel mod as requested by @leocadle in #2943

### DIFF
--- a/NetKAN/nofuel.netkan
+++ b/NetKAN/nofuel.netkan
@@ -9,7 +9,14 @@
     "author"         : "Selfish Meme",
     "resources"      : {
         "homepage"   : "http://forum.kerbalspaceprogram.com/index.php?/topic/130365-105-nofuel-all-tanks-defueled"
-    },
+    },    
+    "install": [
+        {
+            "file": "nofuel",
+            "install_to": "GameData",
+            "filter": ".DS_Store"
+        }
+    ],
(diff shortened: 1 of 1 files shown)


politas:
The "still figuring out CKAN release" is not actually a release, so Netkan won't grab that file. The latest version is the 0.5 one which doesn't have a "nofuel.zip", just a source zip inside of which the folder is called "nofuel-0.5", so that's not going to install. We should also add a filter for the ".DS_Store" file so that doesn't get installed.

Also, the homepage is incorrectly spelled and has an extra www that doesn't work - it should be "http://forum.kerbalspaceprogram.com/index.php?/topic/130365-105-nofuel-all-tanks-defueled"

Zane6888:
I ran the nofuel.netkan through netkan.exe and it gave me a .ckan for v0.6. Netkan didn't seem to care that it wasn't tagged as 'Latest Release' on github (v0.5 and v0.6 are actually the same commit).
The .ckan installed fine but ofc the install contained the ".DS_Store".

politas:
If you add this, it'll filter out that excess file:
"install": [
    {
        "find": "nofuel",
        "install_to": "GameData",
        "filter": ".DS_Store"
    {
]

politas:
Merging with thanks.

PR #2941, opened 2016-01-20
Add Stockalike Mining Extension from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of Suicidalinsanity, to add [Stockalike Mining Extension](https://kerbalstuff.com/mod/1415/Stockalike%20Mining%20Extension) to CKAN.

Please direct questions about this pull request to [Suicidalinsanity](https://kerbalstuff.com/profile/Suicidalinsanity).

PR #3052, opened 2016-02-12
Set up InsanityAerospace Agency mod and add StockalikeMiningExpansion

politas:
StockalikeMiningExpansion shares some files and a folder with Mk2Expansion. This change pulls the shared files out into a separate dependency, which Mk2Expansion also provides.

Shared folder is /GameData/Mk2Expansion
Shared files are /GameData/Mk2Expansion/Agencies

StockalikeMiningExpansion is a much smaller mod, so I have sourced the Agencies folder from there.

Closes #2941

### DIFF
--- a/NetKAN/Mk2Expansion.netkan
+++ b/NetKAN/Mk2Expansion.netkan
@@ -18,7 +18,7 @@
         {
             "find"       : "GameData/Mk2Expansion",
             "install_to" : "GameData",
-            "filter"     : "Agencies"
+            "filter"     : [ "Agencies", "Thumbs.db" ]
         },
         {
             "find"        : "VAB",
--- a/NetKAN/Mk2Expansion.netkan
+++ b/NetKAN/Mk2Expansion.netkan
@@ -8,17 +8,18 @@
         { "name": "FirespitterCore" },
         { "name": "ModuleManager", "min_version": "2.5.3" },
         { "name": "CommunityResourcePack" },
-        { "name": "InterstellarFuelSwitch-Core" },
-	{ "name": "InsanityAerospace" }
+        { "name": "InterstellarFuelSwitch-Core" }
     ],
     "suggests": [
 	{ "name": "FerramAerospaceResearch" }
     ],
+    "provides": [
+	{ "name": "InsanityAerospace" }
+    ],
     "install" : [
         {
             "find"       : "GameData/Mk2Expansion",
-            "install_to" : "GameData",
-	    "filter"     : "Agencies"
+            "install_to" : "GameData"
         },
         {
             "find"        : "VAB",
--- a/NetKAN/Mk2Expansion.netkan
+++ b/NetKAN/Mk2Expansion.netkan
@@ -8,7 +8,8 @@
         { "name": "FirespitterCore" },
         { "name": "ModuleManager", "min_version": "2.5.3" },
         { "name": "CommunityResourcePack" },
-        { "name": "InterstellarFuelSwitch-Core" }
+        { "name": "InterstellarFuelSwitch-Core" },
+	{ "name": "InsanityAerospace" }
     ],
     "suggests": [
 	{ "name": "FerramAerospaceResearch" }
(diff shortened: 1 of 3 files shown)


politas:
Failing because "NetKAN metadata/NetKAN/dummy_ksp is not a vaild install"

Olympic1:
Filter out `Thumbs.db` in `InsanityAerospace.netkan` and `Mk2Expansion.netkan`

NetKAN/Mk2Expansion.netkan
```diff
@@ -13,6 +13,10 @@
     "suggests": [
 	{ "name": "FerramAerospaceResearch" }
     ],
+    "conflicts": [
```
plague006: This creates an implicit conflict between `Mk2Expansion` and `StockalikeMiningExtension`.

`StockalikeMiningExtension` depends `InsanityAerospace` and `Mk2Expansion` conflicts `InsanityAerospace`.

Instead, remove this `conflicts` and `provides` and instead `depends InsanityAerospace` and filter the Agencies folder (as in `StockalikeMiningExtension`).

politas:
I can't see any situation where this would be a problem. If a user has neither mod and selects Mk2Expansion and StockalikeMiningExpansion, they'll install as a pair, and InsanityAerospace won't be required. If they have Mk2Expansion installed and select StockalikeMiningExpansion (which will be a common scenario in the short term), again, InsanityAerospace won't be required. If the have StockalikeMiningExpansion installed and select Mk2Expansion, they'll hit a conflict which can be easily fixed by uninstalling InsanityAerospace. With this setup, no one is ever forced to download a zip file they don't need.

plague006:
> which can be easily fixed by uninstalling InsanityAerospace

To you and I, it's very easy to figure out why. But CKAN won't actually surface the reason for the conflict, only that one exists and that's the real issue. If someone has 100 mods installed, they're actually unlikely to see the conflict without having to dig and search. If we don't have to, it's best not to put the burden on users to learn about relationships and conflicts, but rather to have things work seamlessly.

This is an install of Mining Extension with Insanity chosen to fill the dependency. Now imagine a few days, couple weeks, or couple months later the user wants to install Mk2 expansion and they have no idea which one of their mods is the issue, or even what the issue is. They finally do track down that Insanity is the cause, and it's telling them they have to uninstall Mining as well? It's reasonable for them to think Mining can't coexist with Mk2 Expansion, or at least they'll have to dig around and find out why things are the way they are. The point of CKAN is to make things as easy as they can be.

Screenshot of CKAN

politas:
Why is it telling them they need to uninstall Mining? If they're installing a new mod which provides Insanity, the only conflict should be with Insanity.

plague006:
> I'm worried that current users of Mk2Expansion will be in a weird situation where an installed mod suddenly "needs a dependency" without a new release. Also, I don't think we should force users to download a zip file they have no need of.

While I'm sensitive to the data caps of our users as well as installing extra mods/dependencies, I don't think the 852kB footprint of StockalikeMiningExtension will make or break anyone, and this is far from the first time a mod has been broken into parts.

> How about we install it like I have it here

No, I won't endorse a PR that throws up unhandled exceptions in CKAN. See: http://www.mediafire.com/watch/g0wb6hl3amiaz1k/Insanity.mp4

Not only is Insanity Aerospace not highlighted in red showing that it's in conflict, but trying to uninstall it causes an exception.

If you want to tinker around with it to see for yourself how it works, feel free to test it out on the p6_test branch of CKAN-meta.

politas:
Ah, didn't realise CKAN was stuffing up that badly. Ok, I've put it back to how I originally set it up.

plague006:
The last thing should be resolving the thumbs.db filtering as @Olympic1 (mentioned)[https://github.com/KSP-CKAN/NetKAN/pull/3052#issuecomment-183390041] and then it should be good to go.

plague006:
Forum thread http://forum.kerbalspaceprogram.com/index.php?/topic/130325-105-stockalike-mining-extension-08-release-12016-wip/

politas:
Ref #3091, the Mk2Expansion commits in here are going to conflict now.

politas:
The whole need for the creation of this dependency has been removed, as per forum thread: http://forum.kerbalspaceprogram.com/index.php?/topic/109145-105-mk2-expansion-v1611-update-11132015/&do=findComment&comment=2414116

PR #2938, opened 2016-01-19
Add Vaporo's Blue Galaxy Skybox from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of Vaporo, to add [Vaporo's Blue Galaxy Skybox](https://kerbalstuff.com/mod/1413/Vaporo%27s%20Blue%20Galaxy%20Skybox) to CKAN.

Please direct questions about this pull request to [Vaporo](https://kerbalstuff.com/profile/Vaporo).

PR #3053, opened 2016-02-12
Add Vaporo's Blue Galaxy Skybox (fixed KerbalStuffBot submission) 

politas:
Closes #2938


politas:
Turns out the EnvMap is no longer used, so no requirement for the second option.

Olympic1:
Requires `TextureReplacer`

plague006:
A thought on the whole conflicting with other skyboxes thing, do Kopernicus-based mods usually change the Kerbin skybox? (the ones that change the planet Kerbin anyways) and if they do, we can probably cover our bases by conflicting Kopernicus.

Olympic1:
Kopernicus can't change the skybox

politas:
Yes, I ran a few searches looking for skybox-related metadata before I submitted. I think we're good to go.

PR #3056, opened 2016-02-12
Add The Cube (fixed KerbalStuffBot submission)

politas:
Closes #2889


plague006:
> Recommended Mods: [1.0.5] Kerbal Joint Reinforcement v3.1.4 6/22/15

Olympic1:
oops, how could I miss such a big "banner"

politas:
Ah, I think I saw that when I was first looking into this, but I had decided to wait for the next release, since the metadata is going to be different, then. Then I decided to go ahead with getting this version in and didn't go back for a thorough look.

PR #2961, opened 2016-01-25
Add AUC Aeronautic technoloy from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of Haryr, to add [AUC Aeronautic technoloy](https://kerbalstuff.com/mod/1433/AUC%20Aeronautic%20technoloy) to CKAN.

Please direct questions about this pull request to [Haryr](https://kerbalstuff.com/profile/Haryr).

PR #3059, opened 2016-02-12
Add AUC Aeronautic technoloy (fixed KerbalStuffBot submission)

politas:
Closes #2961

### DIFF
--- a/NetKAN/AUCAeronautictechnology.netkan
+++ b/NetKAN/AUCAeronautictechnology.netkan
@@ -7,6 +7,15 @@
         { "name": "ModuleManager" },
         { "name": "CommunityResourcePack" }
     ],
+    "recommends": [
+        { "name": "RasterPropMonitor" },
+        { "name": "ASET Avionics Pack" },
+        { "name": "ASET Props Pack" }
+    ],
+    "suggests": [
+        { "name": "B9-PWings-Fork" },
+        { "name": "AdjustableLandingGear" }
+    ],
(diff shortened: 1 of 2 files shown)


Olympic1:
The `ADF3internal.cfg` requires `RasterPropMonitor`, `ASETAvionics` and `ASETProps`.

As for the craft files, `adf3 starchild bomber.craft` and `adf3 starchild.craft` require `B9-PWings-Fork` and `AdjustableLandingGear`.

Forum thread says:
> Not included in the pack but required for the IVA:
- RasterPropMonitor
- ASET Avionics Pack v 1.0 
- ASET Props Pack

> PS: the craft that are in the zip file, only the starchild stock is purly stock, for the other craft, you will need: B9 procedural wings and BD adjustable landing gear.

politas:
Ok, so if something is required only for the IVA or for individual craft files, do we put them in depends or recommends? I lean towards recommends, since the mod will work ok (with limitations) without them. If you don't have RPM, it won't have the nice IVA, but a lot of people never use IVA, anyway.

plague006:
From the forum thread http://forum.kerbalspaceprogram.com/index.php?/topic/130683-wip-105-auc-aeronautic-mods-001/ it's availble on curse http://kerbal.curseforge.com/projects/auc-aeronautic-technoloy

politas:
Replaced by KSP-CKAN/CKAN-meta#967

PR #3060, opened 2016-02-12
Add WBIforETT (fixed KerbalStuffBot submission)

politas:
Closes #2962


Olympic1:
Merging with my thanks

plague006:
We should definitely change `supports` to `suggests` or `recommends`. This is a compatibility patch and if someone (for whatever reason) installs it without the mods it patches, we're not presenting users with a suggestion to install the patched mods.

PR #2976, opened 2016-01-28
Add The United Kerbin Church of our Lord and Savior Scott Manley's church (UKCLSSM) from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of amankduffers, to add [The United Kerbin Church of our Lord and Savior Scott Manley's church (UKCLSSM)](https://kerbalstuff.com/mod/1438/The%20United%20Kerbin%20Church%20of%20our%20Lord%20and%20Savior%20Scott%20Manley%27s%20church%20%28UKCLSSM%29) to CKAN.

Please direct questions about this pull request to [amankduffers](https://kerbalstuff.com/profile/amankduffers).

PR #3065, opened 2016-02-13
Add UKCLSSM (fixed KerbalStuffBot submission)

politas:
Closes #2976


politas:
14319 [1] FATAL CKAN.NetKAN.Program (null) - Unexpected character encountered while parsing value: <. Path '', line 2, position 1.

Can anyone work out what NetKAN is complaining about?

plague006:
KerbalStuff is presently having issues, likely resulting in this failure. In spite of multiple attempts I can't download the .zip to review the metadata. Will try again another day.

politas:
#rebuiild

politas:
Nope. This was failing well before KerbalStuff started having issues, and it still is.

Olympic1:
I think the problem is caused by that there is no `KSP Version` on the KS page which means that our bot can't extract the compatible version for the mod. It doesn't extract the version from the changelog page iirc.

plague006:
@politas if you have an alternate source for this, feel free to reopen.

PR #3067, opened 2016-02-13
Add StrutPlus (fixed KerbalStuffBot submission)

politas:
Closes #3008


plague006:
A .version file is included.

politas:
So we should have a vref rather than rely on KerbalStuff?

plague006:
Yep, that's the [convention](https://github.com/KSP-CKAN/NetKAN/search?utf8=%E2%9C%93&q=kerbalstuff+vref).

PR #3068, opened 2016-02-13
Add Snack Parts Updated (fixed KerbalStuffBot submission)

politas:
Closes #3022


politas:
Damn, another one of these weird ones:
FATAL CKAN.NetKAN.Program (null) - Unexpected character encountered while parsing value: <. Path '', line 0, position 0

Is this something about the zip file?

plague006:
KerbalStuff is presently having issues, likely resulting in this failure. Force a rebuild at a later time and it should be fine. JSON is valid and manually looking at how this should install/its relationships looks fine. Good to go once bots give a go-ahead.

PR #3077, opened 2016-02-15
Update to point to the GitHub repository

TheFarix:
Update to point to the GitHub repository instead of the now defunct KerbalStuff.

### DIFF
--- a/NetKAN/TACLS-Declutter.netkan
+++ b/NetKAN/TACLS-Declutter.netkan
@@ -1,12 +1,18 @@
 {
     "spec_version" : 1,
     "identifier"   : "TACLS-Declutter",
-    "$kref"        : "#/ckan/kerbalstuff/1216",
-	"$vref"        : "#/ckan/ksp-avc",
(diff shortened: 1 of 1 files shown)


dbent:
Hi, a few things:

- To fix the break you need a comma after `resources` and before `depends`.
- The version will be pulled from the GitHub tag which currently contains a leading `v`, previous versions of the mod did not contain that. Use `"x_netkan_version_edit": "^[vV]?(?<version>.+)$"` to extract the pure numerical version from the tag.
- The `resources.repository` will be set automatically, no need to hardcode it.
- You'll probably want to update `resources.homepage`, due to the forum migration the link currently points to the main TAC Life Support thread, rather than any particular post.

Since you're updating your own metadata, you may want to consider self-hosting it so you can make changes on your own. Let me know if that's something you're interested in.

dbent:
Build breaking because it can't download dependencies (cause Kerbal Stuff). Checked manually, looks fine, merging. Thanks and let me know if there are any issues.

PR #3080, opened 2016-02-15
Replaced Kerbalstuff by Github

sswelm:
since, KerbalStuff has died, it needs to update from somewhere else

### DIFF
--- a/NetKAN/KSPInterstellarExtended.netkan
+++ b/NetKAN/KSPInterstellarExtended.netkan
@@ -1,7 +1,10 @@
 {
     "spec_version"       : 1,
     "identifier"         : "KSPInterstellarExtended",
-    "$kref"              : "#/ckan/kerbalstuff/791",
+    "$kref"              : "#/ckan/github/sswelm/KSPInterstellar",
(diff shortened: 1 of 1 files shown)


NetKAN/KSPInterstellarExtended.netkan
```diff
@@ -1,7 +1,10 @@
 {
     "spec_version"       : 1,
     "identifier"         : "KSPInterstellarExtended",
-    "$kref"              : "#/ckan/kerbalstuff/791",
+    "$kref"              : "#/ckan/github/sswelm/KSPInterstellar",
+    "$vref" 		 : "#/ckan/ksp-avc"
```
pjf: Oops, we're missing a comma at the end of this line! Can we add one in?

pjf:
Thank you! I think we'll see a lot of these over the next few days.

We seem to be missing a comma on line 5, any chance you can add it in?

dbent:
You also want to add a `resources.homepage` as that will no longer be automatically populated.

sswelm:
I added the comma and made a new pull request

PR #3085, opened 2016-02-16
External NetKAN files for all B9-related mods

blowfishpro:
They're all in their respective Github repositories now

### DIFF
--- a/NetKAN/B9-props.netkan
+++ b/NetKAN/B9-props.netkan
@@ -1,30 +1,6 @@
 {
-    "spec_version" : "v1.2",
-    "identifier"   : "B9-props",
-    "name"         : "B9 Aerospace Props",
-    "$kref"        : "#/ckan/kerbalstuff/132",
-    "license"      : "CC-BY-NC-SA-3.0",
-    "depends" : [
-        { "name": "RasterPropMonitor-Core" },
-        { "name": "FirespitterCore" }
-    ],
-    "recommends": [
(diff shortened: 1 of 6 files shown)


blowfishpro:
Any idea why CI is failing here?  It seems to be trying to install 5.4.0 from Kerbalstuff, but nothing points to Kerbalstuff anymore.  I ran netkan.exe on all the files locally (both the ones I pushed to git and the ones in this PR which link to them) and they all succeeded.

plague006:
For better or worse (better in this case) ckan.exe even with the -c argument will try to install the newest version of the mod. In this case it's taking `R5.4.0` as < `6.0.0`. The fix is to add an epoch to each of the affected mods.
`"x_netkan_epoch": "1",`

plague006:

blowfishpro:
I noticed that the NetKAN bot updated B9AerospaceHX and B9AerospaceLegacy but not B9.  Is there a run log somewhere I could look at to determine the problem?

Olympic1:
@blowfishpro Add an `abstract` to B9 and B9-props.

Olympic1:
@blowfishpro Our bot added your last two mods to CKAN

blowfishpro:
@Olympic1 I added `abstract` this morning, looks like it worked.  Thanks for the help.

Olympic1:
@blowfishpro I'm here to help

PR #3090, opened 2016-02-16
Update kOS netkan

hvacengi:
Update kOS.netkan to point at a netkan file managed by the kOS team.


Olympic1:
Looks good, merging with my thanks

PR #3092, opened 2016-02-16
Renaming for spacedock.info

Y3mo:
SETI-BalanceMod from kerbalstuff renamed to SETI-Rebalance for spacedock.info, I ll adjust my metadata as soon as I have the new spacedock id. 
I can not find any dependencies or references within netkan on the former "SETI-BalanceMod" identifier. 
And with kerbalstuff gone, this is the ideal point in time for a name change.

### DIFF
--- a/NetKAN/SETI-BalanceMod.netkan
+++ b/NetKAN/SETI-BalanceMod.netkan
@@ -1,6 +1,0 @@
-{
-    "spec_version": "v1.2",
-    "identifier": "SETI-BalanceMod",
-    "$kref": "#/ckan/netkan/https://raw.githubusercontent.com/Y3mo/SETI-BalanceMod/master/SETI-BalanceMod.netkan",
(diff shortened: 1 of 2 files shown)


plague006:
>  this is the ideal point in time for a name change.

Not really, existing users won't have an upgrade path so you'll split your userbase. If you're dead-set on changing the `identifier` the best time for that is when 1.1 is released and everyone will assume they need to start from scratch with mods.

Y3mo:
Hm, makes sense.
I did not plan to update the mod before ksp 1.1 and wanted to prevent users from ckan installing it with the current identifier when 1.1 hits and I m not available to make the change.
But since 1.1 will break a lot of mods anyway I guess changing it then will be better than removing the update option now (without knowing a 1.1 release date, might still be weeks away).

Pull request can be closed.
Thank you for your input!

Dazpoet:
Closed by authors request

PR #3093, opened 2016-02-17
kerbalstuff to spacedock

TiktaalikDreaming:
completely unsure of the kref.  :-)

### DIFF
--- a/NetKAN/WernhersOldBits.netkan
+++ b/NetKAN/WernhersOldBits.netkan
@@ -1,7 +1,7 @@
 {
     "spec_version": 1,
     "identifier": "WernhersOldBits",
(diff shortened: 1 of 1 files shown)


plague006:
A `kref` is specifically for automatic indexing via netkan. If you want to plug in a new download link you want to edit https://github.com/KSP-CKAN/CKAN-meta/blob/edc92390fbee48a171b7211e5ba306af91a40337/WernhersOldBits/WernhersOldBits-0.5.ckan

TiktaalikDreaming:
No, you're right.  It's just moving, not sure how best to move it is all.  Probably worth waiting a while till the spacedock people sort stuff out and then re-think it.

PR #3112, opened 2016-02-18
ZeroPointInlineFairings: Update to spacedock

NecroBones:


Olympic1:
runs fine local, merging

PR #3116, opened 2016-02-18
DiscontinuedParts: update to spacedock

NecroBones:


plague006:
:+1: 

Crazy Jenkins log, but it builds fine nonetheless.

techman83:
@plague006 weird, but I noted there was some minor time drift so I've put ntp on the host.

NecroBones:
NTP is a good thing. :)

Issue #2179, opened 2015-08-26
Mods requiring overwriting to complete (incl. KerbinSide modules)


plague006:
To be clear: these mods can all be integrated in "fragile" fashions where there are tons of filters and files explicitly installed, the problem with those is small changes to the files/file structure breaks the whole scheme. These mods are waiting on https://github.com/KSP-CKAN/CKAN/issues/1311 to have easier/more stable install methods.

  Original PR   --   KerbalStuff/Forum link
- [x] #1787 - [Kerbol 4X](https://kerbalstuff.com/mod/970/Kerbol%204X)
- [ ]  #2091 - [Kerbin-Side Ground-Control](https://kerbalstuff.com/mod/1073/Kerbin-Side%20Ground-Control)
- [ ] #2166 - [KerbinSideCORE](https://kerbalstuff.com/mod/1099/KerbinSide%20CORE)
- [ ] #2298 - [KerbinSide Kampus](https://kerbalstuff.com/mod/1160/KerbinSide%20Kampus)
- [ ] #2335 - [KerbinSide Skyways](https://kerbalstuff.com/mod/1173/KerbinSide%20Skyways)
- [ ] #2346 - [KerbinSide KSC2](https://kerbalstuff.com/mod/1178/KerbinSide%20KSC2)
- [x] KSP-CKAN/CKAN-meta#351 - [RemoteTech_XF](http://forum.kerbalspaceprogram.com/threads/108669-0-90-RemoteTech-XF-RemoteTech-v-1-6-3)
- [x] #2395 - [tkw HotRockets fx texture](https://kerbalstuff.com/mod/1200/tkw%20HotRockets%20fx%20texture)
- [x] #2348 - [halo scorpion tank](https://kerbalstuff.com/mod/1179/halo%20scorpion%20tank)
- [ ] #3245 - [Versatile Toolbox System](https://github.com/KSP-CKAN/NetKAN/pull/3245) (overwriting can optimize this install)

plague006:
Per the thread RemoteTech_XF is being integrated into RT proper.

plague006:
tkw HotRockets fx texture and halo scorpion tank were KS mods

plague006:
Kerbol 4X was KS-only.

plague006:
That leaves kerbinside and Versatile Toolbox which can have their installs improved with overwriting. We all know about kerbinside and VTS hasn't updated for 1.1. Closing this and if we ever need later on we can search to find it easily enough.

PR #3118, opened 2016-02-18
Add Kerbin-Side Complete from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of AlphaAsh, to add [Kerbin-Side Complete](https://spacedock.info/mod/150/Kerbin-Side%20Complete) to CKAN.

Please direct questions about this pull request to [AlphaAsh](https://spacedock.info/profile/AlphaAsh).


plague006:
ref https://github.com/KSP-CKAN/NetKAN/issues/2179

politas:
Ref #3155

politas:
I think that rather than leaving an out of date KerbinSide as the only available version, we should just make all the KerbinSide* mods conflict with all other KerbinSide mods, and we can remove the conflicts when we have the overwriting stuff worked out (if ever!)

Dazpoet:
I agree to that solution and will merge this since it seems like the least bad solution for now. Seeing as KerbinSide isn't as widespread I hope this won't lead to a `PlanetPack` level clusterfuck down the line.

PR #3123, opened 2016-02-18
Update BlackRibbon.netkan

Shuudoushi:
Added spacedock and changed the name to reflect what's on spacedock


Olympic1:
Merging with my thanks

PR #3139, opened 2016-02-19
Update and rename GAP.kerbalstuff to GAP.netkan

inigmatus:
my spacedock id is 70


plague006:
Looks good, fails on a KS recommend but tests fine locally.

PR #3143, opened 2016-02-19
Update Deepfreeze

Olympic1:


Dazpoet:
This fails because it's trying to download the old REPOSoftTech-Agencies which points to KS to satisfy its dependency. Since that is fixed in this PR and the other things are just recommends etc I'll classify this as strictly better than what we have currently and roll with it.

PR #3144, opened 2016-02-19
Update DMagic's mods

Olympic1:
Update all of DMagic's mod that were on KerbalStuff


Olympic1:
Runs fine locally

Dazpoet:
It seems the problem is here ``Downloading "https://kerbalstuff.com/mod/99/Contract Window +/download/6.0" (libcurl)`` and why that's happening... I have no idea

Dazpoet:
Ok some further study shows that it does correctly download and install CWP while it's testing that netkan but then it moves over to Orbital sicence and seems to fall back to using the old CKAN-meta version of CWP which directs the download towards KS rather than spacedock.

Dazpoet:
Seeing as the error has been located and the mods are de-facto broken in CKAN right now I have a hard time seeing how a merge of this could make the situation worse. Ergo I'm merging it on the grounds of it working locally and the Jenkins error having been located and deemed Jenkinsy rather than anything else.

PR #3149, opened 2016-02-19
Switch from GitHub to SpaceDock 3th attempt

sswelm:


sswelm:
now what is the problem?

politas:
Lose the trailing "/" on your kref, maybe?

PR #3151, opened 2016-02-19
Spacedocking Wernher's stuff again

TiktaalikDreaming:
I did this, but my github fu is sorely lacking and I think I may have clicky clicked too quickly somewhere because it wasn't done.  If there's two though, they should be the same.


Olympic1:
Merging, closes #3142

PR #3169, opened 2016-02-19
Fix #3163

Olympic1:


Olympic1:
#Rebuild

Dazpoet:
I think all dependencies should exist for download now in CKAN-meta and by the earlier answer I'm going to assume that #Rebuild is a thing again so here goes nothing :)

Dazpoet:
Now it's throwing a 
``The following inconsistencies were found:
RealPlume requires RealPlumeConfigs but nothing provides it``

I'm wondering why it classifies that as a successfull build.

Dazpoet:
This version is strictly better working than the current one so I'm merging it for now so we can move forward. The problem at hand appears to be that RealPlume would like to grab configs which throws a TMP or something to that effect.

PR #3175, opened 2016-02-19
Update and rename Asclepius-clouds.kerbalstuff to Asclepius-clouds.ne…

MrChumley:
…tkan


MrChumley:
I'm not positive that I'm doing this part right.

Olympic1:
Yes you are but it failed because it thinks that Asclepius is still on KS.

Also because these config files will only work if EVE is installed it would be better to move it to `depends`

Olympic1:
Looks good, thanks for updating your mods :+1:

MrChumley:
Thank you.

PR #3178, opened 2016-02-19
Update licensing on SCANsat.netkan

Dazpoet:
When multiple licenses exist we always default to the most restrictive one as per the spec.

Could we put the entire licensearray into a seperate `"licenses"` field perhaps? I'm not sure how netkan handles such a thing nowadays.

Also removed the x_netkan_license_ok so we hardcode the most restrictive one.


Dazpoet:
closing in favour of #3177

PR #3128, opened 2016-02-19
Add AoA Technologies - Aviation Parts from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of martinezfg11, to add [AoA Technologies - Aviation Parts](https://spacedock.info/mod/166/AoA%20Technologies%20-%20Aviation%20Parts) to CKAN.

Please direct questions about this pull request to [martinezfg11](https://spacedock.info/profile/martinezfg11).

PR #3191, opened 2016-02-20
Mk2CockpitExpansion to spacedock. Closes #3128

plague006:


plague006:
Rebasing this just lead to a big pile of nope. It's been pulling commits for the last 10 minutes .

PR #3132, opened 2016-02-19
Add Large Boat Parts Pack from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of Laythe, to add [Large Boat Parts Pack](https://spacedock.info/mod/167/Large%20Boat%20Parts%20Pack) to CKAN.

Please direct questions about this pull request to [Laythe](https://spacedock.info/profile/Laythe).

PR #3192, opened 2016-02-20
Add LargeBoatPartsPack, closes #3132

Olympic1:


plague006:
Looks good, good catch on the license not matching. =)

PR #3203, opened 2016-02-20
GimbalForAll to SpaceDock, closes #3197

Olympic1:


Dazpoet:
I object against this mod since only newbs forget reactioncontrols on their ionprobes and I've totally never lost an entire probefleet in jool orbit due to this issue, nope NEVER <.<

(aka I'm dl'ing this asap)

PR #3204, opened 2016-02-20
Mk2.5 Spaceplane Parts to SpaceDock, closes #3198

Olympic1:
I don't know why the `identifier` `RaginCaucasian` was chosen.


Dazpoet:
I would assume it's due to what's said [here](https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#identifer) about picking the identifier

PR #3160, opened 2016-02-19
Updates for spacedock.info

linuxgurugamer:
Updates for mods uploaded to spacedock.info

PR #3206, opened 2016-02-20
Move to SpaceDock

Olympic1:
Differences from #3160:
* Rebased the branch
* Removed the hardcoded SpaceDock links
* Added forum/repo links
* Added a override to SpacetuxSA (never change an override)

Fixes #3160


Olympic1:
Commenting to force a build

Dazpoet:
passes jsonlint which means nothing should break horribly from merging these. I will classify this as objectively better than what we currently have but would've loved seeing this broken into smaller pieces originally to make it easier to run through the bots. Considering the situation at hand I'll merge this so we can move forward rather than keep poking at these and breking it into smaller, working, pieces.

Dazpoet:
Also I have a large trust in @Olympic1 and their local testing :)

PR #3217, opened 2016-02-21
Add Versatile Toolbox System: USI Pack from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of Nazari, to add [Versatile Toolbox System: USI Pack](https://spacedock.info/mod/204/Versatile%20Toolbox%20System:%20USI%20Pack) to CKAN.

Please direct questions about this pull request to [Nazari](https://spacedock.info/profile/Nazari).

PR #3219, opened 2016-02-21
Add VersatileToolboxSystem

plague006:
Closes https://github.com/KSP-CKAN/NetKAN/pull/3217

### DIFF
--- a/NetKAN/VersatileToolboxSystem.netkan
+++ b/NetKAN/VersatileToolboxSystem.netkan
@@ -1,0 +1,15 @@
+{
+    "license": "CC-BY-NC-SA-4.0",
+    "spec_version": "v1.4",
+    "identifier": "VersatileToolboxSystem",
+    "$kref": "#/ckan/spacedock/204",
+    "depends": [
+        { "name": "KIS" }
+    ],
+    "install": [
+        {
+            "file"       : "MP_Nazari",
+            "install_to" : "GameData"
+        }
+    ]
+}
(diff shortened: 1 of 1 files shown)


plague006:
Really unsure if I should include any USI mods.

politas:
"For compatibility with USI resources, you must install USI Life Support and mods that enable USI fuels(Karbonite)." Not sure what the mod does other than adding USI support. Maybe depends on USI-LS and Recommends Karbonite?
Also "This mod requires Kerbal Inventory System for full functionality"

Olympic1:
> This mod requires Kerbal Inventory System if you want to enable all of the features it was built for:

I should set a `depends` on `USI-LS` (to use all containers) and `KIS` (to use the racks and move parts around) to make this mod fully functional, and set a `recommends` on `Karbonite`.

plague006:
Looking at the SD page alone I'd be inclined to agree we should add more than just KIS as a `depends`. However, if you take a look at the [ksp forum page](http://forum.kerbalspaceprogram.com/index.php?/topic/70312-105-versatile-toolbox-system-added-usi-support/) the overall tone changes. Instead of the mod being named `Versatile Toolbox System: USI Pack` which implies USI is integral, it's now just `[1.0.5] Versatile Toolbox System (Added USI Support)` which is clearly its own mod that happens to support USI and to support USI, it needs specific mods installed.

Further from the forum page:

>Release 3 adds fuel cell parts and support for USI Life Support Resources

This mod is definitely its own thing with or without USI.

plague006:
As a compromise, how about leaving the relationships but hardcoding the `abstract` to include: `For compatibility with USI resources, you must install USI Life Support and mods that enable USI fuels(Karbonite).`

Olympic1:
Then just `recommends` or `suggests` `USI-LS` and `Karbonite`.

Olympic1:
Merging with my thanks

PR #3226, opened 2016-02-21
Custom Asteroids updates

Starstrider42:
This PR moves Custom Asteroids' krefs from KerbalStuff to SpaceDock. In addition, it fixes a mistake I made in the NetKAN file regarding which versions are incompatible with Kopernicus.


Starstrider42:
I don't understand the build log... it says:
> Downloading "https://spacedock.info/mod/210/Custom Asteroids/download/v1.2.0" (libcurl)
Downloading "https://kerbalstuff.com/mod/251/Custom Asteroids/download/v1.2.0" (libcurl)
but when I generated the .ckan files on my machine there were no leftover references to kerbalstuff...

Olympic1:
@Starstrider42, your config files require `CustomAsteroids` it still tries to download from KS because the files aren't updated yet. This will happen for most mods with dependencies, so we need to test this locally, what I will do now.

Olympic1:
Tested locally and installs fine also works in KSP

PR #3229, opened 2016-02-21
Move Chatterer to Spacedock

politas:
Closes #3225


Olympic1:
Merging with my thanks

PR #3240, opened 2016-02-21
Add Versatile Toolbox System: TACLS Pack from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of Nazari, to add [Versatile Toolbox System: TACLS Pack](https://spacedock.info/mod/221/Versatile%20Toolbox%20System:%20TACLS%20Pack) to CKAN.

Please direct questions about this pull request to [Nazari](https://spacedock.info/profile/Nazari).

PR #3242, opened 2016-02-21
Revert "Add VersatileToolboxSystem"

plague006:
Reverts KSP-CKAN/NetKAN#3219


plague006:
Pushing this through before the bots make the identifier "real". Need to resolve this and https://github.com/KSP-CKAN/NetKAN/pull/3240

PR #3249, opened 2016-02-21
StageRecovery to spacedock

plague006:
Closes https://github.com/KSP-CKAN/NetKAN/pull/3237


politas:
Merging with my thanks

PR #3250, opened 2016-02-21
FASA to spacedock, reformat metadata

plague006:
Closes https://github.com/KSP-CKAN/NetKAN/pull/3239


politas:
Same as before, so I guess it should be fine, but I wonder what's in Plugins.

politas:
Ah, Reflection Plugin, which is there as a Recommend.

PR #3252, opened 2016-02-21
BetterBurnTime to spacedock

plague006:
Closes https://github.com/KSP-CKAN/NetKAN/pull/3243


politas:
Merging with my thanks

PR #3245, opened 2016-02-21
Add VersatileToolboxSystem-core, VersatileToolboxSystem-TACLSPack, VersatileToolboxSystem-USIPack

plague006:
Closes #3240

PR #3264, opened 2016-02-22
Remove quotes for boolean values in VersatileToolboxSystem

plague006:


plague006:
Jenkins probably should have caught that.. @techman83 ref https://github.com/KSP-CKAN/NetKAN/pull/3245

Olympic1:
I could build the files with the quotes succesfully using `netkan.exe`.

plague006:
@Olympic1 yep, netkan doesn't have the issue, ckan.exe does (which is why it failed netkan-bot's indexing with "Schema validation failed")

techman83:
@plague006 Ahh, we're assuming that the netkan is producing valid data during testing, but the indexer doesn't want to auto-commit invalid data, so it does a sanity test.

PR #3266, opened 2016-02-22
Update BlacklegIndustries to use github

blackleg:
Update BlacklegIndustries to use github releases


politas:
Looks good to me and the bots are happy. Merging with my thanks!

dbent:
Hi @blackleg since you're updating your own metadata would you be interested in self-hosting it so you don't need to submit PRs to us?

plague006:
@dbent @Olympic1 do either of you have a way of comparing the .netkan and/or generated .ckan against the schema? it's failing @netkan-bot's check

PR #3267, opened 2016-02-22
Move FantomWorks to SpaceDock

politas:
closes #3265

### DIFF
--- a/NetKAN/FantomWorksKAXPartPack.netkan
+++ b/NetKAN/FantomWorksKAXPartPack.netkan
@@ -1,7 +1,7 @@
 {
-    "license": "GPL-3.0",
+    "license": "CC-BY-4.0",
     "identifier": "FantomWorksKAXPartPack",
-    "$kref": "#/ckan/kerbalstuff/565",
+    "$kref": "#/ckan/spacedock/229",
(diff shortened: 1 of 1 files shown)


Olympic1:
From the license file included in the download

> This work is licensed under the Creative Commons Attribution-NonCommercial-NoDerivatives 4.0 International License (CC BY-NC-ND 4.0)

plague006:
Seeing as:

> This part pack provides parts, as an extension for KerbalAircraftExpansion

We should probably `recommends` it.

politas:
Spacedock just said CC-BY-4.0. I checked that the license included was indeed CC rather than GPL, didn't check that deeply, though. KAX is not needed to make this mod function fully, so I'll make it a suggests.

Olympic1:
@politas you're right that it isn't needed for this pack to work, but because it is specifically made for KAX as an expansion we should `recommends` it.

`recommends` means that it isn't dependent to make the mod work, but if installed it will (highly) improve that mod or if it's an extension of it, while `suggests` indicates that the mod is a great thing to install as well but doesn't really add something.

politas:
That's not the way I understood them, but fine, have it your way.

PR #3283, opened 2016-02-23
Update Kopernicus

Olympic1:
* Update forum page.
* Kopernicus only works with MM 2.6.16 or higher.
* Use `suggest` instead of `recommends` because this is only a modders tool.


plague006:
:+1:

PR #3286, opened 2016-02-23
Add Phoenix Industries - ³Sat  from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of MoviesColin, to add [Phoenix Industries - ³Sat ](https://spacedock.info/mod/247/Phoenix%20Industries%20-%20%C2%B3Sat%20) to CKAN.

Please direct questions about this pull request to [MoviesColin](https://spacedock.info/profile/MoviesColin).

PR #3296, opened 2016-02-23
Add PhoenixIndustriesCubeSat

plague006:
Closes https://github.com/KSP-CKAN/NetKAN/pull/3286


politas:
Spacedock and forum thread both say it's "compatible with" MechJeb and Tweakscale, so they should probably be no more than recommends. They're not needed for it to work.

plague006:
Ah yeah, I use a template that has `depends` and forgot to switch it to `recommends`. Good catch.

politas:
Merged with my thanks. I've been wanting to check this one out.

PR #3302, opened 2016-02-24
Update and rename SXT.kerbalstuff to SXT.netkan

DuoDex:
Now points to Github instead of broken Kerbalstuff link.


DuoDex:
@ChucklesTheBeard beat me to it.

PR #3304, opened 2016-02-24
Modify IRnifty/Nifty255 Mods for SpaceDock Rehost

IRnifty:


plague006:
`1890 [1] FATAL CKAN.NetKAN.Program (null) - Could not get the mod from SD, reason: Mod not published..`

IRnifty:
Fixed. Sorry about that.

PR #3311, opened 2016-02-24
Updated "Scatterer" mod to source from Spacedock

logicnotlogical:
Scatterer is now hosted on Spacedock, rather than Kerbal Stuff.


ChucklesTheBeard:
needs to be renamed to .netkan

Olympic1:
LGTM, merging with my thanks

PR #3316, opened 2016-02-24
KSCFloodlight to spacedock - closes #3308

ChucklesTheBeard:


plague006:
I was told there would be thermite? Filter `Thumbs.db` :)

PR #3321, opened 2016-02-24
Kosmodrome -> spacedock - closes #3307

ChucklesTheBeard:
license is on forum thread + spacedock


Olympic1:
Merging with my thanks

PR #3332, opened 2016-02-25
Update FacilityReset

Olympic1:


politas:
Can you add in a conflict with KerbalConstructionTime?

PR #3333, opened 2016-02-25
Update FacilityReset

Olympic1:


politas:
From the forum thread, a conflict with KerbalConstructionTime is called for.

politas:
Needs renaming to .netkan, too

politas:
Looks good now. Merging with my thanks

PR #3343, opened 2016-02-25
Update FogOfTech

Olympic1:


politas:
The forum thread states:
Supported KSP version: 1.0.2 - 1.0.5
Sources: included
So a netkan_override is needed for ksp_version and we should filter src.zip

plague006:
A filter is already in place.

PR #3346, opened 2016-02-26
Change dependency from IFS-core to IFS proper.

plague006:
The files are in mk2expansion's .zips, we should install them.


politas:
Fair enough

PR #3347, opened 2016-02-26
Add CC-Asclepius, Kronkus-clouds and modify relationships

plague006:
@MrChumley it'd be awesome to have your input on this. We were just made aware that your mod has CC contracts and it'd be great to offer them to our users in a way you think makes sense. What I'm currently proposing is adding `ContractConfigurator-Asclepius` as a sperate module and essentially treating it in the same way as `Asclepius-clouds`, so it's a recommendation from the base mod and it requires the base mod to be installed as well as `ContractConfigurator`.

### DIFF
--- a/NetKAN/Asclepius.netkan
+++ b/NetKAN/Asclepius.netkan
@@ -11,15 +11,16 @@
         "MrChumley"
     ],
     "resources": {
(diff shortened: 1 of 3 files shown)


politas:
All seems to test fine and make sense to me. Merging with my thanks

PR #3348, opened 2016-02-26
Update kRPC github and docs links

djungelorm:


plague006:
Look at you all official now with an org. :D
Merging with my thanks.

plague006:
As an aside I notice kRPC installs `VERSION.txt`, if you want a hand converting it to `kRPC.version` we would be able to scrape KSP version compatibility from it automatically (no more manually editing this file).

djungelorm:
Thanks.

Yeah that would be good. Is there documentation for this?

plague006:
It should be http://ksp.cybutek.net/kspavc/Documents/README.htm for the .version file formatting, for CKAN's use of the file documentation is here: https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#vref

The simplest/easiest version file for your mod would be something like:
````
{
  "NAME": "kRPC",
  "VERSION": {
    "MAJOR": 0,
    "MINOR": 2,
    "PATCH": 1
  },
  "KSP_VERSION": {
    "MAJOR": 1,
    "MINOR": 0,
    "PATCH": 5
  }
}
````

and then `"$vref": "#/ckan/ksp-avc"` added to the .netkan so we can make use of it (when it's in a release).

djungelorm:
OK thanks, will add it to the next version of kRPC

PR #3361, opened 2016-02-26
Add SpaceX Barge Lander

Olympic1:
From kdufferskorp


politas:
Merging with my thanks

PR #3363, opened 2016-02-27
Add SimpleNuclear NFT Balanced, closes #3344

Olympic1:
Add SimpleNuclear-NFT-Balanced and conflict with SimpleNuclear-NFT

### DIFF
--- a/NetKAN/SimpleNuclear-NFT.netkan
+++ b/NetKAN/SimpleNuclear-NFT.netkan
@@ -4,7 +4,9 @@
     "$kref": "#/ckan/spacedock/259",
     "license": "CC-BY-NC-SA-4.0",
     "depends": [
-        { "name": "SimpleNuclear" },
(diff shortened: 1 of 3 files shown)


politas:
Only issue I have is that this prevents users installing only one of NearFutureElectrical or NearFuturePropulsion with the SimpleNuclear configs.

I don't think making the NFT mods "recommends" would cause any problems?

Olympic1:
@politas Done

politas:
Merging with my thanks

plague006:
This module fails to install unless a user realizes the 2 `recommends` (which are opt out) conflict with one another.

FWIW if it was misleading, me adding a commit to a PR isn't an endorsement of any kind, I was just fixing a technical error that was blocking Jenkins.

Cleaned up in https://github.com/KSP-CKAN/NetKAN/pull/3374 and https://github.com/KSP-CKAN/CKAN-meta/pull/999.

PR #3365, opened 2016-02-27
Add SimpleConstruction

Olympic1:
Filtering out `InterstellarFuelSwitch.version` and `InterstellarFuelSwitch.dll` because it depends on InterstellarFuelSwitch-Core which includes these files.

### DIFF
--- a/NetKAN/SimpleConstruction.netkan
+++ b/NetKAN/SimpleConstruction.netkan
@@ -1,0 +1,20 @@
+{
+    "spec_version": "v1.4",
+    "identifier": "SimpleConstruction",
+    "$kref": "#/ckan/spacedock/59",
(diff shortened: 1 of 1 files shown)


plague006:
You have IFS filtered but you left `KSPAPIExtensions`'s file `KSPAPIExtensions.dll` and `ExtraPlanetaryLaunchpads`'s `Launchpad.dll` and `EL.version`.

Olympic1:
Those are files from ExtraPlanetaryLaunchpads but can't add a dependency on that because it conflicts with the parts

plague006:
Sounds good. According to the author this can be made EPL-compatible by stripping everything but the Parts folder, but I don't think it would be worth it to our users.

Merging with my thanks.

PR #3366, opened 2016-02-27
Use correct identifier

Olympic1:


politas:
Merging with my thanks

PR #3371, opened 2016-02-28
Update IR-Sequencer

Olympic1:


politas:
May as well ditch that override.

PR #3373, opened 2016-02-28
Update DangIt to spacedock

602p:
Simple change for new host


NetKAN/DangIt.netkan
```diff
@@ -15,6 +15,12 @@
         "homepage"     : "http://forum.kerbalspaceprogram.com/threads/81794",
         "bugtracker"   : "https://github.com/Ippo343/DangIt/issues",
         "repository"   : "https://github.com/Ippo343/DangIt",
-        "kerbalstuff"  : "https://kerbalstuff.com/mod/7"
-    }
+        "x_spacedock"  : "http://spacedock.info/mod/306/DangIt"
```
Olympic1: You don't have to hardcode this, it will automatically get added by to the `.ckan` file.

PR #3375, opened 2016-02-28
Remove karbonite as dependency

plague006:
Per [mod author](http://forum.kerbalspaceprogram.com/index.php?/topic/70312-105-versatile-toolbox-system-added-usi-support/&page=2#comment-2433021)


politas:
Hmm. Guess I should have commented on that when it was added. I nearly did.

PR #3378, opened 2016-02-28
Update BackgroundProcessing

Olympic1:
Moved to SpaceDock and use regex to find the `.dll` if version number changes in the future.


politas:
Looks like we could add a ksp_version override, according to [this forum post](http://forum.kerbalspaceprogram.com/index.php?/topic/88777-102-backgroundprocessing/&do=findComment&comment=2432024)

PR #3379, opened 2016-02-28
Update Quick... Mods

Olympic1:
Updated every Quick... mod except for `QuickStart` because @malahx didn't publish his mod on SpaceDock because it isn't updated for 1.0.5 yet.

`QuickHide` will install as `v3.03` but has the same zipfile as `v3.03a`, so added an epoch for it.


politas:
These all build and install fine, but they all contain .version files we could be using.

plague006:
Scratching my head as I look at the `spec_ver` from `QuickBrake.kerbalstuff`.

`QuickGoTo` should `suggests` `toolbar`.

plague006:
`QuickIVA` should probably `suggests` `toolbar` based on `GameData/QuickIVA/Textures/BlizzyToolBar.dds`

plague006:
Same for `QuickRevert` and `QuickScroll` (both install Blizzytoolbar textures)

plague006:
Looks good to me, merge if you're cool with my commits.

malahx:
Huge thanks for this, I've tried to do it some days ago, but with too many errors ... again thanks.

PR #3381, opened 2016-02-28
Add xScience

Olympic1:
Hardcoded the `author` for the original author (`Brodrick`) and the maintainer (`Z-Key Aerospace`).


politas:
There's a .version file, but I doubt netkan would find it, given the unfortunate file naming.

Olympic1:
The version file is broken because it's missing a comma

politas:
That, too? Merging with my thanks

PR #3382, opened 2016-02-28
Update KerboKatz

Olympic1:
We are allowed by the author to redistribute the KerboKatz mods via CKAN.


plague006:
````
"identifier": "KerboKatzSmallUtilities-EditorCamUtilities",
"version": "1.2.5",

About to install...
 * KerboKatz - SmallUtilities - EditorCamUtilities 1.3.0
````

Olympic1:
After looking into this, I noticed that the `1.3.0` version that we have indexed is the exact same version as the `1.2.5` versions from GitHub, Curse and SpaceDock.

The best way to fix this should be to add an epoch.

politas:
Shame about that epoch; they were all wonderfully in sync, otherwise. Merging with my thanks.

PR #3386, opened 2016-03-01
Get kRPC version info from ksp-avc

djungelorm:
kRPC 0.2.2 was released with a ```kRPC.version``` file. Adding vref field.


plague006:

PR #3387, opened 2016-03-01
Update xScience

politas:
Change homepage and abstract as per authors suggestions in KSP-CKAN/CKAN-meta#1005

### DIFF
--- a/NetKAN/xScience.netkan
+++ b/NetKAN/xScience.netkan
@@ -14,5 +14,13 @@
             "file": "GameData/[x] Science!",
             "install_to": "GameData"
         }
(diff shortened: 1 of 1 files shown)


Dazpoet:
What is the policy on MiniAVC files now? I recall #1455 but it doesn't seem to have ever gotten into a distinct directive.

politas:
My understanding was that we are filtering out the .xml file where one is included. Not sure what that has to do with this mod, though? I can't see any miniAVC files in it.

Dazpoet:
sounds reasonable to me, merging :+1:

PR #3389, opened 2016-03-03
Restore AutoAsparagus - from SpaceDock

LebedevRI:


Olympic1:
Merging with my thanks

PR #3396, opened 2016-03-04
Move SSRMicroSat to Spacedock

politas:


politas:
KSP-CKAN/CKAN-meta#1009 should fix that failure

Olympic1:
Rebuild passed and works locally

PR #3400, opened 2016-03-04
Move GlassPanesandEnclosures to Spacedock

politas:
Fixed Install stanza for changed Directory structure. Much was missing


Olympic1:
It is confusing which license it is because on KerbalStuff he had the GPL license, on SpaceDock he now has CC-BY license and on Curse he has the "All Rights Reserved" license.

Contacted the author about this

politas:
I figured the most recent release is the Spacedock one, so use that license.

Dazpoet:
Seems the author is working on it in accordance with (here)[http://forum.kerbalspaceprogram.com/index.php?/topic/106633-105-glass-panes-and-enclosures-gpe-v1602/&page=4#comment-2442532]

How about safeing it and slapping a restricted on it until it sorts itself out? Iirc the spec says we should always default to the most restrictive license and that would appear to be ARR in this case, although I don't see how a new author could slap ARR onto another modders work... but that's for them to figure out imo.

PR #1229, opened 2015-05-11
AutoAsparagus update

henrybauer:
Add link to forum page
Recommend SmartStage

PR #3403, opened 2016-03-04
AutoAsparagus: recommends SmartStage

LebedevRI:
Pretty much like #1229, just without typos :)


Olympic1:
Merging with my thanks

PR #3404, opened 2016-03-04
Updates for Contracts Window Plus, CapCom, and Progress Parser; a new dependency: Contract Parser

DMagic1:


Olympic1:
Thanks for updating your mods, merging with my thanks

PR #3406, opened 2016-03-05
Updating OctoSat + AtomicAge to SpaceDock

ILM126:
Whoops, did a litter derp there. Actually forgot to add the file during first commit ^_^"


ILM126:
Added another mod that has been causing me some trouble for ages, but didn't know I can change it until yesterday :smile:

Olympic1:
Merging with my thanks

PR #3407, opened 2016-03-05
Added KerbinSideGAP mod

Keniamin:


politas:
Tested locally successfully. Merging with my thanks.

PR #3411, opened 2016-03-06
Updated Endurance Space Exploration System to SpaceDock

ILM126:
Simple as that... Wow..
So I'm thinking about searching for some mods on SpaceDock and updating it soon~

### DIFF
--- a/NetKAN/EnduranceSpaceExplorationSystem.netkan
+++ b/NetKAN/EnduranceSpaceExplorationSystem.netkan
@@ -1,5 +1,5 @@
 {
-    "$kref": "#/ckan/kerbalstuff/536",
+    "$kref": "#/ckan/spacedock/357",
     "identifier": "EnduranceSpaceExplorationSystem",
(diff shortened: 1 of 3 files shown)


politas:
Basically, yes, though while looking at it, there's a few things that could make that mod work better.

The download file has a folder called "Habitat Pack", which is presumably a dependency, but there's nothing in "depends" for it. The modname is "PorkjetHabitats". 

As far as I can tell from the forum thread, KIS and RasterPropMonitor should only be "recommends", and you can add KAS to them, and have TAC-LS and USI-LS in a "suggests" stanza.

ILM126:
Oh, yeah. Thanks ^_^

I'll be sure to check the dependencies, recommends, and suggests of the mods in the future.

politas:
Often, it really is as simple as changing the #kref from kerbalstuff to spacedock. You just happened to hit one that was way more complicated.

politas:
Closed in favour of #3415

PR #3419, opened 2016-03-09
Remove TestFlightRO Config pack.  RealismOverhaul itself now has these and will satisfy the dependency.

jwvanderbeck:


politas:
Can you resubmit with just the changes to TestFlight.netkan? The RO folks have frozen the config netkan so it won't waste resources any more, but still maintains the Identifier.

jwvanderbeck:
I'm sorry but I don't entirely know what you mean?  The CKAN side has always been confusing to me.

Resubmitting this with just the changes to TestFlight.netkan doesn't do me any good, as the driving force here is to remove the TestFlightConfigRO mod, since it no longer exists as of the next version of Test Flight.  The RO Configs are now in RO itself, and @NathanKell has or will be modifying RO so that it satisfies the TestFlightConfig dependency.

NathanKell:
@jwvanderbeck  I believe what @politas is saying is that I already *did* do this stuff. :) Because I did.

jwvanderbeck:
You took care of removing my netkan as well?  If so, then you sneaky bastard you :p

politas:
What he did was rename TestFlightConfigRO.netkan to TestFlightConfigRO.netkan.disabled, which will stop the Netkan bot from trying to find any new versions, while maintaining the existence of the TestFlightConfigRO identifier in case someone wants to try using that name in the future. (Which is admittedly extremely unlikely, but maybe someone will create a "Random Operations" mod in the future and want to have a special configuration for TestFlight to support it.

Additionally, I've put in a change to the current release of TestFlightConfigRO (the .ckan file), which changes the abstract so that it says it is deprecated as of Realism Overhaul v10.9.4. (Well, actually, it turns out I did that on the last-but-one version of TestFlightConfigRO, so I'll just add that change to the latest one as well.)

PR #3423, opened 2016-03-10
Create FTPDeprecated.netkan

NecroBones:


politas:
Missing comma between depends and install

NecroBones:
Comma added. :)

politas:
I see there's no overlapping KSP versions between Fuel Tanks Plus versions < 1.0 and this, so all good. merging with my thanks.

PR #3426, opened 2016-03-12
Move KSPX to Spacedock

politas:


Dazpoet:
I must be blind because I cannot find what versions we already have indexed of this to check we don't get a version problem with the new one with a leading v since I Catherine from the old override it used to not have leading v before?

politas:
There's a v0.2.8.1 and a 0.2.9b, which has max_ksp of 1.0.4, and the restricted licence, means there's no point trying to fix that one. v0.2.9 will become the latest version.

PR #3428, opened 2016-03-13
Move RocketWatch to Spacedock

politas:
This will leave CKAN pointing to a cached version from Kerbalstuff as the latest release. I've left a message for the mod author asking for a preference.


politas:
I have had feedback from the mod author that 1.1.1 and 1.2 are identical. We should be fine to merge this, and I'll keep a weather eye out to see that the mod author bumps the version number sufficiently for the next release.

plague006:
Rather than crossing our fingers that the next release has a version number that doesn't conflict, we can add an epoch to 1) force the download from the more stable/longterm spacedock rather than the backup and 2) ensure that if the next iteration of the mod has the 1.2 version number that our users have an upgrade path.

politas:
We could, but I've discussed it with the author, and they have said they will bump the version number sufficiently.

PR #3430, opened 2016-03-13
Update Identifier Test

dbent:
Permit identifiers to start with digits or hyphen as per KSP-CKAN/CKAN#1570.

@politas @pjf @techman83 @Dazpoet

### DIFF
--- a/t/metadata.t
+++ b/t/metadata.t
@@ -8,7 +8,7 @@ use Test::NetKAN qw(netkan_files read_netkan licenses);
 
 use Data::Dumper;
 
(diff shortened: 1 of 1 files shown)


dbent:
Build failure is because it thinks `metadata.t` is a JSON file.

politas:
Ok, this is a pretty tiny change, and the code looks fine to me. Having @pjf 's agreement that there shouldn't be any issue with the starting character, I'm going to merge with many thanks!

techman83:
:+1:

PR #3436, opened 2016-03-14
RSS: Update netkan for next RSS

NathanKell:
Next RSS will have its cache files in its own folder, no need to deal with that.

**DO NOT MERGE UNTIL NEXT RSS**

### DIFF
--- a/NetKAN/RealSolarSystem.netkan
+++ b/NetKAN/RealSolarSystem.netkan
@@ -24,10 +24,6 @@
         {
             "find"      : "RealSolarSystem",
             "install_to": "GameData"
-        },
-        {
-            "find"      : "Cache",
-            "install_to": "GameData/Kopernicus"
         }
     ]
 }


dbent:
You can use an override and just commit it now (staging changes is one of the main reasons it was created). Just replace `EXPECTED_VERSION_WITH_CHANGE` with, well, the version number you expect will have the change. Don't forget the leading `v` which it seems RSS has in its version number.

```json

{
  "x_netkan_override": [
    {
      "version": "<EXPECTED_VERSION_WITH_CHANGE",
      "override": {
        "install": [
          {
            "find"      : "RealSolarSystem",
            "install_to": "GameData"
          },
          {
            "find"      : "Cache",
            "install_to": "GameData/Kopernicus"
          }
        ]
      }
    },
    {
      "version": ">=EXPECTED_VERSION_WITH_CHANGE",
      "override": {
        "install": [
          {
            "find"      : "RealSolarSystem",
            "install_to": "GameData"
          }
        ]
      }
    }
  ]
}
```

NathanKell:
@dbent Ah, I see! I'd prefer to avoid further overrides if possible--we'll have to add one for 1.1 anyway probably. But good to know!

politas:
Once you have a new release, any old overrides can be removed - Netkan will only re-examine the latest release.

NathanKell:
Ohh! Right. Thanks!

PR #3438, opened 2016-03-15
Updates to latest WBT

Angel-125:
This patch updates Buffalo, DSEV, MOLE, Pathfinder, and WildBlueTools to all point to GitHub and to use the current version of WildBlueTools. Not sure how it's done, but if it is possible to force the netkan to always use the latest WildBlueTools, latest Community Resource Pack, and latest Module Manager, that would really help.


NetKAN/Pathfinder.netkan
```diff
@@ -1,13 +1,20 @@
...
+    "$kref": "#/ckan/github/Angel-125/Pathfinder",
     "$vref": "#/ckan/ksp-avc",
     "license": "CC-BY-NC-SA-4.0",
     "depends" : [
-        { "name": "CommunityResourcePack", "min_version": "0.4.2" },
-        { "name": "WildBlueTools", "min_version": "1.0.4" },
-        { "name": "ModuleManager", "min_version": "2.6.5" }
+        { "name": "CommunityResourcePack", "min_version": "0.4.9" },
+        { "name": "WildBlueTools", "min_version": "1.0.32" },
+        { "name": "ModuleManager", "min_version": "2.6.18" }
```
plague006: missing comma

NetKAN/WildBlueTools.netkan
```diff
@@ -12,11 +12,12 @@
     "license": "CC-BY-NC-SA-4.0",
     "depends": [
         { "name": "CommunityResourcePack" }
```
plague006: missing comma

plague006:
netkan essentially only sees .netkan files in a vacuum. It knows nothing about any other mod or their versions. So when netkan is creating a .ckan it cannot change the min_version of dependencies.

With that all said: in a practical setting it's highly atypical for users to update some mods via CKAN but not all of them. In addition CKAN will always install the latest version of a mod unless someone passes a specific argument via the command-line interface.

politas:
I just put in a change to move Buffalo over, but I think yours are better, though I think the homepages should be redirected to the forum threads, eg,

    "resources": {
        "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/122617-105-buffalo-nasa-inspired-msev"
    },

You could also add this to the mods that it is relevant for:

    "supports" : [
        { "name" : "ConnectedLivingSpace" },
        { "name" : "TACLS" }
    ],

Angel-125:
Ah, that's what it's missing. Thanks for the spot, I don't do these on the regular. :)

Angel-125:
Ok, that should fix the compile issues.

politas:
Ah, the problem now is that the identifiers don't match the filenames. You can't change the identifiers. That's what tells CKAN that these are the same mods. It's the database key.

politas:
Nearly there! WildBlueTools needs 

    "spec_version": "v1.2",

Angel-125:
Ok, just updated that, crossing my fingers...

Angel-125:
Thx for your help! :)

politas:
Thanks for your persistence! It's very much appreciated. Merging now.

politas:
Hey, WildBlueTools, Your DeepSpaceExploration zip file has a Squad folder with a FusionIcon.png file that isn't being installed. Does it need to be?

Angel-125:
Yes it should be installed, that is needed.

politas:
OK, Fixed with #3453

PR #3445, opened 2016-03-17
Move Routine mission manager to Spacedock

politas:
remove obsolete override, filter sources


Dazpoet:
Filter isn't doing what it's supposed to I'm afraid.

plague006:
Can we use `"filter_regexp" : "Licences and Source"` so that we don't accidentally catch any awkwardly named files please.

PR #3446, opened 2016-03-17
Move Davon Throttle Control Systems to Spacedock

politas:
filter sources, fix license


Dazpoet:
As I don't speak regex I'm not sure why the filter isn't working but I can say based on the consoleoutput that it isn't since we're now installing both of them. Don't we need to add brackets around strings or something?

politas:
Brackets are only for where you are doing complex stuff. This is a simple match a string. Don't know why the "." wildcard didn't work.

politas:
That's got it.

plague006:
Can we use `"filter_regexp" : "license and source"` so that we don't accidentally catch any awkwardly named files please.

PR #3448, opened 2016-03-18
TriggerAU flags update

politas:
Move TriggerAU-Flags to Github (following original source)
Add dependencies to other TriggerAU mods for complete installation (for metadata check by @godarklight )


politas:
Dammit Jenkins, apply the commits in order and it'll work!

techman83:
It's a pretty naive bash array that loops over the files in whatever order. Wonder if we can actually do that and whether it would help in all cases?

politas:
Turns out it was my typo at fault.

PR #3454, opened 2016-03-18
Update Docking port alignment indicator

politas:
Set up for Curse scraping when it is implemented. Keeping frozen until then. 
Added missing files installation as discovered through metadata checker by @godarklight


politas:
KSP-CKAN/CKAN#1608 is the change that will stop this from throwing errors.

politas:
Now on Spacedock, so this is not needed.

PR #3459, opened 2016-03-22
Create OrbitalTug.netkan

nli2work:


politas:
Given that you've set the identifier to match the folder, you don't need the install stanza at all. It's failing currently because there's nothing in that install stanza to tell CKAN what to look for in the zip file. 

Remove the whole "install" ztanza, take the comma off the end of the "depends" stanza, and change the "spec_version" from "v1.4" to simply 1 (no quotes), and that should be golden.

politas:
Oh, and it would be lovely if you could replace each of your tabs with 4 spaces. Tabs are problematic, apparently.

PR #3461, opened 2016-03-22
Move KSI Placement Service to Spacedock

politas:
Point homepage back to forum thread, as authors homepage has no information about this mod.
Author name will change; this is correct. Mod author has undertaken a "rebranding" process.


Dazpoet:
It seems we somehow grab a .version file as the repository resource for this, it seems slightly unpractical for someone who'd like to see the repository. Maybe I've just missed that this is how we do things but it felt kinda weird.

politas:
That Spacedock entry is quite wonky.

PR #3462, opened 2016-03-22
Move kOS mutiple archive manager to Spacedock

politas:
Fix broken install; was missing vital git dll that needs to be in GameRoot. 
I checked the C# Regexp syntax carefully.

-   "^" at start of a regexp means the string being matched must be at the start of the string being examined. (It can also be used as a negation when specifying a set of characters eg "[^a-z]")
-   "." matches any character
-   "*" makes the previous element match zero or more times
-   "$" means the string being examined must end with the end of the string being matched

So this matches only an element in the root directory that starts with the letters "git" and ends with the letters "dll"


Dazpoet:
First time I see an install to GameRoot, pretty cool!

politas:
It was the first I'd seen, too. I wanted to be real careful with that regexp.

PR #3463, opened 2016-03-22
Add Jool DIRECT from SpaceDock

politas:
This is an automated commit on behalf of Daelkyr


politas:
A good license, a simple parts pack, and the identifier mactches the only folder in the zipfile. A rare easy Spaceduck submission! Merging now.

PR #3464, opened 2016-03-22
Move Roster Manager to Spacedock

politas:


Dazpoet:
I must admit I don't know what .mdb and .pdb files are, google points me towards some kind of database files which I guess makes sense and seeing as this was deemed safe earlier I'll assume the said can be said for it now so I'll merge it.

PR #3467, opened 2016-03-23
Move Kerbal Tubes to Spacedock

politas:
My local netkan is failing, saying "Could not find KerbalTubes entry in zipfile to install", which is very weird. Adding an install stanza with a find doesn't make a difference.


politas:
Jenkins gets the same error. Am I missing something, or is this a bug in Netkan?

plague006:
Using an install stanza with file rather than find will set you right.

politas:
Ok. I thought that a "file" install was the default. I still don't understand why the "find" method didn't work.

PR #3475, opened 2016-03-24
Move Multiports to Spacedock

politas:


Dazpoet:
Seems a lot of authors are changing license

politas:
Indeed. I choose not to try to play license police, and will trust that they have the right to do so.

PR #3483, opened 2016-03-28
SigmaModExpansions add version override

politas:
According to mod author, latest version is compatible with 1.0.5 as well, see http://forum.kerbalspaceprogram.com/index.php?/topic/123202-105-sigma-outerspacecomms/&do=findComment&comment=2293354
Closes KSP-CKAN/CKAN#1641

### DIFF
--- a/NetKAN/SigmaModExpansions.netkan
+++ b/NetKAN/SigmaModExpansions.netkan
@@ -24,5 +24,15 @@
         }
     ],
     "x_netkan_epoch": "1",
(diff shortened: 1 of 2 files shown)


politas:
http://forum.kerbalspaceprogram.com/index.php?/topic/123049-105-sigma-antennas/&do=findComment&comment=2464663 - SigmaRTExpansion aka Sigma: Antennas also has a compatibility increase over the included .version.

PR #3486, opened 2016-03-28
Create AsphaltTiles.ckan

enceos:


politas:
It looks like you've created a great netkan file to collect all the info from Spacedock. If you submit that file to KSP-CKAN/NetKAN (in the NetKAN subfolder with all the others), new versions will be automatically added to CKAN.

PR #3487, opened 2016-03-29
Add conflicts for Kduffers Wheels packs

politas:
Multiwheels contains the other two packs.
Hopefully we'll get a new source eventually
Ref KSP-CKAN/CKAN-meta#1031


politas:
Yes, Jenkins can't really deal with frozen files.

PR #3490, opened 2016-03-29
Create VensStylePPTextures.netkan

enceos:


politas:
That spec_version needs to be in quotes:

    "spec_version" : "v1.2",

You can keep editing the files in your existing branch just by going to the "Files Changed" tab above and clicking on the pencil, rather than creating new pull requests.

politas:
Excellent! Merging with my thanks

PR #3491, opened 2016-03-29
Create AsphaltTilesWelded.netkan

enceos:


politas:
You state in the description on Spacedock that "welded parts are mod independable." If that is the case, using "recommends" instead of "depends" would be better.

politas:
Also, to use a "find" install directive, you'll need to bump spec_version up to "v1.4".

enceos:
I meant they are independable from the welding mod.

enceos:
Should look good now. I'll change the description on SpaceDock to be more clear.

politas:
OK. Just need to bump spec_version, and it should be fine, then.

enceos:
I changed the "find" to "file"

politas:
Yep. Dependency failure this time

enceos:
It didn't find the AsphaltTiles in the database?

politas:
It's looking for .ckan files, I think. The robot should make one soon.

enceos:
I learned much about CKAN today )

politas:
There's a fair bit to learn. I've been finding different ways to go wrong pretty solidly for the last four months!

enceos:
I think "Adding a Modto CKAN" page needs a link to https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md
Under the "Special Installations" section. If I saw that I woudn't make that many mistakes.

politas:
Feel free to make that addition where you think it would make the most sense. Those pages are using Wiki rules; Be Bold!

politas:
Merged with thanks!

PR #3502, opened 2016-03-30
Add Speed Unit Changer from SpaceDock

politas:
This is a semi-automated commit on behalf of uncreative_user


Dazpoet:
Source needs to be filtered :)

politas:
Shame we can't do that without adding a whole install stanza. :)

PR #3506, opened 2016-03-30
New ModuleManager release is targeted at KSP 1.1

politas:
Fast-tracking to stop people accidentally upgrading to it.


plague006:
:+1:

PR #3512, opened 2016-04-01
Add KSP versions to override for ModularFlightIntegrator

dbent:
`1.1.3.0` should only be compatible with KSP 1.1.0.


plague006:
Yes it would.

PR #3516, opened 2016-04-02
Added KerbalSports.

jrossignol:
For great fishing. :fish: 

http://forum.kerbalspaceprogram.com/index.php?/topic/135728-10511-kerbal-sports-jebediah-kermans-fishing-challenge-v100-2016-04-01/


politas:
Thanks, that looks pretty good, though I think you should make it "name": "Kerbal Sports" to match the forum thread and confuse people less.

PR #3519, opened 2016-04-03
Add Contares-Antennas from Spacedock

politas:


politas:
Will need #3520 to be merged first.

PR #3521, opened 2016-04-03
Set CustomBarnKit, ModularFlightIntegrator, and GCMonitor to use AVC version

blowfishpro:
Should fix some issues with their current CKAN listings (1.1 versions
being installed on 1.0.5 etc)


blowfishpro:
@politas suggests that it might be worth keeping the `x_netkan_jenkins` and `x_netkan_version_edit` fields even if they are now redundant

@dbent mentions that if NetKAN ever starts looking at old versions, the version overrides could be useful.

dbent:
LGTM.

politas:
Checked old versions for CustomBarnKit, and they all have valid .version files.
Likewise for ModularFlightIntegrator.
GCMonitor has a few old versions that don't have .versions, but it's been solid for some time. 

I'm cleaning up the KSP version compatibility of all the old versions based on the .version files.

PR #3523, opened 2016-04-04
Added download link and updated to v1.1

BrianErikson:


politas:
Hi Brian, Netkan will add the download link automatically, so you can remove that line.

BrianErikson:
@politas, I added the download link because there will be multiple releases for different projects on the same repository -- I don't necessarily trust the automatic parser to grab the right download link. Is there another way I should go about doing this? Thanks for your help

plague006:
`asset_match` a search will give some [examples](https://github.com/KSP-CKAN/NetKAN/search?utf8=%E2%9C%93&q=asset_match). As this netkan currently exists we would scrape for (and index) new releases of the mod and each new release would download the exact same file.

BrianErikson:
@plague006, thanks for the information. I was referencing the [example CKAN file](https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#example-ckan-file) in `Spec.md`, so I figured the download property was sufficient. I'll go ahead and implement `asset_match` and re-open a new pull request.

PR #3524, opened 2016-04-04
List a license for DeepSpaceExplorationVessels

plague006:
I'm hoping this resolves the schema error that's appearing on http://status.ksp-ckan.org/


politas:
Github doesn't give us a licence, does it?

plague006:
Nope.

PR #3525, opened 2016-04-04
Change install stanzas to match new archive structure for Firespitter

plague006:


politas:
Would it not make more sense to filter the entire "Plugins" and "Resources" folders, since we're installing them complete in their respective mods? Currently, any file added to either of those folders in new releases will break installs.

NetKAN/FirespitterCore.netkan
```diff
@@ -14,11 +14,12 @@
     },
     "install"  : [
         {
-            "file"       : "Firespitter/Plugins",
+            "find"       : "Firespitter/Plugins",
             "install_to" : "GameData/Firespitter"
         },
         {
-            "file" : "Firespitter/Firespitter.version",
+            "find" : "Firespitter/Firespitter.version",
```
Dazpoet: shouldn't this be just Firespitter.version seeing as we're matching against files? Will there ever be a release with two .version files?

NetKAN/FirespitterCore.netkan
```diff
@@ -1,5 +1,5 @@
 {
-    "spec_version"    : "v1.2",
+    "spec_version"    : "v1.10",
```
Dazpoet: should be 1.16 according the [install-spec](https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#install)

techman83: > should be 1.16 according the install-spec

Looks like we need to add some test for the 1.16 spec changes.

plague006:
Honestly I wasn't looking to overhaul the install process of filtering files vs folders, I just wanted something that would give users the latest version. With that said I always intentionally make my PRs in the @KSP-CKAN repo so that contributors can make changes as they see fit.

politas:
Made sense to me to fix it up, since we're changing the install stanzas anyway.

Dazpoet:
Seems fine to me now and Jenkins seems to be installing the right things when he should.

PR #3529, opened 2016-04-04
Update install stanza for ExtrasolarPlanetsBeyondKerbol-stock

plague006:


politas:
Ah, it's been split into two separate downloads, hasn't it? That'll be much easier on downloads.

plague006:
Nah, both are still included making this a 380 MB download.

PR #3532, opened 2016-04-04
Modify install stanza of KSPInterstellarExtended

plague006:
An attempt to unblock this mod on http://status.ksp-ckan.org/. We're currently not index the most recent version because of it.


politas:
LGTM

PR #3533, opened 2016-04-04
Amended NavballUpDefault due to 1.1 changes

BrianErikson:
KSP 1.1 changes the way the navball is accessed, so any version older than NavballUpDefault-v0.2 (the latest release) will not work on KSP v1.1

### DIFF
--- a/NetKAN/NavballUpDefault.netkan
+++ b/NetKAN/NavballUpDefault.netkan
@@ -5,7 +5,7 @@
     "name"            : "Navball Up Default",
     "abstract"        : "Triggers the Navball to be up by default in the map view.",
     "license"         : "open-source",
-    "ksp_version_min" : "1.0.2",
+    "ksp_version_min" : "1.1.0",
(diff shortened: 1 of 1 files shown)


dbent:
LGTM, thanks. @BrianErikson since you're updating your own metadata would you be interested in self-hosting it so you don't have to submit PRs to us?

BrianErikson:
@dbent, thanks for the merge. As for self-hosting, I'm not quite sure what you mean? It seems all of the other mod authors are updating their `NETKAN` files this way as well unless I'm mistaken.

dbent:
@BrianErikson See [`PlaneMode.netkan`](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/PlaneMode.netkan) as an example. It's just a reference to a `.netkan` file I host in my own repository.

It's not wide spread yet but is currently used by a few mods like RealismOverhaul or the USI mods.

PR #3534, opened 2016-04-04
SmokeScreen 2.6.14 will target KSP 1.1

dbent:


dbent:
Checked locally, merging.

PR #3535, opened 2016-04-05
Update version overrides for 1.1

techman83:
Noticed some build failures due to the 1.1 short version.


politas:
Do you want to clean up those "overridding"s while you're at it?

techman83:
Looks like I broke my history, let me reword that commit message.

techman83:
#3538 contains a KSP 1.1 version and jenkins was happy with it. :dancers:

PR #3536, opened 2016-04-05
EVAHandrailsPackContinued

plague006:
Per http://status.ksp-ckan.org/ this is broken and from local testing I'm really not sure why it's broken.

Anyone with insight on this? @dbent @techman83

````
1303 [1] WARN CKAN.NetKAN.Transformers.SpacedockTransformer (null) - Could not normalize URL: 

2081 [1] FATAL CKAN.NetKAN.Program (null) - After parsing a value an unexpected character was encountered: C. Path 'DOWNLOAD', line 5, position 3.
````

PR #3538, opened 2016-04-05
Revert "EVAHandrailsPackContinued"

plague006:
Reverts KSP-CKAN/NetKAN#3536

New version uploaded with fixed .version file.


politas:
Excellent! It is good when mod authors are responsive.

PR #3539, opened 2016-04-06
Move Mk2Expansion to Spacedock

politas:
Should pick up the 1.1 release there.


techman83:
Failed due to dependency not available for KSP 1.1, merging as will not work in 1.0.5

PR #3540, opened 2016-04-06
Fix Mk2Expansion dependencies

politas:
No longer depends on Firespitter or Klockheed Martian Gimbal


techman83:
Jenkins says yes!

PR #3544, opened 2016-04-06
Add Stockalike Mining Extension from SpaceDock

politas:
This is an automated commit on behalf of SuicidalInsanity


techman83:
LGTM.

PR #3555, opened 2016-04-07
Add Tycho Orbital Shipyards weaponry from SpaceDock

politas:
This is a semi-automated commit on behalf of sashan13


Dazpoet:
I reacted to the identifier on this seeing as TOS is a rather widespread acronym on its own and My first thought was not a KSP mod, how hellbent are we on the identifier should match installed content policy?

politas:
I'm happy to change it to TychoWeaponry, which was what I first thought to set it as.

politas:
#rebuild

PR #3556, opened 2016-04-07
Add Star Wars BD Addon from SpaceDock

politas:
This is a semi-automated commit on behalf of KSP_Bros


techman83:
LGTM, merging with thanks!

PR #3557, opened 2016-04-07
Add MainSailor's Textures (for Procedural Parts) from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of MainSailor, to add [MainSailor's Textures (for Procedural Parts)](https://spacedock.info/mod/487/MainSailor%27s%20Textures%20%28for%20Procedural%20Parts%29) to CKAN.

Mod details:
    name = /mod/487/MainSailor%27s%20Textures%20%28for%20Procedural%20Parts%29
    author = MainSailor
    description = This is your mod listing! You can edit it as much as you like before you make it public.

To edit **this** text, you can click on the "**Edit this Mod**" button up there.

By the way, you have a lot of flexibility here. You can embed YouTube videos or screenshots. Be creative.

You can check out the SpaceDock [markdown documentation](/markdown) for tips.

Thanks for hosting your mod on SpaceDock!
    abstract = A texture pack containing sci-fi and historical textures (and some matching fairings) for Procedural Parts and Procedural Fairings
    license = MIT
    Homepage = http://forum.kerbalspaceprogram.com/index.php?/topic/93524-mainsailors-textures-for-procedural-parts-version-2-now-with-more-history-5th-of-april-2016/&page=1

Please direct questions about this pull request to [MainSailor](https://spacedock.info/profile/MainSailor).

PR #3558, opened 2016-04-08
Add MainSailor's Procedural Parts Texture packs from Spacedock

politas:
Splits out individual packs for those that don't want everything.
PPTextures mods, so KSP version compatibility can be controlled by the dependency's availability.
Closes #3557
Closes KSP-CKAN/CKAN-meta#1045


techman83:
LGTM, merging with thanks!

PR #3568, opened 2016-04-09
Update and rename KerbalFoundries.kerbalstuff to KerbalFoundries.netkan

Dazpoet:
I love this mod <3

### DIFF
--- a/NetKAN/KerbalFoundries.netkan
+++ b/NetKAN/KerbalFoundries.netkan
@@ -1,12 +1,8 @@
 {
     "spec_version"   : "v1.4",
     "identifier"     : "KerbalFoundries",
-    "$kref"          : "#/ckan/kerbalstuff/58",
-    "license"        : "GPL-3.0",
-    "resources":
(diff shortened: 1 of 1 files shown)


Dazpoet:
Woops I found https://github.com/KSP-CKAN/CKAN-meta/pull/1008 and it seems this should not be done so I'm closing it for now :(

politas:
That was only as a stopgap measure. If it's up on Spacedock now, let's scrape that!

Dazpoet:
I assumed this was the incorrect one mentioned in that PR, but sure this is now re-opened if that isn't the case :)

politas:
Ah, yes, I forgot that detail.
Let me check.

politas:
No, you're right, that is the dodgy one. I'm surprised it hasn't been removed from Spacedock.

politas:
Ok, I've left another comment in the KF forum thread (the real one, rather than the old one improperly linked in that Spacedock entry)

PR #3569, opened 2016-04-09
TransferWindowPlanner lacks ksp_version

Dazpoet:
The latest version is marked on github as 1.0.5 compatible and the forum thread for the mod mentions something about UI changes. People are also asking if the mod will get updated/is working so I'm assuming it is not at this time, user ace22 on irc also mentioned it not working. Might have to add ksp_version to the files in CKAN-meta aswell as to ensure that users don't get a really old version instead.


politas:
Here is me thinking it was explicitly set to "ksp_version": "any" by someone clever.

PR #3573, opened 2016-04-10
Add HabTech from SpaceDock

politas:
This is a semi-automated commit on behalf of Benjee10


politas:
That might fail because CLS is not available for 1.1

Dazpoet:
Suggests aren't installed by a basic `ckan install` command so no worries :)

PR #3579, opened 2016-04-11
Add KerwisModuleDepthCharge from SpaceDock

politas:
This is a semi-automated commit on behalf of GeorgeDuty


politas:
I'm concerned that the author name doesn't match up with the authors listed in the part CFG files, especially given the abstract we're getting from Spacedock:

    "abstract": "Is a cool mod ringt\uff1fyep \uff01",

PR #3597, opened 2016-04-14
Force v for ThrottleControlledAvionics

plague006:


plague006:
I'm going full rogue and merging this myself.

PR #3599, opened 2016-04-14
Edit regex to match new folder name in KIEFighter

plague006:
Interesting background to this.. the timing of the folder name's change was such that even though a perfectly working .netkan was merged, an update to the mod was issued that broke the .netkan before we ever indexed a version of the mod.


politas:
That's mean!

PR #3604, opened 2016-04-15
Update KerbinSide relationships

plague006:
@politas

### DIFF
--- a/NetKAN/KerbinSide.netkan
+++ b/NetKAN/KerbinSide.netkan
@@ -10,9 +10,9 @@
     "depends": [
         { "name": "KerbalKonstructs" }
     ],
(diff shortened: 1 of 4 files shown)


plague006:
e8bbc35 passes and that's the important one. I'm going to add another commit to bring KerbinSideAirRace in line with the new relationships but doing so will cause a Jenkins fail.

politas:
Isn't that a virtual package, rather than a metapackage?

plague006:
Yes. I probably said metapackage at some point, I misspoke.

politas:
"provides" : [ "KerbinSide-meta" ],

So if we change that to "Kerbinside-virtual"?

PR #3612, opened 2016-04-18
RN fix recommends for rockets

raidernick:
-I forgot to edit this file when I separate the recommended mods into 2
different mods.

### DIFF
--- a/NetKAN/RN-SovietRockets.netkan
+++ b/NetKAN/RN-SovietRockets.netkan
@@ -35,7 +35,8 @@
     ],
     "recommends"    :
     [
-        { "name"    : "RN-SalyutStations-SoyuzFerries" },
+        { "name"    : "RN-SalyutStations" },
+		{ "name"    : "RN-SovietSpacecraft" },
         { "name"    : "RN-SovietProbes"                }
     ]
 }


raidernick:
I need a second set of eyes to look at this. I am getting a LOT of complaints from people using ckan to install this mod(on ksp 1.0.5) that it's installing it broken. They are telling me that when they install it from ckan the engines provide no thrust. When they install the mod manually it works fine.

I have had 7 different people in the past month mention this to me but I can't see anything wrong with this file. It has worked in the past but suddenly it's become an issue.

politas:
You don't have a main directory for the cfg file to go in? Downloading now to see if I can work it out.

politas:
Looks like it installs everything ok. I can't see any reason why the engines would not provide thrust. You're using Squad's modules for that?

raidernick:
yes

politas:
Are they definitely talking about the engines in this mod?

raidernick:
Yes they specifically mentioned that half of the engines seemed to provide no thrust so the rocket would just fire and sit on the pad. This was not from using a craft file but from building it from scratch. I stopped updating my craft files a long time ago and removed them from the mod so I know that's not an issue anymore.

plague006:
I compared the the ckan install of `RN-SovietRockets` vs. a manual install.

The only 2 files that aren't identical are
FAIL	MD5	Firespitter\Plugins\Firespitter.dll
FAIL	MD5	Klockheed_Martian_Gimbal\Plugins\km_Gimbal.dll

I tried to match km_Gimbal.dll to versions that ckan has indexed (3.0.3-3.0.7) and the md5 didn't match between any of them. I also tried the same with firespitter and found that it matched the 7.1.4 release while CKAN installs 7.1.5 for ksp 1.0.5 users.

PR #3614, opened 2016-04-18
Add KSPChromaControl

plague006:
Copied (with some small revisions) from and closes https://github.com/KSP-CKAN/CKAN-meta/issues/1056. Pinging @Dazpoet to see if there's any precedent on how to handle these types of mods.

I'm also of 2 minds regarding `link2source.txt`. I could see someone making the argument that it's a source and should be filtered, but I could also make the argument that it's in the same vein as a "readme" which we include.


politas:
LGTM, passes Jenkins. Merging with thanks.

PR #3622, opened 2016-04-20
Manually change Contract Parser and Progress Parser to KSP 1.1

DMagic1:
Is including a .version file the simplest way to automate this?


plague006:
Yes it is. Merging with my thanks.

PR #3627, opened 2016-04-20
Update ksp compatibility and asset match for EnvironmentalVisualEnhancements

plague006:


politas:
Looks pretty good. Have to chase up the mods that have this as a dependency, now.

PR #3632, opened 2016-04-20
Update ksp compat

plague006:


plague006:
Smallest of changes and passes Jenkins. Pushing this.

PR #3636, opened 2016-04-20
Move StockVisualEnhancements to Spacedock

politas:
Add epoch


politas:
#rebuild

plague006:
This fails because it's using the old build.sh. Needs a rebase.

politas:
Ok, this has changed completely.

PR #3638, opened 2016-04-20
Move StockVisualEnhancements to Spacedock for KSP1.1

politas:
Rebased


plague006:
Seems to fail for the same reason. Gonna make a quick identical PR to see what's up.

politas:
We need to wait for NetKAN to inflate EVE, I suppose.

PR #3641, opened 2016-04-20
Add epoch.

plague006:
According to https://github.com/Tantares/Taerobee/releases the release is only hours old, but we have a version `2.3` indexed as compatible with `1.0.5`. This may be the same file re-released in which case an override would be accurate, but in the event that the file is different users on `1.0.5` end up installing an incorrect version. An epoch allows us to have both `2.3` with individual compatibilities.


politas:
Are we sure that 2.3 release is actually compatible with 1.0.5?

plague006:
https://github.com/KSP-CKAN/CKAN-meta/commit/96ce6a913d3a6ab1c864ddcd37e68a7729269990 on Feb 22 -- I'm very sure it's not for 1.1. =)

politas:
Would have required some serious transparent mineral latticework visualising, yes.

PR #3644, opened 2016-04-21
Removes ksp_version from FerramAerospaceResearch.netkan

SkyMarshal:
FAR v0.15.6 is currently being reported as KSP 1.0.5 only, when it is KSP 1.1.0 only.


SkyMarshal:
For reference: https://github.com/KSP-CKAN/CKAN-meta/blob/master/FerramAerospaceResearch/FerramAerospaceResearch-3-0.15.6.ckan
https://github.com/ferram4/Ferram-Aerospace-Research/blob/master/GameData/FerramAerospaceResearch/FAR.version

Dazpoet:
Could you also add a vref so we can grab the correct ksp_version range for each version of the mod? :+1:

plague006:
Superceded by https://github.com/KSP-CKAN/NetKAN/pull/3652

PR #3648, opened 2016-04-21
Change Depends on RPM to Recomments

enceos:
DPAI is not dependent on RPM, it can work standalone and adds a new screen if RPM is installed.


Dazpoet:
This sounds resonable and I know I've wondered about this myself a few times. Merging with thanks :+1:

politas:
:+1:

PR #3654, opened 2016-04-21
HaystackContinued to SpaceDock, closes #3646

plague006:
Personally really happy to see this.


politas:
MIT license is only for the source code, according to the forum thread.

plague006:
Good catch on the license.

PR #3660, opened 2016-04-21
Update AlternisKerbolRekerjiggered ksp_version compatibility and install

plague006:


Dazpoet:
Out of sheer curiosity, what is the difference between the `find` and `find_regexp` command in this context which makes the latter better?

Also, damn, that's the biggest conflict stanza I've ever seen :o

plague006:
`find` has to match the folder name exactly while `find_regexp` matches the regular expression. In this case the regular expression is `AlternisKerbol` which will match any folder name that contains that string. The folder name used to be `AlternisKerbol` and is now `AlternisKerbolRekerjiggered` so `find: AlternisKerbolRekerjiggered` would match just as well, but with the regexp we'll be covered in the case that the folder is renamed `AlternisKerbol` in later versions.

Dazpoet:
Thanks for the explanation :+1: 

Also the install looks fine so I'm merging this with my thanks :)

PR #3664, opened 2016-04-21
Changed TriggerAu flags from depends to recommends

WCapelle:
Flags not vital to working of mod.


plague006:
Merging with my thanks.

PR #3666, opened 2016-04-22
Update ksp_ver compat.

plague006:
Closes https://github.com/KSP-CKAN/CKAN-meta/pull/1080


plague006:
`t/metadata.t .. ok` -- all I care about since Jenkins will use the current metadata.

PR #3669, opened 2016-04-22
Achievements to GitHub

plague006:
Closes https://github.com/KSP-CKAN/CKAN-meta/issues/1078


politas:
Looks good to me, passes Jenkins and those look like the right files.

PR #3672, opened 2016-04-22
Add B9 PWings Modified to CKAN

01010101lzy:


plague006:
What you have here is a .ckan file.

A .ckan file is either a .netkan file but inflated with information inflated via a $kref or a file with all metadata manually entered.

plague006:
As this is a fork of `B9-PWings-Fork` you should look at https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/B9-PWings-Fork.netkan for an idea of what the metadata should look like.

PR #3673, opened 2016-04-22
Add KSP-AVC on Github

politas:
Cannot use vref at this time, as the .version file is not correct JSON. Author contacted.
Ref KSP-CKAN/CKAN-meta#1089


plague006:
*chuckle*

Issue #3670, opened 2016-04-22
Mods not updated


leops1984:
Some mods have not yet been updated on CKAN even though the authors have released newer versions, specifically

- [x] Athlonic Electronics / LCD - Launch CountDown (http://forum.kerbalspaceprogram.com/index.php?/topic/39470-11-lcd-launch-countdown-v176-19-apr-2016/)
- [x] Community Tech Tree (http://forum.kerbalspaceprogram.com/index.php?/topic/90530-11-community-tech-tree-24-last-updated-180416/)
- [x] Docking Sounds (http://forum.kerbalspaceprogram.com/index.php?/topic/113390-11docking-sounds-v21-41916/)
- [x] Filter Extensions (http://forum.kerbalspaceprogram.com/index.php?/topic/93955-11-filter-extensions-242-apr-21/)
- [x] Heat Control (http://forum.kerbalspaceprogram.com/index.php?/topic/112027-11-heat-control-more-radiators-updated-18042016-basic-11-release/)
- [x] KerboKatz (http://forum.kerbalspaceprogram.com/index.php?/topic/104505-11-kerbokatz-smallutilities-21042016-partial-11-update/)

politas:
Docking Port Sound FX is already showing up in 1.1

politas:
Filter Extensions is also already in CKAN for 1.1

politas:
All mods now added

PR #3675, opened 2016-04-22
Move HeatControl to Spacedock

politas:
Ref #3670


plague006:
I'm merging this as is because of how ubiquitous MM is and it's unclear what effect the absence of it will have. So I'm acknowledging the contents of the readme:

>DEPENDENCIES

>Required
- None

>Optional
- ModuleManager (v2.6.5)
- Community Tech Tree (v2.0+)

PR #3676, opened 2016-04-22
Mudular Fuel Tanks

Sphinx3k:
First test to contribute hope everything is right


plague006:
What you have here is a .ckan file.

A .ckan file is a .netkan file but inflated with information inflated via a `$kref`.

You need to PR this into ckan-meta and not netkan.

plague006:
Additionally, in cases where we already index earlier versions of a mod it can be significantly easier to take the previous version's .ckan file and modify it rather than starting from scratch.

Sphinx3k:
thanks. I will try to figure everything out. Sorry need to learn how to do it the right way.

PR #3685, opened 2016-04-23
Add SDHI Strobe-O-Matic

sumghai:


plague006:
You can `recommends` or `suggests` `SDHI-SharedAssets` even though it's not yet 1.1 compatible. Until it's compatible users simply won't be presented with the option of installing it.

sumghai:
> You can recommends or suggests SDHI-SharedAssets even though it's not yet 1.1 compatible. Until it's compatible users simply won't be presented with the option of installing it.

I assumed that SDHI-SharedAssets was automatically compatible for all versions, but I guess I'll have to fix that too.

> In addition, x_netkan_force_v is typically for mod that previously used a v in the earlier versions but at some point the author forgot it/changed version schemes. If you think it might happen to you by all means keep it, just want to make sure you're aware it's not necessary.

I'm keeping it in case I do change the versioning scheme for any reason.

sumghai:
Actually, you know what, if you can make SDHI-SharedAssets always compatible with any version of KSP, that'd be great.

plague006:
For `SDHI-SharedAssets` you can set `"ksp_version": "any"` and if you wouldn't mind `"comment": "agencymod"`. I've been slowly standardizing agency-only and flag-only mods.

sumghai:
I'll get to it in a moment, thanks.

politas:
Looks cool!

sumghai:
Thanks @plague006 and @politas :)

PR #3687, opened 2016-04-23
Add IndicatorLights, closes #3684

plague006:


politas:
Worth a release_status of development or testing?

plague006:
I've never once set a `release_status` for 2 reasons:

1. If I'm writing a netkan I consider myself partially responsible for keeping that mod up to date and I won't track a mod to see when/if its release_status changes.

2. We don't surface release_status to users and to my knowledge nobody has ever opened a single ticket to ask that we do.

politas:
Release Status appears in the metadata tab.

PR #3688, opened 2016-04-23
SerialIO to spacedock, closes #3682

plague006:


politas:
Homepage needs to be corrected to http://forum.kerbalspaceprogram.com/index.php?/topic/60281-hardware-plugin-arduino-based-physical-display-serial-port-io

PR #3689, opened 2016-04-23
Move KerboKatz SmallUtilities minimods to github

politas:
These minimods have been updated to KSP1.1
Will move other minimods over to github as they are released.
Ref #3670


plague006:
Need `abstract`s

politas:
Ah, good call. These never had their own abstracts before.

plague006:
Yeah they all shared the same one.

PR #3691, opened 2016-04-23
Updated config files for breaking changes in Custom Asteroids 1.3.

Starstrider42:
This config ensures that the upcoming release of Custom Asteroids 1.3 (which among other changes factors the zip file, as requested by two forum users) will not break CKAN, and that the new ModuleManager dependency is recognized right away.

The removal of `$vref` from the population configs is a temporary workaround for KSP-CKAN/CKAN#1674. The line can be restored once a 1.3 zip is on SpaceDock. (I will need to make a second pull request anyway, to upload `CustomAsteroids-Pops-Stock-Stockalike.netkan`; the file cannot be compiled at present due to a "no files to install" error.)

### DIFF
--- a/NetKAN/CustomAsteroids-Pops-Stock-Inner.netkan
+++ b/NetKAN/CustomAsteroids-Pops-Stock-Inner.netkan
@@ -58,5 +58,6 @@
             }
         }
     ],
+                "ksp_version_max": "1.0.5",
     "x_maintained_by": "Starstrider42"
 }
--- a/NetKAN/CustomAsteroids-Pops-Stock-Inner.netkan
+++ b/NetKAN/CustomAsteroids-Pops-Stock-Inner.netkan
@@ -52,9 +52,9 @@
                         "install_to": "GameData",
                         "filter_regexp": "(?<!Basic Asteroids\\.cfg)$"
                     }
-                ],
+                ]
             }
-        },
+        }
     ],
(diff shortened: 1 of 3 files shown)


Starstrider42:
Is there a way to do the Jenkins build with KSP 1.0.5 other than 1.1.0? I think the absence of a 1.1-compatible release is the only error remaining...

politas:
You could set ksp_version_max to 1.0.5 in order to validate the majority of changes, then have a separate PR to just remove it by itself (which will fail to build, but because it will be such a small change, it'll be easier to gauge the impact.)

Starstrider42:
Ok, it's done.

politas:
It doesn't. There are two checks that are run for each commit.

politas:
You reverted it before anyone else could see the results.

politas:
If you want to do that, copy and paste the successful build report into the conversation.

Starstrider42:
Why not just provide a link? http://ci.ksp-ckan.org/job/NetKAN/2832/

And I don't see what you mean by "two checks are run". The last commit (the revert) is the only one with one associated build.

politas:
To be honest, I've no idea, then. @techman83 might be able to answer that.

Happy with that link, merging with many thanks for your patience.

techman83:
@techman83 isn't 100% sure either. Part of the GitHub jenkins plugin probably. #notajenkinsexpert

PR #3679, opened 2016-04-22
Add Mk3 Hypersonic Systems from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of nestor_d, to add [Mk3 Hypersonic Systems](https://spacedock.info/mod/562/Mk3%20Hypersonic%20Systems) to CKAN.

Mod details:
    name = /mod/562/Mk3%20Hypersonic%20Systems
    author = nestor_d
    description = 
[[http://imgur.com/a/KTiFT]]
Kerbal Engineers at N5 Aerodynamics and Space found the lack of sleeker and more diverse components for  Mk3 spaceplanes disturbing, so they set out to remedy just that! This mod adds a few parts geared toward making Mk3 SSTOs more aerodynamic and look cooler!

 
NEW VERSION!
Includes reworked textures and models, plus four brand new parts, redone IVA, and new particle effects!

 
Parts included are:

1. Mk3 Hypersonic Cockpit - A stylish command pod that fits five kerbals with a much pointier shape for less drag

2. Mk3 radial attachment fuselage extensions - This essentially turns the Mk3 fuselage into a lifting body, much like the Mk2. Three parts fall into this category: A wing edge root extension, a lifting body extension and a RAM intake.

3. Hybrid high-bypass turbofan/ramjet 2.5m engine - Much like Mk2 spaceplanes use the 1.25 jet-engines, it seems only natural that Mk3 spaceplanes could rely on proportionally large jet engines. This engine is just that.

4. A 2.5m shock cone intake. Following the same logic, 2.5m intakes are needed for larger spaceplanes.

5. A 2.5 switchable LF/LFO 10m long tank - There's plenty of 2.5m tanks, but this one allows for fuel selection, very useful for spaceplanes.

6. Radial attachment scramjet engine - This is a rather large scramjet with its own intake. Like real life scramjets it will not produce thrust until it's going at high speeds.

New parts:
7. Mk3 Hypersonic Drone Core. A sleek drone core designed for super high speeds

8. Mk3 Hybrid high-bypass turbofan/ramjet. A larger version of the size 2 engine with moar thrust.

9. Mk3 Ram Intake. A big jet engine is gooing to need lots of air.

10. Mk3 Aerospike rocket. Because this parts pack is geared toward making spaceplanes, it makes sense to have a rocket engine suitable for the purpose.
[[https://www.youtube.com/watch?v=R7qlc4msCYw]]

[...]

PR #3693, opened 2016-04-23
Mk3HypersonicSystems to spacedock, closes #3679

plague006:


techman83:
LGTM. Merging with thanks!

plague006:
@techman83  Why/how did Jenkins install this correctly when, from all account, it shouldn't have? For comparison I created https://github.com/KSP-CKAN/NetKAN/pull/3773. The difference in the file tree being output is drastic: 
`- GameData/Mk3HypersonicSystems/Parts/Mk3Aerospike/Aerospike.mbm` in this PR's Jenkins run
`- GameData/Mk3HypersonicSystems/GameData/Mk3HypersonicSystems/Parts/Mk3Aerospike/Aerospike.mbm`  In #3773's Jenkins run.

I took a look at the files in the spacedock changelog: version 1.1 (version before this Jenkins run) has a folder structure identical to 1.1.0 (this run) and 1.1.0.1 (the run in 3773). We haven't previously indexed 1.1.0 so it shouldn't be matter of an old cached file.

techman83:
Jenkins doesn't cache files at all (It'd be nice if it cached in the same job, but the CKAN client eats symlinks).

So I'm not sure. I could write a line to check for GameData folders within the GameData folders easy enough.

PR #3695, opened 2016-04-23
Update KSPLogiRGB.netkan

Kaeltis:
Changed homepage to forum thread


politas:
Excellent! Thanks for that! Merging now.

PR #3696, opened 2016-04-23
New Kerbin-Side packs on Spacedock

politas:
Closes #3681, #3678, #3674, #3671


politas:
Checked with AlphaAsh, and Kerbin-Side-Core is not a core in the sense we mean.

plague006:
You need the word "closes" before each number.

PR #3699, opened 2016-04-23
Additional shared file in Contares mods

politas:
/GameData/Contares/Patches/CONTARES_TweakScale.cfg is shared across Contares mods.
Closes KSP-CKAN/CKAN#1676

### DIFF
--- a/NetKAN/Contares-Antennas.netkan
+++ b/NetKAN/Contares-Antennas.netkan
@@ -13,7 +13,7 @@
         {
             "find": "Contares",
             "install_to": "GameData",
-            "filter": "Agencies"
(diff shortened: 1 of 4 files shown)


Dazpoet:
I see that these two are also in at least two of the mods

```
- GameData/Contares/Patches/CONTARES_KIS.cfg
- GameData/Contares/Patches/CONTARES_TAC.cfg
```

politas:
I think that's Ok for now, because those two mods conflict with each other. CA1 is a sub-pack of Contares.

Dazpoet:
My mistake, this looks fine then. Thank you for all the hard work :D

PR #3702, opened 2016-04-23
Fix install for DefaultActionGroups

plague006:


Dazpoet:
Jenkins is spazzing out something fierce and spewing `tar` in his testing so I'm pinging @techman83 just to be sure.

The install looks fine though!

plague006:
The timestamps get borked like that from time to time.

techman83:
Seems that box was missing ntp. Sorted that out.

PR #3703, opened 2016-04-23
Update SDHI SMS to v3.1.1 (KSP 1.1.0)

sumghai:


sumghai:
NetKAN admins, hold off on this until @Starwaster has officially updated AnimatedDecouplers

plague006:
if we don't merge this v3.1.1 will continue to be available to KSP 1.0.5 users. If we do merge this v3.1.1 will become unavailable since the dependency on `AnimatedDecouplers` won't be met (users won't be able to incorrectly install the mod).

sumghai:
In that case, can we temporarily delist until this is resolved?

Alternatively, CKAN by default only shows mods compatible with the current version of KSP (1.1) anyway.

plague006:
CKAN lists mods compatible with a user's version of KSP. If I merge this it'll have the following effect:

`v3.1.1` will no longer be available to KSP 1.0.5 users.
`v3.1.1` will not be available to KSP 1.1 users until a KSP 1.1-compatible `AnimatedDecouplers` release is indexed.

That's the desired effect, right?

sumghai:
Yes, that's right.

plague006:
Merging with my thanks.

sumghai:
Addendum - most users will install SDHI SMS manually, using an unofficial (and therefore not listed on CKAN) 1.1-compatible version of AD.

PR #3714, opened 2016-04-24
Update for 1.1, primary moved to SpaceDock

lOmicronl:
Updates Near Future Propulsion.

Added version checking (hope I did that right?), removed some deprecated relationships.

### DIFF
--- a/NetKAN/NearFuturePropulsion.netkan
+++ b/NetKAN/NearFuturePropulsion.netkan
@@ -1,16 +1,13 @@
 {
-    "spec_version"   : "v1.4",
-    "identifier"     : "NearFuturePropulsion",
-    "$kref"          : "#/ckan/github/ChrisAdderley/NearFuturePropulsion",
-    "name"           : "Near Future Propulsion",
-    "abstract"       : "Advanced engine technologies in a stockalike aesthetic",
-    "author"         : "Nertea",
-    "release_status" : "stable",
-    "license"        : "CC-BY-NC-SA-4.0",
-    "ksp_version"    : "1.0.5",
-    "resources": {
-        "homepage": "http://forum.kerbalspaceprogram.com/threads/52042-0-21"
-    },
+    "spec_version"        : "v1.4",
+    "identifier"          : "NearFuturePropulsion",
+    "$kref"               : "#/ckan/spacedock/557",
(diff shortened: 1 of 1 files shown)


politas:
Very happy to see these turn up! Some of my favourite mods. Assuming the information on Spacedock is the same, you could dispense with the name, author and abstract.

lOmicronl:
*shrug* It's kinda always been in there. Also, I like writing custom abstracts, so that they are shorter and more to the point. :P

NetKAN/NearFuturePropulsion.netkan
```diff
@@ -1,16 +1,13 @@
...
-    },
+    "spec_version"        : "v1.4",
+    "identifier"          : "NearFuturePropulsion",
+    "$kref"               : "#/ckan/spacedock/557",
+    "name"                : "Near Future Propulsion",
+    "abstract"            : "Advanced electric engines in a stockalike aesthetic.",
+    "author"              : "Nertea",
+    "license"             : "CC-BY-NC-SA-4.0",
+    "$vref"               : "#/ckan/ksp-avc",
+    "x_netkan_license_ok" : true,
```
plague006: `x_netkan_license_ok` is redundant when we have a hardcoded `license`

plague006:
> I like writing custom abstracts, so that they are shorter and more to the point. :P 

If you have the author's blessing then that's cool, or if you're changing the mod author's incorrect or unhelpful `abstract` then we're happy to have better information for our users. However if you're just changing `abstract` because you prefer your writing style we would prefer to have the mod author's own words.

lOmicronl:
I'm sort of "officially" maintaining the CKAN listings forn Nertea, who has no interest in doing so himself. He's pretty much told me to go do hatever I think is best in that regard.

As for the license thing - okay, makes sense I guess. I copied the x_netkan thing from the SpaceDock guide without really understanding what it does, as it isn't explained very well there. The reason all these files have hardcoded licenses is because at some point in the past they weren't being pulled properly from Kerbalstuff, and since it worked fine after they were in there, I never bothered to remove them again.

I'll go and remove the extraneous line. Afterwards things should be good to go?

plague006:
:+1:

PR #3715, opened 2016-04-24
Update NearFutureSolar-Core for 1.1, primary moved to SpaceDock

lOmicronl:


lOmicronl:
Not sure why the build check failed. JSONLint says the syntax is valid...

plague006:
You can't use `find` on a file without including `"find_matches_files": true` in the install stanza.

NetKAN/NearFutureSolar-Core.netkan
```diff
@@ -0,0 +1,17 @@
+{
+    "spec_version"        : "v1.4",
+    "identifier"          : "NearFutureSolar-Core",
+    "$kref"               : "#/ckan/spacedock/559",
+    "name"                : "Near Future Solar Core",
+    "abstract"            : "This is the Near Future Solar plugin stand-alone, for adding functionality to other mods. It contains no parts and does nothing by itself.",
+    "author"              : "Nertea",
+    "license"             : "CC-BY-SA-NC-4.0",
```
plague006: `"CC-BY-NC-SA-4.0"`

lOmicronl:
Thanks, I think I fixed it by changing the "find" to a "file"... yep, the build suceeds now.

In regards to the license - how odd! I thought it was the wrong way around, because SpaceDock displays it differently on the mod pages, so I went and changed it everywhere. Guess CKAN didn't like that! ;)

plague006:

PR #3717, opened 2016-04-24
Update NearFutureConstruction for 1.1, primary moved to SpaceDock

lOmicronl:


lOmicronl:

NetKAN/NearFutureConstruction.netkan
```diff
@@ -1,11 +1,14 @@
...
-    "identifier"     : "NearFutureConstruction",
-    "$kref"          : "#/ckan/kerbalstuff/348",
-    "release_status" : "stable",
-    "license"        : "CC-BY-NC-SA-4.0",
+    "spec_version"        : "v1.4",
+    "identifier"          : "NearFutureConstruction",
+    "$kref"               : "#/ckan/spacedock/563",
+    "abstract"            : "Structural components and truss segments for building deep space vessels and bases. Includes various skeletal adapters and docking ports.",
+    "author"              : "Nertea",
+    "license"             : "CC-BY-SA-NC-4.0",
```
plague006: You want `CC-BY-NC-SA-4.0`

lOmicronl:
Now it complains about B9PartSwitch, which - as far as I can tell - points towards an external Netkan file. Is that error expected in such a case? Will it still work?

Or does the error mean that there is no 1.1 compatible release at the end of that external Netkan file? Which would be kind of odd, since Nertea bundled a 1.1 compatible version with the direct SpaceDock download, and he'd have to have gotten it from somewhere.

plague006:
There's no 1.1 release on github.

plague006:
Tests fine locally. As a heads up this won't be available via CKAN until a version of `B9PartSwitch` is available.

Dazpoet:
I see in the B9PartSwitch repository that @blowfishpro has what appears to be a ready to go 1.1 version of the .dll but no release yet.

blowfishpro:
I was waiting for something that actually requires it to be released.  I'll package up a release today.

blowfishpro:
New version released, should be picked up by NetKAN on the next run

PR #3718, opened 2016-04-24
Update NearFutureElectrical-Core for 1.1, primary moved to SpaceDock

lOmicronl:


lOmicronl:

NetKAN/NearFutureElectrical-Core.netkan
```diff
@@ -1,16 +1,13 @@
...
-    "resources": {
-        "homepage": "http://forum.kerbalspaceprogram.com/threads/52042-0-21"
-    },
+    "spec_version"        : "v1.4",
+    "identifier"          : "NearFutureElectrical-Core",
+    "$kref"               : "#/ckan/spacedock/558",
+    "name"                : "Near Future Electrical Core",
+    "abstract"            : "This is the Near Future Electrical plugin stand-alone, for adding functionality to other mods. It contains no parts and does nothing by itself.",
+    "author"              : "Nertea",
+    "license"             : "CC-BY-SA-NC-4.0",
```
plague006: `"CC-BY-NC-SA-4.0"`

plague006:
The failure is caused because netkan gets too many "hits" for .version files. It detects `B9PartSwitch.version`, `CRP.version`, and `NearFutureElectrical.version` and doesn't know which one to use. I'll think on this a bit to see if I can come up with an elegant workaround.

plague006:
Because of how `$vref` works there isn't an elegant solution for this case. You'll need to hardcode `ksp_version`.

lOmicronl:
Many thanks for your super patient assistance in the face of my helpless flailing around the spec. The AVC thing was another thing I thought I should add because the guide makes it sound like a great idea, but clearly I wasn't aware of what it would cause to happen.

plague006:
No fault to you for the AVC thing. It generally *is* a good idea to use it, this is just a very niche scenario where this particular metadata isn't installing the .version file for this mod so netkan scans the entire .zip for any and all .version files.

PR #3720, opened 2016-04-24
Update NearFutureElectrical-DecayingRTGs for 1.1, primary moved to SpaceDock

lOmicronl:


lOmicronl:

plague006:
Same issue as https://github.com/KSP-CKAN/NetKAN/pull/3718

plague006:
Because of how `$vref` works there isn't an elegant solution for this case. You'll need to hardcode `ksp_version`.

plague006:
Tests fine locally, merging with my thanks.

PR #3722, opened 2016-04-24
Use AVC integration for UbioWeldContinued

Darael:
...rather than a hardcoded ksp-version setting.

Should help to address UbioWeldingLtd/UbioWeldContinued#38


plague006:
Looks good, merging with my thanks. (As a note in case you're not aware: this won't make `2.3.0_Beta` of the mod available via CKAN. @netkan-bot doesn't index releases that are tagged as pre-release).

Darael:
Yeah, I'd got that.  I'll pass it on, though - I don't think I'd made it entirely clear.

PR #3741, opened 2016-04-25
Update TWR1.netkan

SirDiazo:
KSP 1.1 update.


plague006:
Merging with my thanks.

PR #3742, opened 2016-04-25
Add ION RCS from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of Jplrepo, to add [ION RCS](https://spacedock.info/mod/586/ION%20RCS) to CKAN.

Mod details:
    name = /mod/586/ION%20RCS
    author = Jplrepo
    abstract = Contains a ION RCS thrusters and Pulsed Plasma Thrusters RCS pack of parts. These parts used to be distributed as part of my AmpYear mod. But due to popular demand I have split them off to a separate parts pack.
    license = GPLv3
    Homepage = 
    description = ION RCS.   
Pulsed Plasma Thruster RCS.   

[[http://imgur.com/a/P9Pcc#0]]

**Ion RCS**
Thrusters that use Xenon-gas and electricity for incredibly efficient RCS. Some limitations:   
Has half the Isp of normal Ion engines.   
One quarter of the thrust of normal monopropellant RCS thrusters.   
Available in one-way linear thruster, 4-way ION RCS block and 4-way 45degrees ION RCS block.   
Integrated to TechTree at ionPropulsion.  

**Pulsed Plasma Thrusters RCS**    
Each thruster contains a fixed amount of Teflon and uses this along with Electric Charge for incredibly efficient propulsion for small craft or maneuvering.   
Some limitations:   
Has roughly quarter the ISP of normal RCS thrusters.   
One eighth of the thrust of normal monopropellant RCS thrusters.   
Only linear thrusters are available.   
Integrated to TechTree at advFlightControl.   
* See (http://en.wikipedia.org/wiki/Pulsed_plasma_thruster).    

Please direct questions about this pull request to [Jplrepo](https://spacedock.info/profile/Jplrepo).

PR #3743, opened 2016-04-25
Add IONRCS, close #3742

plague006:


Dazpoet:
Somehow github hates my workcomputer so I cannot look at this correctly....

However I recall us having removed the RepoSOFTtech agency and things in other mods so I guess there's a core package somewhere which this should depend on rather than including the files?

politas:
REPOSoftTech-Agencies is the one Daz is thinking of.

plague006:
Awesome catch Daz.

PR #3711, opened 2016-04-24
HideEmptyTechNodes to SpaceDock, closes #3707

plague006:
Closes https://github.com/KSP-CKAN/NetKAN/pull/3747

PR #3747, opened 2016-04-25
Create HideEmptyTechTreeNodes.netkan

jyanjyanjyan:
This is the mod author. It looks like there's some other pull requests already set up (using the old KerbalStuff netkan file it looks like), but please use this one. I changed the name of the mod slightly, so this file has the correct identifier. It also contains the correct "depends" and "suggests" fields.

It also follows the instructions on your "Adding a mod to CKAN" webpage, so I think is more up to date? For example, the website says to use "spec_version" : 1, but the old netkan file has 1.4. I'm not sure if that matters, but hey.

### DIFF
--- a/NetKAN/HideEmptyTechTreeNodes.netkan
+++ b/NetKAN/HideEmptyTechTreeNodes.netkan
@@ -1,0 +1,13 @@
+{
+    "spec_version"   : 1,
+    "identifier"     : "HideEmptyTechTreeNodes",
+    "$kref"          : "#/ckan/spacedock/577",
+    "x_netkan_license_ok": true,
+
+    "depends" : [
+        { "name" : "ModuleManager" }
+    ],
+    "suggests" : [
+	{ "name" : "CommunityTechTree" }
+    ]
+}


plague006:
> I changed the name of the mod slightly, so this file has the correct identifier

`identifier` is not a field to be changed as this is an internal CKAN ID which allows mods to be properly identified, installed, and upgraded. `name` is what users see via the CKAN client. With that said, I see you want CTT from a `suggests` to a `recommends` and the licensing of the mod has changed so I'll update my PR with that. If you would prefer this PR merged let me know and I'll give you line comments on the changes that will be required.

Ref. https://github.com/KSP-CKAN/NetKAN/pull/3711/files

jyanjyanjyan:
Hmmm I originally thought KerbalStuff was abandoned, so I wanted to try to start fresh with SpaceDock and update the identifier (as I did my internal source code) to match the title on the forum page. Even though it's hidden from users, would it be possible to use the new identifier for consistency's sake and delete the old netkan from the KerbalStuff days?

jyanjyanjyan:
Oh also, the CTT is changing from a recommends to a suggests, not the other way around.

plague006:
Ah yes I miswrote which direction CTT is changing, sorry about that.

The `identifier` =/= the netkan. The netkan file is frozen by using a .kerbalstuff file extension but the `identifier` is still in place in our extant metadata in both https://github.com/KSP-CKAN/CKAN-meta/blob/master/HideEmptyTechNodes/HideEmptyTechNodes-0.3.ckan and https://github.com/KSP-CKAN/CKAN-meta/blob/master/HideEmptyTechNodes/HideEmptyTechNodes-0.2.ckan. There are currently CKAN users out there with your mod installed under the `HideEmptyTechNodes` identifier and if we were to create a new identifier for your mod that strands the userbase without a path to upgrade (a lot of people don't uninstall old mods when upgrading ksp).

jyanjyanjyan:
Fair enough. Was just kind of an OCD thing (but not really OCD). Okay, please merge the PR. Several questions, though.

Would merging fix the remaining changes that need to be made on [3711](https://github.com/KSP-CKAN/NetKAN/pull/3711/files)? For example, I don't see 1) "requires : ModuleManager", 2) the min version should probably be deleted, and 3) I deleted the "homepage" part because I figure CKAN would automatically take the correct forum page from SpaceDock.

Also, 4) my license is still MIT, is it not? It's included in my plugin's GameData folder, and it's listed as MIT on SpaceDock.

One last question: on SpaceDock I reuploaded all the old versions as well, referencing them to use the correct version of KSP. But I had also changed the ZIP file names from HideEmptyTechNodes-## to HideEmptyTechTreeNodes-##. Will this cause any issues since we are using the same identifier?

plague006:
> Fair enough. Was just kind of an OCD thing (but not really OCD).

I hear you, there's a lot of legacy `identifier`s that could be improved to help with searching (when we want to find the right files to modify) but changing them is something we try to avoid at all costs.

> Okay, please merge the PR. Several questions, though.

I'm not sure which PR you mean but I'll answer based on the assumption you're asking about #3711 

> 1) I don't see "requires : ModuleManager", 2) the min version for MM should probably be deleted

The `min_version` for MM is inside of the `depends` (`requires` isn't a thing). It's legacy metadata so I can remove the `min_version`.

> 3) The homepage

Yep that's legacy too and while the link works, having it dynamically scraped from the spacedock page will mean that if you ever create a new thread we won't have to change the metadata directly.

> 4) my license is still MIT, is it not? It's included in my plugin's GameData folder, and it's listed as MIT on SpaceDock.

Yes the MIT is included in the mod and spacedock while the KSP forum thread links to http://creativecommons.org/licenses/by-nc/4.0/ -- in cases of confusion we err on the side of the more restrictive license but seeing as you're telling me directly it's under MIT then I'll take that as canon.

> on SpaceDock I reuploaded all the old versions as well

Netkan only scrapes for the most recent version of a mod's release.

jyanjyanjyan:
Oh, you're right about the license. Very sorry for the confusion, please use the creative commons license. I will update the license in the mod and SpaceDock page once the mod appears on CKAN, if that's okay,

PR #3748, opened 2016-04-25
Pathfinder compatibility override

plague006:


politas:
LGTM, merging with thanks

PR #3749, opened 2016-04-25
New Version for KSP 1.1

Sphinx3k:


Sphinx3k:
I hope this was the right way. Still trying to figure out everything.

NetKAN/AtmosphereAutopilot.netkan
```diff
@@ -1,11 +1,11 @@
 {
-    "spec_version" : "v1.4",
+    "spec_version" : "v1.5.2",
```
plague006: `spec-version` doesn't change. `spec_version` is the version of the CKAN specification that a metadata file is using (for example if you're using the `find` function you need `"v1.4"`.

plague006:
`ModularFuelTanks-5.7.0.netkan` has the contents of a .ckan file not a .netkan. A netkan requires a `$kref` to scrape information from a source.

In addition https://github.com/KSP-CKAN/CKAN-meta/blob/master/ModularFuelTanks/ModularFuelTanks-5.7.0.ckan is available already.

plague006:
ffeba995

plague006:
Closing this due to the referred PR/issue/commit.

PR #3757, opened 2016-04-25
Fix install for ItsTheLittleThings

plague006:
Another convoluted regex to cover all kinds of capitalization changes.

In this case the folder went from `Its the little things` to `Its the little Things`

This regex covers all cases of caps changing as well as spaces between words and the addition of an apostrophe `'` to `Its`.

I'm a regex scrub so there are probably 100 more efficient ways to do this.

### DIFF
--- a/NetKAN/ItsTheLittleThings.netkan
+++ b/NetKAN/ItsTheLittleThings.netkan
@@ -1,11 +1,11 @@
 {
     "identifier": "ItsTheLittleThings",
-    "spec_version": 1,
+    "spec_version": "v1.10",
     "$kref": "#/ckan/spacedock/462",
     "license": "CC-BY-SA-2.0",
     "install": [
         {
-            "file": "its the little things/game data/Its the little things",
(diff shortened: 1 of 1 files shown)


dbent:
`(?i)game\s*data/it'?s\s*the\s*little\s*things` should work.

`(?i)` is a [miscellaneous construct](https://msdn.microsoft.com/en-us/library/x044wc7s%28v=vs.110%29.aspx), basically options that configure how the regex is parsed. In this case it makes all the letters match case-insensitively.

`\s` is a whitespace character
`*` is equivalent to `{0,}` i.e. "any amount of the previous"
`+` is equivalent to `{1,}` i.e. "at least one of the previous"
`?` is equivalent to `{0,1}` i.e. "either 0 or 1 of the previous"

However, because `\` needs to be escaped in JSON, it would be:

`(?i)game\\s*data/it'?s\\s*the\\s*little\\s*things`

plague006:
Ah `(?i)` isn't something I'd run across and is exactly the kind of time-saver I was looking for. As for using `?` that's something I totally should've remembered. Thanks a bunch.

techman83:
Awesome, LGTM. Merging with Thanks.

plague006:
Since this was authored the directory name changed to `itstheLittleThings`. Handy thing that regex.

PR #3758, opened 2016-04-25
Fix install for RLA-Stockalike

plague006:


techman83:
LGTM, merging with thanks!

PR #3734, opened 2016-04-24
FinalFrontier to spacedock

plague006:
http://spacedock.info/mod/580

PR #3762, opened 2016-04-26
Update FinalFrontier to 1.1.0

SmarterThanMe:
Please let me know if I've stuffed anything up.

### DIFF
--- a/NetKAN/FinalFrontier.kerbalstuff
+++ b/NetKAN/FinalFrontier.kerbalstuff
@@ -20,8 +20,8 @@
             "delete" : [ "ksp_version" ],
             "override" :
             {
-                "ksp_version_min" : "1.1.0",
-                "ksp_version_max" : "1.1.0"
+                "ksp_version_min" : "1.1",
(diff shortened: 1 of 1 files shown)


plague006:
https://github.com/KSP-CKAN/NetKAN/pull/3734 is already open to use spacedock rather than github.

Issues with this PR (don't make any changes until you read the last issue):
When using a github kref you must hardcode an `abstract` and should hardocde a `name`, `author`, and a `homepage` in the the `resources`. The `install` stanza is incorrect and would result in `GameData/GameData/Nereid/ModName` (that one is our fault, it was already written that way when you edited the file). Netkan overrides are for when mods have faulty metadata from the mod's host (ex. spacedock/kerbalstuff) or in their .version file.

The biggest issue though is that there are no binary releases in the FinalFrontier repo: https://github.com/Nereid42/FinalFrontier/releases

SmarterThanMe:
Thanks plague006 for comprehensively aschooling me on how to do this. :)

I'll close in favour of #3734.

PR #3765, opened 2016-04-26
Add B9-PWings-Modified, closes #3672

plague006:


techman83:
LGTM, merging with thanks!

PR #3775, opened 2016-04-26

DuoDex:
Correct author.


plague006:
Merging with my thanks.

PR #3776, opened 2016-04-26
Create VesselMover.netkan

DuoDex:
Adds Vesselmover back to the CKAN.


plague006:
Need to delete `VesselMover.kerbalstuff` as part of this.

plague006:

PR #3778, opened 2016-04-27
Add ExceptionDetector back into CKAN

godarklight:
Updated because people have been asking me :)


plague006:

PR #3779, opened 2016-04-27
StockVisualEnhancements fix

politas:
Use bundled EVE and conflict with the full EVE.

### DIFF
--- a/NetKAN/StockVisualEnhancements.netkan
+++ b/NetKAN/StockVisualEnhancements.netkan
@@ -5,15 +5,9 @@
     "license"        : "CC-BY-NC-SA",
     "release_status" : "stable",
     "x_netkan_epoch" : "1",
-    "depends": [
-        { "name" : "EnvironmentalVisualEnhancements" }
-    ],
-    "provides": [
-        "EnvironmentalVisualEnhancements-Config"
-    ],
     "conflicts": [
         { "name": "SVE-Textures" },
-        { "name" : "EnvironmentalVisualEnhancements-Config" }
+        { "name" : "EnvironmentalVisualEnhancements" }
     ],
     "recommends": [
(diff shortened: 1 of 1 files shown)


DuoDex:
See also PR #3784

Nhawks1717:
politas just in case you didn't see my comment on the forums. The current version of SVE has the latest release of EVE bundled in it.

politas:
So, is SVE anything more than an EVE configuration set?

DuoDex:
This most recent config will break SVE. SVE currently provides its own surefire working EVE plugin, as EVE updates break config packs from time to time. Installing EVE _over_ SVE causes some impressive graphical and performance issues, and that's exactly what you seem to be proposing to do.

politas:
The conflict with EnvironmentalVisualEnhancements-Config, shouldn't have allowed any other EVE config to install, but ok, let's go back to using the bundled EVE and conflicting with EVE.

pjf:
Just checking the conversation vs the code. The code changes have this depend upon EVE rather than conflicting with it. Is that correct?

politas:
No, it's installing the bundled EVE and conflicting with EVE

plague006:
@Nhawks1717 should CKAN only be installing the spacedock release or should we make each quality level available via the github releases?

politas:
That would require Github releases.

Nhawks1717:
So is it fixed now or no? I'm a bit confused.

DuoDex:
The current state of the config should work. At the moment SVE provides both configuration and base - and conflicts with anything that provides either configuration _or_ base.

Nhawks1717:
Okay, I'll let the thread peps know this'll be merged soon. Thanks for the hard work guys.

plague006:

PR #3785, opened 2016-04-27
Set RSSTextures to be 1.1-compatible

aw1621107:
RSS has been released for 1.1, and there aren't any texture changes, so
the textures for 1.0 should also work for 1.1


plague006:
https://github.com/KSP-CKAN/NetKAN/pull/3788

PR #3788, opened 2016-04-27
Update RSS-Textures netkan for 1.1 4096

NathanKell:
Thanks to @politas !


politas:
I rather doubt KSP 1.1 is going to go over 100 patch levels.

PR #3794, opened 2016-04-28
Merged MusicMute and MusicMute Continued

nimobeeren:
I have spoken with @mattolenik and we've decided to merge both our forks
into one CKAN entry, the old KerbalStuff version can also be removed.


plague006:
Because `MusicMuteContinued` already has a 1.1 release I would recommend using that identifier so that the userbase doesn't fragment / has an upgrade path. `identifier` isn't something we surface to users and you can/should stick with the `MusicMute` `name`.

nimobeeren:
@plague006 That's probably better, should I change the identifier and the file name to MusicMuteContinued(.netkan)?

plague006:
Exactly right.

nimobeeren:
Should be good now, feel free to ask if my intentions are not clear.

plague006:
You need to add the following:
````
"install": [
        {
            "find"       : "MusicMute",
            "install_to" : "GameData"
        }
    ]
````

**Comment by plague006** [comment]
Without an install stanza CKAN will try to find a folder with a name that matches `identifier` and install that. If it can't find that folder name it fails. This install stanza tells it what folder to install.

File: NetKAN/MusicMuteContinued.netkan [hunk 1]
```diff
@@ -1,12 +1,18 @@
 {
     "spec_version" : 1,
```
plague006: The `1` needs to be changed to `"v1.4"` in order to use `find` in the install stanza.

plague006:

nimobeeren:
Phew, great. Thanks again :)

PR #3797, opened 2016-04-28
Move PartCommander to SpaceDock

seanmcdougall:
I've released a new version with KSP 1.1 support via SpaceDock and would like to get CKAN updated.  Thanks!


plague006:
:+1: 
If at some point you'd like, we can make use of the information in your .version file by adding the line `"$vref": "#/ckan/ksp-avc",`. It can be useful if you want to set a `ksp_version_min` and `ksp_version_max` since spacedock (afaik) doesn't allow compatibility ranges.

PR #3801, opened 2016-04-29
Update HangarExtender to 1.1

aw1621107:
Alewx has released 3.4.8, which is 1.1-compatible


plague006:
If the mod is compatible with 1.1.1 you should use `ksp_version_min` and `ksp_version_max` of `1.1.0` and `1.1.1` respectively.

plague006:
Merging this now, 1.1.1 compatibility can be addressed in the future.

PR #3808, opened 2016-04-29
Update KerbalRenamer for KSP11

NathanKell:


plague006:
How I feel when you don't delete your branches

PR #3810, opened 2016-04-29
Create OIFC.netkan

Orum:
This will require my other mod's listing, ODFC.netkan, to be approved in order to function correctly (for the dependency).


plague006:
In cases like this you can PR both .netkans together and Jenkins will test them accurately.

Orum:
Ah, thanks, I'll keep that in mind.

plague006:

PR #3817, opened 2016-04-29
SimpleConstruction fix install

politas:
This hasn't been getting updates, presumably due to a folder structure change and find not doing so great with the root directory.

### DIFF
--- a/NetKAN/SimpleConstruction.netkan
+++ b/NetKAN/SimpleConstruction.netkan
@@ -12,7 +12,7 @@
     ],
     "install": [
         {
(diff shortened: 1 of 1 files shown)


politas:
Seems to work. It's certainly finding 1.7

plague006:
There's nothing wrong with the install stanza, we have 1.7 indexed (https://github.com/KSP-CKAN/CKAN-meta/tree/master/SimpleConstruction). The problem is forcing a v. Can't modify branch since it's your private forking so closing for https://github.com/KSP-CKAN/NetKAN/pull/3818.

PR #3818, opened 2016-04-29
Force v to SimpleConstruction

plague006:


politas:
D'Oh! Why didn't I spot that? Good catch!

PR #3814, opened 2016-04-29
Add Jettison from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of BiffSpace, to add [Jettison](https://spacedock.info/mod/621/Jettison) to CKAN.

Mod details:
    name = /mod/621/Jettison
    author = BiffSpace
    abstract = Adds a button to the right-click menu of parts with resources, allowing you to dump the resource
    license = GPLv3
    Homepage = 
    description = **Jettison - v 1.0 for KSP 1.1.1**

This mod allows individual resource tanks to be emptied from the right-click menu.

RELEASE (2016-4-29), v 1.0.0.6 

**Overview**

Resources may be dumped from tanks using the right-click menu. This is for the cases where you have brought along monoprop in ships that don't use it or oxidizer in tanks feeding nuclear engines, or you need to get rid of extra mass by any means.

A button is added to the right click menu of any tank with resources, except for electric charge, ablator, solid fuel, and ore (which already has this functionality in stock). For tanks holding both liquid fuel and oxidizer, clicking once empties the oxidizer, and a second click empties the liquid fuel. Other mod resources should be treated as stock ones, but if there are different types of mod resource held in the same tank, all will be emptied with one click.

**Installation**

The Jettison folder needs to be unzipped and put in your KSP\GameData folder. 

Jettison requires Module Manager, which is not included in the zip file. This should go in the GameData folder. 

**Code**

The source code is in the Source folder in GameData\Jettison.

**Licence**

Jettison, its associated files and the source code are all released under the GPL 3 licence, text here: http://www.gnu.org/licenses/gpl.txt.

Please direct questions about this pull request to [BiffSpace](https://spacedock.info/profile/BiffSpace).

PR #3823, opened 2016-04-29
Add Jettison, closes #3814

plague006:


politas:
Needed moduleManager, I think it's good to merge.

PR #3831, opened 2016-04-30
New KerboKatz SmallUtilities

politas:


plague006:
Where does `RecoverAll` fit into this? https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/RecoverAll.netkan

politas:
Ah, good catch, he's moved that one into the SmallUtilities group. I'll fix and resubmit.

politas:
Checked with mod author - they want a clean break on this mod, and will mark the old one as deprecated.

PR #3833, opened 2016-04-30
Fixed configs being installed to wrong folder.

Starstrider42:
This is a fix to a minor issue where installing Custom Asteroids configs would leave the dummy "Standard Setup" folder in GameData. All configs not associated with another mod should go into `GameData/Custom Asteroids/config`.

### DIFF
--- a/NetKAN/CustomAsteroids-Pops-Stock-Inner.netkan
+++ b/NetKAN/CustomAsteroids-Pops-Stock-Inner.netkan
@@ -1,5 +1,5 @@
 {
-    "spec_version": 1,
+    "spec_version": "v1.2",
     "identifier": "CustomAsteroids-Pops-Stock-Inner",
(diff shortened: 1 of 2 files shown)


politas:
Seems an awkward way to do the installs, rather than just finding the specific file in each case, but it works, so I'm merging with thanks!

Starstrider42:
So should it look like? (The original config wasn't mine, so I'm just making incremental changes).

politas:
If I were wanting to match a single file, I'd do it like this:

    "find": "Standard Setup/config/Basic Asteroids.cfg",
    "find_matches_files": true,
    "install_to": "GameData/CustomAsteroids/config"

Which would require bumping the spec_version to "v1.16"
(This method of using a negative filter_regexp would have been the only way to do it before v1.16, so it's entirely understandable)

Starstrider42:
Ok, thanks for the tip.

PR #3835, opened 2016-04-30
Update SDHI Strobe-O-Matic compat to KSP 1.1.2

sumghai:
This only uses stock part modules, therefore no plugin compatibility issues. Should be able to be updated ASAP.


plague006:

PR #3836, opened 2016-05-01
Update SDHI SMS compat to KSP 1.1.2

sumghai:
Dependencies should be a-OK as well.

If not, we'll deal with them.


sumghai:
Darn, looks like it's RealChutes, which was just updated for KSP 1.1.2 moments ago.

plague006:
Given the actual change in this PR is a single digit I'm not even going have Jenkins rebuild. #fullYOLO

PR #2263, opened 2015-09-11
1.0.4 override for TestFlight

plague006:
Closes #2262

Issue #2262, opened 2015-09-11
TestFlight in 1.0.4


NathanKell:
Works ok per me and per Tyler Raiz's video series. Looks like it can have its metadata overridden...

PR #3838, opened 2016-05-01
Remove netkan override, version should be grabbed from AVC

jwvanderbeck:


politas:
You do realise that override is only for release 1.3.1.1, which specified being compatible with 1.0.2 only, while the next release was a 1.0.5 recompile. 1.3.1.1 is the correct release to be available for users wanting to run KSP 1.0.2 - 1.0.4. Making this change is unlikely to have any effect, because Netkan only inflates the latest release, anyway.

jwvanderbeck:
I have no idea.  I didn't add that override, don't know why it was there, and don't see a reason for it to be there.  CKAN is confusing enough for me to worry about without having extra things added by someone else.

plague006:
I wrote the override in https://github.com/KSP-CKAN/NetKAN/pull/2263 on September 15, 2015 in response to #2262. At the time the current `ksp_version` was `1.0.4` and we were informed that your mod was compatible with it in spite of the version information being provided via the .version file (`ksp_version` `1.0.2`). To best serve our user base version `1.3.1.1` of your mod was marked with a compatibility range from `1.0.2` (provided by your .version file) up to `1.0.4` (provided by users of your mod via @NathanKell in #2262). 69 days after the override was written you released `1.4.0.2` of your mod marked as compatible with ksp `1.0.5` at which point the override was no longer "active" as `1.3.1.1` was no longer the current version of your mod.

I hope that helps clear things up for you.

plague006:
As there's no technical need to remove old overrides and could lead to incorrect metadata in the future I'm closing. If you feel this was in error please feel free to comment or re-open.

PR #3849, opened 2016-05-01
Test ColorCodedCans

plague006:
Fails indexing with `Cannot find central directory`. Hopefully Jenkins will shed some light.


plague006:
I'm going to merge this to see if it has any effect/see if it triggers an immediate update.

PR #3854, opened 2016-05-01
Update RSSeaDragon.netkan

NecroBones:
Adding RSB suggestion.


plague006:

PR #3865, opened 2016-05-01
Update HyperEdit.netkan

dbent:
Add a dummy URL fragment to bypass NetKAN's caching mechanism.


plague006:
Was able to test locally as I had an old cache lying around and this did what it says on the tin. #cachebusted

PR #3859, opened 2016-05-01
Update InterstellarFuelSwitch-Core.netkan

sswelm:
updated dependencies

PR #3866, opened 2016-05-01
removed all depandencies

sswelm:
The core dll does not have external mo dependencies,


NecroBones:
Supersedes  #3859

plague006:

PR #3867, opened 2016-05-01
Add override for SmartStage-v2.9.2

plague006:
Opened https://github.com/xytovl/SmartStage/issues/5 to try to validate the proposed compatibility with the author.


xytovl:
The .version file for SmartStage now has min and max versions in order to ignore minor KSP version, the override should not be required.

plague006:
:+1:

PR #3869, opened 2016-05-02
RN mod updates for ksp 1.1.2

raidernick:
-Update all netkan files for mod updates to new ksp version
-Switched reflection plugin to texture replacer as reflection plugin is
dead and doesn't work in 1.1+
-Craft files are being maintained again so they are now installed by
default again
-MM CFG files are now in the mod directories instead of GameData
directly, so I removed those lines as they are not needed any longer
-Removed KM gimbal as the stock module now contains all the
functionality km gimbal had.

PR #3870, opened 2016-05-02
For Jenkins test

plague006:
Will merge #3869 manually if this passes.


plague006:
Dependencies finally met.

PR #3873, opened 2016-05-02
Add vref and update authors for NRAP

plague006:


plague006:
Unclear on why `1.1.99` is being replaced with `1.1.0` when `build.sh` says `KSP_VERSION_DEFAULT="1.1.2"`.

techman83:
We should remove those testing files from NetKAN/CKAN-meta, all the build scripts live [here](https://github.com/KSP-CKAN/xKAN-meta_testing) now.

techman83:
LGTM, merging with thanks!

plague006:
That explains that.

PR #3875, opened 2016-05-02
Fix install CommunityNavballDockingIndicator

plague006:


politas:
Shame we don't have a find_matches_all option, eh? LGTM.

PR #3876, opened 2016-05-02
Mark latest CommunityResourcePack as explicitly compatible with KSP 1.1.2-1.1.99

dewiniaid:
Mark latest CommunityResourcePack as explicitly compatible with KSP
1.1.2-1.1.99.

Hopefully future versions of CRP will have an updated .version file.   The mod contains configs only, so nothing from 1.1.1 to
1.1.2 should break anything.


dewiniaid:
Also, most of RoverDude's mods have already been updated for 1.1.2 and depend on this (which he owns), so I'm fairly certain it's safe to say it's compatible.

plague006:
You'll have to PR directly to the file listed in the `$kref` and not to us. https://github.com/BobPalmer/CKAN/

PR #3879, opened 2016-05-02
RN naming fix

raidernick:
-Fix naming error in Soviet Spacecraft, it still said salyut, now the
name is correct.


plague006:

PR #3881, opened 2016-05-02
Attempting to add KSP-AVC versioning support to these two

DMagic1:


plague006:

PR #3883, opened 2016-05-02
create netkan file for davon supply mod on spacedock

PrivateFlip:
Hello I have created this pull request to add this file to NetKAN. A similar file "DavonSupplyMod.kerbalstuff" already exists which this file should replace. The .netkan file itself is tested with the ckan.exe and netkan.exe utilities. I am the owner of the mod this file refers to.

Yesterday a similar pull request was made by me but it did not show up in the history, so I suspect I did something wrong, hence I'm trying once more.


NetKAN/DavonSupplyMod.netkan
```diff
@@ -0,0 +1,17 @@
+{
+    "spec_version": 1,
+    "identifier": "DavonSupplyMod",
+    "x_netkan_license_ok": true,
```
plague006: You can remove `x_netkan_license_ok` because you have a hardcoded license.

plague006:
Thanks. This pull request has been integrated manually.

PrivateFlip:
Thanks for accepting the pull request. You removed the license line as well like you requested me to do; sorry for that, only checked back on the request today. I will try to be a little more attentive in the future. Keep up the great work!

PR #3885, opened 2016-05-02
Added KSP-AVC support of my StockMods

malahx:
Hello, I've added support of KSP-AVC on my mods. I don't know why but at this time, I don't see StockRT on the CKAN list, perhaps something goes wrong? As I can see, my last StockRT update is indexed on the metadata, I don't know what I can do to correct this.

Thanks :)

### DIFF
--- a/NetKAN/StockNoContracts.netkan
+++ b/NetKAN/StockNoContracts.netkan
@@ -1,6 +1,7 @@
 {
     "spec_version": 1,
     "$kref": "#/ckan/spacedock/111",
+    "$vref" : "#/ckan/ksp-avc",
     "identifier": "StockNoContracts",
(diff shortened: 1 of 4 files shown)


malahx:
Perhaps, it's the RemoteTech versionning which need a "v"? v1.6.0 ?

plague006:
This is failing because your mod depends on RT but there's no 1.1.2-compatible RT release. The only changes made that could potentially be an issue are `$vref` with .version files that aren't JSON-compliant but since the metadata tests passed that rules this possibility out.

So as a heads up `StockRT` won't be visible anywhere within the client until RT has a ksp 1.1.2 release. Ref https://github.com/KSP-CKAN/CKAN/issues/1695

PR #3886, opened 2016-05-02
Update depandancies

sswelm:


sswelm:
dependencies should be correct

plague006:

Issue #3890, opened 2016-05-03
Metadata standardization


plague006:
@politas @Dazpoet @techman83 @dbent @Olympic1 

With us moving towards a real mirrorkan (https://github.com/KSP-CKAN/CKAN/issues/1682) and having file hashes in the metadata (https://github.com/KSP-CKAN/CKAN/pull/1703) it's time to stop letting PRs into CKAN-meta unless it can't be avoided.

To that end I've scraped the metadata for all the identifiers we currently have and where .netkans didn't previously exist I have opened #3892 and #3893 to create either a .netkan if the mod meets certain criteria or a .frozen for mods that don't.

Speaking for myself in the past things have been a little unclear and inconsistent in how extensions were applied and when .netkans were deleted. I'd love for that to change moving forward.

`.frozen` should be applied to any former `.netkan` that no longer passes the indexer's test and can't be made to work. This will also include "placeholder" files to represent identifiers that are claimed. `.netkan` files should no longer be deleted in any case other than when we are de-indexing a mod fully and completely (no leftover references to the identifier anywhere in the metadata).

A little to-do list mostly to remind myself but I'd be ecstatic if someone else does it before me:

- [x] Add/modify PR template in ckan-meta to make submitters aware that they should PR to netkan.
- [x] Add/modify issue template in ckan-meta to redirect all issues to this repo.
- [x] Modify "[Adding mod](https://github.com/KSP-CKAN/CKAN/wiki/Adding-a-mod-to-the-CKAN)" wiki entry to reflect new metadata submission process.
- [x] Check to see if CONTRIBUTING/README files need updates (and update if applicable).

dbent:
`.ckangenerator` could just be `.netkan`, there's not much need to distinguish them.

plague006:
Yeah I don't know why I'd forgotten that the plan was to just PR the .netkan and have the indexer inflate the CKAN. I had in mind that we would still PR netkan-generated .ckan files. Will edit OP.

dbent:
LGTM

plague006:
For the sake of completeness I'm also going to create files for virtual modules.

techman83:
LGTM. The indexer attempts to inflate all files with the NetKAN extension in the NetKAN directory. Anything else gets ignored. So frozen is more of a convention than a hard requirement (a sane one, though I'd be happy with .doge :P).

Moving away from hand crafting metadata is a GoodThing™ :+1:

plague006:
[...]

PR #3893, opened 2016-05-03
Add netkans based on ckan-meta .ckans.

plague006:
These mod represent the set of ckan-only mods that have had more than 1 release in their history. The contents should represent the most recent .ckan we have. I validated the JSON of all files but did with netkan/ckan. I'll let Jenkins chew through as much of it as I can.

Part of #3890


plague006:
Fails Jenkins, but we're probably going to want to just modify the testing method. I'll leave that to @techman83 to deal with whenever he deems fit.

dbent:
It's not particularly harmful since they'll be overwritten if necessary anyway, but `download_size` doesn't need to be in any of these. (nor `x_generated_by`, but that's minor).

techman83:
Yeah, the testing expect certain things for NetKANs that don't apply for those. We'll just need to add some logic around when a [v|k] is and isn't needed.

plague006:
Oh yeah there's extra metadata inside because the .netkans here are just verbatim copies of the most recent .ckan we have on file. I don't have the right tool or knowledge set to automatically pull out the un-needed bits. We'll just have to modify them as they're used/as we go.

politas:
I can probably write up a script to strip out all the download_size and x_generated_by lines.

politas:
There we go.

plague006:
I thought about doing it that way too. Problem is I didn't have a way to remove the field while not breaking JSON validity. In your case you changed 36 files and 35 of them are no longer JSON valid. :S

techman83:
The JSON seems valid, it's the metadata tests that expect vref/kref's because they are netkans. We've changed the rules. Just need to change that test a little, if it has "x y + z" kref/vref isn't required. I'll take a closer look when I get back.

politas:
Ok, that does the same thing, but also cleans up the excess commas.

politas:
Ok, that ^^^ is a whole other category of netkans that will need to be sorted - The ones we have an old frozen netkan file pointing to kerbalstuff for.

plague006:
Unless there's a way to automate tracking down those fringe cases I see no issue with simply fixing them as new releases are issued.

plague006:
I was too lax with which mods I turned into .netkans. More will be added as frozen in another PR.

plague006:
Removed astro pack because watching Jenkins download the same (very large) file over and over was torture.

plague006:
Passes all Jenkins metadata tests which is what matters. Fails to install because an old version of `Toolbar` wasn't available to meet a dependency (entirely expected a dependency to be unavailable at some point).

This is ready to rumble.

techman83:
Awesome work @plague006, merging with epic thanks. Any issues we can resolve individually.

PR #3895, opened 2016-05-03
Create Historian-expanded.netkan with fixed missing comma

Aelfhe1m:


plague006:
You can fix this or #3894 whichever you prefer. Both are failing because the filename isn't identical to the `identifier` you set.

plague006:
You're able to pass local testing because neither netkan.exe nor ckan.exe contain all the testing needed to ensure compliance. You just need to rename `Historian-expanded.netkan` to `Historian-Expanded.netkan` or change the `identifier` from `Historian-Expanded` to `Historian-expanded`.

PR #3897, opened 2016-05-03
Patch 3

Aelfhe1m:


plague006:
You're able to pass local testing because neither netkan.exe nor ckan.exe contain all the testing needed to ensure compliance. You just need to rename `Historian-expanded.netkan` to `Historian-Expanded.netkan` or change the `identifier` from `Historian-Expanded` to `Historian-expanded`.

Aelfhe1m:
Thanks plague006.

plague006:

PR #3900, opened 2016-05-03
Add SMURFF

dbent:
FYI, @Kerbas-ad-astra

I assume SMURFF is compatible with KSP 1.1.2? In which case you'll probably want to update your `.version` file in the repository.

### DIFF
--- a/NetKAN/SMURFF.netkan
+++ b/NetKAN/SMURFF.netkan
@@ -1,0 +1,30 @@
+{
+    "spec_version": 1,
+    "identifier": "SMURFF",
+    "$kref": "#/ckan/github/Kerbas-ad-astra/SMURFF",
+    "$vref": "#/ckan/ksp-avc",
+    "x_netkan_version_edit": "^[vV]?(?<version>.+)$",
+    "name": "SMURFF",
+    "abstract": "Brings fuel mass fractions closer to reality, while staying simple.",
+    "license": "Apache-2.0",
+    "resources": {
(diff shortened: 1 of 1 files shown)


plague006:
> Suggestions

>Other addons that bring "real-ish" capabilities and challenges to Kerbal Space Program include:

>    AntennaRange, to "enforce and encourage antenna diversity". Relaying is handled automatically, but only if you've used the right antennas for the job. (We'll see how the 1.1 antenna system is when that comes.)
        I'll also advertise my AntennaRange Relays contract pack (for Contract Configurator), to give some guidance and financial support for deploying relays.
    SCANSat, to make biome and elevation maps (handy for planning landings) and require just a bit more effort when scanning for resources.
    USI Life Support, for a life support system which is simple and forgiving (unless you configure it to kill Kerbals).
        USI Kolonization Systems, for ISRU that allows self-sustaining colonies (with some effort).
        You should also consider picking up Extraplanetary Launchpads and OSE Workshop (and Kerbal Inventory System, OSE's dependency), as an extension of NASA's real-world interest in in-situ manufacturing and repair. (And to give your bases something new to do.)

dbent:
@plague006 I've added basically every mod mentioned as a `suggests` including the "recommendations" which probably shouldn't be installed by default.

plague006:

plague006:
Looks good.

Kerbas-ad-astra:
I haven't updated for 1.1.2 yet because I'm waiting on CryoEngines to officially update.

PR #3902, opened 2016-05-03
Update SigmaBinary-core.netkan

Sigma88:


NetKAN/SigmaBinary-core.netkan
```diff
@@ -1,18 +1,18 @@
 {
-    "spec_version" : "v1.4",
+    "spec_version" : "v1.5.1",
```
plague006: This should still be "v1.4". It's the version of the [CKAN spec](https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md) required to accurately read metadata.

Sigma88:
oh sorry about that
:)
see, I shouldn't try to do these things :D
better leave it to the pros ;)

plague006:
I always appreciate a mod author's efforts =)

PR #3904, opened 2016-05-03
RN us probes update

raidernick:
-Update us probes config, remove broken dependency
-Change bdanimation as it was abandoned? and update the code for 1.1.2,
if the author returns this can be changed back but the mod was
completely broken in 1.1+


plague006:

PR #3910, opened 2016-05-04
DatedQuickSaves to spacedock

plague006:


politas:
Cool. I was wondering what the deal was with two PRs adding MagiCore. Merging now.

plague006:
Two PRs making an identical change means they can be merged in any order without relying on one another.

PR #3911, opened 2016-05-04
SensibleScreenshot to spacedock

plague006:


politas:
Files in GameData?  grumble, grumble

PR #3912, opened 2016-05-04
Update VSR netkan to mark next version for KSP1.1

NathanKell:


plague006:
Please expand the existing `x_netkan_override`. Multiples overwrite one another (iirc).

NathanKell:
@plague006 how does one do that exactly? And I fear that may affect more than this, I think I added another override recently...

NathanKell:
Ah, no, I think I only did that here after all.

NathanKell:
Thanks!

plague006:
Picture worth a thousand words. =)

NathanKell:
Indeed!

PR #3921, opened 2016-05-04
Add KerbitingSystemPlanetpack

plague006:
First/only mod that uses the folder name `System`. Similarly the first/only mod that installs to `Kopernicus/Config`. With that established, this shouldn't conflict with any extant mod in CKAN.


plague006:
My mistake, forgot the tag in place. Had it up while discussing the installation with the mod author.

politas:
Cool.

PR #3923, opened 2016-05-04
Fix install for RFStockalike

plague006:


politas:
Fixed, indeed. Merging

PR #3924, opened 2016-05-04
Update install for KerbinSideGAP

plague006:


politas:
Looks about right. Merging with thanks.

PR #3936, opened 2016-05-05
Update asset_match for InfernalRobotics

plague006:
That last asset_match I wrote was true facepalm. :(


plague006:
As this is a highly sought after mod and this is a very small change, merging.

politas:
I would probably go with `^IR.*Core\\.zip` to ensure that it starts with "IR", but I enjoy complex regexps.

plague006:
"IR" could so easily change to "InfernalRobotics" is the reason I balked at it.

politas:
Ok, then `^I.*Core\\.zip`

PR #3938, opened 2016-05-05
Add conflict, bump epoch for RN-SovietRockets

plague006:
This will add a conflict to `FASA` and add an epoch. The epoch bump will appear to the CKAN client as if there's a new version of `RN-SovietRockets` so users who want to upgrade will have to remove FASA.


plague006:
Got a +1 from mod author in IRC, passes Jenkins, merging.

politas:
:+1:

PR #3937, opened 2016-05-05
Add Phoenix Industries EVA Survival Suit from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of MoviesColin, to add [Phoenix Industries EVA Survival Suit](https://spacedock.info/mod/659/Phoenix%20Industries%20EVA%20Survival%20Suit) to CKAN.

Mod details:
    name = /mod/659/Phoenix%20Industries%20EVA%20Survival%20Suit
    author = MoviesColin
    abstract = Militarized looking EVA suit loosely based on The Martian.
    license = Creative Commons Attribution-NonCommercial-NoDerivatives 4.0 International License
    Homepage = 
    description = Loosely based on The Martian, these space suit textures give a militarized, utilitarian look to your Kerbals' EVAs. 

Needs TextureReplacer. Simply choose a color variant and drop the associated files into the Default folder in Texture Replacer.

Please direct questions about this pull request to [MoviesColin](https://spacedock.info/profile/MoviesColin).

PR #3940, opened 2016-05-05
Add PhoenixIndustriesEVASurvivalSuit suite

plague006:
So.. even though I'm submitting this PR doesn't mean it has my endorsement. I'm writing it to get it out of the way but I definitely want to take time to think about if we want to add 7 unique mods for this or if it's unreasonable. Will add the virtual mod's .frozen after Jenkins runs the .netkans.

Ref https://github.com/KSP-CKAN/NetKAN/pull/3937


politas:
That's some serious 32-bit thinking.

plague006:
While I disagree that suits are "supposed to go" in a particular folder (the TextureReplacer OP specifies that the default folder is a viable option) I'm closing this from a utilitarian perspective. I think we're doing more users a disservice by adding 7 mutually-conflicting entries in the mod list than we are doing a service to users by adding these.

PR #3951, opened 2016-05-06
Create PreciseManeuver.netkan

### LINKED ISSUE #3945: PreciseManeuver plugin add
Hi there. I'm the author of the PreciseManeuver plugin (a fork of the PreciseNode plugin) and I think it's ready for CKAN. I tried to write the netkan module for it:
```
{
    "spec_version" : 1,
    "identifier"   : "PreciseManeuver",
    "$kref"        : "#/ckan/github/radistmorse/KSPPreciseManeuver",
    "name"         : "Precise Maneuver",
    "abstract"     : "A tool that gives you a much better control over your maneuvers",
    "license"      : "BSD-2 clause",
    "ksp_version"  : "1.1.2",
    "resources": {
        "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/128303-11x-precise-maneuver-editor-ver-20-pre-17-04-2016/"
    }
}
```
But I encountered several problems. First of all, netkan.exe segfaults while processing it, both on win10 and fedora linux. Second, is there a way to provide a range of supported KSP versions? Like 1.1.1-1.1.99.

plague006:
Credit and thanks to @radistmorse

Closes #3945

### DIFF
--- a/NetKAN/PreciseManeuver.netkan
+++ b/NetKAN/PreciseManeuver.netkan
@@ -1,0 +1,15 @@
+{
+    "spec_version" : 1,
+    "identifier"   : "PreciseManeuver",
+    "$kref"        : "#/ckan/github/radistmorse/KSPPreciseManeuver",
+    "name"         : "Precise Maneuver",
+    "abstract"     : "A tool that gives you a much better control over your maneuvers",
+    "license"      : "BSD-2-clause",
+    "ksp_version"  : "1.1",
+    "resources": {
+        "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/128303-11x-precise-maneuver-editor-ver-20-pre-17-04-2016/"
+    },
+    "conflicts": [
+        { "name": "PreciseNode" }
+    ]
+}


dbent:
@radistmorse Is there a reason you're tagging your releases with the short commit hash? They're unordered, effectively random, numbers which means we'll basically have to bump the epoch with every other release to make sure users get the latest version.

Tags are typically "friendly" strings, i.e. the released version number, "1.0", "2.0", etc. So you can do a `git checkout 1.0` and know you're getting the exact code as it was for that version. The commit hash will always be available anyway, I don't see a need to use it as the tag.

radistmorse:
@dbent  No, no particular reason, it's just the easiest thing to do using github's web interface: you provide a commit, and it does the tag and a release automatically. I just didn't care about tags. I can change it to something meaningful, sure. If it will make the life easier.

radistmorse:
@dbent I changed the tag for the release yesterday, but it still lists the old one as "latest". Is this something I should be concerned about?

plague006:
https://github.com/KSP-CKAN/NetKAN/pull/3956 will resolve the version ordering.

radistmorse:
Hm, the version got screwed again. I'm new to this so I just don't know: should I maybe remove all the releases except for the latest? They do all have random tags, so maybe that's the problem.

dbent:
@radistmorse  The GitHub API is returning the `68a26c7` tag as the most recent because it says the `2.0` tag release was [published at 2016-04-08T14:17:15Z](https://api.github.com/repos/radistmorse/KSPPreciseManeuver/releases), which is kinda weird. Any ideas?

If you delete the `68a26c7` tagged release it should work correctly. Let us know when you do so we can bump the `epoch` again.

radistmorse:
@dbent Ok, I removed all the tags and releases that were made in 2016 except for the latest. Let's hope that'll help.

dbent:
Great, I've bumped the epoch in  #3983 next run of the bot in (~30 minutes) should take care of it. Thanks for your help.

dbent:
I forgot merges automatically trigger netkan now, should be good to go as of this [commit](https://github.com/KSP-CKAN/CKAN-meta/commit/a8e71a271c1169ad3197ceea8b4e7e2c7f01f773).

PR #3952, opened 2016-05-06
Fix install for KerbitingSystemPlanetpack

plague006:


politas:
Did it change between when we added it and now? I could have sworn it was installing files ok

plague006:
Yep it changed.

PR #3953, opened 2016-05-06
Add GalacticNeighborhood

plague006:


politas:
Ooh, this mod is far more fragile than I realised. Looks like everything's in order. Merging with thanks

PR #3956, opened 2016-05-07
Add epoch

plague006:


politas:
That's a force_v, not an epoch. Is that what you wanted?

plague006:
Wanted an epoch, hit the wrong macro button. Thanks for the catch.

politas:
Description matches change. Merging with thanks

PR #3958, opened 2016-05-07
Update DestructionEffects.netkan

Wicloz:
This mod has been tested to work with KSP 1.1.2


plague006:
If there hasn't been a new release of the mod specifically for 1.1.2 it's best to use `"ksp_version_min": "1.1.0"` and `"ksp_version_max": "1.1.2"`. Feel free to use `"ksp_version_max": "1.1.99" if you believe the mod will be patch-immune.

Wicloz:
Thanks for telling, that wasn't in the guide ...
It should be patch immune so I changed it.

plague006:

PR #3967, opened 2016-05-08
Update Kerbalism

dbent:
Updates Kerbalism to account for the new "profiles" setup in 0.9.9.5.

`Kerbalism.netkan` has been updated to *recommend* (a profile isn't required for Kerablism to function) a virtual package `Kerbalism-Profile`. The install has been updated to filter out any profile files that ship in the distribution archive. Also explicitly supported mods have been added.

`Kerablism-Profile-Default.netkan` has been created which *provides* `Kerablism-Profile`, *conflicts* with `Kerablism-Profile`, and installs the `Default.cfg` profile file which ships in the distribution archive.

### DIFF
--- a/NetKAN/Kerbalism-Profile-Default.netkan
+++ b/NetKAN/Kerbalism-Profile-Default.netkan
@@ -1,0 +1,21 @@
+{
+    "spec_version": "v1.2",
+    "$kref": "#/ckan/spacedock/533",
+    "identifier": "Kerbalism-Profile-Default",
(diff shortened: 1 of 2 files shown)


dbent:
On a related note, `Kerablism` no longer *conflicts* with `CommunityResourcePack`, that will have to be updated by @BobPalmer. There is already a PR (BobPalmer/CKAN#5) which does that.

politas:
Nicely done! Very neat and tidy. Merging with thanks!

PR #3968, opened 2016-05-08
Metadata file for Hooligan Labs Airships

dunclaw:
Initial Metadata file for Hooligan Labs Airships


politas:
Hi there! I'm very interested in this, as I was very fond of this mod in earlier versions of KSP. I know that JewelShisen is determined not to have HLAirships on CKAN, and has loaded HLAirships up on Spacedock now with an "All Rights Reserved" license (I'm not sure what the legal position is about reverting HLAirships to ARR having clearly received significant code contributions from yourself and others, including the original development by Hooligan Labs, but that's for you other contributors to take up). 

If you plan to maintain this as an ongoing fork from the clean MIT-licensed code, could you create a new forum thread, so that people aren't getting confused and going to JewelShisen for support? The homepage thread you have listed here is a locked thread, which will end up directing users to JewelShisen's thread, and I'm assuming that the codebases will start to diverge.

There is an existing identifier, "HooliganLabsAirships", which I think could be reasonably used, given that this is the direct continuation of the original Hooligan Labs code, under its original license, and JewelShisen is certainly not wanting to use the identifier. To switch to that, rename this file to "HooliganLabsAirships.netkan", change the identifier, and delete "HooliganLabsAirships.frozen" from the NetKAN folder in your branch.

There's also a little confusion in your github source and zip file naming, as to whether this is HLAirships or HLAirshipsRedux?

dunclaw:
I'll talk to the other contributors and get back to you. I updated the code for the plugin from the original source, but I'm not sure of the history of the art assets. I'd rather just have one common project that we all contribute to instead of different forks.

dunclaw:
Talked to the other contributors and they agreed to release to CKan.
I updated the netkan to point to the Spacedock release point and the ARR license and latest forum thread.

politas:
Ok, if that's the route you're taking, you will need to change 

    "x_netkan_license_ok": true,

to 

    "license": "restricted",

And you can then remove the `"name"`, `"abstract"`, `"ksp_version"` and `"resources"` elements, as all that data can be drawn from Spacedock

politas:
Might also be good to talk to JewelShisen about release version naming consistency. On Spacedock, there is "V_5.0" followed by "5.1". Won't be an issue immediately, because netkan won't inflate that older release, but if it had, "V_5.0" would be considered the latest release, since "V" comes after "5". There's a method we have for fixing such issues, but it's a manual process every time the version prefixes change in a way that breaks the ordering.

dunclaw:
Thanks - this is my first attempt at creating a Ckan release file. I'll make the updates

dunclaw:
OK, I think I've sorted out the naming and it should be good to go.

politas:
No sorry, the identifier can't be changed from one to the other. It has to be 'HooliganLabsAirships.netkan' with an install stanza.

dunclaw:
We standardized on the identifier of "HLAirships" for the directory, plugin and identifier. Will that not work?

dunclaw:
Leave a comment

politas:
We don't abandon old identifiers. They're mostly hidden from the users, so they aren't that critical. Have a look at #4053. I think that's about right.

dunclaw:
I've removed the hard dependency on Toolbar.
That install stanza passes the local check

plague006:

PR #3962, opened 2016-05-08
Create new TACLS netkan

NathanKell:
Even if your submission tested fine locally there are several errors that only the bots will note and reject. This should be expected.
Working on 3960

PR #3972, opened 2016-05-08
TACLS - New Maintainer

### LINKED ISSUE #3960: TACLS maintainer change
KSP-RO group ( @danfarnsy in particular) has taken over maintenance of TACLS.
We have released a temporary, 1.1.2 compatible version, and will soon release a fuller version.
Going forward the kref should point to https://github.com/KSP-RO/TacLifeSupport ( in particular the new KSP1.1.2 release is https://github.com/KSP-RO/TacLifeSupport/releases/tag/v0.11.3 ) but all old releases should point to their old locations.

Can that be done?

politas:
TAC Life Support has been taken over by the KSP-RO group
Closes #3960 
Closes #3962


NathanKell:
Thank you for this. I evidently operated under the faulty assumption that the text of the .frozen netkan was still correct, and further that the 'author' field described the current maintainer, rather than the history of prior authors.

politas:
That would only be the case if this were a fork of the original, rather than a takeover. (Though it would still be better to maintain the previous author details)

NathanKell:
Ah, I see.

NathanKell:
Ah, the above comment was supposed to be @plague006 . Sorry to you both!

plague006:
No apologies needed, thanks for helping us keep the metadata up to date.

plague006:
Still has no `name`.

PR #3973, opened 2016-05-08
RN us rockets update

raidernick:
-Update to install craft files now that I include them with latest
release


politas:
Looks all fine, merging with thanks!

PR #3980, opened 2016-05-09
Move DockingPortSoundFX to Github

### LINKED ISSUE #3970: Docking Port sounds FX now point to Github release?
Have been looking at installing Docking Port sounds FX. 
I think the NetKan is pointing to spacedock. Which does not look like it has been updated. There is also no link to Spacedock from the mods forum thread. The latest release appears to be on Github here:

https://github.com/TedThompson/Docking-Sounds/releases

Should the netkan point to the github entry?

politas:
Closes #3970


plague006:
Author is `tg626`

politas:
Thanks for catching that!

PR #3982, opened 2016-05-09
Update StockBugFixModules

dbent:
[New releases](https://github.com/ClawKSP/KSP-Stock-Bug-Fix-Modules/releases).

- Removes the `asset_match` since now only a single zip file is published in the release.
- Strip leading `v` from version cause I like neatness.
- Updates `ksp_version` to 1.1.2. (Unfortunately, no `.version` file).
- Updates `resources.homepage` to new forum URL
- Removes `ModuleManager` as a dependency since it's no longer needed.
- Updates the `install` stanza to account for new directory structure.
- Removes `StockBugFixPlus.netkan` since it no longer exists separately.


dbent:
Looks like `StockBugFixPlus` became the new `StockBugFixModules`, closing this for now.

PR #3986, opened 2016-05-09
force v for ContractConfigurator-ContractPack-SCANsat

plague006:
Jenkins failure because ksp_version_max is 1.2.0 and dependencies can't be met because of that.


politas:
LGTM, presuming it was working before.

PR #3988, opened 2016-05-09
New mod: RSS DateTime Formatter

jwvanderbeck:
Add new mod, RSS DateTime Formatter.  I am the mod author.


politas:
Looks all fine, passes Jenkins. Merging with many thanks!

PR #3993, opened 2016-05-10
Update B9-PWings-Fork.netkan

DuoDex:
This contains wings, not tanks and other parts.


politas:
Sounds fair to me. Merging with thanks!

PR #3996, opened 2016-05-10
KerbalConstructionTime - depend MagiCore, recommend KRASH

magico13:
Changes required for KCT 1.3.3.7 for KSP 1.1.2


politas:
Looks all fine. Merging with thanks

PR #4001, opened 2016-05-10
Add Epoch to B9-PWings-Fork

politas:
Mod went from 2.1 to 0.40.3 when it upgraded for 1.1


plague006:
Ah ok, 2.1 was an unofficial mirror of B9 PWings while now they're issuing their own releases under 0.40.X

PR #4010, opened 2016-05-11
Make CTT 1.1-1.1.2 compatible

NathanKell:
CTT had explicit 1.1.0 metadata from spacedock; this changes current release to min/max to work in 1.1.2. @nertea you might want to update on spacedock too :)


politas:

PR #4028, opened 2016-05-11
Update PersistentThrust.netkan

bld:
Latest version now works with KSP 1.1


plague006:

PR #4031, opened 2016-05-11
Create Kerbcam-continued.netkan

DuoDex:
Thank you autofill.


plague006:
Having a `ksp_version` of `1.1.99` means your mod won't be compatible with KSP until that is the current version.
You want `ksp_version` `1.1` which is compatible with all 1.1.X releases.

DuoDex:
Yeah but the console is saying "module contains no files to install" which is a separate error.

plague006:
I wasn't speaking to the error you're getting. You're getting that error because you're not targeting a folder or file inside the zip.

plague006:
`KerbCam`

plague006:
You've listed the license as BSD 3-clause but the license written in `LICENSE.txt` is BSD 2-clause.

DuoDex:
It should be noted the license on the forums is listed as BSD2.

politas:
Can you make the case usage consistent?

    "identifier"   : "KerbCam-Continued",

(and the matching filename. GitHub sorts files according to strict ASCII, so this would not appear in the NetKAN folder until way down after ZPIF-4-Stock, which means currently it wouldn't appear at all)

    "name"   : "KerbCam",

politas:
Brilliant! Merging with thanks!

PR #4041, opened 2016-05-12
Update TACLS.netkan, add CRP dependency

danfarnsy:
We removed TACLS's organic resource definition in favor of CRP dependency.


politas:
Looks good to me. This is for the current release, I presume?

PR #4042, opened 2016-05-12
Update TACLS.netkan, add CRP dependency

danfarnsy:
Removed TACLS resource definitions in favor of adding CRP dependency


politas:
Looks good to me. This is for the current release, I presume?

danfarnsy:
Yes, it's for the current release. Had a user reporting some NRE's when CRP wasn't installed. Oops. Thanks again.

PR #4044, opened 2016-05-13
Taerobee is ok for 1.1.2.

NathanKell:


NathanKell:
Also, a problem: even before this PR, I noticed in CKAN-meta the .ckan built from this shows version 1.0.5 not 1.1.0. What's up with that?

plague006:
You're looking at the wrong file. If you notice `"x_netkan_epoch": "1",` in the netkan that means the newest version of the mod in CKAN has a `1:X.X.X` version (or `1-X.X.X` in the .ckan filename). So you were looking at [Taerobee-2.3.ckan](https://github.com/KSP-CKAN/CKAN-meta/blob/master/Taerobee/Taerobee-2.3.ckan) when [Taerobee-1-2.3.ckan](https://github.com/KSP-CKAN/CKAN-meta/blob/master/Taerobee/Taerobee-1-2.3.ckan) is the current version.

NathanKell:
Ah, I see. Thanks!

PR #3561, opened 2016-04-08
Add Master Tech Aerospace from SpaceDock

politas:
This is a semi-automated commit on behalf of DFGRhys

PR #4054, opened 2016-05-14
Split off QuizTechAeroPack-internals

plague006:
I've no idea if `InternalsMod` will be a thing, but they're definitely their own class and nearly certain to not be broken by any change to modding. And if they do break we have a comment to search for.

Unblocks https://github.com/KSP-CKAN/NetKAN/pull/3561


politas:
Looks good to me. We should document those "type" comments somewhere.

PR #4055, opened 2016-05-14
InterstellarFuelSwitch-Core fix install

plague006:
CC @sswelm


politas:
LGTM. Merging with thanks!

sswelm:
damn, this is actually a zip mistake

PR #4058, opened 2016-05-14
Airports database for Real Solar System / KSC Switcher

miki-g:
some real airports locations as new ksc

... my first "mod", maybe all ok? :)


NetKAN/RSS-Airports.netkan
```diff
@@ -0,0 +1,30 @@
+{
+    "spec_version" : v1.4,
```
plague006: `v1.4` is a string and needs quotes (`"v1.4"`) around it.

PR #4070, opened 2016-05-15
Add VensStockPartRevamp-CTTPatch

plague006:
Before this is merged...

@NathanKell would you like to get in touch with the author about integrating this into VSPR itself?

### DIFF
--- a/NetKAN/VenStockRevamp.netkan
+++ b/NetKAN/VenStockRevamp.netkan
@@ -9,6 +9,9 @@
     "depends"      : [
         { "name" : "ModuleManager" }
     ],
+    "suggests"      : [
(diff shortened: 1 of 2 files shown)


NathanKell:
Alas I don't have any special means of contact, I just post on the repo like anyone else. :]

plague006:
Oh.. I pinged you because you're listed as the author in our metadata.. but that might be an error in the way we scrape github for authors, it seems to only grab the user who issued the release.

NathanKell:
I do have collab rights on the repo, so I can push MM fixes and in extremis make releases.

politas:
LGTM, Merging now

PR #4074, opened 2016-05-15
KerbinSideGAP mod dependencies updating

Keniamin:
I've updated dependencies to fit new packages structure that AlphaAsh had introduced.

Am I right, that this affects only further releases of my mod, while current ones (v1.0-v1.2) remains as they are and so they will be still available for KSP 1.0.5 users?

### DIFF
--- a/NetKAN/KerbinSideGAP.netkan
+++ b/NetKAN/KerbinSideGAP.netkan
@@ -7,8 +7,9 @@
     "license": "CC-BY-NC-4.0",
 
     "depends": [
+        {"name": "KerbinSide-virt"},
         {"name": "ContractConfigurator", "min_version": "1.9.6"},
-        {"name": "KerbinSide"}
+        {"name": "ContractConfigurator-KerbinSideJobs", "min_version": "2.0"}
     ],
     "recommends": [
         {"name": "NavUtilities"},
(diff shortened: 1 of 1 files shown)


politas:
No, changing the metadata will affect the current release and future releases.

Keniamin:
This mean I can not have releases with different dependencies? Or "current and further" means just v1.2 and later, while the previous ones (v1.0 and v1.1) will not be affected?

politas:
Yes, "current and further" means just v1.2 and later, while the previous ones (v1.0 and v1.1) will not be affected.

Keniamin:
Ok, got it. I've released "dummy" v1.3 to let v1.2 remain unchanged for 1.0.5 users. Thank you for explanation. Now my pull request can be merged.

plague006:
Instead of a dummy release which can be confusing to users, you could simply keep the .netkan as is but add an override

````
    "x_netkan_override" : [
        {
            "version" : "<=1.2",
            "override" : {
                "depends": [
                    {"name": "ContractConfigurator", "min_version": "1.9.6"},
                    {"name": "KerbinSide"}
                ]
            }
        }
    ]
````

Keniamin:
Ok, thanks, I'll keep that in mind if I'll have to make another change. But now I'm anyway planning to release new feature as soon as this request be merged. That will be new, v1.4 release, so dummy v1.3 is absolutely temporary.

PR #4077, opened 2016-05-16
Split YongeTechTreesPlugin from ETT

politas:
Will probably get a separate Spacedock entry soon.

### DIFF
--- a/NetKAN/ETT.netkan
+++ b/NetKAN/ETT.netkan
@@ -11,7 +11,8 @@
         }
     ],
     "depends": [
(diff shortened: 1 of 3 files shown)


plague006:
If you want custom version numbers per release your best bet at the moment is to take the output .ckan from `YongeTechTreesPlugin.netkan` and PR that into CKAN-meta. And rename `YongeTechTreesPlugin.netkan` to `YongeTechTreesPlugin.frozen` to hold the identifier.

plague006:

PR #4078, opened 2016-05-16
Remove executable tags from .netkan files

politas:


plague006:
Keep it classy Linux-istas, keep it classy. :D

PR #4079, opened 2016-05-16
Tabs to spaces and trim trailing spaces

plague006:


politas:
Tab police on the job!

PR #4081, opened 2016-05-17
RealisticRemodel1x fix install

plague006:
The dashes are gone for now but they may be back, and in greater numbers.

### DIFF
--- a/NetKAN/RealisticRemodel1x.netkan
+++ b/NetKAN/RealisticRemodel1x.netkan
@@ -13,7 +13,7 @@
     ],
     "install": [
         {
-            "find_regexp": "RealisticRemodel-1x",
+            "find_regexp": "RealisticRemodel.?1x",
             "install_to": "GameData"
         }
     ]


politas:
If they may be back in greater numbers, then a .* would be called for!

plague006:
Well if you're playing joke killer I can play that game, I never said the dashes would be consecutive. Right --p-o-l-i-t-a-s--? :D

politas:
Interestingly, did you know that full stops (aka periods) in gmail addresses are ignored? So First.Last@gmail.com is the same as FirstLast@gmail.com, or indeed F.i.r.s.t.L.a.s.t@gmail.com

plague006:
I did know that, earlier this year either I gave out the wrong address or someone miswrote my address but I got an email at one point to firstnamelastname@gmail.com when what I signed up for was firstname.lastname@gmail.com. Needless to say it was a "wtf" moment before a google search put me at ease.

PR #4082, opened 2016-05-17
Remove build and test files

plague006:
FACT: Every test we perform on metadata slows down our ability to process requests.
FACT: Every test we perform puts strain on Jenkins.
FACT: These test files are useless.

Thank you CKAN contributor, but our test files are in another repository.
https://github.com/KSP-CKAN/NetKAN/pull/3873#issuecomment-216126076


techman83:
LGTM. Merging with Thanks.

PR #4084, opened 2016-05-17
ShowFPS add v to version

plague006:


techman83:
LGTM, merging with thanks!

PR #4085, opened 2016-05-17
Add Space Launch System Stock Dimension from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of Sobol, to add [Space Launch System Stock Dimension](https://spacedock.info/mod/706/Space%20Launch%20System%20Stock%20Dimension) to CKAN.

Mod details:
    name = /mod/706/Space%20Launch%20System%20Stock%20Dimension
    author = Sobol
    abstract = Stock Dimension version mods Space Launch System
    license = GPLv3
    Homepage = http://forum.kerbalspaceprogram.com/index.php?/topic/126067-space-launch-system-part-pack/
    description = This is your mod listing! You can edit it as much as you like before you make it public.

To edit **this** text, you can click on the "**Edit this Mod**" button up there.

By the way, you have a lot of flexibility here. You can embed YouTube videos or screenshots. Be creative.

You can check out the SpaceDock [markdown documentation](/markdown) for tips.

Thanks for hosting your mod on SpaceDock!

Please direct questions about this pull request to [Sobol](https://spacedock.info/profile/Sobol).

PR #4086, opened 2016-05-18
Add SpaceLaunchSystemStockDimension from Spacedock

politas:
Replaces #4085

### DIFF
--- a/NetKAN/SpaceLaunchSystemStockDimension.netkan
+++ b/NetKAN/SpaceLaunchSystemStockDimension.netkan
@@ -13,7 +13,7 @@
         { "name": "LaunchPad39B" },
         { "name": "KerbalKonstructs" }
     ],
-    "conflicts" [
+    "conflicts" : [
         { "name": "SpaceLaunchSystemPartPack" }
     ],
     "install": [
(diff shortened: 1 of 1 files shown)


politas:
SpaceLaunchSystemPartPack also needs some attention, if my reading of the forum post is correct. I've approached the author.

techman83:
LGTM, merging with thanks!

plague006:
If I'm understanding properly, this is the version to install for stock KSP while `SpaceLaunchSystemPartPack` will be for RO? "Stock dimension" = stock version of the game?

I think we'll want to install the VAB folder, we'll also want to conflict `SpaceLaunchSystemPartPack`, and if I'm correct that this is for the stock version maybe change the ID to `SpaceLaunchSystemPartPack-stock` or something similar.

politas:
I was thinking of maybe overriding the name on the PartPack version, but I've got to work out what the deal is, first. Like I said, I've approached the author. It sounds like the PartPack is for either RealismOverhaul (which may actually mean RSS) _or_ Stock. I have a suspicion that it may be broken as we currently have it.

politas:
To get the RealismOverhaul version, apparently you need to install a bunch of files from a Dropbox folder. I don't think we have any way to set that up.

politas:
Without fully understanding what the author means by "Stock Dimension", I'm leaving it as is. Feel free to hold off on merging while I keep trying to get info from Sobol.

plague006:
If/when you get confirmation that stock dimension = stock aero model and stock Kerbin then this is ready to merge in my book.

politas:
OK, I'm going to merge this, because I'm fairly certain that this one will work fine, and it's PartPack we need to fix up.

PR #4089, opened 2016-05-18
Update CryoTanks

dbent:
- Unfreeze
- Repoint to SpaceDock
- Add comment about how CryoTanks is sourced from KerbalAtomics.
- Add `$vref` which points directly to the KerbalAtomics version file.
- Switch from `find` to `file`, some reason `find` wasn't working, think the archive might be borked in some way, will have to look into that.


techman83:
LGTM. Merging with thanks!

Issue #4088, opened 2016-05-18
ProceduralProbes has moved to CurseForge


HiddenKn:
https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/ProceduralProbes.frozen
used to be hosted at KerbalStuff but seems to have moved to CurseForge at
http://kerbal.curseforge.com/ksp-mods/233798-proceduralprobes

See http://forum.kerbalspaceprogram.com/index.php?/topic/118112-104-proceduralprobes/ for more info

HiddenKn:
See http://forum.kerbalspaceprogram.com/index.php?/topic/118112-104-proceduralprobes/ for more info

PR #4091, opened 2016-05-18
Unfreeze ProceduralProbes, move download

plague006:
Closes https://github.com/KSP-CKAN/NetKAN/issues/4088


techman83:
LGTM, merging with thanks!

PR #4095, opened 2016-05-18
Update PilotAssistant.netkan

dbent:
Fix override, no leading `v` (thankfully).


techman83:
LGTM, merging with thanks!

PR #4096, opened 2016-05-18
Update DeployableEngines

dbent:
- Unfreeze.
- Repoint to SpaceDock.
- Add comment about how DeployableEngines is sourced from KerbalAtomics.
- Add `$vref` which points directly to the KerbalAtomics version file.


techman83:
LGTM, merging with thanks.

PR #4097, opened 2016-05-18
Update KerbalAtomics

dbent:
- Move to SpaceDock
- Use `.version` file
- Remove `force_v` (was never necessary) and bump the `epoch`
- Integrate override dependency changes
- Requires dependency in #4096


dbent:
#rebuild

techman83:
LGTM, merging with thanks!

PR #4098, opened 2016-05-18
Update CryoEngines

dbent:
- Unfreeze.
- Use KSP-AVC all around.
- Unify epoch of all the main mod and patches.
- Add dependency to `B9PartSwitch` remove dependency on `InterstellarFuelSwitch-Core`
- Depends on #4096


techman83:
#rebuild

techman83:
Well Rebuild actually works. Cool. Passes now that dependencies are met. LGTM, merging with thanks!

PR #4099, opened 2016-05-18
Fixed for SpaceDock

Magion:


dbent:
Hi @Magion, you'll also want to change the filename from `WernherChecker.frozen` to `WernherChecker.netkan` so the index will actually read it.

Magion:
Why it was renamed to .frozen before?

dbent:
Because when KerbalStuff fell off the face of the earth we didn't want to spend time making requests we knew were going to fail.

LGTM, merging, thanks.

Magion:
Aaah, sure! That makes sense.

PR #4100, opened 2016-05-18
Update NearFutureSpacecraft

dbent:
- Unfreeze.
- Redirect to SpaceDock.
- Use KSP-AVC.


plague006:
I forget on a monthly basis that we can target specific .version files inside $vref :(

PR #4102, opened 2016-05-18
Add kOSPropMonitor

dbent:
Closes #4094


plague006:

PR #4104, opened 2016-05-18
DIRECTLaunchVehicles fix install

plague006:


Olympic1:
LGTM, merging with my thanks

PR #4109, opened 2016-05-20
Change version to min/max

NathanKell:


techman83:
LGTM, merging with thanks!

PR #4112, opened 2016-05-20
Freeze FarFromKerbin

plague006:
Mod author is taking the summer off from development, and decided to upload an empty .zip to spacedock which is stuck in the indexer.


politas:
A weird choice.

PR #4113, opened 2016-05-20
KSP Version Update for SigmaBinary-DunaIke

politas:
Needs to be kept in sync with SigmaBinary-Core


plague006:
As this is tied directly to SigmaBinary-core's ksp compatibility, I'd happily merge a `ksp_version:any`

politas:
There's no "API level" testing between the two though, as they're released as a single entity. We have no way of knowing whether a newer version of -DunaIke will work with an older version of -core.

PR #4115, opened 2016-05-21
Add CommunityTerrainTexturePack

dbent:
New dependency for OuterPlanetsMod 2.0.

PR #4116, opened 2016-05-21
Update OuterPlanetsMod

dbent:
- Use KSP-AVC
- Add dependency on new `CommunityTerrainTexturePack` in #4115 
- Update minimum verison of Kopernicus


plague006:
Passes netkan, the $vref having broken JSON would be my only concern on this PR and it passed that. Having Jenkins rebuild just because I like check marks.

plague006:
:white_check_mark:

PR #4124, opened 2016-05-22
CustomAsteroids compatibility fixes

Starstrider42:
Pushing some queued corrections to the CustomAsteroids compatibility data. This should be the last PR I make for a while, don't worry.


plague006:

PR #4132, opened 2016-05-23
Update KAX

dbent:
Release only on Curse.


plague006:
For our mirroring, I think we should set this to a `restricted` license.

>Specifically, you may not:
Distribute unmodified versions of KAX, in part or whole, without permission
Use KAX for any commercial purpose

dbent:
Bleh, when did that change. Updated, although due to the GitHub [issues](https://status.github.com/messages) may take a while for the change to appear.

dbent:
Alright, GitHub has finally caught up.

plague006:

techman83:
> For our mirroring, I think we should set this to a restricted license.
> 
>> Specifically, you may not:
Distribute unmodified versions of KAX, in part or whole, without permission
Use KAX for any commercial purpose

Good catch @plague006 - definitely something we need to vigilant of.

PR #4138, opened 2016-05-24
NavHud KSP Version fix

politas:
Match CKAN to forum thread


politas:
I have personally tested in 1.1.2, and it's working fine there. Given the simplicity of this mod, the Unity upgrade would have been important, but most KSP updates aren't going to affect it.

PR #4141, opened 2016-05-25
Fix install for KWRocketry-CommunityFixes and -interstage

plague006:
Change in casing for `GameData`, switch from `file` paths to `find`.

@linuxgurugamer I was under the impression `KWRocketry-CommunityFixes` was discontinued in favour of `KWRocketryRedux` but seeing as you're continuing to issue new releases, should we bump the ksp_version compatibility of `KWRocketry-2.7.0-community`?

### DIFF
--- a/NetKAN/KWRocketry-CommunityFixes-interstage.netkan
+++ b/NetKAN/KWRocketry-CommunityFixes-interstage.netkan
@@ -1,5 +1,5 @@
 {
-    "spec_version"  : "v1.2",
+    "spec_version"  : "v1.16",
     "identifier"    : "KWRocketry-CommunityFixes-interstage",
     "$kref"         : "#/ckan/spacedock/51",
     "$vref"         : "#/ckan/ksp-avc",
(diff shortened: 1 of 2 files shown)


linuxgurugamer:
No, please don't do that.

I updated the old community fixes for those people still running 1.0.5; 
if anything, please change it's compatiblity from 1.1 down to 1.0.5


JBB

plague006:
`KWRocketry-2.7.0-community` will stay locked to `1.0.5`.

The KSP forum thread, the sapcedock page, the included .version file and the remote version file are all in agreement that this is for 1.1. I'll happily merge any PR for an override if you choose to write one.

linuxgurugamer:
OK, so I've done is this:

1.  Update the .version file to reflect only up to 1.0.5, new version 
for the community fixes is 0.4.14.1

2.  Update the thread (I'll ask Sal_vager to unlock it)

3.  Update Spacedock

4.  Update the KSP-AVC entry

Will I need to write a PR in that case, and if so, what needs to be written?


Thanks



JBB aka Linuxgurugamer

politas:
#rebuild

politas:
So the two previous versions that are currently marked as 1.1 compatible should be cut back to just 1.0.5 as well?

politas:
This is looking fine, so regardless of the action required for the previous releases, this can be merged.

linuxgurugamer:
Yes

PR #4142, opened 2016-05-25
Update KIS & SurfaceLights

ihsoft:
Both live on Guthub but release via CurseForge so, taking full control on it.


ihsoft:
It looks tag `license` is required when using `#/ckan/netkan` even though the remote `.netkan` file has it. Wiki should be updated with the list of the required fields.

plague006:

And re: updating our wiki, which page mentions metanetkans?

ihsoft:
https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#ckannetkanurl

plague006:
Ah, my confusion was because you said wiki page and that's a file. I'll poke the person who'd know best about metanetkans and see needs changing.

PR #4144, opened 2016-05-25
Fix install for Races

plague006:
Every time I see the identifier I think it's a mod that adds ethnic Kerbals...


politas:
"Oh, let me submit this to CKAN and then change the folder structure on my very next release!" 
Grumble, grumble, grumble

PR #4146, opened 2016-05-26
Add Advanced SAS For All from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of rodmcnew, to add [Advanced SAS For All](https://spacedock.info/mod/725/Advanced%20SAS%20For%20All) to CKAN.

Mod details:
    name = /mod/725/Advanced%20SAS%20For%20All
    author = rodmcnew
    abstract = Enables all advanced SAS functionality on all command pods and probe bodies in career mode.
    license = GPLv3
    Homepage = https://github.com/rodmcnew/KSP-Mod-AdvancedSASForAll
    description = This is your mod listing! You can edit it as much as you like before you make it public.

To edit **this** text, you can click on the "**Edit this Mod**" button up there.

By the way, you have a lot of flexibility here. You can embed YouTube videos or screenshots. Be creative.

You can check out the SpaceDock [markdown documentation](/markdown) for tips.

Thanks for hosting your mod on SpaceDock!

Please direct questions about this pull request to [rodmcnew](https://spacedock.info/profile/rodmcnew).

PR #4147, opened 2016-05-26
Add Advanced SAS For All from Spacedock

politas:
Replaces #4146


techman83:
LGTM. Merging with thanks!

PR #4155, opened 2016-05-28
Update SAVE.netkan

dbent:
Manually specify `resources.homepage` since it's not specified on SpaceDock.


politas:
Good catch, thanks. Merging

PR #4156, opened 2016-05-28
Add SMURFF-Extras-VenStockRevamp

dbent:
SMURFF now comes with optional patches in its `Extras` directory. This adds a package for the only one so far: `SMURFF-Extras-VenStockRevamp`. For future extensibility I figure its best to group the patches by mod/general functionality. If that does not prove to be granular enough we can split them in the future.


politas:
I'm left wondering why this patch can't just be added to the main mod with a "needs" modifier, but that's not our responsibility.

PR #4157, opened 2016-05-28
Update AdvancedJetEngine.netkan

dbent:
AJE has new maintainers and a new forum thread. Repository remains camlost2/AJE for now, however.

I also changed the license from `LGPL-2.1` to `LGPL-2.0`, as far as I can tell from the [README history](https://github.com/camlost2/AJE/commits/solver_redo/README.md) it's always been LGPL-2.0. Will have to go back and update old meta.

The haphazard order of the properties was really annoying me so I also just reformatted it.


plague006:

PR #4161, opened 2016-05-29
KerbinSideGAP mod dependencies updating

Keniamin:
Updated required version of Contract Configurator since I want to use new "HasCrew.excludeKerbal" feature.


politas:

PR #4160, opened 2016-05-29
Add Contract Pack: Sounding Rockets from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of inigma, to add [Contract Pack: Sounding Rockets](https://spacedock.info/mod/728/Contract%20Pack:%20Sounding%20Rockets) to CKAN.

Mod details:
    name = /mod/728/Contract%20Pack:%20Sounding%20Rockets
    author = inigma
    abstract = Contracts for USI Sounding Rockets
    license = MIT
    Homepage = 
    description = Contract Pack: Sounding Rockets 1.0

A simple contract pack developed by @inigma for USI Sounding Rockets by @RoverDude

SpaceDock Download

Description:

A simple set of 6 altitude contracts requiring the use of USI Sounding Rockets. Perfect for any fledgling beginner space program.

Please direct questions about this pull request to [inigma](https://spacedock.info/profile/inigma).

PR #4163, opened 2016-05-29
Add ContractConfigurator-SoundingRockets from Spacedock

politas:
Replaces #4160


politas:
.version file is fixed. Good to merge, now.

PR #4166, opened 2016-05-30
Change install path to match new download structure

Raptor831:


plague006:
Installing `GameData` into `GameData` will result in `KSP/GameData/GameData/Modfolder` which is not likely your intent.

Raptor831:
Thanks for the info, changed it to just place the 3 folders explicitly.

plague006:
Change both instances of `file` to `find` and bump the `spec_version` to `v1.4` and that'll have the effect you're looking for.

Raptor831:
That look about right?

plague006:

PR #4169, opened 2016-05-30
Fix install StarWarsBDAddon

plague006:
Ref https://github.com/KSP-CKAN/CKAN/issues/1477


politas:
It is truly amazing how many extensions to BDArmory are only compatible with a KSP version that BDArmory is not compatible with. This looks fine, given the dependency.

PR #4164, opened 2016-05-30
Add TWSA's Payload Adapters! from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of tristonwilson12, to add [TWSA's Payload Adapters!](https://spacedock.info/mod/729/TWSA%27s%20Payload%20Adapters%21) to CKAN.

Mod details:
    name = /mod/729/TWSA%27s%20Payload%20Adapters%21
    author = tristonwilson12
    abstract = This is a small mod that includes 1 part. its a payload adapter thats compatible with tweakscale! Payload adapters ar what go between the Payload, and the Farings on a rocket!
    license = CC BY-NC 4.0
    Homepage = 
    description = This is a small mod that includes 1 part. its a payload adapter thats compatible with tweakscale! Payload adapters ar what go betweek the Payload, and the Farings on a rocket!

Heres some pics! (the one all the way to the left is the standered size, 2.5m, the other 2 or Tweakscaled)

Please direct questions about this pull request to [tristonwilson12](https://spacedock.info/profile/tristonwilson12).

PR #4171, opened 2016-05-31
Add TWSA-Payload-Adapters from Spacedock

politas:
Replaces #4164


plague006:
I see "highly recommended" and not "required" in the forum thread OP, but I'm guessing you have a reason for your interpretation. ¯\_(ツ)_/¯

politas:
There's only the one part. The only way to get different sizes is with Tweakscale.

PR #4107, opened 2016-05-19
Add MalemuteRover

Olympic1:
Requested by @ArnePeirs for https://github.com/BobPalmer/CKAN/pull/6

PR #4173, opened 2016-06-01
NetKAN File for USI Malemute Rover

Cphoenix:
Someone be sure to give this a good look-over. First time submitting metadata to CKAN.

Edit: I'm aware of this PR: https://github.com/KSP-CKAN/NetKAN/pull/4107

However that PR's .netkan fails to be converted properly by netkan.exe to a .ckan file. This one works.


politas:
Take a look at [the proposed target of that metanetkan](https://github.com/BobPalmer/CKAN/pull/6/files) - it's a better match with RoverDude's other mods and is more in keeping with the way he has set up his self-hosted metadata. I'd be happy to merge this if it looked more like that .netkan.

Cphoenix:
@politas 

Okay, I was confused for a bit there. I gather that #4107 is pointing the Netkan bot to where the **actual** metadata file will be once RoverDude merges that to his CKAN repo. Whoopsie; didn't realize RoverDude was managing his metadata files (I was sleepy at the time, I blame that :stuck_out_tongue: ).

Two options: I can pretty much copy that target netkan in that PR you linked, or close this and wait for RoverDude to accept that PR. I suggest the latter so RoverDude is in control of the metadata file that gets used.

Either way I'm going to go add a few lines to the [wiki guide](https://github.com/KSP-CKAN/CKAN/wiki/Adding-a-mod-to-the-CKAN#before-you-start) to mention this possibility in the "before you start" bit. The whole "check that the mod author doesn't make their own CKAN files and you can just stick a link to it" bit would've prevented this mistake. As it is that bit is buried in the metadata spec.

politas:
I'd actually suggest the former. We can always switch to the self-hosted netkan when RoverDude merges it. I was about a day away from doing that, already.

Cphoenix:
Very well. I've done a straight copy-paste of the other netkan file to this one.

politas:
The Filename needs to be renamed to `MalemuteRover.netkan` - No "USI-" and no version number.

Cphoenix:
Fixied. >.>

politas:
Cool. Merging with thanks

PR #4174, opened 2016-06-01
Fix bad forum thread link for ContractConfigurator-KerbalAcademy

severedsolo:


politas:
Closing in favour of #4175, which rebased this change

PR #4180, opened 2016-06-02
Make Rocketdyne-F1 Netkan

### LINKED ISSUE #4179: Rocketdyne F1 works in 1.1.x
Or at least it does on my machine when I played with it. I *think* its metadata just needs to be changed so its max KSP version is 1.1.99.

politas:
Need to update KSP version compatibility
Closes #4179


techman83:
LGTM, merging with thanks!

PR #4181, opened 2016-06-02
Add DarkSideTechnology

Cphoenix:
Netkan.exe actually fails to transform this properly but after reading issue 1469 (https://github.com/KSP-CKAN/CKAN/issues/1469) and Program.cs in the CKAN repo I gather that this will actually work in reality. (Or at least I assume so, and that netkan is just being persnickety.)


politas:
We use strict JSON, so you'll need to remove the comma on the "license" line.

politas:
I would also add a suggests list for CommunityTechTree, USI-LS and Snacks

politas:
Looks all fine, now. Merging with thanks!

PR #4194, opened 2016-06-04
Add Kerbal Research and Development

Cphoenix:
@mmoench

**With some caveats.**

This is based on the mod author's own [netkan](https://github.com/mmoench/KRnD/blob/1.8/KRnD.netkan).

The netkan.exe application fails to validate this with two errors:

```1273 [1] WARN CKAN.NetKAN.Transformers.GithubTransformer (null) - No releases found for mmoench/KRnD
1427 [1] FATAL CKAN.NetKAN.Program (null) - JSON deserialization error
```

However the file itself looks valid to me, after adding install directives, changing the KSP version to 1.1.2 (the latest release is compatible with it), and a laundry list of conflicts.

The conflict list is me erring well on the side of caution, as mm mentioned that this mod is likely to cause weird behaviour when combined with or otherwise be incompatible with other mods that edit part stats at runtime. I've gotten the impression that CKAN is meant to be very strict about stuff like that (like any package manager really ought to be).


Cphoenix:
Whoa there. What have we here? I think I must've derped hard with the command line. Excuse me a sec.

PR #4195, opened 2016-06-04
Add Kerbal Research & Development

Cphoenix:
@mmoench

**With some caveats.** (Sorry about the borked PR earlier. I done goofed.)

This is based on the mod author's own [netkan](https://github.com/mmoench/KRnD/blob/1.8/KRnD.netkan).

The netkan.exe application fails to validate this with two errors:

```
1273 [1] WARN CKAN.NetKAN.Transformers.GithubTransformer (null) - No releases found for mmoench/KRnD
1427 [1] FATAL CKAN.NetKAN.Program (null) - JSON deserialization error
```

However the file itself looks valid to me, after adding install directives, changing the KSP version to 1.1.2 (the latest release is compatible with it), and a laundry list of conflicts.

The conflict list is me erring well on the side of caution, as mm mentioned that this mod is likely to cause weird behaviour when combined with or otherwise be incompatible with other mods that edit part stats at runtime. I've gotten the impression that CKAN is meant to be very strict about stuff like that (like any package manager really ought to be).


politas:
There are no release packages on GitHub. If you can figure out an asset_match regexp that will use the Source Code (zip), you're doing better than I.

dbent:
>  If you can figure out an asset_match regexp that will use the Source Code (zip), you're doing better than I.

There's [`use_source_archive`](https://github.com/KSP-CKAN/CKAN/pull/1587) which I apparently never documented.

politas:
I am contacting the mod author in an attempt to sort out scraping for this mod.

politas:
Ah, that does the trick nicely. We can use that to get the mod author's original netkan working. I'm unsure about the laundry list of conflicts. Would any other @KSP-CKAN/wranglers like to have input as to whether we need them?

dbent:
They may not be strict `conflicts` (i.e. things break when these two mods are used together) and more a warning (KSP-CKAN/CKAN#211) (things may not work entirely correctly when these two mods are used together).

plague006:
As the @mention hasn't summoned the author I'd suggest PMing them on the KSP forums with the list of conflicting mods, describing what a conflicts relationship means, and asking if they think the relationships should be included or not.

Cphoenix:
> They may not be strict conflicts (i.e. things break when these two mods are used together) and more a warning (KSP-CKAN/CKAN#211) (things may not work entirely correctly when these two mods are used together).

Indeed in this case I think a warning, not an outright block of installation, would be more appropriate given what the author has said in the mod's forum thread. I found myself wishing there was one while adding those conflicts.

> As the @mention hasn't summoned the author I'd suggest PMing them on the KSP forums with the list of conflicting mods, describing what a conflicts relationship means, and asking if they think the relationships should be included or not.

I'll leave that to @politas since he has already contacted them.

plague006:
Needs spec_version `"v1.4"`

Cphoenix:
Related [issue on KRnD repo](https://github.com/mmoench/KRnD/issues/1).

PR #4198, opened 2016-06-04
Add EasyVesselSwitch v1.0.0

ihsoft:


ihsoft:
Please, let me know if any info is required but missing.

plague006:
Looks good to me, when CKAN 1.18 is released I'll change the license to `Unlicense` for specificity, merging with my thanks.

ihsoft:
> I'll change the license to Unlicense for specificity

Should I change it in my repository as well or it will work without the change?

plague006:
It'll work without any change on your end.

PR #4200, opened 2016-06-04
Add KerbalResearchAndDeveleopment 1.3 for KSP1.0.5

politas:
Based off #4195, this will get the old release supporting KSP 1.0.5 into CKAN

### DIFF
--- a/NetKAN/KRnD.netkan
+++ b/NetKAN/KRnD.netkan
@@ -11,6 +11,17 @@
     "depends" : [
         { "name" : "ModuleManager" }
     ],
+    "conflicts" : [
+        { "name" : "B9AerospaceProceduralParts"},
+        { "name" : "InterstellarFuelSwitch"},
+        { "name" : "ProceduralAirships"},
+        { "name" : "ProceduralDynamics"},
+        { "name" : "ProceduralFairings"},
+        { "name" : "ProceduralParts"},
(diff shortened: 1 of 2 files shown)


politas:
I have a change ready for another PR after this is merged that will have the current and future releases working.

Cphoenix:
I know the identifier is supposed to be short, but isn't KRnD a little _too_ short? I know it's not something I would search for if looking to add a new netkan for it, as in the forum thread it's mostly referred to by it's full name. And most netkans it seems have the full or almost full names for most mods.

politas:
KRnD matches the folder name the mod will install. Searching for existing netkans by name is a fundamentally flawed method. Much better to fire up a CKAN client, set the filter to "All" and use the filter by name box to search for unusual name elements.

politas:
It's also the identifier the mod author picked, and the github repository name.

Cphoenix:
Eh, fair enough. It's probably just me having that initial "Ewwwww, why abbreviate when you can be verbose but specific" reaction.

plague006:
`homepage` link is dead

politas:
Done. Have you seen the latest episode of Silicon Valley? I'm interested to know why the hating on tabs.

politas:
Merging, then I will create a new PR to switch to the self-hosted netkan

PR #4204, opened 2016-06-05
Update DecoupleFromHeatshield.netkan

jtac:
I've updated DecoupleFromHeatshield to work with newest version of KSP


jtac:
All checks passed

politas:
This change is unnecessary. version 1.1.2 has already been picked up by the Netkan robot. We only need to revisit the metadata if the file structure or relationships change.

PR #4207, opened 2016-06-05
Fix homepages

politas:
Seems the latest forum update has changed the URL requirements for threads.


plague006:
Why not just put the names of the mods rather than a random character? To make the links descriptive in addition to functional?

politas:
My desire for minimisation, I suppose. The smallest URL to get the required page. Adding a bunch of text that may not reflect the thread's title in the future seems wrong to me.

dbent:
LGTM

PR #4212, opened 2016-06-07
Add DraggableMenu

Cphoenix:
It's for 1.0.5 but might as well in case it gets updated.

### DIFF
--- a/NetKAN/DraggableMenu.netkan
+++ b/NetKAN/DraggableMenu.netkan
@@ -1,0 +1,6 @@
+{
+    "spec_version" : 1,
+    "identifier"   : "DraggableMenu",
+    "$kref"        : "#/ckan/spacedock/311",
+    "x_netkan_license_ok": true
+}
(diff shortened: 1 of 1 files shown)


Cphoenix:
Actually I'm being silly and this became stock in 1.1 but still, can't hurt for anyone using 1.0.5.

politas:
Needs an install stanza in order to filter Sources:

    "install": [
        {
            "file": "DraggableMenu",
            "install_to": "GameData",
            "filter": "Sources"
        }
    ]

Cphoenix:
Fixed

politas:
My mistake, that needs to be `"file": "GameData/DraggableMenu",`

politas:
Oh, and this also needs this inserted:

    "conflicts": [
        { "name": "MenuStabilizer" }
    ],

Cphoenix:
Added/fixed. And no worries; I'm in class right now. I'm just blindly copy-pasta'ing. Should've caught the GameData bit myself.

politas:
I think that does it. We probably won't see any further versions of this, but I'm sure there are people still using 1.0.5 that may find it useful.

PR #4183, opened 2016-06-02
Add Soviet Engine Pack from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of Dragon01, to add [Soviet Engine Pack](https://spacedock.info/mod/739/Soviet%20Engine%20Pack) to CKAN.

Mod details:
    name = /mod/739/Soviet%20Engine%20Pack
    author = Dragon01
    abstract = A high quality engine pack first made by BobCat, then updated by DECQ and Dragon01.
    license = MIT
    Homepage = 
    description = The famous Soviet Engine Pack updated for 1.1.2, by the combined effort of Dragon01 and DECQ. Some features (including balance and career) still WIP.

Please direct questions about this pull request to [Dragon01](https://spacedock.info/profile/Dragon01).

PR #4218, opened 2016-06-08
Add SovietEnginePack

politas:
Given the confusion in the forum threads, the new maintainers' decision to use a different name, and the restarted versioning, I decided make this a fork of SovietEngines rather than a continuation.
Replaces #4183


dbent:
Should this `conflicts` with `SovietEngines` if/when that is updated?

politas:
Good call

dbent:
LGTM

PR #4220, opened 2016-06-08
Opt spaceplane parts

Cphoenix:
Added new $kref to SpaceDock, x_netkan_license_ok set to false (because
it's **not** in the enum in the spec), and removed unnecessary whitespace
from the depends list.

The mod is being maintained, but I really wish there was a way to add the **HUGE** caveat that comes with the recent update, and that is that the next update, whenever that happens, will break everyone's ships.


Cphoenix:
I am going to murderize the GitHub Desktop app. So bloody unintuitive.

politas:
IT seems the only way to cleanly sync a fork with the upstream is using git: https://help.github.com/articles/syncing-a-fork/

Cphoenix:
Good lord. Thanks for that. I knew there was a way to fix this silliness but I'd totally forgotten how. Everything I found when searching suggested rebasing as a solution. 😒

politas:
The only way I can think to do it via the website is to delete your fork of the repository and make a new one.
The easy way is to always start your new branches from the KSP-CKAN/NetKAN master fork, rather than your own one. It will create a branch based on the upstream master in that case.

Cphoenix:
Actually, I do create them from the KSP-CKAN/NetKAN master. And yet, it still tries to merge all my commits from my own master fork instead. It's...extremely odd. I'm just going to eschew the desktop app as much as possible and stick to command line. (At least it'll be good practice for future comp. sci. courses. As long as you guys don't mind these little muck-ups.)

PR #4211, opened 2016-06-06
Add Game of Kerbals from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of RobForrest, to add [Game of Kerbals](https://spacedock.info/mod/749/Game%20of%20Kerbals) to CKAN.

Mod details:
    name = /mod/749/Game%20of%20Kerbals
    author = RobForrest
    abstract = Medieval pack for KSP
    license = MIT
    Homepage = 
    description = GAME OF KERBALS

Medieval pack for Kerbal Space Program.

Credit for cfgs/porting into game goes to TMasterson5

Models:

Rob Forrest
-Airship
-Bow
-Catapult
-Pitch Bombs
-Sword
-Wildfire Jars

TMasterson5
-Cannon
-Galleon

OTHERS:
-Castle - Kaos3D - http://tf3dm.com/3d-model/castle-28614.html
-Horse - 3dregenerator - http://tf3dm.com/3d-model/horse-883.html
-Row Boat - ahmetsalih - http://tf3dm.com/3d-model/boat-37697.html

Textures:

Rob Forrest
-Airship
-Bow
-Catapult
-Pitch Bombs
-Sword
-Wildfire Jars
-Galleon
-Row Boat
- Knights of all houses (requires Texture Replacer)
	- Arryn, Barratheon, Bolton, Forrest, Frey, KING, Knight, Lannister, Martell, Stark, Targaryen, Tully, Tyrell.

TMasterson5
-Cannon

OTHERS:
-Castle - Kaos3D - http://tf3dm.com/3d-model/castle-28614.html
-Horse - 3dregenerator - http://tf3dm.com/3d-model/horse-883.html

Please direct questions about this pull request to [RobForrest](https://spacedock.info/profile/RobForrest).

PR #4222, opened 2016-06-08
Add GameofKerbals

politas:
Replaces #4211


dbent:
#rebuild

dbent:
Looks like the `TextureReplacer` directory was removed from the download?

politas:
Maybe someone pointed out that the house logos they were using were copyrighted images?

dbent:
LGTM

PR #4192, opened 2016-06-03
Add Destruction FX from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of Concodroid, to add [Destruction FX](https://spacedock.info/mod/744/Destruction%20FX) to CKAN.

Mod details:
    name = /mod/744/Destruction%20FX
    author = Concodroid
    abstract = An updated version of BD's mod.
    license = CC0 1.0 Universal
    Homepage = 
    description = This is a s**imple mod** which adds **Destruction** Effects to your Kerbal Craft. It is an updated version of BahudmoD's **Destruction FX. **

Please direct questions about this pull request to [Concodroid](https://spacedock.info/profile/Concodroid).

PR #4226, opened 2016-06-09
DestructionEffects fixup

politas:
Following advice [here](http://forum.kerbalspaceprogram.com/index.php?/topic/99088-10xdestruction-effects-flames-and-smoke-on-joint-breaks/&do=findComment&comment=2622536).
Replaces the current DestructionEffects with a pure continuation.
Makes DestructionEffects-SemiContinued a separate, conflicting mod.
The Spacedock release appears to be someone re-releasing the DestructionEffects here without giving credit. (CC-0, so they can do that, but this source seems better.)
Replaces #4192

### DIFF
--- a/NetKAN/DestructionEffects-SemiContinued.netkan
+++ b/NetKAN/DestructionEffects-SemiContinued.netkan
@@ -1,0 +1,24 @@
+{
+    "spec_version"      : 1,
+    "identifier"        : "DestructionEffects-SemiContinued",
+    "$kref"             : "#/ckan/github/Wicloz/DestructionEffects-Semi-Continued",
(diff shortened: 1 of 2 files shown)


techman83:
LGTM, merging with thanks!

plague006:
Considering the .netkan was authored and maintained by the author you should probably give a courtesy ping to @Wicloz since you're forcing users away from their branch.

Wicloz:
If there are better destruction effect versions available, my version should not be used. It was meant as a temporary fix, and it seems we now have a better one. I would like for the .netkan file for DestructionEffects-SemiContinued to be removed, as it is now obsolete.

PR #4243, opened 2016-06-12
Update OPT SpacePlane Parts

Cphoenix:
Self-explanitory. Now points to SpaceDock as it should.

And suggestions for helping prevent users from incurring a massive, massive headache when this mod gets its next major update and breaks all the parts would be welcome. The best answer I've seen is [this](http://forum.kerbalspaceprogram.com/index.php?/topic/90246-the-comprehensive-kerbal-archive-network-ckan-package-manager-v1161-13-mar-2016/&do=findComment&comment=2626210), which would involve stopping this from updating itself (so instead of using $kref, build a custom file), and then creating a new identifier for the new version that breaks things.

Ideally, we need the functionality discussed here: KSP-CKAN/CKAN#60


politas:
I would make a new identifier for this and call it OPTSpacePlaneParts-Beta or some such, leaving the original identifier for the ongoing version when it comes out..

Cphoenix:
Could do that, but wouldn't that still update if the author pushes the save-breaking update to SpaceDock?

politas:
If you used a $kref, yes. But if you just have a download link to the current release zip file, no scraping will happen.

Cphoenix:
Wait, what's K R&D doing up there? You know what, I'm just going to kill this branch and make a new PR. Silly me. Thanks for putting-up with me while I feel my way around git. (Don't worry, I reading [the documentation](https://git-scm.com/) too. :wink: )

PR #4245, opened 2016-06-12
Add EvaFollower

dbent:


politas:
Seems to do the job. Not sure what the x_netkan_version_edit is doing? 
Merging with thanks

PR #4240, opened 2016-06-11
Add Tribute to HarversteR from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of k0spwn, to add [Tribute to HarversteR](https://spacedock.info/mod/760/Tribute%20to%20HarversteR) to CKAN.

Mod details:
    name = /mod/760/Tribute%20to%20HarversteR
    author = k0spwn
    abstract = A ( very very ) simple mod that rename the Drills into Drill-O-Matic Resource HarvesteR
    license = MIT
    Homepage = 
    description = I saw that there was a petition to rename the "'Drill-O-Matic' Mining Excavator" into "Drill-O-Matic Resource HarvesteR" so I made a simple modulemanager patch that does it ! ( Aslo for the small drill )

Please direct questions about this pull request to [k0spwn](https://spacedock.info/profile/k0spwn).

PR #4246, opened 2016-06-12
Add TributetoHarvesteR

politas:
Replaces #4240


techman83:
LGTM, merging with thanks!

PR #4255, opened 2016-06-14
Update Workshop.netkan

dbent:
Most recent release does not have a top-level `GameData` directory. Just use default install information.


dbent:
Basic change, merging myself.

PR #4256, opened 2016-06-14
Update BWA-X-ROMBUS.netkan

Lupinarius:
CKAN still shows this as incompatible with 1.05, so I'm guessing the .5 is needed after the 1.0 in the ksp_version_min statement. I thought that the lower 1.0 version number would be inclusive of versions in between but maybe not? If you can see any other reason this shows up as not compatible with KSP-1.0.5, please let me know so I can fix it, thank you and sorry about any extra work this has caused.


politas:
No, 1.0 should be fine. I see it as compatible in my CKAN client. Have you restarted CKAN? KSP compatibility changes (and a few other types, too) in a mod release only take effect after a restart of CKAN.

PR #4258, opened 2016-06-15
Add PSA-Greenhouse

Cphoenix:
See [forum thread](http://forum.kerbalspaceprogram.com/index.php?/topic/141704-112-psa-greenhouse-for-usi-life-support-v-1123-2016jun13/).

### DIFF
--- a/NetKAN/PSA-Greenhouse.netkan
+++ b/NetKAN/PSA-Greenhouse.netkan
@@ -1,0 +1,16 @@
+{
+    "spec_version" : 1,
+    "identifier"   : "PSA-Greenhouse",
+    "$kref"        : "#/ckan/spacedock/767",
+    "license"      : "GPL-3.0",
+    "depends" : [
+        { "name" : "ModuleManager"},
+        { "name" : "USI-LS"}
+    ],
+    "install" : [
+        {
+            "find"       : "PanarchistSpaceAlliance",
+            "install_to" : "GameData"
+        }
+    ]
+}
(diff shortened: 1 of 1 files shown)


politas:
#   Failed test 'PSA-Greenhouse - spec_version v1.4+ required for install with 'find''
#   at t/metadata.t line 139.

Cphoenix:
Hope that fixes it.

politas:
To have a spec_version other than 1, it needs to be entered as a version string, eg `"v1.4"`

politas:
Since it has MM configs for USI-UKS (resource consumer) and CLS, it might be best to make USI-LS a "Recommend" rather than a depend, and suggest those two.

Cphoenix:
Actually you were right. There isn't an MM config for CLS. The CLS module is in the part configs themselves, line 92 in Greenhouse1.cfg and line 95 in Greenhouse2.cfg. I'll add USI-UKS-Shared and KeepFit to suggests, and CLS to depends?

politas:
I _think_ that CLS module ref won't cause problems if CLS isn't installed, but maybe ask in the forum thread or a direct message to the author? Make it a recommend at least.

Cphoenix:
Will do. Waiting on response from author then.

Cphoenix:
Alright, fixed-up the spec version correctly this time, and moved around dependencies. The only real hard one is MM.

politas:
Lokos good, passes Jenkins. Merging with thanks!

PR #4189, opened 2016-06-03
Add Telemachus 2.0 from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of wonderbreadninja, to add [Telemachus 2.0](https://spacedock.info/mod/742/Telemachus%202.0) to CKAN.

Mod details:
    name = /mod/742/Telemachus%202.0
    author = wonderbreadninja
    abstract = Telemachus 2.0 is a updated version of the masterpiece created by RICHARD BUNT that allows you to track vital vehicle systems in a separate web browser on your computer or on other computers via your local network.  This allows you to run a "Mission Control" style setups to help govern your mission.  With Telemachus, you will be able to not only see the data, but you can also control your ship via the web browser as well!
    license = BSD
    Homepage = http://forum.kerbalspaceprogram.com/threads/24594
    description = **Telemachus 2.0**

This is a updated version of the masterpiece created by ***RICHARD BUNT*** that allows you to track vital vehicle systems in a separate web browser on your computer or on other computers via your local network.  This allows you to run a "Mission Control" style setups to help govern your mission. 

With Telemachus, you will be able to not only see the data, but you can also control your ship via the web browser as well!  This can create some interesting co-op situations in which others can act as your "mission Control" as you play.

Please direct questions about this pull request to [wonderbreadninja](https://spacedock.info/profile/wonderbreadninja).

PR #4261, opened 2016-06-17
Update Telemachus to use SpaceDock

Cphoenix:
Because it's easier to manage SpaceDock NetKANs than GitHub ones.

Alternative to #4189 .

Odd that this fails checks. O.o Works fine for me.

### DIFF
--- a/NetKAN/Telemachus-2.netkan
+++ b/NetKAN/Telemachus-2.netkan
@@ -6,8 +6,10 @@
     "conflicts"    : [
         { "name" : "Telemachus" }
     ],
-    "install"      : [
-        { "file"       : "GameData/Telemachus"},
-        { "install_to" : "GameData" }
+    "install" : [
+        {
+            "file"       : "GameData/Telemachus",
+            "install_to" : "GameData"
+        }
     ]
 }
--- a/NetKAN/Telemachus-2.netkan
+++ b/NetKAN/Telemachus-2.netkan
@@ -1,0 +1,13 @@
+{
+    "spec_version" : 1,
+    "identifier"   : "Telemachus",
+    "$kref"        : "#/ckan/spacedock/742",
+    "license"      : "BSD-4-clause",
+    "conflicts"    : [
+        { "name" : "Telemachus-2" }
+    ],
+    "install"      : [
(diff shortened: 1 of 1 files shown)


politas:
You are aware that this "Telemachus 2.0" is not by the original author?

politas:
I'm really not happy with this Spacedock release taking over the Telemachus identifier. I think we should handle it as a separate mod, conflicting with Telemachus. Unfortunately, the author has put it on Spacedock with the name "Telemachus 2.0", which I think is very confusing.

Cphoenix:
Better? Not sure why it's failing with JSON deserialization error after the transform. $kref should be valid. Maybe I borked something...

politas:
I stared at it and stared at it, and then I saw it.  

    "install"      : [
        { 
            "file"       : "GameData/Telemachus",
            "install_to" : "GameData" 
        }
    ]

instead of what you have, that comes out as:

    "install": [
      {
        "file": "GameData/Telemachus"
      },
      {
        "install_to": "GameData"
      }
    ],

(Which is two incomplete install stanzas)

Cphoenix:
:confused: Okay, I have to admit I don't see it. Isn't what I have syntactically the same as what's in the spec? It gives this as an example of a valid install stanza:

```
"install" : [
    {
        "file"       : "GameData/ExampleMod",
        "install_to" : "GameData"
    }
]
```

Forgive me if I'm being obtuse. That's...very weird. Is there some clear fix that's sailing over my head here?

politas:
Each bit inside {} is a separate set of instructions, a JSON structure which has to fit the rules for an install stanza, so it needs something to say what to grab from the zip file, and something to say where to put it. 

What you have is an array of two separate install stanza structures. One which says "grab this GameData/Telemachus file or folder" but doesn't say where to put it, and then an entirely separate one that doesn't say what to grab, but says that undefined nothingness has to go in GameData.

Cphoenix:
I get how JSON works. :wink:  What threw me was that what you quoted in the code block above was a valid install stanza, but the file itself is wrong. I stupidly assumed you copied that from the file directly (stupid me and my assumptions).

Fixed now.

politas:
LGTM. Merging with thanks

PR #4252, opened 2016-06-13
Add Launch Clamp Fuel Pumping (LCFP) from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of WillThe84th, to add [Launch Clamp Fuel Pumping (LCFP)](https://spacedock.info/mod/769/Launch%20Clamp%20Fuel%20Pumping%20%28LCFP%29) to CKAN.

Mod details:
    name = /mod/769/Launch%20Clamp%20Fuel%20Pumping%20%28LCFP%29
    author = WillThe84th
    abstract = Adds a fuel pumping capability to the stock launch clamps.
    license = MIT
    Homepage = 
    description = This mod adds a module manager file to add a fuel pump button to the stock launch clamps, it will also increase the mass and cost of the launch clamps for balancing reasons.

Please direct questions about this pull request to [WillThe84th](https://spacedock.info/profile/WillThe84th).

PR #4262, opened 2016-06-18
Add Launch Clamp Fuel Pumping

Cphoenix:
Alternative to #4252.

Just needed to add a depends for MM.


politas:
LGTM, merging with thanks!

PR #4266, opened 2016-06-19
Set epoch due to past version weirdness

### LINKED ISSUE #4264: Historian Expanded Failing To Update
Despite #4257 it appears that Historian Expanded still isn't updating correctly. At least for me, it's not appearing after refreshing the CKAN application.

It [appears](https://github.com/KSP-CKAN/CKAN-meta/tree/master/Historian-expanded) that the correct metadata files are available. Perhaps the incorrect 1.2.5 file (not v1.2.5) needs to be deleted?

[status.ksp-ckan.org](http://status.ksp-ckan.org/) reports that it was last checked and inflated 3hrs ago (at time of posting this issue) but last indexed 13 days ago. O.o

politas:
A dot after the v? Really?
Closes #4264


politas:
Minor change with an immediate and long-awaited impact. Merging now.

PR #4280, opened 2016-06-24
RN fix for new ksp version

raidernick:
-Update files for new ksp version

### DIFF
--- a/NetKAN/AntaresCygnus.netkan
+++ b/NetKAN/AntaresCygnus.netkan
@@ -11,7 +11,7 @@
     },
     "$kref"         : "#/ckan/github/KSP-RO/Antares-Cygnus",
     "license"       : "CC-BY-NC-ND-3.0",
-    "ksp_version"   : "1.1.2",
+    "ksp_version"   : "1.1.3",
     "install"       : [
     {
         "find"      : "KK_Antares",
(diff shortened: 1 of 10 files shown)


raidernick:
it seems to be stalled because of firespitter, which seems to work in 1.1.3 just fine, it can probably be marked as compatible.

politas:
@BobPalmer is maintaining the metadata for FirespitterCore. He would need to update that.

Bob?

raidernick:
@politas since I don't see any issues with FS can I simply update the metadata for 1.1.3 so I can finish this pr?

politas:
Sorry, I've been somewhat busy. If these are new releases targeted only at 1.1.3, then this should be fine to merge. As long as there will still be working releases for anyone using 1.1.2. If you're saying the existing 1.1.2 releases are also fine for 1.1.3, then it would be better to switch to Min /Max KSP versions. (As I believe one of these already is)  Sorry, commenting on my phone, so I have limited abilities.

raidernick:
there now they are min/max versions. Yes they still work with 1.1.3 I made no changes so even with FS metadata not fixed I want this merged.

politas:
Of no one else gets to it, I'll merge it this evening

BobPalmer:
FYI - new Firespitter is up

PR #4294, opened 2016-06-26
Removes WildBlueIndustries Mods from CKAN database

Angel-125:
Licensing on my mods is now out of date; I would like for my mods to be de-listed from CKAN Thanks.


TeddyDD:
Better way to de-list mods is to change extension to .frozen instead of removing netkans. This way It will be easier to add your mods back if you decide to do so. If you are happy with that solution I will send PR to rename your ,netkans in a minute.

Angel-125:
So long as the .frozen extension results in the mods being de-listed from CKAN and unavailable, I'm amiable to your proposal, thanks.

TeddyDD:
Ok, thanks. I'll prepare PR then.

politas:
Freezing is merged.

Issue #4295, opened 2016-06-26
PersistentRotation update not being displayed.


MarkusA380:
I have PR submitted on SpaceDock and it is published to CKAN. But the updated version does not appear on CKAN.

Blu3wolf:
https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/PersistentRotation.netkan
I can see the 1.0.2 version as the latest version on CKAN. Perhaps it didnt like your versioning changing from three digits to two digits?
Perhaps someone with more know how can chime in.

Blu3wolf:
The CKAN meta file doesnt include any versions after 1.0.2, so presumably its something to do with the process that creates new CKAN files from the NetKAN files and external data from SpaceDock.

Blu3wolf:
In version 1.0.2, your downloaded file structure was GameData/PersistentRotation. Now, its PersistentRotation/GameData/PersistentRotation. It looks like you didnt update the NetKAN file to reflect that (and no one else caught it either). Should be an easy fix now.

Blu3wolf:
This should hopefully be fixed now?

Blu3wolf:
Can download in CKAN now, should be okay to be closed I think.

politas:
Closing as per above. Please reopen if problems persist, @MarkusA380

PR #4300, opened 2016-06-30
Update Persistent Rotation directory structure

Blu3wolf:
Fix directory structure to reflect new changes since 1.1 as per #4295


MarkusA380:
Thank you for the help. I did not know that could've been a problem. :)

politas:
Merging with thanks!

PR #4312, opened 2016-07-01
Increment version number for Distant Object Enhancement

Blu3wolf:


politas:
Why did you take the settings out of the settings mod and put it in the DLL mod?

Blu3wolf:
Because up until now, thats where its been. 

The RSS setup provides a flare config for RSS, and the default setup provides a flare config for Stock. Neither provide a settings file. Its pretty easy to add the settings file to the default setup, but then it would need to be added to the RSS setup as well, and I dont intend on modifying that.

PR #4320, opened 2016-07-02
Update Distant Object Enhancement

Blu3wolf:
Carries on from #4312 

* increments version number

* Filters Settings.cfg from DOEdefault, preventing installation from trying to overwrite files


politas:
Tests fine locally. Merging with thanks!

PR #4321, opened 2016-07-02
Just another netkan for a SETI mini mod

Y3mo:
Along the lines of SETI-RemoteTechConfig.

Best regards,
Yemo/Y3mo


Y3mo:
No idea why that does not work, it is an exact replica of the one for SETI-RemoteTech.netkan.

Y3mo:
Retracted due to bug with the mod.

PR #4323, opened 2016-07-03
Update EPL 5.4.0 for 1.1.3

Blu3wolf:
Updates fixed .NetKAN file for EPL to point at 5.4.0 instead of 5.3.2


politas:
Merging with thanks. We can edit the .ckan for EPL 5.3.2 to restrict it.

PR #4324, opened 2016-07-03
Update toolbar

Blu3wolf:
Changes Toolbar KSP version to KSP version min and max, 1.1.2 and 1.1.3 respectively


Blu3wolf:
Echo for attention by @Dazpoet  or @politas

politas:
No replies from Blizzy78 on the forum thread or to PMs to say there is a new release, and reports that it's working. Merging with thanks.

PR #4329, opened 2016-07-04
Override DRE versioning until DRE actually updates

NathanKell:


politas:

PR #4330, opened 2016-07-04
fix homepage url

miki-g:
bad and ugly forum - changed thread URL ( after change of title )

### DIFF
--- a/NetKAN/RSS-Airports.netkan
+++ b/NetKAN/RSS-Airports.netkan
@@ -25,6 +25,6 @@
        }
     ],
     "resources": {
-        "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/139827-112-airports-fictional-ksc-for-real-solar-system-ksc-switcher/"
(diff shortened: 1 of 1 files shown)


politas:
You've added a comment= bit at the end that will opena  specific comment rather than the main thread. If you want a link that will remain unchanging, the forums currently ignore everything after the first "-", (but sometimes you need to have _a_ character there, whatever it may be.) So for example, this should work:

    "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/139827-*"

politas:
Oh, and it looks like Jenkins is saying your .version file is not strict JSON.

miki-g:
.version file I fixed yesterday immediately after the jenkins test
but I do not know how to run jenkins again without changing the file .netkan

thanx for sugestion with * in url :) ... #comment is "ok" ... for better focus on first post (scrolling away forum banner, menu, ... ) ... or it is bad idea? :)

politas:
#rebuild

politas:
Merged with thanks! Always good to learn a new trick!

PR #4331, opened 2016-07-04
fix homepage url

miki-g:
bad and ugly forum - changed thread URL ( after change of title )

### DIFF
--- a/NetKAN/RSS-FictionalKSC.netkan
+++ b/NetKAN/RSS-FictionalKSC.netkan
@@ -29,6 +29,6 @@
     ],
 
     "resources": {
-        "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/139827-112-airports-fictional-ksc-for-real-solar-system-ksc-switcher/"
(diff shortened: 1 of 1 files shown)


politas:
AS in #4330, you've added a comment= bit at the end that will open a specific comment rather than the main thread (I believe it's the OP.) If you want a link that will remain unchanging, the forums currently ignore everything after the first "-", (but sometimes you need to have a character there, whatever it may be.) So for example, this should work:

    "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/139827-*"

And also like #4330, netkan is complaining about your .version file:

    After parsing a value an unexpected character was encountered: ". Path 'KSP_VERSION', line 14, position 3.

miki-g:
.version file I fixed yesterday immediately after the jenkins test
but I do not know how to run jenkins again without changing the file .netkan

thanx for sugestion with * in url :) ... #comment is "ok" ... for better focus on first post (scrolling away forum banner, menu, ... ) ... or it is bad idea? :)

politas:
#rebuild

politas:
I hadn't thought about using the comment link that way; that's quite clever!

politas:
Merging with thanks! Always good to learn a new trick!

PR #4335, opened 2016-07-04
Make SXT KSP version generic

NathanKell:


politas:

PR #4336, opened 2016-07-05
VSR add ver override to 1.9.4 too

NathanKell:


politas:
Can 1.9.2's compatibility be restricted?

NathanKell:
They both should be for both versions--there are serious issues in 1.9.4 that break FAR handling of engines, so RO players (and anyone else using FAR) need to use 1.9.2 for now.
I am hoping Ven can upload a fix soon, or if not I'll see if I'm allowed to post a 1.9.5 that selectively reverts the problematic bits.

PR #4348, opened 2016-07-05
Add min/max version to ancient outdated KCT-RP0 conf

NathanKell:


NathanKell:
Says RP-0 isn't available for those versions, which is really odd.

politas:
So let me get this straight - This KerbalConstructionTime-RP-0-Config is no longer required, and all the old releases should have KSP version restrictions applied to stop this appearing in people's CKAN client?

We'll need to do that in CKAN-meta direct. There are thirteen .ckan files that all need to be changed.

NathanKell:
Ah, quite right.

NathanKell:
And yes, you understand correctly.

politas:
Ok, sorry to take so long getting back to this. @NathanKell , can you change this PR to rename the file to `KerbalConstructionTime-RP-0-Config.frozen`?

NathanKell:
Done so! No worries.

politas:
Thanks for the super-prompt response!

NathanKell:
:)

PR #4351, opened 2016-07-06
RN dependency meta update for TR and BD

raidernick:
-These are still compatible and fully working in ksp 1.1.3


raidernick:
@politas texture replacer is not showing up at all in ckan, not even as incompatible. This is preventing my mods from showing up at all in any filter. I know texture replacer is 1.1.+ compatible and still works fine. Can you look at the metadata for that and see if there is anything obviously wrong with it causing it to not show up at all?

politas:
Yes, that would be it. As far as I know, that's the only reason for a mod with a .ckan marked as compatible to disappear from all filters.

PR #4355, opened 2016-07-06
Updated dependancies

sswelm:
I added FilterExtensions to dependancies, although technically not required, it prevents the user from using the deprecated parts


politas:

PR #4363, opened 2016-07-06
Update MagiCore to work on 1.1.2/3 both

NathanKell:
Can be reverted once https://github.com/magico13/MagiCore/pull/2 is accepted. As of now nothing that depends on MagiCore (KCT, mods that req KCT) will show in CkAN due to this mismatch.

Hopefully @magico13 can get to the PR soon, but I see no harm in doing this until then.


NathanKell:
My record recently has been rather better, so I'm going to go ahead and merge this. Hope that's ok! :)

PR #4367, opened 2016-07-07
Freeze StockVisualEnhancements

politas:
Author has requested removal, and installation is wrong


TeddyDD:
Here is fixed netkan for medium resolution variant of this mod (If anyone interested) https://gist.github.com/TeddyDD/8d32965e71bae7dd7d4a0ba11a0aab82

politas:
The Scatterer config files in SVE have been modified,so you need to take them into account. This is part of why I froze the mod. CKAN installation was broken.

TeddyDD:
Yeah, you right. That means Scatterer would have to be spitted into scatterer, scatterer-confing and scatterer-sunflare. I can do this later, but I have to compare all three editions of SVE with Meld to avoid any surprises.

politas:
I was actually thinking of just doing the same thing with Scatterer as we do for EVE in SVE

politas:
I'm checking with the new mod maintainer about any changes in the other bundled mods.

PR #4377, opened 2016-07-11
Update MagiCore.netkan

plague006:
Ref https://github.com/magico13/MagiCore/issues/1


techman83:
#rebuild

politas:
Looks good to me. Merging with many thanks!

PR #4398, opened 2016-07-14
Update SpaceXLegs.netkan

svendii:
Relink to new SpaceDock page

PR #4399, opened 2016-07-14
Rename SpaceXLegs.netkan to KerbalReusabilityExpansion.netkan

svendii:


politas:
Nno, #4398 is on the right track. Changing the identifier breaks people's upgrade path.

PR #4400, opened 2016-07-14
FASTCORPAutogarage

linuxgurugamer:
closes #4260


politas:
Needs a dependency on RasterPropMonitor for full functionality, so we would normally include that.

Also, given the need for changes to the settings ffile, can we override the abstract to include a request to check the homepage for proper configuration?

politas:
Do those "/n"s work in the client?

linuxgurugamer:
Not sure.  I think so, but will be checking once it gets accepted.  If not, i'll redo it

politas:
Just tested, and it does indeed work beautifully!
Merging with thanks!

PR #4402, opened 2016-07-14
2KASJ0523Planetarysystem.netkan

linuxgurugamer:
closes #4265

### DIFF
--- a/NetKAN/2KASJ0523Planetarysystem.netkan
+++ b/NetKAN/2KASJ0523Planetarysystem.netkan
@@ -1,7 +1,7 @@
 {
     "$kref": "#/ckan/spacedock/774",
     "spec_version": "v1.4",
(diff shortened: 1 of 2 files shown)


politas:
Hmm. I thought we'd changed that "No starting with a number" test. @techman83?

techman83:
We might have done in the testing library, but the schema regex doesn't allow it.
```
^[A-Za-z][A-Za-z0-9-]*$
```
Changing it to this will allow it to start with anything alpha-numeric:
```
^[A-Za-z0-9-]*$
```

politas:
Ok, the schema is now in sync with the spec. @linuxgurugamer , can you strip the leading "a" from the filename and identifier?

politas:
Looks good to me. Glad the schema change worked! Merging with thanks!

PR #4405, opened 2016-07-15
TimeControl Epoch

politas:
Version string weirdness


politas:
Minimal change, passes Jenkins and does what's needed. Merging.

PR #4411, opened 2016-07-15
Added metadata for AT-Utils library.

allista:
Added ModuleManager and AT-Utils as dependencies to
ThrottleControlledAvionics metadata.

### DIFF
--- a/NetKAN/AT-Utils.netkan
+++ b/NetKAN/AT-Utils.netkan
@@ -1,0 +1,27 @@
+{
+    "spec_version" : 1,
+    "identifier"   : "AT-Utils",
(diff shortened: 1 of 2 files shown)


allista:
Could someone check what's the problem with the build here? Is my metadata at fault or the other patches that got into the same built?

politas:
About Ils is failing because it has no releases on Github, only a pre-release.

allista:
Thanks for the reply. Fixed that.

politas:
The server ModuleManager resides on is having some issues at the moment, which Sarbian is aware of. This will probably pass testing when that's sorted out.

linuxgurugamer:
#rebuild

politas:

PR #4408, opened 2016-07-15
Quickfinecontrol

linuxgurugamer:
closes #4275

PR #4414, opened 2016-07-17
Quickcursorhider

linuxgurugamer:
Look for problems


politas:
This got added in the earlier PR, so would need to be re-based

linuxgurugamer:
I don't understand, sorry

politas:
This PR is trying to add a new file that has already been created. It was included in #4408. Thus the conflict.

linuxgurugamer:
Oh, I didn't realize that.

No need for this, the files are identical

PR #4422, opened 2016-07-18
Add For Science to conflicts list for DMagic Orbital Science

DMagic1:
For Science doesn't properly interact with Orbital Science experiments. This can lead to simply thinking the experiments can't be run, to preventing experiments from being transmitted while using Remote Tech.


politas:

PR #4423, opened 2016-07-19
DmagicOrbitalScience conflict with ForScience

politas:
For Science doesn't properly interact with Orbital Science experiments. This can lead to simply thinking the experiments can't be run, to preventing experiments from being transmitted while using Remote Tech.
Closes #4422


politas:
Nevermind, Dmagic raised a PR, not an issue.

PR #4225, opened 2016-06-08
Add Year2200 from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of matterbeam, to add [Year2200](https://spacedock.info/mod/722/Year2200) to CKAN.

Mod details:
    name = /mod/722/Year2200
    author = matterbeam
    abstract = A complete resource and nuclear energy mod allowing you to play with technologies of the next 200 years.
    license = MIT
    Homepage = http://forum.kerbalspaceprogram.com/index.php?/topic/139310-112-year-2200-a-nuclear-future-testing/
    description = Currently in testing.

Adds fusion and fission nuclear reactors that produce MWt or MWe.
Adds thermal and electric engines.
Adds nuclear fuels and water.
Replaces stock ISRU with a more realistic, yet stockalike solution.
Includes SimpleConstruction: build rockets using stock parts!

Please direct questions about this pull request to [matterbeam](https://spacedock.info/profile/matterbeam).

PR #4425, opened 2016-07-21
Add Year2200

politas:
Fixed Spacedock submission
Closes #4225

Waiting on confirmation from author that this submission was intentional, and clarification of whether SimpleConstruction should indeed be a conflict, or a dependency.


politas:
Ok, passes without the RSS issues. I'll add that back and then this can be merged.

PR #4445, opened 2016-07-26
Update SolarSailNavigator.netkan to KSP 1.1

bld:
Module has been modified and compiled to work with KSP 1.1 and Unity 5


politas:

PR #4450, opened 2016-07-28
Update Stock Bug Fix Modules & StockPlus

DinCahill:
This is my first time editing NetKAN.

The name of the zip, and folder within, changed from "StockBugFixPlus" to "StockBugFixControllerPlus". Should it be handled as I have done, or should the netkan file be renamed?


politas:
We certainly don't want to rename the netkan or change the identifier. This looks fine to me. Merging with thanks!

PR #4452, opened 2016-07-28
Update compatibility for Aviation Lights

DinCahill:
The author reckons this still works. Seems to work for me, too.


politas:

PR #4453, opened 2016-07-28
Add VesselMover Continued

DinCahill:
VesselMover has a new maintainer, so here's a new netkan file.

I'm not sure whether KSP-AVC should be a suggests or supports.

http://forum.kerbalspaceprogram.com/index.php?/topic/144421-*/


politas:

PR #4456, opened 2016-07-29
Updated Pledna, with Kopernicus

linuxgurugamer:
Closes #4444 

The original, while accepted, was incorrect. This one, generated by my new script, properly includes Kopernicus

### DIFF
--- a/NetKAN/Pledna.netkan
+++ b/NetKAN/Pledna.netkan
@@ -1,0 +1,19 @@
+{
+  "license": "CC-BY-NC-SA",
+  "identifier": "Pledna",
+  "x_via": "Automated SpaceDock CKAN submission",
+  "$kref": "#/ckan/spacedock/862",
+  "$vref": "#/ckan/ksp-avc",
+  "depends": [
+    {
+      "name": "Kopernicus"
+    }
+  ],
+  "install": [
+    {
+      "find": "Pledna",
+      "install_to": "GameData"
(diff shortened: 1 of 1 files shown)


politas:
Is this your own script you're using, not the [webtool developed by Blauerdaemon](http://forum.kerbalspaceprogram.com/index.php?/topic/143239-ckan-metadata-webtool)?

linuxgurugamer:
My own script, it does more
The "x_via" is there because I'm going to give this to the Spacedoc people to use instead of whatever they are currently using

linuxgurugamer:
Actually, I should say that they are different.  My script is fully automated, and is intended for use  by Spacedock when mods are uploaded and the CKAN box is checked.  It works by examining the zip file which is uploaded (right now I download the zip manually) and determining what's in the file and where things are supposed to go, etc

Blauerdaemon's is a manual input, which can actually provide more information, but requires the user's input

politas:
Interesting. How is it determining dependencies?

politas:
Tested locally, works fine. Merging with thanks!

linuxgurugamer:
It starts by having a local database of all active netkan ids and directories associated with them.  Then it examines the zip file and does some clever stuff :D

linuxgurugamer:
There are several special cases, such as ModuleManager, ContractPacks & Kopernicus whch have special code there

PR #4284, opened 2016-06-25
Add Half Size RSS from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of Pap, to add [Half Size RSS](https://spacedock.info/mod/787/Half%20Size%20RSS) to CKAN.

Mod details:
    name = /mod/787/Half%20Size%20RSS
    author = Pap
    abstract = Half Size Real Solar System (5x stock scale)
    license = CC-BY-NC-SA
    Homepage = http://forum.kerbalspaceprogram.com/index.php?/topic/142336-113-half-size-rss-v12/
    description = **DOWNLOAD THESE TEXTURES OR THE GAME WILL NOT LOAD!**

[https://github.com/pap1723/HalfRSS-Textures/releases](http://)

HalfRSS is a mod that uses the great work from @NathanKell and creates our Real Solar System in a smaller scale in KSP. At half the size of the RSS, HalfRSS is still 5 times larger than stock KSP and provides a huge challenge to players, but can be achieved with stock parts. I always loved the idea of playing in our Real Solar System, but I am not a good enough player to be able to accomplish that on the grand RSS scale. The guys over at Stock Size RSS gave me the inspiration for this with their awesome work, but I found that size to be easier than I hoped for.

I am hoping that this is a good balance between stock sizes and full size RSS. You can play this without using Realism Overhaul and without having to create super huge launchers to get anywhere.

**SPECIAL THANKS:**
* NathanKell and the entire Real Solar System group for creating almost everything that is released in this mod
* CaptRobau for letting me use his planets Thatmo and Slate from the Outer Planets mod as two of the Dwarf Planets that we know almost nothing about
* OhioBob for creating the atmospheres that are present in the mod
* The entire Kopernicus team for creating the mod that makes this all possible! @Thomas P. @NathanKell @KillAshley @Teknoman117, Bryce Schroeder
* ImkSushi and @pozine for their work on Real Solar System Expanded that provided Ceres and Vesta

**DOWNLOAD:**
[https://github.com/pap1723/HalfRSS-Textures/releases] - These are the textures that are necessary to download

Included in the download:

Half RSS Mod files
Module Manager (by @sarbian, @swamp_ig, @ialdabaoth) See thread for details, license, and source: http://forum.kerbalspaceprogram.com/threads/55219
Kopernicus - See thread for details, license, and source: http://forum.kerbalspaceprogram.com/threads/114649 
[...]

PR #4458, opened 2016-07-29
HalfSizeRSS

linuxgurugamer:
created by new automated script
hand verified
Closes #4284


NetKAN/HalfSizeRSS.netkan
```diff
@@ -0,0 +1,25 @@
+{
+  "license": "CC-BY-NC-SA",
+  "identifier": "HalfSizeRSS",
+  "x_via": "Automated SpaceDock CKAN submission",
+  "$kref": "#/ckan/spacedock/",
```
politas: This is unlikely to work

linuxgurugamer:
Bug in script, fixed, regenerated and updated

linuxgurugamer:
#rebuild

politas:
This going to need at least one more netkan created for the Half RSS textures, so that can be a dependency.

linuxgurugamer:
Look like he has 4 files in that zip, including the basic HalfRSS netkan.  Should I put them in all one pull?

linuxgurugamer:
I've taken his files and fixed the syntax errors, also merged his basic .netkan with mine (he was m issing several dependencies).  So now I'm just waiting to hear the best way to do this.

ayan4m1:
It looks like this mod has other dependencies hard-included in its SpaceDock download. I'm still learning about NetKAN packaging but I think Kopernicus and ModuleManager should be dependencies rather than copy-pasted into the mod ZIP.

Also, just checking where the mod is supposed to be sourced from - GitHub or SpaceDock - because right now the author-supplied netkans and the ones you have provided disagree over the source location for the mod. Both locations are including other mods without specifying them as dependencies.

linuxgurugamer:
The version I have on my system has all the dependencies, and for now, I'm using the author-supplied source location.
The reason Kopernicus & ModuleManager are in the zip is for those people who aren't using  CKAN.  CKAN ignores those directories and my netkan properly includes them as dependencies.
Again, I'm waiting on advice on whether I should put all 4 netkan files into a single commit or not.

politas:
So, I gather their are different netkans for the different texture quality levels - We would want them all to have a "provides" line and a "conflicts" section for a common "HalfRSSTextures" virtual netkan, and then have the main HalfRSS.netkan depend on that virtual netkan.

linuxgurugamer:
I KNOW THAT!

My question is how to submit, as a single commit or multiple commits

politas:
Oh, sorry. Don't think it really matters, as long as they're added to this PR's branch.

linuxgurugamer:
ok.  I'll add these as one commit later this evening

DinCahill:
CI failure incoming!

DinCahill:
I don't actually know how to fix the error, though. I just ran NetKAN quicker than Jenkins. You might want to do the same to save time. You can also check the JSON with jsonlint.

linuxgurugamer:
Not sure what the problem is with the HalfRSS.netkan, can someone take a look?

techman83:
@DinCahill yeah, Jenkins has to spin up a container and it does a little more than just run NetKAN on the file :smile: 

@linuxgurugamer it's barfing on this NetKAN
```
Running NetKAN for NetKAN/HalfRSS-Textures2048.netkan
...
4738 [1] FATAL CKAN.NetKAN.Program (null) - After parsing a value an unexpected character was encountered: ". Path 'DOWNLOAD', line 5, position 3.
```

DinCahill:
Should we run NetKAN through the debugger to see what's happening? I haven't done that with C# before, so I'll just do something else for now :P

linuxgurugamer:
@techman83 It isn't barfing on that.  I have been running netkan.exe locally, it's barfing on the HalfRSS.netkan:
$ netkan.exe HalfRSS.netkan
544 [1] FATAL CKAN.NetKAN.Program (null) - After parsing a value an unexpected character was encountered: ". Path 'DOWNLOAD', line 5, position 3.

Found the problem, the .version file is in the wrong place.  Once I removed the vref line, it works

linuxgurugamer:
Ok.  I've opened an issue with @pap1723 about the bad .version file, hopefully when he fixes it he will let me know and I'll update the netkan.
All 4 netkans are passing my local netkan.exe check, i believe the problem now is that the each depend on each other, and it's a race condition.  I suspect it will need manual intervention at this point

techman83:
@linuxgurugamer aha - cool. The CI process puts all the metadata in the same PR into the local repository it creates. So it should be available. The build failure analyser reports this:

```
At least one GameData directory has been found within the GameData or a directory within a mods Installed Directory.
```
It also reports 
```
HalfRSS requires HalfRSS-Textures but nothing provides it
```
Which if you've dropped the `$vref` may be version related.

politas:
No, it's just that netkan's testing each .netkan file separately rather than all at once, so the interlinked dependencies are doomed to failure. Let me test the install locally.

linuxgurugamer:
The alternetive would have been to them up without any dependencies, and then update once accepted.
Thanks for testing locally

politas:
Yep. Missing comma on the "DOWNLOAD" line in HalfRSS.version
That'd do it!

techman83:
@politas not quite. It [generates](https://github.com/KSP-CKAN/xKAN-meta_testing/blob/master/NetKAN/build.sh#L306), then [injects](https://github.com/KSP-CKAN/xKAN-meta_testing/blob/master/NetKAN/build.sh#L345) all the build files before running [ckan](https://github.com/KSP-CKAN/xKAN-meta_testing/blob/master/NetKAN/build.sh#L371)

linuxgurugamer:
@politas That's why I removed the vref

politas:
Yep, just confirmingthat is what was wrong. All files test fine locally. ckan's created, all install fine.
Merging with thanks!

politas:
@techman83, Ah, could it be because there are multiple providers of the dependency, and it needs input to choose one?

linuxgurugamer:
Great.  I was chatting with the author earlier today, once he gets the version file fixed, I'll update the netkan for HalfRSS

pap1723:
Here is the corrected .VERSION file for you. Thanks for all of the help getting this together.
[HalfSizeRSS.zip](https://github.com/KSP-CKAN/NetKAN/files/400218/HalfSizeRSS.zip)

linuxgurugamer:
@pap1723 You need to get the corrected .version file into the release on giithub (and Spacedoc), Netkan/CKAN download from there.  The current config is downloading from the HalfSizeRSS release directory on Github

techman83:
@politas aha, yes that'll do it!

techman83:
@linuxgurugamer @pap1723 once that's done ping me and I'll nudge the indexer.

pap1723:
OK, files are updated at Spacedock and Github!

linuxgurugamer:
@techman83 I just updated the netkan

linuxgurugamer:
or not :D

linuxgurugamer:
I updated my branch, then realized you had merged.

linuxgurugamer:
@techman83 Can you add the vref to the netkan and redo it?  If you need me to update it, let me know

linuxgurugamer:
NM, I just did it on a new branch/pull

PR #4438, opened 2016-07-23
Add Soviet Space Program Career Conversion from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of tjsnh, to add [Soviet Space Program Career Conversion](https://spacedock.info/mod/858/Soviet%20Space%20Program%20Career%20Conversion) to CKAN.

Mod details:
    name = /mod/858/Soviet%20Space%20Program%20Career%20Conversion
    author = tjsnh
    abstract = Converts a [new] career mode game into the old Soviet Space program.
    license =  Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International License 
    Homepage = 
    description = This mod has a number of dependencies , and special installation instructions. Please view the forum link above. 

Please direct questions about this pull request to [tjsnh](https://spacedock.info/profile/tjsnh).

PR #4459, opened 2016-07-29
SovietSpaceProgramCareerConversion

linuxgurugamer:
Closes #4438

Created by shell script
Not sure if the Subassemblies should be installed to the root or not.  There is also another file in the root of the zip file, RemoteTech_Antennas.cfg

### DIFF
--- a/NetKAN/SovietSpaceProgramCareerConversion.netkan
+++ b/NetKAN/SovietSpaceProgramCareerConversion.netkan
@@ -4,13 +4,9 @@
   "x_via": "Automated SpaceDock CKAN submission",
   "$kref": "#/ckan/spacedock/858",
    "conflicts" : [ { "name" : "RemoteTech" } ],
-  "install": [
+   "install" : [
     {
-      "find": "Subassemblies",
-      "install_to": "GameData"
-    },
-    {
-      "find": "Ships",
+      "file": "Ships",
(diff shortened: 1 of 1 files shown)


politas:
The forum thread has special install instructions:

>Installation Instructions: 
Unzip the contents of "gamedata" into your "gamedata" like most other mods.
If you wish, copy the craft files from the ships and subassemblies folders into your save.
If you are using remotetech - and this is VERY important - you need to delete a couple of files (sorry) in order for things to work correctly. Consider moving them off somewhere so you can replace them later if you uninstall this mod.

>- Delete gamedata/RemoteTech/RemoteTech_Antennas.cfg and replace it with the one in the base leve of the zip.
>- Delete gamedata/Contares/Patches/CONTARES_RT2.cfg
>- Delete gamedata/Tantares/_Extra_RemoteTech.cfg

CKAN won't do any of the that stuff for RemoteTech, so we'll have to just leave that RemoteTech config file uninstalled and mark this mod as conflicting with RemoteTech.
If there are SPH/VAB folders, we can install the craft files in Ships, otherwise, just leave them out.

linuxgurugamer:
@politas seems to be good, please check

politas:
The contents of Ships should go in Ships rather than GameData

PR #4460, opened 2016-07-29
AdditionalProgressionContracts

linuxgurugamer:
created by new automated script
hand verified
Closes #4437

### DIFF
--- a/NetKAN/AdditionalProgressionContracts.netkan
+++ b/NetKAN/AdditionalProgressionContracts.netkan
@@ -1,7 +1,7 @@
 {
   "license": "CC-BY-NC-SA-4.0",
   "identifier": "AdditionalProgressionContracts",
-  "x_via": "Automated SpaceDock CKAN submission",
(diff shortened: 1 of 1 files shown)


politas:
Ah. Thanks to KSP-CKAN/CKAN#1477, that find doesn't work. Try this instead:

    {
        "find": "additionalprogressioncontracts",
        "install_to": "GameData/ContractPacks"
    }

linuxgurugamer:
So, if I understand this problem correctly, if there is no GameData directory (as in this case), and then there is a directory (such as ContractPacks) with only another directory in it, it will fail?

What would happen if there are two directories below the top dir?  or a directory and a file?

Thanks

linuxgurugamer:
Also, what would happen if there WAS a GameData, such that the file contained:
GameData/ContractPacks/packname/(allfiles)

politas:
The existence of GameData is irrelevant. We can't find a directory that only contains subdirectories, not files. (For now; it's a bug I'm planning to try to fix soon. :)

politas:

PR #4463, opened 2016-07-29
KepsilonKeridani

linuxgurugamer:
Closes #4403


politas:
LGTM. Merging with thanks!

PR #4385, opened 2016-07-12
Add Kerbal Cities Pack from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of lemlur, to add [Kerbal Cities Pack](https://spacedock.info/mod/839/Kerbal%20Cities%20Pack) to CKAN.

Mod details:
    name = /mod/839/Kerbal%20Cities%20Pack
    author = lemlur
    abstract = ever fancied a kerbin wide city network (planned) or smashing up moe than just ksc (planned but im bufggered if i know how ill do it)  just trying to fly through narrow buildings and around tight corners for the hell of it? well we have the city for you
    license = MIT
    Homepage = 
    description = **This mod is in early alpha!** ever wanted to fly through tight buildings and narrowly avoid smashing into stuff? ever wanted to have the possibility of a kerbin wide city network (final release) or even the ability to smash up Kerbal kosntucts  buildings  ( want to do this but ill be blowed if i know how!)
all this i hope to achive i you will support me in developing this mod by downloading or even a little donation! this prject started from kid donatons from the ksp comunity (had to purchase the blender plugin fo the heaft price of £80! phew! but the kerbalers saved the day by donating £105 to my campaign linked above despite the target being  £30! you guys are amazing!  

**i would like to thank:**

Nathan Johnson for kindly donating £10 to get this rolling

Billy Winn (AKA Jatwaa) who is now cheif youtuber and co deving with me a little! for his donation of £20

and Anon for donating and amazing £70, whoever you are, you are a legend!

Please direct questions about this pull request to [lemlur](https://spacedock.info/profile/lemlur).

PR #4464, opened 2016-07-29
KerbalCitiesPack

linuxgurugamer:
Closes #4385


politas:
Ok, there's an existing kerbalCitiesPack.frozen that you should delete (assuming this is a new source for the same mod)

linuxgurugamer:
I also saw that this file had the issue of a dir with only a subdir inside, so I changed the "find" to a "file"

politas:
This does not look like the same Kerbal Cities Pack that Amankduffers made.

politas:
Could use a homepage resource declaration:

    "resources": {
       "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/144012-kerbal-cities-pack"
   },

politas:
Great stuff! Merging with thanks!

PR #4378, opened 2016-07-11
Add Lost Twins from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of electricpantz, to add [Lost Twins](https://spacedock.info/mod/837/Lost%20Twins) to CKAN.

Mod details:
    name = /mod/837/Lost%20Twins
    author = electricpantz
    abstract = Adds a brown dwarf, 3 planets and 1 moon
    license = MIT
    Homepage = http://forum.kerbalspaceprogram.com/index.php?/topic/143522-lost-twins/
    description = This planet pack adds a small brown dwarf binary companion and it's two moons.
If you want Pics, referr to the Forum post for this mod. :P

Please direct questions about this pull request to [electricpantz](https://spacedock.info/profile/electricpantz).

PR #4465, opened 2016-07-29
LostTwins

linuxgurugamer:
Closes #4378


politas:
Hmm, surprisingly, that is working, despite KSP-CKAN/CKAN#1477. Merging with thanks!

PR #4473, opened 2016-08-01
Build env testing

techman83:
Lets see if we can coax Jenkins back into happiness.


techman83:
#rebuild

techman83:
#rebuild

techman83:
Yay, all seems to be good now :dancers:

PR #4475, opened 2016-08-01
SnapDock

linuxgurugamer:
closes #4472


politas:
Can't quite work out from the description what this mod does, but this seems to install it correctly. Merging with thanks!

PR #4478, opened 2016-08-01
BDArmoryContinued

linuxgurugamer:
closes #4467

### DIFF
--- a/NetKAN/BDArmoryContinued.netkan
+++ b/NetKAN/BDArmoryContinued.netkan
@@ -7,6 +7,14 @@
     "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/144775-113-bdarmory-continued-v01110-prerelease-28-jul-2016/",
     "repository": "https://github.com/PapaJoesSoup/BDArmory"
   },
+  "provides": [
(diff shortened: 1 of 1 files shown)


politas:
This will need a "provides" and "conflicts" for "BDArmory", or it will be largely useless. There are an awful lot of mods that depend on BDArmory. I'm sure people will be very pleased to see it back!

linuxgurugamer:
Given that this is replacing BDArmory, would it make sense to change the ID to BDArmory and elete the old file?

politas:
BahamutoD could always come back and take up where he left off. This is how we handle functional replacement forks.

linuxgurugamer:
Understood.  Will get to it later this evening

linuxgurugamer:
closes #4467

DinCahill:
Do we need the repository and homepages lines? They should get scraped from SpaceDock.

politas:
Ah, the Spacedock entry has been updated. No, we don't need it, but it won't hurt. This looks like it will do the job, now, and people have been waiting on BDArmory for 1.1.3 a long time. Merging with many thanks!

Issue #4278, opened 2016-06-24
Request for de-listing Sigma88 mods


Sigma88:
I would like to request the de-listing of my mods from CKAN

the interested mods are:

- [Sigma Dimensions](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/SigmaDimensions.netkan)
- [Sigma Binary](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/SigmaBinary-core.netkan)
- [Galactic Neighborhood](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/GalacticNeighborhood.netkan)
- [Sigma: OPM Recolor](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/SigmaOPMRecolor.netkan)
- [Sigma: Jool Recolor](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/SigmaJoolRecolor.netkan)
- [Sigma: PluronKhato](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/SigmaPluronKhato.netkan)
- [Sigma: Antennas](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/SigmaRTExpansion.netkan)
- [Sigma: OPM Tilt](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/SigmaOPMExpansion.netkan)
- [Sigma: EveRecolor](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/SigmaStockRecolor.netkan)
- [Sigma: OuterSpaceComms](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/SigmaModExpansions.netkan)
- [Sigma Binary - DunaIke](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/SigmaBinary-DunaIke.netkan)

If you want to freeze them or remove them completely that's your choice.
Before I come back to ckan I want to see how the attitude evolves and I want to consider the possibility of managing my metadata myself.

The longer this request gets ignored, the less likely I am to come back to ckan in the future.

None of my mods will get any updates untill this has been solved.

Sigma88:
I see you had time to add new mods to ckan but not time to respect your policy about de-listing.

I'll take care of that by myself if this cannot be achieved asking nicely

after all this is totally legal right?

Sigma88:
Freezing is fine for me as well

[...]

PR #4479, opened 2016-08-02
Update RealisticRemodel1x.netkan

WillThe84th:


DinCahill:
You need commas on the end of L12 and L13 to make this valid JSON.

The name of the mod on SpaceDock is still "Realistic Remodel 1x". Should the identifier be changed? (I'm new and don't know the full implications)

This depends on SigmaDimensions, but Sigma88's mods have been removed from CKAN (#4278). I don't know how to handle this.

@politas

politas:
It's much easier for everyone if the identifier is not changed. SigmaDimensions should be coming back to CKAN, but may be a while.

Olympic1:
Closing in favor of #4718

PR #4482, opened 2016-08-03
Add LightFixer! from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of daniel_l, to add [LightFixer!](https://spacedock.info/mod/874/LightFixer%21) to CKAN.

Mod details:
    name = /mod/874/LightFixer%21
    author = daniel_l
    abstract = Ever wondered why the sunlight is so dull? Try this! It amps up the sunlight intensity and disables ambient lighting (Good luck fumbling about in the dark!) Now everything looks exactly why the Nasa pictures with super bright lit areas and pitch black in the shade. (Requires Kopernicus.)
    license = CC-BY-SA
    Homepage = 
    description = Tired of your sunlight being so washed out and boring? Always wondering why you can see in the dark in space? Why the daylight is so dull and the night is ghost lit? Download the LightFixer today! (Needs [Kopernicus](http://forum.kerbalspaceprogram.com/index.php?/topic/140580-113-kopernicus-release-1-july-4/))

Please direct questions about this pull request to [daniel_l](https://spacedock.info/profile/daniel_l).

PR #4483, opened 2016-08-03
Add LightFixer!

DinCahill:
The mod is just a `Config.cfg` which edits Kopernicus using ModuleManager, which is why I added ModuleManager as a hard dependency, rather than getting it transitively through Kopernicus.

Closes #4482


politas:
LGTM, author has had time to object. Merging with thanks!

PR #4485, opened 2016-08-03
VehiclePrefabsReadytoGo

linuxgurugamer:


politas:
Should this have a BDArmory dependency? It looks like it should from the screenshots.

linuxgurugamer:
I don't think so, but I'll examine the file this evening.  What I remember is that there is about 130 different models

linuxgurugamer:
I looked at the config files of some of the military planes, and unless I'm missing something, they don't need BDArmory.  
Can someone else please take a look?

linuxgurugamer:
I just saw BahaSP in some of the config files, so I'll get it fixed

politas:
The BahaSP references I can find all look like engine FX files, rather than weaponry.

Olympic1:
BahaSP engine effects depend on`BDAnimationModules`

politas:
@linuxgurugamer, can you add that dependency?

linuxgurugamer:
yes

politas:
Looks all good now. Merging with thanks!

PR #4486, opened 2016-08-03
Updated AntennaRangePatches for spacedock

linuxgurugamer:


linuxgurugamer:
I had a few requests for this, so if it can get accepted quickly, that would be great
Thanks

DinCahill:
I tried to install this on my machine, but it said AntennaRange is incompatible. Indeed, it's listed on SpaceDock as for 1.1.2. However, the author and others say it works fine on 1.1.3. Why didn't this fail CI?

politas:
Looks good to me. Will work with the available versions of AntennaRange

PR #4489, opened 2016-08-04
Increase Atomic Age KSP Compatibility

politas:
Forum users report this mod is working (mostly) fine in KSP 1.1


DinCahill:
The zip file contains an "AtomicAge v3.1" directory (which is confusing since the zip is Atomic_Age-3.2.zip), which contains GameData and Ships directories. The mod in GameData is installed, but not the Ships. Should we be installing the ships?

politas:
We often don't

politas:
But now we will

DinCahill:
LGTM. Thanks :rocket: :shipit:

PR #4491, opened 2016-08-04
Added vref to HalfRSS

linuxgurugamer:


linuxgurugamer:
@politas The only change here is the added vref to the file, I guess you still need to do this locally

politas:
Need to check the zip file contents / installation: "GameData directory found within GameData".

linuxgurugamer:
I did, and there is no GameData inside Gamedata.  Also, when I run netkan locally, it is fine.  Could there be a problem with one of the other files?

politas:
Local installation of generated CKAN seems fine. No second-level GameData. Merging with thanks!

linuxgurugamer:
Any idea why it thinkgs there is a 2nd level GameData?

politas:
I think that might be a failure to recognise the actual issue why it is failing.

politas:
It doesn't throw a "nothing provides", because there are mods providing it, but it doesn't succeed attempting the installation, either. @techman83 might be able to pick apart what happened.

linuxgurugamer:
@techman83 If you find the problem (if there IS a problem), can you please let me know what the issue is/was?

techman83:
Well it found a folder called gamedata within the game data folder. I should make it report where it found it for easier diagnosis.

https://github.com/techman83/xKAN-meta_testing/blob/master/NetKAN/build.sh#L383

techman83:
And this is what it found. It looks like a mistake in the zip, which should be innocuous.
```
GameData directory found within GameData
Path: dummy_ksp/GameData/./HalfRSS/Planets/Earth/GameData
```

politas:
Hmm. Didn't find that going through the zip files.

techman83:
It's quite buried. The zip the build used has it.

politas:
Oh, yeah, when I say I didn't find it, I meant that I didn't go that deep thoroughly enough. As you say, it's not an issue.

PR #4492, opened 2016-08-05
Add Planet Wiki - OPM Edition from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of jandcando, to add [Planet Wiki - OPM Edition](https://spacedock.info/mod/881/Planet%20Wiki%20-%20OPM%20Edition) to CKAN.

Mod details:
    name = /mod/881/Planet%20Wiki%20-%20OPM%20Edition
    author = jandcando
    abstract = Planet Wiki is an in-game reference sheet full of information on every celestial body in the game, including the Outer Planets!
    license = CC BY-NC-SA 4.0
    Homepage = http://forum.kerbalspaceprogram.com/index.php?/topic/141072-11x-planet-wiki-v3-july-13-2016/
    description = I made a little "mod" that I have wanted in the game for a while now: Planet Wiki, a KSPedia-based mini-wiki that contains useful information regarding the planets such as science multipliers in different situations and the border between "In Space High" and "In Space Low." The kind of stuff you normally have to go to the KSP wiki to find. Here are its pages.

So far the pages are populated with the following information:

A picture of the planet

Pictures of the planet's parent body and moons

The planet's in-game description

The science multiplier in different situations at the planet

The altitude border between different situations at the planet

Orbital Period

The planet's orbital inclination

Semi-major axis needed for synchronous orbit

The phase angle for optimal transfer

delta-v needed to (safely) get to the surface from Kerbin's surface

delta-v needed to land/take off

delta-v needed to escape

Surface gravity in g's

Height of the atmosphere

Number of biomes

The highest and lowest point on the surface, with their coordinates

The Sidereal day length

Efficiency of solar panels on planet's surface (as opposed to Kerbin's surface)

(Moons are included, too!)

Minimum and maximum distances that the planets can be from Kerbin

Advice on the actual design is something I'd be grateful for, since I don't want this to look like an unreadable mess. Please post any ideas below for information you sometimes need to know for your missions that isn't easily accessible/not available in the stock game. All suggestions are welcome!

Please direct questions about this pull request to [jandcando](https://spacedock.info/profile/jandcando).

PR #4493, opened 2016-08-05
Add PlanetWiki-OPM Edition 

politas:
closes #4492


DinCahill:
The author says this works on all KSP 1.1.x. Spacedock lists "1.1.3", so CKAN will restrict it to 1.1.3. What's the best way of getting it on all 1.1.x?

politas:
The best way would be for the author to include a .version file, but other than that, we can either hard code the ksp versions, or use an override. I might look at doing a dependency on OPM that sneaks in an effective ksp min version.

politas:
Override added.

PR #4494, opened 2016-08-05
Added resources and KSP-AVC recommendation

sswelm:


politas:

PR #4495, opened 2016-08-05
Fixed Contract Configurator Requirement

pap1723:


DinCahill:
Is the mod broken without this patch, or is it just clean-up?

pap1723:
It won't install with Contract Configurator through CKAN wihtout the update. It shows as conflicting mods.

DinCahill:
I'm out right now, but I'll look at this later tonight if no one else does.

politas:
Yes, indeed. The highest version released of ContractConfigurator is 1.2.5

PR #4497, opened 2016-08-06
Add MinKSP override to KerbalAlarmClock

### LINKED ISSUE #4283: Some help fixing KAC metadata
Heya,

Have had some issues with people using KSP1.1.2 and the latest version of KAC.

What I need to do is change the version min and max values so KAC 3.6.3 is the last version for 1.1.2 and 3.7 and above is the 1.1.3 and above setup. I think I need to set the max version in 3.6.3 and teh min version in 3.7 and all future ones in netkan and in my .version files going forward.

Can I lean on someone to make the changes to netkan repo's? and do I need to also change any existing .version files in github so they stay right

Dave

politas:
closes #4283

### DIFF
--- a/NetKAN/KerbalAlarmClock.netkan
+++ b/NetKAN/KerbalAlarmClock.netkan
@@ -19,9 +19,18 @@
         { "name" : "TriggerAu-Flags" }
     ],
     "install" : [
-            {
-                "find"       : "KerbalAlarmClock",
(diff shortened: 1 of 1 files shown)


DinCahill:
I've not used `x_netkan_override` before, but it looks right! Thanks. :shipit:

PR #4500, opened 2016-08-07
RN dependency update anim dec fix

raidernick:
-Fix animated deocuplers to mark as compatible with 1.1.3 which it is.
-Remove x86 and x64 versions as he got rid of these once 1.1 dropped,
there is only 1 updated version for x64 and x86


politas:
Can you add new AnimatedDecouplers-x64.frozen and AnimatedDecouplers-x86.frozen files?

raidernick:
@politas did i do that correctly?

politas:
That looks fine to me, but Jenkins doesn't seem to be finding the new netkan. I'll test it when I get home.

politas:

PR #4501, opened 2016-08-07
HullcamVDSContinued

linuxgurugamer:


DinCahill:
Are we supposed to also `provide` HullcamVDS when making a Continued version? Not that we have anything which depends on HullcamVDS. Just asking.

politas:
If it's providing a DLL and directory with the same name as the original, then we should use a "provides".

politas:
Actually, KerbalTubes depends on HullCamVDS, though admittedly, the last version of that is for KSP 1.0.5 so far.

linuxgurugamer:
I'm out of town, will update this tomorrow

PR #4503, opened 2016-08-08
RN add new dependencies

raidernick:
-Added module manager as a dependency to all mods that always needed it
-Added animated decouplers to 2 mods that now use it


politas:

PR #4510, opened 2016-08-11
Update FerramAerospaceResearch.netkan

politas:
Activate staging for future releases as per KSP-CKAN/CKAN#1679


techman83:
@politas was dropping the epoch deliberate?

politas:
No, it wasn't.

techman83:
Cool. LGTM.

PR #4511, opened 2016-08-11
Activate Staging for main SVE mods

politas:
Activate staging as per KSP-CKAN/CKAN#1679


techman83:
LGTM, merging with thanks!

PR #4519, opened 2016-08-16
Set KAX KSP version to match Curse

### LINKED ISSUE #4518: KAX works in 1.1.3
KerbalAircraftExpansion.netkan only allows KAX in 1.1.2 and below, but it works in 1.1.3 (based on an extensive checking process of sticking a few bits on and seeing what happens, ahem, but others report it works in the thread).

politas:
closes #4518


DinCahill:
LGTM. Thanks :shipit:

PR #4526, opened 2016-08-17
Fix Installation for ContractConfigurator-HistoricMissions*

politas:
Discovered these were failing  on http://status.ksp-ckan.org/ Folders have been switched around again. Will probably happen again in the future.


politas:
#rebuild

DinCahill:
I noticed that the Notes.txt in the regular and RSS version says there 500 and 700 contracts respectively. Each one actually appears to have 605 and 606 files in the Missions folder respectively. You've written "over 600" and "over 500" in the netkans.

Other than that, looks good!

DinCahill:
LGTM. Thanks :shipit:

PR #4528, opened 2016-08-18
Fix Installation for MilitaryVehicleChassis

politas:
Discovered this was failing on http://status.ksp-ckan.org/ - Completely messed up zip file structure, this is ugly as heck, and will probably be wrong for the next release. Staging activated.


DinCahill:
Woo, staging! LGTM. Thanks :shipit:

PR #4529, opened 2016-08-18
Fix Installation for KerbalCitiesPack

politas:
This was right when #4464 was checked by Jenkins, but got broken before it was committed.


DinCahill:
LGTM. Thanks :shipit:

PR #4534, opened 2016-08-19
Fix install on MethaneRockets

politas:
Discovered this was failing on http://status.ksp-ckan.org/

### DIFF
--- a/NetKAN/MethaneRockets.netkan
+++ b/NetKAN/MethaneRockets.netkan
@@ -3,14 +3,14 @@
     "$kref": "#/ckan/spacedock/337",
     "identifier": "MethaneRockets",
     "license": "restricted",
-    "release_status": "development",
+    "release_status": "stable",
     "depends": [
         { "name": "RealPlume" },
         { "name": "ModuleManager" }
     ],
     "install": [
         {
-            "file": "MR",
+            "file": "Methane_Rockets",
             "install_to": "GameData"
         }
     ]


DinCahill:
I couldn't install this, because RealPlume needs another package for its config:

```
CKAN.TooManyModsProvideKraken: Too many mods provide RealPlumeConfigs:
RealPlume-StockConfigs v0.10.9
* RealismOverhaul v11.3.1
```

Here's the relevant output from Jenkins:

```
2675 [1] ERROR CKAN.RegistryManager (null) - Loaded registry with inconsistencies:

RealPlume requires RealPlumeConfigs but nothing provides it
- EngineLighting 1.4.5
- MethaneRockets 0.6
- ModuleManager 2.6.25
- RealPlume 2:v10.5.1
- SmokeScreen 2.6.17.0
2640 [1] ERROR CKAN.RegistryManager (null) - Loaded registry with inconsistencies:

RealPlume requires RealPlumeConfigs but nothing provides it
```

How should we deal with this?

politas:
Strictly, that's a problem with RealPlume, not with this mod. We'll have to look into that separately.

DinCahill:
It works if I install RealPlume first and select the the stock config. So we're good to merge this, and then fix RealPlume? I don't know how to do that one :smile:

politas:
RealPlume is fine, nothing needs to be done there as far as I can see.

PR #4525, opened 2016-08-17
Add Pirates of the Keribbean from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of GagaX, to add [Pirates of the Keribbean](https://spacedock.info/mod/902/Pirates%20of%20the%20Keribbean) to CKAN.

Mod details:
    name = /mod/902/Pirates%20of%20the%20Keribbean
    author = GagaX
    abstract = This is a fun mod, it adds parts for creating pirate spaceships. So far, there is only dozen or so parts, enough to create simple spaceplanes and rockets. 
    license = Pirates of the Keribbean by GagaX is licensed under a Creative Commons Attribution-NonCommercial-NoDerivatives 4.0 International
    Homepage = 
    description = 

Where there is space there are space pirates.

This is a fun mod, it adds parts for creating pirate spaceships. So far, there is only dozen or so parts, enough to create simple spaceplanes and rockets. 

Pirates have unique high-tech (magic) propulsion system, so they never run out of fuel (always active fuel generator).

So raise your black flags and off to explore Kerbol system. 

Instalation:

1. extract PoK.zip to GameData folder

2. have fun!

Please direct questions about this pull request to [GagaX](https://spacedock.info/profile/GagaX).

PR #4540, opened 2016-08-20
Add Pirates of the Keribbean

DinCahill:
Closes #4525.


politas:
homepage needs to be wrapped in a resources tag:

    "resources": {
        "homepage": "url"
    },

PR #4544, opened 2016-08-21
CriticalTemperatureGauge mod repo changed

formicant:


politas:
The second commit would have updated the existing PR if you'd left it open, by the way. GitHub is pretty cool that way. This looks all fine to me, and Jenkins is happy, too! 

Merging with thanks!

PR #4545, opened 2016-08-21
Fix RSSExtrasolar installation and details

politas:
Found to be broken via Netkan status page. Quite a few changes to the organisation on this one. Enabled staging


politas:
Strange error Jenkins is giving, there. Testing locally without RSS dependency to avoid the huge installation overhead.

politas:
Looks good to me. Thanks, me from two weeks ago!

PR #4549, opened 2016-08-23
Prep for next release

Olympic1:
Unfreeze for next release


politas:
Given there's only a single release of LTechUnofficial, and there are unlikely to be many people still running on 0.90, I'm not going to hold this up due the identifier change, and will just delete the old one. I very much doubt anyone has been waiting on a new version of this mod before upgrading their from 0.90 to 1.1.3.

Merging with thanks!

PR #4560, opened 2016-08-25
Fix Contares

Olympic1:


politas:
Jenkins seems happy with that and it appears to be adding the required files. Merging with thanks!

PR #4564, opened 2016-08-27
Address TACLS version inconsistencies

phardy:
TACLS has changed hands a couple of times this year, and as a result the versioning for it is a bit of a mess. A quick grep of my up-to-date registry.json shows these versions:

* "v0.12.2"
* "v0.12.1"
* "v0.12.0"
* "v0.11.3"
* "V0.12.3" (note capitalised V)
* "0.12.2"
* "0.12.1"

The most recent release was V0.12.3 a few days ago, but it's getting put well down the ordered version list.

At this point I figured the best approach is to just increase the epoch and strip the V from the start. But would be happy to be corrected.

Bonus sly change: Added current author to authors field.

### DIFF
--- a/NetKAN/TACLS.netkan
+++ b/NetKAN/TACLS.netkan
@@ -1,10 +1,12 @@
 {
     "spec_version"   : "v1.4",
     "identifier"     : "TACLS",
-    "author"         : ["KSP-RO Group/danfarnsy", "Taranis Elsu" ],
+    "author"         : [ "Jplrepo", "KSP-RO Group/danfarnsy", "Taranis Elsu" ],
     "name"           : "TAC Life Support (TACLS)",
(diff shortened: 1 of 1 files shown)


phardy:
Oh, I just realised @politas has started talking to the author about this too. Feel free to reject this then. :)

ahmedcharles:
I'm not sure that the change that JPLRepo is planning will actually fix the issue?

politas:
It's such an easy thing for @JPLRepo to fix by changing the release tag to a lower case "v", that I'd like to hold off on working around it.

JPLRepo:
It has been resolved already. I've spoken to @politas and resolved it and going forward will ensure it follows a compatible format.

politas:
Thanks, @JPLRepo!

PR #4565, opened 2016-08-27
Create AustrianSpaceAgency.netkan

brama0309:


ayan4m1:
Hash and size matches, validates against the JSON schema, looks good to me but I have little experience here.

politas:
Hi @brama0309 ,

Thanks for submitting your mod metadata to the CKAN! What you're doing is how I first got involved in CKAN metadata!

We actually need the original .netkan file that it seems you have written in order to generate this file, not the generated .ckan, and these are the only contents required:

    {
        "spec_version": 1,
        "identifier": "AustrianSpaceAgency",
        "$kref": "#/ckan/github/brama0309/AustrianSpaceAgency",
        "abstract": "Containing Flags for the Austrian Space Agency",
        "license": "MIT",
        "ksp_version": "1.1.3"
    }

All of the other data will be collected by the netkan process, so will get automatically updated as you change things on your GitHub Repo. In fact, you could even drop the `"abstract"` line, and Netkan will use your Repo's readme.md (but that would be slightly different to what you have here).

politas:

PR #4567, opened 2016-08-28
RN soviet rockets update

raidernick:
-Update config for changes to folder structure of mod


raidernick:
@politas it failed because I did not have a released made yet when I created this PR so it could not find the new path in the existing release files. I was planning on sending the release through as soon as this was merged to avoid having people making broken installs.

politas:
Tests fine on the rebuild. Seems to install everything. Merging with thanks!

PR #4570, opened 2016-08-28
Update AustrianSpaceAgency.netkan

brama0309:
Added Mod recommendation

### DIFF
--- a/NetKAN/AustrianSpaceAgency.netkan
+++ b/NetKAN/AustrianSpaceAgency.netkan
@@ -6,7 +6,7 @@
     "author"       : [ "brama0309" ],
     "license"      : "MIT",
     "ksp_version"  : "any",
-    "comment": "flagmod",
+    "comment"      : "flagmod",
     "resources": {
         "homepage": "https://github.com/brama0309/AustrianSpaceAgency"
     },
--- a/NetKAN/AustrianSpaceAgency.netkan
+++ b/NetKAN/AustrianSpaceAgency.netkan
@@ -5,13 +5,24 @@
     "name"         : "Austrian Space Agency",
     "author"       : [ "brama0309" ],
     "license"      : "MIT",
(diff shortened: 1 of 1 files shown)


politas:
CKAN recommends are intended to refer to mods that have some particular synergy. The recommended mod should have a particular interaction with this mod, which I don't believe is the case here. If you wish to add some TextureReplacer assets to your mod, then setting TextureReplacer as a dependency would be appropriate, but while this is just a flags mod, this isn't really appropriate.

brama0309:
I´ve added some Suit Textures with the ASA Flag on them, and for this TextureReplacer is needed, sry for not mention it in the first comment.

politas:
Those would need to be installed in the GameData/TextureReplacer/Suits directory, in that case, and TextureReplacer would need to be in a `"depends"` block rather than `"recommends"`.

brama0309:
I´ve changed it! Thank you for your help and patience!!

politas:
Now you'll need to add the `"install"` instructions:

    "install": [
        {
            "find":"GameData/AustrianSpaceAgency",
            "install_to": "GameData"
        },
        {
            "find":"Suits/AustrianSpaceAgency",
            "install_to": "GameData/TextureReplacer/Suits"
        }
    ]

politas:
Oh, and could you change the ksp_version and add a comment?

    "ksp_version": "any",
    "comment": "flagmod",

politas:
Looking good, but you've slipped back to recommends instead of depends.

politas:
Ah, sorry, forget to mention this change required up the top:

    "spec_version": "v1.4",

politas:
Thanks for your patience! Merging now

brama0309:
Thank you!

Am 28.08.2016 23:56 schrieb "Myk" <notifications@github.com>:

PR #4571, opened 2016-08-28
VesselView

linuxgurugamer:
I'm taking over this mod, updating the file so existing installs will work


DinCahill:
Nice! LGTM :shipit:

PR #4583, opened 2016-08-29
Swag 7

Thomdare:
Adds APModpack.netkan

### DIFF
--- a/NetKAN/APModpack.netkan
+++ b/NetKAN/APModpack.netkan
@@ -1,0 +1,26 @@
+{
+    "spec_version": 1,
+    "identifier"  : "APModpack",
+    "name"        : "Airship Pirates KSP Modpack",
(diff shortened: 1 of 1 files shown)


politas:
My understanding is that BDDMP.dll is/was a replacement for the BDArmory.dll to work with DMP?

The old BDDMP mod included a modified version of BDArmory. Has this been rewritten to now work in combination with BDArmory, rather than replacing it?

I realise that this is a pack of mods that everyone needs in order to play together, have you considered adding conflicts with mods known to cause problems, or considered to be cheating? It might be worth adding a conflicts/provides to BDDMP, at least, in case anyone gets confused and thinks they need it.

Thomdare:
@politas Yes, however, the new BDDMP.dll is outdated and I simply needed a mod as the core mod. Since this one hasn't been officially updated for a long time, I just used it.

politas:
Merging with thanks!

PR #4594, opened 2016-09-05
Latest TACLS update moved config. Fixes #4593

### LINKED ISSUE #4593: Realism Overhaul and TACLS
First, apologies for not being able to do more at present, and therefore begging you all to do so. :)
(1.2 is OMGBUSY atm)

TACLS recently updated and changed the location of its config file. RO also moved its file. However, neither Config netkan entry (those which provide TACLS-Config) were updated, causing conflicts.

Can someone unfreeze the TACLS-stock config and point it at the new correct place? And put the TACLS netkan back to requiring a config? (I'm not sure why that was _ever_ dropped). The RO TACLS config netkan appears to still be correct.

Olympic1:
Fixes the problem where RO tries to overwrite the TACLS config.

I'm not good with `filter_regexp` so I filtered it my way that also works.

Fixes KSP-RO/RealismOverhaul/issues/1368 and #4593


Olympic1:
Well this will keep failing because it ignores the config setup we're using.

Olympic1:
Tested this on three different PC's using the metadata from my own repo and worked three times.

NathanKell:
:heart:

politas:
This seems unnecessarily complex, given that we're only really talking about a single file, LifeSupport.cfg

PR #4597, opened 2016-09-06
TACLS-Config Simplify installs

politas:
Simplifies #4594 
Not entirely sure about the TACLS filter, though. Let's see what Jenkins says

### DIFF
--- a/NetKAN/TACLS-Config-Stock.netkan
+++ b/NetKAN/TACLS-Config-Stock.netkan
@@ -1,5 +1,5 @@
 {
-    "spec_version"   : "v1.4",
+    "spec_version"   : "v1.16",
     "identifier"     : "TACLS-Config-Stock",
     "author"         : ["KSP-RO Group/JPLRepo", "Taranis Elsu" ],
     "name"           : "TAC Life Support (TACLS) - stock config",
@@ -13,8 +13,9 @@
     "conflicts"      : [ { "name" : "TACLS-Config" } ],
     "install" : [
         {
-            "find"          : "ThunderAerospace/TacLifeSupport/LifeSupport.cfg",
-            "install_to"    : "GameData/ThunderAerospace/TacLifeSupport"
+            "find"               : "ThunderAerospace/TacLifeSupport/LifeSupport.cfg",
+            "find_matches_files" : true,
+            "install_to"         : "GameData/ThunderAerospace/TacLifeSupport"
         }
     ]
 }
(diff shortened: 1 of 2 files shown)


politas:
Ok, I don't have a clue what's happening. I give up.

NetKAN/TACLS-Config-Stock.netkan
```diff
@@ -13,10 +13,8 @@
     "conflicts"      : [ { "name" : "TACLS-Config" } ],
     "install" : [
         {
-            "find"          : "GameData/ThunderAerospace/TacLifeSupport",
-            "install_to"    : "GameData/ThunderAerospace",
-            "filter"        : [ "PluginData", "Textures", "MiniAVC.dll", "TacLifeSupport.dll", "TacLifeSupport.version" ],
-            "filter_regexp" : "MM_*"
+            "find"          : "ThunderAerospace/TacLifeSupport/LifeSupport.cfg",
+            "install_to"    : "GameData/ThunderAerospace/TacLifeSupport"
         }
```
techman83: This needs:
```
	    "find_matches_files" : true
```

NetKAN/TACLS.netkan
```diff
@@ -16,9 +16,9 @@
     ],
     "install" : [
         {
-            "find"          : "GameData/ThunderAerospace",
-            "install_to"    : "GameData",
-            "filter_regexp" : "LifeSupport\\.cfg$"
+            "find"       : "GameData/ThunderAerospace",
+            "install_to" : "GameData",
+            "filter"     : "LifeSupport/LifeSupport.cfg"
         }
```
techman83: Also looks like this will need to be
```
"filter"     : "TacLifeSupport/LifeSupport.cfg"
```

techman83:
LGTM, merging with thanks!

politas:
Hmm. That should not be installing `GameData/ThunderAerospace/TacLifeSupport/LifeSupport.cfg` but it is.

techman83:
I reckon you probably need a full path on that filter. There should be only one LifeSupport.cfg in the zip I would expect, so should filter just the file without the path. Or use a filter_regexp.

Olympic1:
Ah, I forgot to add `"find_matches_files" : true,`, that's why it didn't work for me and why I added all those filters. I was busy for a long time (and still are), so I don't now every change that is added to CKAN.

politas:
Yeah, unfortunately there are multiple LifeSupport.cfg files in the zip.

PR #4599, opened 2016-09-06
Move SovietConversion to Spacedock

politas:
Whole bunch of changes and reviving the old identifier
Closes #4459


politas:
#rebuiild

PR #4601, opened 2016-09-06
Add KSP version override for ABLaunchers

politas:
Forum thread describes the current release as being ok for all of KSP 1.1
Will fix dependency required by #4599


Olympic1:
LGTM, merging

PR #4608, opened 2016-09-09
Update CrowdSourcedScience to be 1.1.x-compatible

EoD:
As CrowdSourcedScience is bascially compatible with almost all versions of KSP, this fix should be safe. 

See also https://github.com/DuoDex/CrowdSourcedScience/issues/49 .


EoD:
I am not sure why the checks are failing

PR #4609, opened 2016-09-09
Update CrowdSourcedScience to be 1.1.x-compatible

EoD:
As CrowdSourcedScience is bascially compatible with almost all versions of KSP, this fix should be safe. According to the changelog this is 1.0.2 until 1.1.x

See also DuoDex/CrowdSourcedScience#49 .


politas:
Seems fair enough. Merging with thanks!

Issue #4607, opened 2016-09-08
BDArmory - New developer team


Olympic1:
**BahamutoD** has taken a leave of absence to attend to real life, so a new team has taken over most/all of his mods. I ll update the netkans for the continued versions.

[New thread](http://forum.kerbalspaceprogram.com/index.php?/topic/155014-bda)
- [x] BDArmory
- [x] VesselMover
- [x] VesselSwitcher
- [x] Destruction Effects
- [x] BDAnimation Modules
- [x] Burn Together

**These mods are still being worked on or postponed:**
- [ ] Improved Chase Camera **(pre-release)**
- [x] BDynamics Mk22 Cockpits **(pre-release)**
- [x] Adjustable Landing Gear **(After release of KSP 1.2)**
- [x] Camera Tools **(WIP)**
- [ ] Baha Parts for Extraplanetary Launchpad **(WIP)**

Olympic1:
Leave this open after PR is merged to remind me/us to check this after 1.2

politas:
Would you like to talk to PapaJoe about releasing that BTforCommandModules.zip from BurnTogether as a separate download?

Olympic1:
Sent a PM regarding this

Olympic1:
PapaJoe is willing to rename the MM config to a `.txt` and add an installation doc, but will not include the config directly.

politas:
That's confusing. What was his reasoning? I didn't mean putting the cfg file directly in BurnTogether, I meant taking that zip file and making it a separate mod entirely.

Olympic1:
Thats what I said too but he wants the releases to be the same without many changes.

> I want to keep the distribution as it is because if Bahamuto comes back he can continue on the update without many changes.

politas:
Hmm. I don't think our "as" keyword will rename a file, unfortunately. I believe it's just for renaming a directory.

PR #4611, opened 2016-09-10
BahamutoD (BDA) Continuations

Olympic1:
Reference: #4607


politas:
These all seem to work. Excellent work! Merging with thanks!

Issue #4620, opened 2016-09-14
NetKAN curating


rkagerer:
This is my first issue opened for NetKAN; please go easy on me.  I would like to suggest some modifications to the master "database" of metadata, as well as submit a couple of mods for your consideration.

1) Please update LanderControl mod to indicate that it is compatible with KSP 1.0.5 through 1.1.3.  It is a simple part, and I have used it in all versions since 1.0.5 with no issues.  Here's the relevant file: https://github.com/KSP-CKAN/CKAN-meta/blob/8a218c712846babb7ac1e658c4c27ffa83021ec6/LanderControl/LanderControl-1.0.1.1.ckan

Also note [Curse indicates](https://mods.curse.com/ksp-mods/kerbal/238277-lander-orientation-improvement#t1:changes) it supports 1.1.3.

2) There's a more recent version (2.3.3) of @Alexw's and @girka2k's UbioWeldContinued at https://github.com/UbioWeldingLtd/UbioWeldContinued/releases which is more up to date than the 2.3.2 one in CKAN.  Also I propose this mod should be marked as KSP 1.1.3-compatible (it is currently marked with 1.1.2).  I have been using it extensively for quite some time in 1.1.3, and while there are a few rough edges (which I'm convinced are equally present regardless of whether you're on KSP 1.1.2 vs 1.1.3) it works well enough that I think it should be marked as compatible.

3) The LightsOut mod on CKAN is somewhat out of date.  While the mod owner @nodrog6 has not published updates in some time, there have been fixes by the community since it was released.  For details see the last few pages of its [forum thread](http://forum.kerbalspaceprogram.com/index.php?/topic/102558-105-lightsout-v015-daynight-mode-in-vabsph/&page=4).

I am happy to submit my fork of it for inclusion in CKAN.  Here is my CKAN file: https://github.com/rkagerer/KSP-Fwiffo-Repository/blob/master/CKAN-meta-fwiffo/LightsOut-Fwiffo/LightsOut-Fwiffo-v0.1.5.1.ckan

4) I'd also like to submit the following skybox mod.  More information [is here](http://forum.kerbalspaceprogram.com/index.php?/topic/78778-raredens-real-8k-skybox-for-texture-replacer/&do=findComment&comment=2508959).  Here is my CKAN file; please note the download link in it should be changed from "rawgit.com" to "cdn.rawgit.com" and status can be changed from "development" to "stable":
[...]

PR #4624, opened 2016-09-14
Add FwiffoRaredenSkybox

politas:
cdn.rawgit.com leads to a corrupt file after a very long period (I think curlsharp must be timing out) rawgit.com URL works fine, though.
ref #4620


rkagerer:
Sorry is the issue with "cdn.github.com" or "cdn.rawgit.com"?  Assuming the latter has no known issues, could we make it that, please?

politas:
Sorry, it was cdn.rawgit.com that had the problem.

Olympic1:
Downloads and installs fine, merging with my thanks

PR #4643, opened 2016-09-17
Add PartOverhauls.netkan

benjwgarner:
Commit e1a1575 has already been merged, so I'm not sure why it's in this pull request, but it should merge just fine because it's identical.

PR #4645, opened 2016-09-17
Add PartOverhauls.netkan

benjwgarner:
The problems with #4643 have been resolved.

I am hosting the .zip as the original .zip was 15 times larger because it contained the source. It also seemed to be throttled to 72KB/s, so it took way too long to download. See link for original source:

http://forum.kerbalspaceprogram.com/index.php?/topic/147582-update-12-pre-release-is-here

### DIFF
--- a/NetKAN/PartOverhauls.netkan
+++ b/NetKAN/PartOverhauls.netkan
@@ -1,0 +1,29 @@
+{
+    "spec_version"   : 1,
+    "name"           : "Part Overhauls",
+    "abstract"       : "Porkjet's in-progress part overhauls",
+    "identifier"     : "PartOverhauls",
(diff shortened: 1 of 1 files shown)


benjwgarner:
Weird. It worked with `netkan.exe --verbose`, but Jenkins seems to think that it includes over 100 commits by other contributors.

Also, I've now encountered an even less useful error message than "unspecified error". Jenkins says:
"No problems were identified. If you know why this problem occurred, please add a suitable Cause for it." 😜

politas:
The trouble you're having now is that you've specified a `ksp_version_min`, but no `ksp_version_max`. If you are pretty sure this will stay working throughout the 1.2 KSP series, please set `"ksp_version_max: "1.2.99"`  Otherwise, change `ksp_version_min` to just `ksp_version` to set compatibility to just the the current release.

PR #4665, opened 2016-09-28
Create FMRS.netkan

jacktose:
Based on FMRS.frozen and netkan tutorial template.
Makes a good .ckan and installs from GUI and CLI, in my test.
Should handle KSP 1.1.x.


jacktose:
Looks like it failed because there's an FMRS.frozen.
Maybe I'm not understanding the system. Is it not good to replace that with a working .netkan? Do I need to remove the .frozen in my PR?

jacktose:
I guess that did it. This PR should be ready, unless someone can tell me I got it all wrong.

Olympic1:
This works and installs fine. Merging with my thanks

PR #4683, opened 2016-10-11
Update USI mods

Olympic1:
Remove deprecated mods
Add Konstruction

Looks like github wants to rename the file instead of creating a new one. Lazy github 💤


politas:
Odd, but looks fine. Merging with thanks!

PR #4685, opened 2016-10-12
RCSBuildAid 0.9

m4v:


politas:
We actually have automated updating for Curse now. Would you like to switch over to that, @m4v?

m4v:
@politas sure, lets try it.

PR #4690, opened 2016-10-12
ExperimentTracker for 1.2

derrrey:


politas:
This is not actually required. CKAN is getting the KSP version from Curse correctly, and the old homepage URL will continue to work.

politas:
We do need a force_v though, since "1.3.1" is less than "v1.3"

PR #4707, opened 2016-10-13
Updated depandancies for KSP 1.2

sswelm:


sswelm:
I don't understand what the problem is ...

techman83:
At a guess, tweakscale not being available for 1.2
```
4931 [1] ERROR CKAN.RelationshipResolver (null) - Dependency on TweakScale found but it is not listed in the index, or not available for your version of KSP.
Module TweakScale required but it is not listed in the index, or not available for your version of KSP.
```

sswelm:
But where can I view this relationship resolver?

Olympic1:
Ignoring tweakscale, it installs fine, merging

sswelm:
Great, but KSPI is still not avialable on CKAN, I guess there is still a dependancy check which prevent is from showing up

Olympic1:
If tweakscale is update it will show up. I see that @pellinor0 is working on it.

sswelm:
true

PR #4708, opened 2016-10-13
RN updated for ksp 1.2

raidernick:
-Update all mods for ksp 1.2
-Removed all unneeded dependencies, I now only use bahaSP and sometimes
texture replacer for my mods. Firespitter and animated dec are no longer
needed.
- Updated all craft files so they can be installed by default again

### DIFF
--- a/NetKAN/RN-USProbesPack.netkan
+++ b/NetKAN/RN-USProbesPack.netkan
@@ -4,14 +4,14 @@
     "name"          : "US Probes Pack",
     "author"        : "Raidernick",
     "abstract"      : "This pack will include pretty much all probes ever launched by NASA and the US military. It is a long term project and I will be adding to it slowly over time.",
-    "comment"       : "Ships explicitly not installed by author request (he no longer maintains them)",
+    "comment"       : "Ships installed again",
     "resources": {
         "homepage"  : "http://forum.kerbalspaceprogram.com/threads/93389"
     },
     "$kref"         : "#/ckan/github/KSP-RO/USProbesPack",
     "license"       : "CC-BY-NC-ND-3.0",
     "ksp_version_min"   : "1.1.1",
-    "ksp_version_max"   : "1.1.3",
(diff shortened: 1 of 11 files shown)


raidernick:
@politas this may fail to validate as smokescreen and texture replacer aren't officially updated yet. They are both not REQUIRED for my mods to work, only for pre-launch smoke effects and some reflections on probes. The mod will function perfectly fine without them.

politas:
If they aren't required, you could move them into "recommends" rather than "depends". Otherwise, your mods will disappear from the CKAN list when this is merged.

raidernick:
@politas I meant the mods will "work" without the plugins, but for the probes it is essential that they are reflective, like the sputnik, it is not meant to be matte. I could change it, but I don't want to give the option for the users to disable installing it. Smokescreen I will consider. Can you give me a few days to think about this before merging it. I'll ping you again on here when/if I decide to make changes and have it merged.

politas:
Can't see a reason not to move TextureReplacer to another maintainer if the original author is MIA

raidernick:
@politas I made the changes to TR and changed smokescreen to recommended. Let me know if that looks good. If all is acceptable and this will allow my mods to show up for ksp 1.2, this is good to merge.

NetKAN/RN-USProbesPack.netkan
```diff
@@ -26,8 +26,10 @@
     "depends": [
         { "name": "BDAnimationModules" },
         { "name": "TextureReplacer" },
-        { "name": "FirespitterCore" },
-		{ "name": "ModuleManager" },
-        { "name": "SmokeScreen" }
+		{ "name": "ModuleManager" }
+    ]
```
Olympic1: Forgot a comma here

raidernick:
@Olympic1 added it

Olympic1:
Thx. Jenkins loves it, everything installs fine locally, so merging with my thanks

PR #4709, opened 2016-10-14
EVE 1.2 maintainer move

### LINKED ISSUE #4684: EVE for KSP 1.2
The original author is AFK, so I have temporarily (and unilaterally) taken responsibility for maintenance releases. The release for KSP is here:

https://github.com/WazWaz/EnvironmentalVisualEnhancements/releases/tag/EVE-1.2-1

There are now a number of EVE config mods (and probably more on the way). If there is anything that I could do to smooth the way users choose which configs to use, please let me know. All too often in the past, users ended up with both the simple default configs *and* other configs (eg. SVE), which is of course a pain for everyone. Possible CKAN already handles this, in which case, bravo!

politas:
Closes #4684


Olympic1:
LGTM

PR #4724, opened 2016-10-14
kerbalLaunchFailure

linuxgurugamer:


Olympic1:
`KerbalLaunchFailure` matches an existing frozen identifier: [NetKAN/KerbalLaunchFailure.frozen](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/KerbalLaunchFailure.frozen)

PR #4742, opened 2016-10-15
Support KSP 1.2 for PreciseManeuver

MarkSRobinson:


Olympic1:
CKAN does not support pre-releases. Some are stable, some not, but in the case of the latter people get mad that their mod isn't working.

PR #4759, opened 2016-10-16
Update PreciseNode.netkan

BasharMilesTeg:
This results in a 4 digit version (1.2.4.0) rather than the previous 3 digit (1.2.3)


politas:
Here's me thinking I knew all the ways to make a .netkan file work!  Looks good, merging with thanks!

PR #4762, opened 2016-10-16
Add CommunityCategoryKit

### LINKED ISSUE #4761: USI Life Support Update
Good morning. Tried to do an update this morning. All of them worked fine accept for the update to USI Life Support Update.
I got the following error message:
"Module CommunityCategoryKit required but it is not listed in the index, or not available for your version of KSP."

Olympic1:
A new mod of BobPalmer and JPLRepo

Closes #4761

### DIFF
--- a/NetKAN/CommunityCategoryKit.netkan
+++ b/NetKAN/CommunityCategoryKit.netkan
@@ -1,0 +1,7 @@
+
+    {
+        "spec_version" : "v1.16",
+        "identifier" : "CommunityCategoryKit",
+        "$kref" : "#/ckan/netkan/https://raw.githubusercontent.com/BobPalmer/CKAN/master/CommunityCategoryKit.netkan",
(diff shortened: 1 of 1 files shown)


Dazpoet:
Seems to be doing what it's supposed to but it only lists 1 author while your PR mentions two of them.

Dazpoet:
So now that this is showing up in CKAN I was very confused to see it being named Community Resource Pack rather than what it should be called...

Looking back at the log I also see that it was acctually printed out incorrectly. However I don't know why since the [source-material](https://github.com/BobPalmer/CKAN/blob/master/CommunityCategoryKit.netkan) lists the right name and the bot [generates the correct identifier](https://github.com/KSP-CKAN/CKAN-meta/commit/9bafe4c82a4c32c2eba3870a2a9d2b542dd1f6cf) aswell...

PR #4765, opened 2016-10-16
Updated/added metadata for my mods.

allista:
Namely:

* Updated AT-Utils
* Updated ThrottleControlledAvionics
* Added ConfigurableContainers
* Defrosted and updated Hangar


Olympic1:
LGTM, merging with my thanks.

@allista Are you maybe interested in maintaining your mods on your own repo, like the USI mods do. This way you could easy change things around.

allista:
**I have no objection** to the idea if you deem it beneficial. It just never occurred to me. 

Though today's experience with building errors shows that local a check with `netkan+ckan install -c` routine is somehow not enough to catch everything. I.e. this way I was able to install everything locally before I've made this pull request.

Also, I don't expect frequent changes to these files, and here I have an established routine.

Olympic1:
No problem, use what you finds the best way to do things, was just suggesting.

PR #4712, opened 2016-10-14
Fix SVE

Olympic1:
Testing these using `netkan.exe` downloads v1.0.83 instead of v1.1.1. Lets see if Jenkins does it too

Closes #4711

PR #4768, opened 2016-10-17
Scatterer Sunflare configuration has moved

politas:
SVE Sunflare was fixed in #4712, but the original and Astroniki are broken. 
Scatterer users will likely have to uninstall/reinstall to move file ownership.


Olympic1:
LGTM

PR #4776, opened 2016-10-18
Add temporary conflicts for KEI

plague006:
See commit comments for backstory.


Olympic1:
Oh, thought I pushed that already. Apparently no, merging

PR #4779, opened 2016-10-18
Revert SVE to old regex

Olympic1:
Switched back to the old kref regex.

Question: Is there a way to include regex inside a vref, because with this way we'll always need update it with every release. If it isn't possible, I'll switch back to using `ksp_version`.


politas:
It shouldn't be too difficult to code a way to use regex in $vref. Given that the ability to code a file location in the $vref was never documented, I'm not even sure whether it might already be there.

politas:
I'll merge this, can you open a new issue for the regex question? I don't have the time to investigate it now.

PR #4780, opened 2016-10-18
TextureReplacer - Filter skybox

Olympic1:
Filter the EnvMap of TR and seperate into a selectable skybox.


Olympic1:
Poodmund:
> Please be aware @Olympic1 that RangeMachine is removing these files from his future releases of TR: https://github.com/RangeMachine/TextureReplacer/issues/1#issuecomment-254634616

PR #4794, opened 2016-10-20
Update Kopernicus.netkan

dbent:
[Kopernicus updated for KSP 1.2.0](http://forum.kerbalspaceprogram.com/index.php?/topic/140580-120-kopernicus-release-2-october-20/).

Might want to add any mod with manually set `ksp_version` to the staging queue so new versions don't get generated for the wrong KSP versions.


politas:
That's good thinking. Want to add that to this one?

Olympic1:
Add staging

PR #4796, opened 2016-10-21
Change IndicatorLightsCommunityExtensions relationships from recommends to supports

dbent:
I think `supports` is the more appropriate relationship here over `recommends` since users installing this mod may not even care or be interested in any of the mods it supports except maybe one or two. It's basically a set of compatibility patches, it doesn't encourage using any particular mod.


politas:
LGTM, Merging

PR #4797, opened 2016-10-21
Add name to ConfigurableContainers-Core

dbent:
To differentiate it from `ConfigurabableContainers`.


politas:
If this is the core mod, with different configurations possible, would it not be better to have the .version file in this one with the $vref, and set the "config" mod to ksp_version "any"?

PR #4801, opened 2016-10-21
Updated Extraplanetary Launchpads to 1.2.0

bennie98:


Olympic1:
You need to add a `ksp_version` if you're using `dowload`.

@politas, @ayan4m1, We should also check for `vref`s if no `ksp_version` is given

bennie98:
I did verify this using netkan and it did produce a correct ckan file with the correct ksp version and mod version.

Olympic1:
That is true, but Jenkins is more stubborn then netkan.exe

bennie98:
So what is line should I add? 
ksp_version, version or both?

Olympic1:
"ksp_version": "1.2.0",

bennie98:
This should fix it.

Olympic1:
Yep, merging

PR #4802, opened 2016-10-21
Fix far forum link

bennie98:


politas:
Should do the job. Merging with thanks!

PR #4803, opened 2016-10-21
Update RealisticAtmospheres

dbent:
- Switch to using `curse` `$kref`.
- Remove auto-populated properties.
- Add override for latest version for KSP versioning.


Olympic1:
LGTM, merging

PR #4804, opened 2016-10-21
RCSBuildAid switch to Auto-updating

politas:
Ref #4685 
Ping @m4v - With this change, as long as your AVC .version file is correct, CKAN should stay updated.


m4v:
> as long as your AVC .version file is correct

What needs to be correct? version data? download link?

politas:
The KSP version data is the only thing that the Netkan inflater robot uses, from my understanding .

PR #4805, opened 2016-10-21
Updated Mod requirements for KSPI-E

sswelm:


sswelm:
failed because: 4600 [1] ERROR CKAN.RelationshipResolver (null) - Dependency on CommunityTechTree found but it is not listed in the index, or not available for your version of KSP.

well, CommunityTechTree has just updated for KSP 1.2, CKAN doesn't appear to have updated yet.

sswelm:
Untill CTT get's properly updated, I have removed the minimum version requirement, I will update it later

sswelm:
still the same problem :(

sswelm:
Finally, we are all green!

PR #4807, opened 2016-10-21
Reduce ProbesPlus dependencies

puetzk:
The upstream zip includes the FireSpitter plugin and resources,
and the DMagic animation plugin, but not their respective parts;
these are apparently the hard dependencies

DMagicOrbitalScience is needed "for some science experiments to work",
but much of the mod functions well without it; demote to recommends.

The firespitter parts do not seem relevant.


PhineasFreak:
I was about to make that suggestion to the KSP forum CKAN thread! Also, ProbesPlus requires only the FSTextureSwitch2 module so the FirespitterResources dependency can easily be removed.

politas:
@puetzk, do you want to remove the FirespitterResourcesConfig dependency as well?

puetzk:
I kept the dependency on FirespitterResourcesConfig  bcause the upstream CA_ProbesPlus_BETA_0.14.1.zip did include GameData\FireSpitter\Resources\FireSpitter.cfg as well as the plugin.

I don't know **why** they bundled it (looking closer, FSCoolant doesn't seem to be referenced from any of the parts), but leaving it made the CKAN "depends" list match what the author bundled. 

I think @PhineasFreak should probably raise that question with the author on the mod's [forum thread](http://forum.kerbalspaceprogram.com/index.php?/topic/131145-wip-coatl-aerospace-probesplus-dev-thread-v0141-beta-102016/) or [github repo](https://github.com/raveloda/Coatl-Aerospace/commits/master/GameData/Firespitter). But I don't think CKAN should drop it unless @raveloda removes the resources from the release.

PhineasFreak:
OK, on it!

raveloda:
The Firespitter usage has not changed in my mod. Should be the same as it was last time, with just the DLL. I added the resources because I thought Firespitter needed it. I included Firespitter "as is" with no modifications as to not violate any license.

politas:
Excellent! Merging with thanks!

PR #4812, opened 2016-10-22
SVE-Terrain now has multiple downloads

Olympic1:
Freezing `SVE-Terrain.netkan` and adding low and high resolution downloads.


politas:
Looks good! Merging

PR #4816, opened 2016-10-23
Mod’s owner changed

formicant:


SirDiazo:
Confirming transfer of ownership.

politas:
Needs something to tell CKAN which KSP version is supported. Either explicit "ksp_version", a range of  "ksp_version_min" to "ksp_version_max", or a $vref if you include an AVC .version file

formicant:
I added a .version file.

politas:
Excellent! Merging with thanks!

PR #4818, opened 2016-10-24
Fix link for WasdEditorCameraContinued

plague006:


politas:
I think a slash is the only character that doesn't work, now. Merging with thanks!

PR #4845, opened 2016-10-28
Update ModularFuelTanks.netkan

Dazpoet:
User GeekArt came onto irc asking for this to be updated. Based on #3890 I assume this is how we're supposed to do it now?

I think I poked the right parameters :)


politas:
Looks about right to me. Merging with my thanks!

PR #4847, opened 2016-10-28
RealisticRemodel1x Fix install

politas:
Dependency on SigmaDimensions will prevent this from appearing, but that should return at some stage


Olympic1:
That was why I created #4718 but was never merged

politas:
Must have missed that one, sorry.

PR #4854, opened 2016-10-30
Make VensStylePPTextures KSP_version "any"

politas:
Its functionality is far more dependant on Procedural Parts versions, all of which it will work for currently.


Olympic1:
LGTM, merging

PR #4856, opened 2016-10-30
Update AsteroidDay

Olympic1:
Use curse `kref` instead of just `download`.

### DIFF
--- a/NetKAN/AsteroidDay.netkan
+++ b/NetKAN/AsteroidDay.netkan
@@ -1,5 +1,5 @@
 {
-    "spec_version": 1,
+    "spec_version": "v1.10",
     "identifier": "AsteroidDay",
     "$kref": "#/ckan/curse/232196",
     "abstract": "Squad partnered with the B612 Foundation to bring you this mod which will give you four new parts, a new experiment, and a unique contract.",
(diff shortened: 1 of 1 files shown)


politas:
I've noticed a problem with the Curse krefs for some mods - the version numbers aren't being handled properly. This seems to be one of them. It looks like Curse doesn't actually have version numbers explicitly, and if the download zip and the mod name are the same, we get what looks like a random number. LanderControl is one example, and KerBalloons does the same thing. It was doing up KerBalloons that I noticed, so I stopped working on that change. This seems to be another example (it's trying to make version 98).

Olympic1:
Added a version number but the dowload is corrupted, works fine locally though

politas:
Won't that force the same version number for every release?

Olympic1:
Yes, used override to fix it.

politas:
98 is just the last three digits of one part of the download URL (098). I think we should stick to manually updating this mod and make the new release version 1.2, to match what we've done previously. Otherwise, we're going to have to put an epoch on it half the time it updates. Ditch the kref.

Olympic1:
Done

politas:
LGTM!

PR #4858, opened 2016-10-30
Create KerbalAtomics-NFECompatibility.netkan

lOmicronl:
Created on request of a user. Sorry if it doesn't work right away, I did most of this from memory because I'm really pressed for time. Will fix problems in the next few days if there are any.

### DIFF
--- a/NetKAN/KerbalAtomics-NFECompatibility.netkan
+++ b/NetKAN/KerbalAtomics-NFECompatibility.netkan
@@ -3,6 +3,7 @@
     "identifier": "KerbalAtomics-NFECompatibility",
     "$kref": "#/ckan/spacedock/710",
     "name": "Kerbal Atomics - NFE Compatibility Patch",
+    "author": "Nertea",
     "abstract" : "Optional, experimental patch that makes Kerbal Atomics nuclear engines utilize the Near Future Electrical reactor plugin.",
     "license": "CC-BY-NC-SA-4.0",
     "depends": [
@@ -10,6 +11,6 @@
         { "name" : "NearFutureElectrical" }
     ],
     "install" : [
-        { "find" : "NearFutureElectricaNTRs", "install_to" : "GameData/Extras" }
+        { "find" : "NearFutureElectricaNTRs", "install_to" : "GameData/KerbalAtomics/Extras" }
     ]
 }
--- a/NetKAN/KerbalAtomics-NFECompatibility.netkan
+++ b/NetKAN/KerbalAtomics-NFECompatibility.netkan
@@ -1,0 +1,15 @@
+{
+    "spec_version": "v1.4",
+    "identifier": "KerbalAtomics-NFECompatibility",
+    "$kref": "#/ckan/spacedock/710",
+    "name": "Kerbal Atomics - NFE Compatibility Patch",
+    "abstract" : "Optional, experimental patch that makes Kerbal Atomics nuclear engines utilize the Near Future Electrical reactor plugin.",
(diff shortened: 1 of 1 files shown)


Olympic1:
@lOmicronl could you update the PR follwing the requested changes

lOmicronl:
Neither change is necessary. The folder structure as I have written it reflects the structure created during manual installation today. The "author" field should be auto-populated via Spacedock, as far as I remember, and older Near Future metadata has it as a left-over from times where this didn't work (Kerbalstuff went down etc.)

I mean, you're free to change it if you want, but I see no reason to.

It's quite unfortunate that Fraz had to doublepost here, when this question has already been solved in PM on the KSP forums more than a week ago.

Olympic1:
Well you have to think for other users too. If I would see a folder called `Extras` I don't know which mod that is from untill I check it, Installing it into `KerbalAtomics` lets me know where it is from.

lOmicronl:
...I'm at a loss for words right now, to be honest.

You know, I've been doing a lot of CKAN stuff for Nertea's mods in the past. I did so voluntarily, even though Nertea didn't support CKAN installs (and still doesn't today). And I'm having trouble recalling any pull requests I made that didn't have someone come in and ask for corrections. This is all fine and dandy if I'm actually making a mistake. The spec is like an arcane tome of strange symbols that changes everytime you look away, of course I will get it wrong now and then. But what's with this completely inconsistent caviling?

Once I was told (fairly bluntly, too) to stop putting in redundant fields; then, a bit later, I was told to stop removing redundant fields, because once something is in the metadata it should stay there for some unspecified reason. Now you ask me to *include* a redundant field? Once I was told that the entire point of CKAN was absolute consistency in installing mods, and that I shouldn't get creative with install instructions. Now you ask me to deviate from the folder structure that the mod author ships in his download, because a user might potentially disagree with how it was chosen? Once I was told to update the metadata for a mod, only to find that, by the time I had drafted up the changes and went to upload them, the exact same person who had asked me to do it, had already gone and implemented the changes themselves (without notifying me of course). And now you keep a pull request open for weeks, unwilling to pull it in before I make some uttermost trivial change that anyone with a Github account, even completely unaware of the CKAN spec, could have done in half a minute?

I just don't get it. Sure, I'll go and change the install path so it is more pleasing to the eye, just give me a minute. But I hereby give up trying to understand what you guys want.

Olympic1:
I greatly appreciate your help with updating Nertea's metadata. I agree, the spec does change every now and then which lets people make mistakes because of those changes. Even people who have been updating metadata for a long time can make mistakes if they are not aware of the spec changes.

I can't apologize for the person who bluntly told you to stop adding/removing reduntant fields but adding redundant fields isn't bad, they're just needed if the information where scraping from doesn't match from the forums or author.

Regarding the `Extras` folder: Nertea shipped this folder seperate so that people who don't want these functions don't have to worry about them. The meaning of that folder is to install it inside the mod folder (in this case KerbalAtomics) so that it doesn't stay in the `GameData` folder when KerbalAtomics is removed. That's why I asked you to change the `install_to` path.

fraz86:
I apologize for any inconvenience, Omicron. I actually posted here first, before it occurred to me to PM you. Being relatively unfamiliar with github, I wasn't sure what effect my post would have (if any), or if you would even see it here, thus why I decided to send the PM. I wasn't trying to be redundant.

Likewise, it was not my intent to nitpick over trivial issues, but rather to simply note differences upon comparison to patches for the other extras patches (I had no idea whether or not the differences were important).

PR #4862, opened 2016-11-01
update QuickMods with a new directory

malahx:
Hello, I've changed the default directory of all my QuickMods. I want to know if it will break some things on CKAN, thanks ;)


Olympic1:
Every spec needs to be changed to `"spec_version" : "v1.2",`

Olympic1:
Thx, for updating. Merging with my thanks

PR #4874, opened 2016-11-03
DSD Mobile Labs

MrKiel:
DSD Mobile Labs is now being maintained by me, with endorsement from the original author JDCollie.

PR #4879, opened 2016-11-03
Update to original DSD Mobile Processing Labs

MrKiel:
As Discussed in https://github.com/KSP-CKAN/NetKAN/pull/4874

Modified original netkan instead of creating a new one (my bad, sorry).


MrKiel:
Formatting Error

PR #4881, opened 2016-11-03
Update DSD Mobile Labs

MrKiel:


politas:
Please maintain the "DSDMobileProcessingLabs" identifier and filename, this marks the mod as a continuation of the original codebase.

politas:
Thanks, that looks fine now. Merging with thanks!

PR #4883, opened 2016-11-04
ExtraPlanetaryLaunchpads new release

politas:
Extended KSP version as per forum thread.
KIS is now a dependency rather than a recommend, as per forum thread

### DIFF
--- a/NetKAN/ExtraPlanetaryLaunchpads.netkan
+++ b/NetKAN/ExtraPlanetaryLaunchpads.netkan
@@ -11,10 +11,10 @@
         "repository" : "https://github.com/taniwha-qf/Extraplanetary-Launchpads"
     },
     "depends"        : [
-        {"name" : "ModuleManager"}
+        {"name" : "ModuleManager"},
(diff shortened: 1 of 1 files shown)


Dazpoet:
Shouldn't this be targetting 1.2.1 as that was the only change to it against the earlier version as per the [EPL forum thread](http://forum.kerbalspaceprogram.com/index.php?/topic/54284-12-extraplanetary-launchpads-v553/&page=170#comment-2836162)?

politas:
The main post describes it as being for "1.2", and I don't read that post as being exclusive to 1.2.1, but inclusive of.

>The only change from 5.5.2 is to update the version check for KSP 1.2.1

So there were no changes in code. I am wondering if there might be a .version file we can use in there now, though.

politas:
Awwww, yis! Very cool, let's give more control to Taniwha!

politas:
Whoops, Bash replaced $kref and $vref with the contents of non-existent environment variables.

Dazpoet:
LGTM 👍

PR #4886, opened 2016-11-04
Image Viewer Continued

linuxgurugamer:


NetKAN/ImageViewerCont.netkan
```diff
@@ -0,0 +1,14 @@
+{
+  "spec_version": "v1.4",
+  "install": [
+    {
+      "install_to": "GameData",
+      "find": "ImageViewer"
+    }
+  ],
+  "$vref": "#/ckan/ksp-avc",
```
politas: "$vref": ""#/ckan/ksp-avc/GameData/ImageViewer/ImageViewerCont.version",

linuxgurugamer: Done, but why was it needed?

Olympic1: If there are multiple version files in a download, ckan doesn't know which one to pick and fails. Adding a path to the vref tells which file to use.

politas:
Merged with my thanks!

PR #4889, opened 2016-11-05
Include RSTKSPUtils in TACLS

danfarnsy:
TACLS now requires a .dll located in REPOSoftTech/RSTKSPUtils
Added "supports" line for CommunityCategoryKit

### DIFF
--- a/NetKAN/TACLS.netkan
+++ b/NetKAN/TACLS.netkan
@@ -14,9 +14,14 @@
     "suggests" : [
         {   "name" : "KSP-AVC" }
     ],
+    "supports": [
+        {
+            "name": "CommunityCategoryKit"
+        },
     "install" : [
         {
             "find"          : "GameData/ThunderAerospace",
+            "find"          : "GameData/REPOSoftTech/RSTKSPUtils",
             "install_to"    : "GameData",
             "filter_regexp" : "TacLifeSupport\\/LifeSupport\\.cfg$"
         }
--- a/NetKAN/TACLS.netkan
+++ b/NetKAN/TACLS.netkan
@@ -21,11 +21,16 @@
     "install" : [
         {
             "find"          : "GameData/ThunderAerospace",
+            "install_to"    : "GameData",
+            "filter_regexp" : "TacLifeSupport\\/LifeSupport\\.cfg$"
+        }
+        {
             "find"          : "GameData/REPOSoftTech/RSTKSPUtils",
             "install_to"    : "GameData",
             "filter_regexp" : "TacLifeSupport\\/LifeSupport\\.cfg$"
         }
     ],
+    
     "resources" : {
         "homepage" : "https://github.com/KSP-RO/TacLifeSupport/wiki",
         "bugtracker" : "https://github.com/KSP-RO/TacLifeSupport/issues",
--- a/NetKAN/TACLS.netkan
+++ b/NetKAN/TACLS.netkan
@@ -27,7 +27,6 @@
         {
             "find"          : "GameData/REPOSoftTech/RSTKSPUtils",
             "install_to"    : "GameData",
-            "filter_regexp" : "TacLifeSupport\\/LifeSupport\\.cfg$"
         }
     ],
     "resources" : {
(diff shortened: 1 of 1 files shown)


politas:
Needs a whole separate install stanza, but is that folder not from another mod, originally?

politas:
Have to check, but I think that folder will conflict with AmpYearPowerManager.

danfarnsy:
Just checked, it looks like it's fine. @JPLRepo wrote that utility and included it in TACLS, but I just checked his library of released mods, and it doesn't appear to be anywhere else yet. But there's a good reason to expect it will be a common utility for his other mods in the future. Right now, the only place to find it publicly released is in TACLS, even though it has a separate folder structure from TACLS.

Do we need to create a separate NetKAN file and point to that location within the TACLS .zip?, then include "depends" in the TACLS NetKAN?

danfarnsy:
Sorry, when I said, "it's fine," I realize I should be more specific: After a through check, there appear to be no other mods from JPLRepo using REPOSoftTech which have the /RSTKSPUtils in them.

NetKAN/TACLS.netkan
```diff
@@ -14,12 +14,20 @@
     "suggests" : [
         {   "name" : "KSP-AVC" }
     ],
+    "supports": [
+        {
+            "name": "CommunityCategoryKit"
+        },
```
Olympic1: Remove `,` and add `],` to close the array.

NetKAN/TACLS.netkan
```diff
@@ -14,12 +14,20 @@
...
     "install" : [
         {
             "find"          : "GameData/ThunderAerospace",
             "install_to"    : "GameData",
             "filter_regexp" : "TacLifeSupport\\/LifeSupport\\.cfg$"
         }
```
Olympic1: Add a `,` after `}`

NetKAN/TACLS.netkan
```diff
@@ -14,12 +14,20 @@
...
+        {
+            "find"          : "GameData/REPOSoftTech/RSTKSPUtils",
+            "install_to"    : "GameData",
```
Olympic1: Remove `,`

JPLRepo:
Sorry guys. Yes RSTKSPUtils is included in DeepFreeze and will be included in several other mods next versions. What do we need to do?

danfarnsy:
@politas @Olympic1 Okay, this probably needs to be closed. We need a separate NetKAN file to point to RSTKSPUtils and then have TACLS and DeepFreeze (and others in the future) list it as a dependency. @JPLRepo has indicated he doesn't have time for CKAN overhead. Do either of you have interest in taking this on? In the mean time, the CKAN install for TACLS 0.12.6 likely needs to be de-listed since it lacks the RSTKSPUtils.

politas:
@JPLRepo, any chance you could release the RSTKSPUtils as a standalone mod? That'd make it much easier.

danfarnsy:
FWIW, @JPLRepo and I just discussed this in a PM: "Would this mean it has to be distributed as a separate mod? as I don't want to do that."

Is there a way to have a separate NetKAN listing for the utility which points to the relevant Github location? Otherwise the intent is to distribute it with the mods in manual installation rather than having a separate release.

politas:
Certainly, though we can also do what we do with the Reposoft Agencies folder, sourcing it from one mod to use in both.

danfarnsy:
Okay, that sounds ideal, though it will likely complicate your auto-crawler's updating mechanism. We could probably ask that @JPLRepo include a version file for the utility so this doesn't otherwise become an ongoing support issue on either his end or yours.

politas:
It's easier if he doesn't include a separate version file, actually. We'll just use the version number of DeepFreeze as the version number for the Utils.

danfarnsy:
Perfect. Thanks for your help!

politas:
Sorry it's taken so long to get to this. I saw @JPLRepo's comment in #ckan the other day, but got distracted by easier tasks.

politas:
Ok, we'll need to add the dependency and filter to all @JPLRepo mods as the dll gets included in them.

JPLRepo:
Hey guys. Sorry... there are reasons I don't want to distribute it as a separate mod.
It probably won't be around for that long.
Sob: Wish it was easier. Actually didn't want to cause work for anyone.

PR #4890, opened 2016-11-05
Update Kopernicus.netkan

### LINKED ISSUE #4888: Kopernicus
Kopernicus was updated to 1.2.1-1 and the metadata needs to be updated. The new version can be found as the first file in the original github link in the metadata. CKAN is a god send!!!! Thanks for all the work.

Dazpoet:
I think this is what's needed to resolve #4888 but I'm not sure how this staging thing works so I might be totally off.


politas:
Failing because no ModularFlightIntegrator for 1.2.1, yet. #4892 will make it work.

politas:
#rebuild

Olympic1:
LGTM, merging

PR #4891, opened 2016-11-05
Create SigmaBinary.netkan

Rethkir:
Sigma Binary is a plugin that allows for binary systems. It has recently been updated to work with KSP 1.2.1.

### DIFF
--- a/NetKAN/SigmaBinary.netkan
+++ b/NetKAN/SigmaBinary.netkan
@@ -1,0 +1,23 @@
+{
+    "spec_version" : 1,
+    "identifier"   : "SigmaBinary",
+    "$kref"        : "#/ckan/github/Sigma88/Sigma-Binary",
+    "install": [
+       {
+           "file"       : "GameData/Sigma",
+           "install_to" : "GameData"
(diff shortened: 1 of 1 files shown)


Rethkir:
I noticed that all of Sigma88's mods have been frozen. Why is this?

politas:
Sigma88 asked for all of their mods to be removed from CKAN temporarily. That decision has not been reversed to my knowledge.

Rethkir:
Shouldn't they be on the Opt Out list in that case? I didn't see them there.

On Nov 5, 2016 11:46 AM, "Myk" <notifications@github.com> wrote:

Sigma88 asked for all of their mods to be removed from CKAN temporarily.
That decision has not been reversed to my knowledge.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/KSP-CKAN/NetKAN/pull/4891#issuecomment-258619923>, or mute
the thread

politas:
They demurred being on the opt-out list. A polite message occasionally would probably not annoy them too much.

PR #4894, opened 2016-11-05
Create RSTKSPUtils, set relevant filters and dependencies

politas:
Ref #4889


politas:
TACLS has a dependency on CommunityResourcePack that is holding it back from KSP 1.2.1

politas:
New CRP should be out today.

politas:
#rebuild

Dazpoet:
Seems fine to me :)

PR #4899, opened 2016-11-06
Future Weapons - remove BDA version restriction

### LINKED ISSUE #4898: Future Weapons Updated
All right, I am the author of Future Weapons, which is now updated to 1.2.1.
It used to be on CKAN (I can still find it, but using the old version) it may however be linked to Spaceport, rather than Spacedock. 

That needs to be updated, and it also has the dependency BDArmory, which is for now defunct. Instead, it now uses BDArmory Continued.

Requesting some help to update those two, and get it working again.

politas:
Closes #4898 (hopefully)
Not sure what was blocking 3.3 being inflated.

### DIFF
--- a/NetKAN/FutureParts.netkan
+++ b/NetKAN/FutureParts.netkan
@@ -4,7 +4,7 @@
     "license": "CC-BY-SA-2.0",
     "identifier": "FutureParts",
     "depends": [
-        { "name": "BDArmory", "min_version": "v0.9.9.1" }
+        { "name": "BDArmory" }
     ],
     "install": [
         {


Olympic1:
It was released on SD 43 minutes ago and the bot still needs to run

politas:
Oh, I thought webhooks made Spacedock updates happen almost immediately.

Olympic1:
They changed some stuff to reduce load. Untill then we can keep this open if it doesn't get inflated on the next run.

Olympic1:
Closing as its added now

PR #4900, opened 2016-11-07
Remove suggested mod from Kopernicus.netkan

Rethkir:
KittopiaTech is discontinued and out of date. Removing suggestion from code.


Olympic1:
KittopiaTech isn't discontinued, there was a release on 23th October https://github.com/Kopernicus/KittopiaTech/releases

Olympic1:
But apparently the netkan isn't updated, fixing now

PR #4902, opened 2016-11-07
Remove HideEmptyTechNodes.netkan dependency

jyanjyanjyan:
Remove Module Manage dependency, as it is no longer needed (mod author here).


Olympic1:
LGTM, merging with my thanks

jyanjyanjyan:
If it matters, this is only for v0.6 of the mod, which I uploaded to SpaceDock a couple of hours ago. Any version before 0.6 will still need Module Manager. Do I need to change anything in the NETKAN file to reflect that? v0.6 is the only version for KSP 1.2.1 so far. Thanks in advance.

Olympic1:
It only removed the MM dependency for v0.6, so no extra work to do

PR #4870, opened 2016-11-02
OPTSpacePlanes New release with legacy split

politas:
Mod author requested change via Forum message.

PR #4906, opened 2016-11-07
Update OPTSpacePlaneParts

Olympic1:
Main mod was moved to a new SD page and OPT Legacy was added


politas:
Already created #4870

Olympic1:
Oh

PR #4496, opened 2016-08-05
Add Modified MK22 IVA [ASET Avionics]  from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of speedio, to add [Modified MK22 IVA [ASET Avionics] ](https://spacedock.info/mod/411/Modified%20MK22%20IVA%20%5BASET%20Avionics%5D%20) to CKAN.

Mod details:
    name = /mod/411/Modified%20MK22%20IVA%20%5BASET%20Avionics%5D%20
    author = speedio
    abstract = Modified interior of the MK22 cockpit created by BDynamics. Using ASET Avionics for analog gauges
    license = CC BY-SA 2.0
    Homepage = http://forum.kerbalspaceprogram.com/index.php?/topic/129625-105-modified-mk22-iva-aset-avionics/
    description = See the forum page for more details. (Mod website, forum thread)

 Dependencies

BDynamics Mk22 Cockpits v1.0
ASET Props Pack
ASET Avionics Pack
RasterPropMonitor
Module Manager 

Please direct questions about this pull request to [speedio](https://spacedock.info/profile/speedio).

PR #4910, opened 2016-11-08
BDynamics Mk22 Cockpit KSP version override

politas:
KSP compatibility to match forum thread.
Should enable #4496 to build


Olympic1:
LGTM, merging

PR #4917, opened 2016-11-09
Wrong install directory

jonasAhnstrom:


politas:
Indeed. Merging with thanks!

PR #4933, opened 2016-11-13
SearchAndRescue to Dropbox

politas:
Mod has been removed from Spacedock


Olympic1:
There are 2 `license`s in the file on line 2 and 8. Otherwise it looks good

PR #4934, opened 2016-11-13
Tweakable Everything Continued

linuxgurugamer:


politas:
[ToadicusTools](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/ToadicusTools.netkan) is TweakableEverything?

Dazpoet:
ToadicusTools is a [dependancy](https://github.com/KSP-CKAN/NetKAN/search?utf8=%E2%9C%93&q=Toadicustools) for various other mods. I'm guessing we might want to start indexing it from this mod instead seeing as the older link is outdated?

linuxgurugamer:
Right now, those mods are not compatible with 1.2.1.  ToadicusTools is included in this mod, when I release VOID (or maybe sooner), I'll break it out and fix this netkan, but for now, this is self contained for 1.2.1

politas:
Currently, this is only installing ToadicusTools:

>Showing 5 installed files:
- GameData/ToadicusTools
- GameData/ToadicusTools/Plugins
- GameData/ToadicusTools/Plugins/ToadicusTools.dll
- GameData/ToadicusTools/ToadicusTools.version
- GameData/ToadicusTools/MiniAVC.dll

politas:
There'll be a lot of users who have an outdated copy of ToadicusTools installed already, which will stop this from installing. Is there some source we can get a clean copy of ToadicusTools? If we use an http $kref, we can use the internal ToadicusTools.version for version number.

linuxgurugamer:
Ok.  I'll get an updated version of the tools out this evening and update the netkan for this, as well as add a new one for the tools.  Will call it ToadicusToolsContinued to avoid confusion

linuxgurugamer:
#4356

linuxgurugamer:
#rebuild

Dazpoet:
Seems to install the right things but I noticed that the old Toadicustools ran under the BSD-3-clause license while this one is listed as unknown. Seeing as I know very little about the finer points of FOSS-licensing I'll leave it to somebody else to figure that one out though.

linuxgurugamer:
It should be BSD, that's what I specified on Spacedock

NetKAN/TweakableEverythingCont.netkan
```diff
@@ -9,13 +9,12 @@
...
         }
     ],
-    "install": [
-        {
-            "file": "GameData/ToadicusTools",
-            "install_to": "GameData"
-        }
-    ],
     "depends": [
+    	{ "name": "ToadicusToolsCont" },
```
politas: You've called it "ToadicusToolsContinued"

politas:
You only put "BSD" rather than "BSD-3-Clause"

linuxgurugamer:
I'm puzzled, i just looked at the current netkan, it shows BSD-3-Clause

linuxgurugamer:
I just updated Spacedock.
FYI, when creating a mod on Spacedock, it shows BSD as one of the choices, which is what I had checked

NetKAN/ToadicusToolsContinued.netkan
```diff
@@ -0,0 +1,14 @@
...
+    {
+      "install_to": "GameData",
+      "find": "ToadicusTools"
+    }
+  ],
+  "$vref": "#/ckan/ksp-avc",
+  "$kref": "#/ckan/spacedock/1069",
+  "x_via": "Automated Linuxgurugamer CKAN script",
+  "identifier": "ToadicusToolsContinued",
+  "license": "unknown"
```
Dazpoet: This is the line in question, it says unknown while the old version of ToadicusTools said BSD-3-clause

linuxgurugamer:
Oh, ok. 
Sorry.  I just updated it

Dazpoet:
Aww yis! Merging with many thanks and I'm going to turn off KSP and download this mod through CKAN the second it indexes :D

PR #4951, opened 2016-11-16
remove epoch again

AndyMt:
Unintentionally used a "V" prefix in version 1.7.1.
Latest Version 1.7.2 doesn't have it as "most" earlier versions.
I promise to not use a "V" prefix again.

### DIFF
--- a/NetKAN/GravityTurnContinued.netkan
+++ b/NetKAN/GravityTurnContinued.netkan
@@ -7,7 +7,6 @@
     "abstract": "Automated highly efficient launches",
     "ksp_version": "1.2",
     "license": "GPL-3.0",
-    "x_netkan_epoch": 1,
     "resources": {
         "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/148448-12161-gravityturn-continued-automated-efficient-launches/",
         "repository": "https://github.com/AndyMt/GravityTurn/"


Dazpoet:
I'm afraid that removing the epoch will make all versions released from now on be counted as older than the version(s) released under the new epoch. Might be able to solve it with an override but I'm not quite sure how.

AndyMt:
Ok, I see. It just looks so ugly in CKAN now with the 1: in front. It shows 1:1.7.2 at the moment.

AndyMt:
Another question if I may: if I tag a release as a "prerelease" in Github ([like V1.7](https://github.com/AndyMt/GravityTurn/releases/tag/V1.7)) I don't want them to show up at all on CKAN. Haven't found anything about that topic in the CKAN documentation, only a reference in issue #403: https://github.com/KSP-CKAN/CKAN/issues/403

Dazpoet:
If you look at which versions of the mod we have indexed [so far](https://github.com/KSP-CKAN/CKAN-meta/tree/master/GravityTurnContinued) you will be glad to see that we do not index pre-releases :)

politas:
There are a lot of mods on CKAN with epochs making the version numbers look odd. CKAN users are accustomed to them.

politas:
AndyMt, are you cool with this PR being closed?

AndyMt:
Yes. I wasn't aware that my version of the mod was on CKAN already, because until 1.7.2 it still was work in progress. I specifically asked at some point in the dev thread to NOT put new versions on CKAN. One reason was that I wanted to figure out myself how to do that and what processes and guidelines to follow. 

Please don't get me wrong: I very much appreciate the work all of you do and the excellent service you provide. For mod authors and users alike :-).

PR #4955, opened 2016-11-16
Update SDHI Strobe-O-Matic for KSP 1.2.1

sumghai:
Strobe-O-Matic pretty much always works with all updated versions of KSP.

Still working on an update for Service Module System.


politas:
If it works pretty generally without updates, you could set a range like:

    "ksp_version_min": "1.0",
    "ksp_version_max": "1.2",

Which would enable it for all versions of ksp from 1.0.x to 1.2.x

sumghai:
@politas - Good to know, thanks!

PR #4956, opened 2016-11-16
Update deps

Olympic1:
* Update Kopernicus min version
* Add CustomBarnKit
* Remove Kopernicus Expansion support
* Rename `Karbonite` to `CommunityResourcePack`


Dazpoet:
came here to do just this, you're awesome olympic1 :)

PR #4957, opened 2016-11-17
RLAContinued

linuxgurugamer:


linuxgurugamer:
When he gets a .version file in, I'll get this updated

linuxgurugamer:
#rebuild

Dazpoet:
Are the configs in the extras folder installed by standard or should they be Broken out into their own thing? Looks like they fit in with a normal install but I'm just making sure.

PR #4963, opened 2016-11-19
Octosat Continued

linuxgurugamer:


Dazpoet:
I can't check right now but I suspect we might not want to install the shellscript in the partsfolder? Assuming that a.sh is a shellscript?

linuxgurugamer:
It won't hurt, but I'll get it removed
Since this is mine, I'm not going to merge, but if someone could, that would be fine.

linuxgurugamer:
Updated, shell script is out

politas:
LGTM. Merging

PR #4968, opened 2016-11-20
Create NSSC.netkan

Real-Gecko:
Add Not So Simple Construction, Simple Construction addon.


politas:
LGTM. Merging with thanks!

PR #4972, opened 2016-11-21
Manually adding Simple Logistics mod

Real-Gecko:
Automates power and resources distribution between landed nearby vessels. Power distribution range 2000m. Resource distribution range 150m.

Engineer required for power distribution, must be in any nearby landed vessels.

Resource distribution happen automagically.

Distributed resources: LfOx, MonoProp, Ore.

### DIFF
--- a/NetKAN/NSSC.netkan
+++ b/NetKAN/NSSC.netkan
@@ -12,8 +12,14 @@
          },
          {
              "name": "ModuleManager"
+         },
+         {
+             "name": "KIS"
          }
      ],
+    "recommends"     : [
+        {"name" : "KAS"}
+    ],
     "conflicts": [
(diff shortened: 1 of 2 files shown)


politas:
You should make new git branches for each PR. The first two commits in this PR have already been merged in #4968.

"ksp_version" : "any"    -  are you sure about this? This mod will definitely not need any future updates to work with possible changes to USI tools or KSP? And it will definitely work with all past versions of USI Tools/KSP?

Real-Gecko:
Oh, sorry, I missed the fact that old fork was not deleted, will redo right now. And about version: yes, it's a set of MM patches, nothing more. If MM an USI Tools available, then everything will work, if not then it won't break anything - patches won't be applied.

Issue #4971, opened 2016-11-20
Near Future Space Craft report 1.2.1 ...


Bit-Fiddler:
Thanks for wanting to bring an issue to our attention!

Before you open a ticket, please do a quick search of the existing issues to see if it's already been brought up.

Please note that this repository is strictly for metadata (issues relating to incorrect information about mods, missing versions, or incorrect installations). If you want to suggest something or report a bug relating to CKAN as a whole, open the issue in: https://github.com/KSP-CKAN/CKAN

politas:
It'll appear again when B9PartSwitch is updated

Bit-Fiddler:
not sure if this will be seen as this has been closed...   but B9 part switch is updated and installed by CKAN yet the Near Future Space Craft is still not showing on the install list for me.

Dazpoet:
We're missing NearFutureProps which was accidentally dropped from the package, see http://forum.kerbalspaceprogram.com/index.php?/topic/47786-121-near-future-technologies-nov-18-many-updates/&page=280#comment-2855894

A workaround _might_ be to create an override and allow the last version of it to be used for this one aswell. I'll re-open the issue since it wasn't fixed by the proposed fix.

PR #4987, opened 2016-11-25
Remove dependeny from NearFutureSpacecraft.netkan

Rethkir:
The current version of NearFutureSpacecraft does not seem to rely on this dependency. It runs without installing NearFutureProps, which is not current.


Dazpoet:
The author has stated that the lacking dependency was a [mistake](http://forum.kerbalspaceprogram.com/index.php?/topic/47786-121-near-future-technologies-nov-18-many-updates/&page=280#comment-2855894) and as such I think it would be unwise to remove it from our metadata.

I recommend checking #4971 to track this issue

linuxgurugamer:
Should this be closed?

PR #4985, opened 2016-11-25
Add Wenchang Satellite Launch Center from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of IcecoveryStudio, to add [Wenchang Satellite Launch Center](https://spacedock.info/mod/1091/Wenchang%20Satellite%20Launch%20Center) to CKAN.

Mod details:
    name = /mod/1091/Wenchang%20Satellite%20Launch%20Center
    author = IcecoveryStudio
    abstract = Chinese Wenchang Satellite Launch Center
    license = CC BY-SA 4.0
    Homepage = 
    description = ##A replica of Wenchang Satellite Launch Center, a southern and newly built launch facility in People's Republic of China, which serves flight missions of CZ-5, CZ-7, and future heavy launch vehicles like CZ-9.
##Our members:

###Models: ladeng
###Textures:ladeng
###Import: ladeng
###Configs: ladeng
###Technical Support:

##DEPENDENCIES:

###Kerbal Konstructs

##LICENSE:

###CC BY-SA 4.0

Please direct questions about this pull request to [IcecoveryStudio](https://spacedock.info/profile/IcecoveryStudio).

PR #4989, opened 2016-11-25
Update KerbalKonstructs.netkan

Dazpoet:
Based on http://forum.kerbalspaceprogram.com/index.php?/topic/151818-12x-kerbal-konstructs-0985-21112016/ it seems this mod is under new maintainer and I caught it while working on #4985 which depends on this.


politas:
We should override the author to include the previous maintainer?

Dazpoet:
That should do it :)

PR #4993, opened 2016-11-26
Update BlueHawkIndustries.netkan

### LINKED ISSUE #4977: Blue Hawk industries will not download for 1.2.1.   it reports the file is missing.

Dazpoet:
Closes #4977

Very unsure on the version string. I think this will cound as a new epoch? The mod used to follow a 2.1-2.2-...-7.1 etc version scheme but now it's based on version 1.0-1.1-...-1.5 so a new epoch is needed.

This PR does not fix the borked 1.2 version we have indexed in CKAN-meta nor does it fix any other broken versions which might exist in there.


politas:
That seems to sort it out. Do we have archives of the 404ing versions we can switch to?

Dazpoet:
I must admit that I don't know nor do I know how I would get such information.

PR #4997, opened 2016-11-29
Updated per github README -  KISS

linuxgurugamer:


linuxgurugamer:
Very strange, I just saw that this mod became available in CKAN.  So before merging this, please check to see if it's necessary

Dazpoet:
version v2.0.0 shows up in my CKAN client, but it seems you're trying to make 2.1.0 show up?

politas:
Once you go "v", you can't go back (without an epoch). Needs either 

    "x_netkan_force_v" : true,

or

    "x_netkan_epoch" : 1,

linuxgurugamer:
so I'm puzzled.  This file has an X_netkan_override, but it also has a $kref.

It seems that CKAN sees the 2.0, but it's up to 2.1 on the server

I'm simply trying to get it updated properly, if this isn't the correct thing, either tell me what, or do it.
Thanks

politas:
The x_netkan_override was for an earlier version, probably before the AVC file was added to the zip.
KerbalImprovedSaveSystem-v2.0.0 comes after
KerbalImprovedSaveSystem-2.1.0

Forget about changing the x_netkan_override, what we need is 

    "x_netkan_force_v" : true,
or

    "x_netkan_epoch" : 1,

linuxgurugamer:
I had deleted the repo, so will be submitting a new one to replace this, with the change suggested by Politas

PR #4999, opened 2016-11-29
Update SimpleConstruction.netkan

Real-Gecko:
- Updated license
- Removed `x_netkan_force_v` to avoid odd version numbering
- Removed installation directives.

### DIFF
--- a/NetKAN/SimpleConstruction.netkan
+++ b/NetKAN/SimpleConstruction.netkan
@@ -19,11 +19,11 @@
     "install" : [
         {
             "find": "SimpleConstruction",
-            "install_to": "GameData",
+            "install_to": "GameData"
         },
         {
             "find": "ExtraplanetaryLaunchpads",
-            "install_to": "GameData",
+            "install_to": "GameData"
         }
     ]
 }
--- a/NetKAN/SimpleConstruction.netkan
+++ b/NetKAN/SimpleConstruction.netkan
@@ -3,12 +3,27 @@
     "identifier": "SimpleConstruction",
     "$kref": "#/ckan/spacedock/59",
     "license": "MIT",
-    "x_netkan_epoch": 1,
+    "x_netkan_version_edit": {
+        "find": "[Version_]?",
+        "replace": "",
+        "strict": false
+    },
     "depends": [
         { "name": "ModuleManager" },
         { "name": "InterstellarFuelSwitch-Core" }
     ],
     "conflicts": [
         { "name": "ExtraPlanetaryLaunchpads" }
+    ],
+    "provides"  : [ "ExtraPlanetaryLaunchpads" ],
+    "install" : [
+        {
+            "find": "SimpleConstruction",
+            "install_to": "GameData",
+        },
+        {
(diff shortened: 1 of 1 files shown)


Real-Gecko:
Connected with https://github.com/KSP-CKAN/CKAN-meta/pull/1238

Dazpoet:
Does the client know what to do nowadays when we flat out remove a lot of metadata like in https://github.com/KSP-CKAN/CKAN-meta/pull/1238 ? I recall it being an issue but I'm not sure how it's now. Another solution could be the addition of an [epoch](https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#epoch) using the [x_netkan_epoch](https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#x_netkan_epoch)

Real-Gecko:
So, if I make `"epoch": 1` that will solve current version numbering problem and there will be no need for https://github.com/KSP-CKAN/CKAN-meta/pull/1238?

politas:
Yes, that's correct. Deleting metadata from the repository doesn't update the current installed version metadata, so any users who have say "v3.4" installed will never see a "3.5" release as an upgrade, even if the repository now contains only "3.3", "3.4", "3.5". Otherwise, users would not be able to usefully install from a .ckan file.

Changing the epoch is the right way to reset the versions if you really feel it is necessary to not have "v" on the CKAN version numbers.

Real-Gecko:
Did that.
Prior to version 3.1 version numbering [was ok](http://spacedock.info/mod/59/SimpleConstruction#changelog), like `2.2`, `3.0`, but starting from version 3.1 MatterBeam started to add something like this: `Version_3.1`, reason is unknown to me, but it now caused problems with new version. I have `3.4.1` released already, but client still thinks that `Version_3.3` is major:
screenshot_20161201_024836

Real-Gecko:
So?

Real-Gecko:
So, I think I found a better solution by adding "x_netkan_version_edit", but now build fails with error
`CKAN.FileNotFoundKraken: Trying to install SimpleConstruction vVersion_3.3, but it's not downloaded or download is corrupted` as it tries to get deleted version from Spacedock.
But I also noticed that install section is like this
```
About to install...

 * SimpleConstruction vVersion_3.3
 * Module Manager 2.7.5
 * Interstellar Fuel Switch Core 2.2.9
 * TweakScale - Rescale Everything! v2.3.2
```
Why does it tries to install TweakScale?

politas:
Versions are now 404, so...

Real-Gecko:
Thanks!

politas:
Ah, damn. Thought this was the other one.

Real-Gecko:
Emm, version 3.4.1 is not doubled in CKAN-meta:
```
 		SimpleConstruction/SimpleConstruction-3.4.1.ckan
		SimpleConstruction/SimpleConstruction-v3.4.1.ckan
```

politas:
Yeah, sorry, I really didn't mean to merge this change. We need an epoch or we can't ditch the "v"

PR #5002, opened 2016-12-03
SimpleConstruction Add epoch

politas:
ref #4999

### DIFF
--- a/NetKAN/SimpleConstruction.netkan
+++ b/NetKAN/SimpleConstruction.netkan
@@ -8,6 +8,7 @@
         "replace": "",
         "strict": false
     },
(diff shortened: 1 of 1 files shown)


Real-Gecko:
So, we need to set epoch anyway? Will we need to remove `x_netkan_version_edit`?

politas:
x_netkan_version_edit can stay. Shouldn't make a difference. People are still going to have local metadata saying vVersion<whatever>, and so CKAN will not upgrade until the KSP version forces it.

Real-Gecko:
OK, nice :)

Real-Gecko:
Let's rock!

Issue #5009, opened 2016-12-06
Changes in Kerbalism metadata


ShotgunNinja:
Hi,

starting from Kerbalism version 1.1.5 for KSP 1.2.1 (to be released in the following days), multiple profiles can cohexist in the same user installation.

As a consequence of this, the package Kerbalism-Profile-Default is now obsolete and has no reason to keep existing. I don't know how to proceed about this, but I doubt that is as simple as deleting the metadata associated.

In addition, the package Kerbalism has a special provision: files with extension .cfg in the Profile/ subfolder are deleted post-installation. This is not necessary anymore, for the same reason specified above.

Dazpoet:
pinging @dbent who did the last update to account for profiles. There are so many fancy netkan tools nowadays there's probably a much easier way to resolve this than the ones I know off.

ShotgunNinja:
Thanks a lot!

politas:
Thanks for letting us know ahead of time, @ShotgunNinja !

dbent:
Just to note @ShotgunNinja users upgrading that already have `Kerbalism-Profile-Default` installed will have to uninstall it before they can upgrade to `Kerbalism` 1.1.5 since the `Profiles/Default.cfg` file will now be provided by `Kerbalism` and will produce a conflict on upgrade.

PR #5010, opened 2016-12-06
Update Kerbalism in preparation for 1.1.5 release

dbent:
See KSP-CKAN/NetKAN#5009

- Freeze Kerbalism-Profile-Default.
- Move `recommends` and `install` directive into an override targetting
  versions less than 1.1.5.

@ShotgunNinja @Dazpoet


dbent:
The net effect of this is that when Kerbalism 1.1.5 is released the metadata won't include the `recommends` for `Kerbalism-Profile` and the default install process of looking for a directory with the name `Kerbalism` will be used since we no longer have to filter the profile configs.

politas:
LGTM. Merging with thanks!

PR #5016, opened 2016-12-08
Update Kopernicus.netkan

dbent:
@techman83 Is staging working correctly? `Kopernicus` has `x_netkan_staging` set to `true` because it has manually specified KSP version information but it seems like the most recent update was automatically [comitted](https://github.com/KSP-CKAN/CKAN-meta/commit/1259de3d40c52e8c297635b7c5c9a930a8ec7512) rather than staged.

@KSP-CKAN/wranglers


politas:
LGTM

PR #5018, opened 2016-12-10
SVE Override broken AVC KSP versions

politas:
The AVC .version files in the current releases have ksp_min as 1.2 and ksp_max as 1.2.1, so for 1.2.2 we're offering the previous release which has been removed from GitHub.


politas:
#rebuild

PR #5024, opened 2016-12-12
Create UnmannedBeforeMannedChallenge.netkan

Y3mo:
Another mini mod with self-administered netkan.

Thank you very much,
Y3mo

### DIFF
--- a/NetKAN/UnmannedBeforeMannedChallenge.netkan
+++ b/NetKAN/UnmannedBeforeMannedChallenge.netkan
@@ -1,0 +1,6 @@
+{
+    "spec_version": 1,
+    "identifier": "UnmannedBeforeMannedChallenge",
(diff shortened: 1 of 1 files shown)


Y3mo:
Ok, I give up, no idea whats wrong with it.

Dazpoet:
I think Jenkins might be doing a dumdum but let's ask @techman83 to be sure

politas:
`Extracted 1.99.99 as KSP version.
Overriding '1.99.99' with '1.2.1'
Creating a dummy KSP '1.2.1' install`

I think that's an issue with our build script

politas:
#rebuild

techman83:
The build script should probably favor the ksp_version_min if present over the ksp_version_max.

It was indeed not available for 1.2.1 :-)
```
"ksp_version_max": "1.99.99",
"ksp_version_min": "1.2.2",
```

PR #4995, opened 2016-11-29
Add Horizon Aeronautics - Zenit 3SLB Reduzx from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of Delta8930, to add [Horizon Aeronautics - Zenit 3SLB Reduzx](https://spacedock.info/mod/1095/Horizon%20Aeronautics%20-%20Zenit%203SLB%20Reduzx) to CKAN.

Mod details:
    name = /mod/1095/Horizon%20Aeronautics%20-%20Zenit%203SLB%20Reduzx
    author = Delta8930
    abstract = This mod was originally created by the wonderful Stubbles, who at that time created one of the best rocket packs out there. This pack contains the Zenit-3SLB booster, updated for KSP 1.2.X by MeCripp, who also included support for Animated Decouplers and Procedural Fairings. Unfortunately, after the release of v1.2, meant for KSP 0.23.5, all work on this project was terminated. After a long hiatus, @MeCrippand I are excited to bring this mod back to the KSP Community.   The Zenit booster is a medium-heavy launch vehicle, capable of carrying heavy payloads to Geostationary orbit. This Zenit can bring ~10-15T to Geosynchronous orbit. It is compatible with FAR, KWRocketry, MechJeb, and DeadlyReentry. 
    license = CC BY-NC
    Homepage = http://forum.kerbalspaceprogram.com/index.php?/topic/152728-12x-horizon-aeronautics-zenit-3slb-redux/
    description = HORIZON AERONAUTICS

is proud to present....

The mighty Zenit 3SLB rocket!

Background information: This mod was originally created by the wonderful @stubbles, who at that time created one of the best rocket packs out there. This pack contains the Zenit-3SLB booster, updated for KSP 1.2.X by @MeCripp, who also included support for Animated Decouplers and Procedural Fairings. Unfortunately, after the release of v1.2, meant for KSP 0.23.5, all work on this project was terminated. After a long hiatus, @MeCrippand I are excited to bring this mod back to the KSP Community. 
 
The Zenit booster is a medium-heavy launch vehicle, capable of carrying heavy payloads to Geostationary orbit. This Zenit can bring ~10-15T to Geosynchronous orbit. It is compatible with FAR, KWRocketry, MechJeb, and DeadlyReentry. 

Please direct questions about this pull request to [Delta8930](https://spacedock.info/profile/Delta8930).

PR #5034, opened 2016-12-13
DecalStickers

linuxgurugamer:
Closes #5014

PR #5035, opened 2016-12-13
HorizonAeronauticsZenit3SLBReduzx

linuxgurugamer:
Closes #4995

### DIFF
--- a/NetKAN/DecalStickers.netkan
+++ b/NetKAN/DecalStickers.netkan
@@ -1,0 +1,25 @@
+{
+  "spec_version": "v1.4",
+  "install": [
+    {
+      "install_to": "GameData",
+      "find": "blackheart"
+    }
+  ],
+  "depends": [
+    {
+      "name": "FirespitterResourcesConfig"
+    },
+    {
+      "name": "ModuleManager"
+    },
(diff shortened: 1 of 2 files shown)


Dazpoet:
It gets stuck on a .version file in the _MACOSX folder. I'm not sure why it's looking there tbh since the CKAN-spec clearly states that [`Only .version files that would be installed for this mod are considered.`](https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#vref) and the find directive should only work against the [top-most matching directory](https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#install) which shouldn't be in the _MACOSX folder since that makes everything one layer deeper. Not sure how to ping on this but I'm guessing @dbent since it seems NetKAN isn't doing what the spec says it should be doing.

On another note your branch is funny since this PR also contains #5034 but that shouldn't be an issue.

Dazpoet:
Oh and maybe we want to also install the Ships folder?

dbent:
That's weird.

linuxgurugamer:
I'm going to redo this, adding filters to ignore the MAC specific folders

PR #4947, opened 2016-11-15
Eve fix

Dazpoet:
As requested in #4946 

Closes #4946

PR #5037, opened 2016-12-13
Make EVE compatible with 1.2.2

### LINKED ISSUE #5030: Environment visual enhencement not compatible with 1.2.2 in CKAN
Environment visual enhencement is not compatible with 1.2.2 in CKAN, can you add it in compatibility list?
It cause problems with some depedencies

Dazpoet:
closes #5030


Dazpoet:
Fails on them depending on eachother, pretty sure this should work and be correct. It's basically a repeat of  #4947 which ran into the exact same issue.

dbent:
LGTM

PR #5044, opened 2016-12-14
Add BlacklegIndustriesMM.netkan

blackleg:


politas:
So this adds or changes TAC Life Support settings for the parts in Tantares, is that right? It would be good to have that in the mod abstract. I think this is failing due to https://www.blackleg.es not responding, (Looks like that server is not running https?)

blackleg:
@politas I changed the plugin identifier and add info in the abstract

politas:
```
#   Failed test 'BlacklegIndustriesMM_TantaresTACLS: CKAN identifiers must consist only of letters, 
     numbers, and dashes, and must start with a letter or number.'
#   at t/metadata.t line 26.
#                   'BlacklegIndustriesMM_TantaresTACLS'
#     doesn't match '(?^:^[A-Za-z0-9-]+$)'
```
No underscores allowed in identifiers, sorry.

politas:
Thanks, @blackleg. I don't understand why Jenkins is complaining about the spec_version on two other mods, still. This is all fine now as far as I can see. ¡Uniendo con muchas gracias!

techman83:
@politas it'd be a fork of the repo from before those files were fixed. That test runs over all Netkans in the repo.

politas:
@techman83, ah, that explains it. I was baffled.

PR #5055, opened 2016-12-18
Updated dependancies

sswelm:


sswelm:
mention something about CommunityTechTree not beeing updated yet for KSP 1.2.2, which should not matter

Dazpoet:
Aye the bot is being a douche, merging this with many thanks for continuosly keeping your dependencies updated :)

techman83:
Poor bot!
```
5487 [1] ERROR CKAN.RelationshipResolver (null) - Dependency on CommunityTechTree found but it is not listed in the index, or not available for your version of KSP.
```
Latest CommunityTechTree supports 1.2.1. This states 1.2.2 as the KSP version, so won't show up in CKAN due to dependency not being available.

Bot was being correct :)

Dazpoet:
Aye that was what I meant by it being a douche and not doing what we want it to do rather than just doing what we Tell it to do :p

PR #5071, opened 2016-12-22
RF Stockalike provides RealPlume configs

blowfishpro:
OP also strongly recommends RealPlume since stock effects are broken in some cases, so that is reflected as well


Dazpoet:
LGTM :)
Seems the RO train has started picking up speed 👍

PR #5077, opened 2016-12-24
Create MagicSmokeIndustriesModelRework

Sheepingtin:
Magic Smoke Industries Infernal Robotics - Model Rework - Working with KSP 1.2.2

### DIFF
--- a/NetKAN/MagicSmokeIndustriesModelRework.netkan
+++ b/NetKAN/MagicSmokeIndustriesModelRework.netkan
@@ -1,0 +1,29 @@
+{
+    "spec_version": 1,
+    "identifier"  : "MagicSmokeIndustriesModelRework",
+    "name"        : "Magic Smoke Industries Infernal Robotics - Model Rework",
+	"author"      : ["ZodiusInfuser","Ziw"],
+	"abstract"    : "More robotic parts!",
+    "version"     : "1.1.2",
+    "ksp_version" : "1.2.2",
+    "license"     : "GPL-3.0",
+	    "depends": [
+        { "name": "MagicSmokeIndustries" }
+        ],
+        "recommends": [
+        { "name": "MagicSmokeIndustriesLegacy" }
+        ],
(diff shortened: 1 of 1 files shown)


LemonzPC:
Tested and working in game with my last edit to the main netkan file. The legacy pack has been delayed do to compatibility with ckan, Fixes are waiting for approval.

politas:
I believe this should be depending on `"InfernalRobotics"` rather than `"MagicSmokeIndustries"`

NetKAN/MagicSmokeIndustriesModelRework.netkan
```diff
@@ -0,0 +1,29 @@
...
+    "spec_version": 1,
+    "identifier"  : "MagicSmokeIndustriesModelRework",
+    "name"        : "Magic Smoke Industries Infernal Robotics - Model Rework",
+	"author"      : ["ZodiusInfuser","Ziw"],
+	"abstract"    : "More robotic parts!",
+    "version"     : "1.1.2",
+    "ksp_version" : "1.2.2",
+    "license"     : "GPL-3.0",
+	    "depends": [
+        { "name": "MagicSmokeIndustries" }
```
politas: I believe this should be "InfernalRobotics" rather than "MagicSmokeIndustries"

Dazpoet:
We do have these old ones laying about:
https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/IR-Model-Rework-Core.frozen https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/IR-Model-Rework-Expansion.frozen https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/IR-Model-Rework-Utility.frozen
but seeing as this PR is only one file I guess the author might have dropped the old notation of splitting it into multiple parts?

politas:
Ah, I think we need to look at changing those instead.

PR #5088, opened 2016-12-27
Redoing the VesselView updates

linuxgurugamer:


linuxgurugamer:
Politas, I redid this entire PR becuase I had to make more changes and didn't have the original around anymore

politas:
No worries. Am I right in thinking that the two "UI" mods are only installing a single file each?
If so, you could use find_matches_files, eg

    "find" : "VesselViewRPM.dll",
    "find_matches_files" : true,
    "install_to" : "GameData/VesselView"

rather than matching the folder and excluding all the other files.

linuxgurugamer:
There is a problem with this.  When you now try to install the VesselViewer, CKAN then presents two lines for the VVesselViewUI.  Multiple issues:

1. The mod name for each is identical, probably because it's pulling the name from spacedock
2. Only one can be selected
3. Selecting and installing gets a CKAN error

PR #5095, opened 2016-12-28
Create AoATech.netkan

DuoDex:


politas:
You'll need to remove all those "+" symbols at the start of each line, and the identifier should match the filename.

Also, are you sure this isn't [Mk2CockpitExpansion](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/Mk2CockpitExpansion.netkan) moved to a new Spacedock ID? It looks like it might be, and if so, it would be best to keep the old identifier/filename.

politas:
Closing in favour of #5098

PR #5099, opened 2016-12-29
Added metadata for GroundConstruction.

allista:
It is a reference to the actual .netkan file maintained in its repository.


politas:
Looks good. Thanks for supporting the CKAN!

PR #5102, opened 2016-12-30
Move KAX to Curse for KSP 1.2.1 - 1.2.2 and the future

politas:
This should save us work on future updates


Dazpoet:
"x_netkan_epoch": 3 <--not something I see every day

politas:
> "x_netkan_epoch": 3 <--not something I see every day

We'll have more of them now thhat epochs are hidden by default and we're using them to force reinstalls when there are changed dependencies.

Issue #5105, opened 2016-12-31
KerbalHacks/License.txt


pkrasimirov:
About to install...

 * Kerbal Hacks: Droptank "Wrapper" 1.0(cached)

Oh no! We tried to overwrite a file owned by another mod!
Please try a `ckan update` and try again.

If this problem re-occurs, then it maybe a packaging bug.
Please report it at:

https://github.com/KSP-CKAN/NetKAN/issues/new

Please including the following information in your report:

File           : GameData/KerbalHacks/License.txt
Installing Mod : KerbalHacksDroptankWrapper 1.0
Owning Mod     : AnimatedStationScreens
CKAN Version   : v1.22.0-0-g31057f3 (beta)

Your GameData has been returned to its original state.

linuxgurugamer:
Please notify the author of the mod about this

pkrasimirov:
I found this: http://forum.kerbalspaceprogram.com/index.php?/topic/134889-12x-kerbal-hacks-procedural-parts-textures-asphalt-tiles-unusual-parts-other-hacks/&page=6#comment-2895527

**[1.2.x] Kerbal Hacks**
> If you upload them to SpaceDock, I'll make sure the CKAN files get generated correctly.
> 
> These look (in the words of Tony the tiger) GRRRREEEAAAAAATTTTTTTT!!!!!!!!!
>
> 
>
> Oh, and FYI, there are a number of single-part mods which do have their own thread.
> Edited December 21 by linuxgurugamer

Should I write in the forum thread?

pkrasimirov:
@linuxgurugamer, I see you fixed this with #5110, thanks.  Happy New Year!

linuxgurugamer:
Problem was fixed, thanks

PR #5110, opened 2016-12-31
Specifying the install folder

enceos:
to omit the license.txt


pkrasimirov:
This solves #5105, thanks.

PR #5119, opened 2017-01-01
Octosat Continued

linuxgurugamer:
Added new dependency


politas:
Please add an epoch bump to force the new dependency install

linuxgurugamer:
#rebuild

linuxgurugamer:
Added, but don't understand why it was needed

politas:
I'd like to try to avoid the ModuleNotFound Kraken bug by making sure that any change in mod dependencies comes with a forced "upgrade"

PR #5121, opened 2017-01-01
Make the Grindy Sciences conflict and add vrefs

politas:


Dazpoet:
Also both mods depend on modulemanager ;)

politas:
Good catch!

politas:
#rebuild

Dazpoet:
LGTM

PR #5125, opened 2017-01-03
Added name field to avoid conflicts since all pull from the same spac…

linuxgurugamer:
…edock

file
Removed toolbar dependency


linuxgurugamer:
#rebuild

linuxgurugamer:
Currently VesselViewer CANNOT be installed because these two files were missing the name fields, therefore the CKAN.exe was getting confused as to the mod names, etc.

PR #5135, opened 2017-01-05
Vesselview, updated install directory for UI dlls

linuxgurugamer:


politas:
To force CKAN clients to re-install, this will either need a re-release of the mod or a epoch bump.

PR #5140, opened 2017-01-05
Deleted per author's request

### LINKED ISSUE #5073: It appears that Goodspeed and GPOSpeedPump refer to the same mod
I did a quick search for a prior issue regarding this but did not find one. However, if this is a duplicate, feel free to close the issue.

linuxgurugamer:
Closes #5073


Dazpoet:
Maybe we should add a 
```
"provides" : [ "GPOSpeedPump" ],
"comment" : "see https://github.com/KSP-CKAN/NetKAN/issues/5073"
```

So people who somehow have the wrong one installed don't get stuck. I think that a provides field might give them a way of updating in the future?

When I found https://github.com/KSP-CKAN/NetKAN/pull/1105 I was sure I was the one who screwed up but then I saw https://github.com/KSP-CKAN/NetKAN/pull/3649 and will instead blame @politas for not properly knowing about every mod we index before merging 😈 👍

politas:
Mea culpa!

PR #5098, opened 2016-12-29
AoA Technologies

linuxgurugamer:
Closes #5041

PR #5144, opened 2017-01-06
AoATechnologies

linuxgurugamer:
Closes #5098 

Removed the ASET_Props dependency for now, author included it in the mod
ASET_Props is not available at all in CKAN, since kerbalstuff went down. 
If/when it is available again, I'll update this
Same issue with ASETAvionics


politas:
LGTM

PR #5145, opened 2017-01-07
DockingCameraKURSstyle- Remove force-v, add epoch

### LINKED ISSUE #5134: DockingcameraKURSstyle: v1.2.0.2 is wrongly determined by ckan to be more recent than 1.2.1.1
Seems to be the `v` in front of v1.2.0.2 that's causing things to break.

politas:
Closes #5134 
(Hopefully for good this time)


Dazpoet:
otherways it can probably be solved with an overly complex regex

PR #5157, opened 2017-01-10
KIS trad fr.netkan

modFR:
How i know if it's good


modFR:
I am your tutorial
I would like to post on ckan a patch that translates KIS
it does not work ?

modFR:
I need help please
https://spacedock.info/mod/1157/KISFrenchTranslationPatch

PR #5171, opened 2017-01-16
allow WarpEverywhere on KSP v1.2.x

lamont-granquist:
this mod is simple, works on 1.2.0/1.2.1/1.2.2 and is likely to not
need recompiles for patch revs of KSP.

Signed-off-by: Lamont Granquist <lamont@scriptkiddie.org>


politas:

PR #5176, opened 2017-01-17
Zenit3SLBReduzx

linuxgurugamer:


linuxgurugamer:
`8193 [1] FATAL CKAN.NetKAN.Program (null) - Too may .version files located: HorizonAeronautics_V1.32/GameData/AnimatedDecouplers/AnimatedDecouplers.version, __MACOSX/HorizonAeronautics_V1.32/GameData/AnimatedDecouplers/._AnimatedDecouplers.version
Build step 'Execute shell' marked build as failure`


It gets stuck on a .version file in the AnimatedDecouplers folder. I'm not sure why it's looking there tbh since the CKAN-spec clearly states that Only .version files that would be installed for this mod are considered. and the find directive should only work against the top-most matching directory which shouldn't be in the _MACOSX folder since that makes everything one layer deeper. Not sure how to ping on this but I'm guessing @dbent since it seems NetKAN isn't doing what the spec says it should be doing.

politas:
I think that Netkan is downloading and looking for .version files before it gets to the filtering part of the instructions. You can specify the name of the .version file to force the correct file. eg.

    "$vref": "#/ckan/ksp-avc/horizonaeronautics.version/",

linuxgurugamer:
Thanks, I'll fix that, but isn't this non-spec?

linuxgurugamer:
This file doesn't have a .version, so I just removed the $cref line

politas:
The spec was missing some details there. I've updated it. 
Ah, not duplicate .versions, but no .version at all!

politas:
Sorry, meant to merge this earlier, but fatigue drove me off to bed.

PR #5179, opened 2017-01-17
Updated dependancies and added recommends

sswelm:


politas:

PR #5206, opened 2017-01-25
Added conflicts due to issues with Kerbalism for KVV & Hullcam

linuxgurugamer:


politas:
Can you add n `"x_netkan_epoch"` bump to force reinstallation? Changing dependencies like this always causes grief, otherwise.

dbent:
> Can you add n "x_netkan_epoch" bump to force reinstallation? Changing dependencies like this always causes grief, otherwise.

It shouldn't since v1.20.0 and KSP-CKAN/CKAN#1730.

politas:
>It shouldn't since v1.20.0 and KSP-CKAN/CKAN#1730.

Maybe not, but it still does. We've had many reports.

dbent:
So they should upgrade...

politas:
I don't think KSP-CKAN/CKAN#1730 is working in all cases.

dbent:
> I don't think KSP-CKAN/CKAN#1730 is working in all cases.

I'd need to know what cases to fix it. There shouldn't be any situation that a epoch bump would handle that KSP-CKAN/CKAN#1730 wouldn't.

politas:
Have a look through the various issue reports on Stock Visual Enhancements.

PR #5222, opened 2017-01-30
Added recommend line to a number of netkans (mine)

linuxgurugamer:


linuxgurugamer:
This shouldn't need a version bump, it's only adding a "recommends" line, but I'd like someone to verify that it won't cause a problem

politas:
I am almost completely certain that adding recommends to mods will not cause any problems. CKAN cannot tell the difference between a newly added recommend and a recommend you have chosen to ignore.

PR #5226, opened 2017-02-01
Create KISFrenchTranslationPatch

modFR:
Translation patch for the mod KIS (Kerbal Inventory System)

Installation:

Step 1 : Just Drag and Drop the content from the Gamedata Folder (The one who is in Zip file of course) into your KerbalSpaceProgram/Gamedata Folder.

Step 2 : Enjoy !

PS: This is the first version of this translation patch so if you see some mistakes let me know :) ===> ksptranslationfr@gmail.com

Patch made by yrou ( http://forum.kerbalspaceprogram.com/index.php?/profile/145128-yrou/ ) and translated in french by me ( http://forum.kerbalspaceprogram.com/index.php?/profile/168212-iceolator88/ )

### DIFF
--- a/KISFrenchTranslationPatch
+++ b/KISFrenchTranslationPatch
@@ -1,0 +1,7 @@
++{
+ +    "spec_version": "v1.4",
+ +    "x_via": "Automated SpaceDock CKAN submission",
+ +    "license": "All Rights Reserved",
+ +    "identifier": "KISFrenchTranslationPatch",
+ +    "$kref": "#/ckan/spacedock/1157"
+ +} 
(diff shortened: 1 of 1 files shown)


KISFrenchTranslationPatch
```diff
@@ -0,0 +1,7 @@
+{
+    "spec_version": "v1.4",
+    "x_via": "Automated SpaceDock CKAN submission",
+    "license": "All Rights Reserved",
```
politas: Will need to be `"restricted"` rather than "All Rights Reserved".
Also, this should depend on KIS, please add the following, making sure that all but the last stanza have commas at the end:

    "depends": [
        { "name": "KIS" }
    ]

Olympic1: pinging @modFR

PR #5228, opened 2017-02-02
updated version for SVE due to update

linuxgurugamer:

PR #5229, opened 2017-02-02
Update SVE Scatterer Config and Sunflare

dbent:
- Filename changed so change `asset_match` to match on anything with "scatterer" in it (case-insensitively)
- Remove duplicate `abstract` properties

Should close #5228


dbent:
Tested locally, merging.

PR #5231, opened 2017-02-03
Test Flight Continued

linuxgurugamer:


Olympic1:
The RO team is maintaining this, they're not so active on it but they released a beta on January 5th.
https://github.com/KSP-RO/TestFlight

Blu3wolf:
Might it be worth adding a conflicts with for TestFlight itself if that is still going to be updated?

linuxgurugamer:
There already is a conflicts in there for TestFlight

Blu3wolf:
Im clearly blind! Whoops

PR #5232, opened 2017-02-03
fix mess with HazardTanksTextures and VensStylePPTextures

TeddyDD:
Those mods depends on Procedural Parts. I also added missing resource to HazardTanksTextures.

HazardTanksTextures dependency on VensStylePPTextures is hack since both have one common file that should be extracted to separate package. I'm not sure if it's worth a trouble.

### DIFF
--- a/NetKAN/HazardTanksTextures.netkan
+++ b/NetKAN/HazardTanksTextures.netkan
@@ -5,11 +5,12 @@
     "license": "CC-BY-NC-4.0",
     "spec_version": "v1.4",
     "ksp_version": "any",
+    "resources": {
+        "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/134889-KerbalHacks"
+    },
     "depends": [
-        { "name": "VensStylePPTextures" }
-    ],
-    "suggests": [
-            { "name": "ProceduralParts" }
+        { "name": "VensStylePPTextures" },
+        { "name": "ProceduralParts" }
     ],
     "provides": [
         "PPTextures"
--- a/NetKAN/VensStylePPTextures.netkan
+++ b/NetKAN/VensStylePPTextures.netkan
@@ -7,7 +7,7 @@
     "resources" : {
         "homepage" : "http://forum.kerbalspaceprogram.com/index.php?/topic/134889-KerbalHacks"
     },
-    "suggests": [
+    "depends": [
         { "name": "ProceduralParts" }
     ],
     "install": [


linuxgurugamer:
Seems that VensStylePPTextures is not yet marked as 1.2.2 compatible, I've pinged @enceos about it

enceos:
It's not a mess. Procedural Parts is not 1.2.2 compatible, people can't install it through CKAN and have to do it manually. I removed the dependency on Procedural Parts until it is updated. For now those who want Hazard Tanks will have to download VensStylePPTextures as well.

TeddyDD:
> people can't install it through CKAN

I can install VensStylePPTextures and HazardTanksTextures via CKAN but THEY WILL NOT WORK since they depends on Procedural Parts. This is wrong and will confuse users.

zrzut ekranu z 2017-02-06 15-01-39

> I removed the dependency on Procedural Parts until it is updated.

It's called making a mess. You assume that user install Procedural Parts manually and then install textures via CKAN and this should never be the case.

> For now those who want Hazard Tanks will have to download VensStylePPTextures as well.

That's what I said. But techincally its not true. You can download Hazard Tanks and istall it manually without VensStylePPTextures. I also pointed what could be done to make this possible in CKAN but I think it's not worth the trouble.

enceos:
@TeddyDD  People who use Procedural Parts usually know their way around mods. You certainly do.

TeddyDD:
> People who use Procedural Parts usually know their way around mods.

That's just unfounded claim. We don't know who will try to install this packages (keyword: usually) Also following this way of thinking, why don't you mark Procedural Parts as 1.2.2 compatible? If we can have broken packages with removed dependencies in repo then we can have half working, incompatible mods as well. 

Advanced user can easily download .ckan files, edit and install them via CKAN. But please don't mess with end users because its convenient for you.

BTW:
 I wrote netkan.exe wrapper that makes extremely easy to create and maintain local ckan repo and pipe custom packages to CKAN client. I might release it in near future.

enceos:
I chose the way which is acceptable for the majority of users and don't care if someone doesn't visit my thread to get info on what they're actually installing.

TeddyDD:
What worries me is that CKAN allows broken metadata nowadays 💢 
Well, that's your mods so I do not have nothing more to say for now.

Maintainers: Fell free to close this PR

TeddyDD:
@linuxgurugamer https://github.com/TeddyDD/kure It has no docs just lame asciicast but It's quite simple to use. If you have any question feel free to ask.

Olympic1:
Closing on request

PR #5236, opened 2017-02-04
Fixes #5235 Stock Visual Terrain

### LINKED ISSUE #5235: Stock Visual Terrain not up to date in CKAN - NetKAN issue?
SVT no longer has low and high resolution downloads. Could be why CKAN isnt getting the new release?

Perhaps #4812 could just be reverted?

Blu3wolf:
#5235 not sure if this is an appropriate identifier for the mod? Guidance would be appreciated.


Blu3wolf:
I see Ive been beaten to the punch by a day, should have checked both Issues AND Pull Requests... looks like @linuxgurugamer s version is better than mine, covers the mac as well.

Issue #5256, opened 2017-02-11
Bad netkan file: ASETProps-CE.netka


linuxgurugamer:
there is a bad netkan:  NetKAN/NetKAN/ASETProps-CE.netkan, the mod was deleted from Spacedock when the full ASETProps was released.  Would deleting this or renaming it be the right thing to do?

techman83:
Looks like @Dazpoet sorted it, freezing is usually the way to go. Avoids future identifier clashes.

linuxgurugamer:
Yup. Thanks

PR #5257, opened 2017-02-11
Rename ASETProps-CE.netkan to ASETProps-CE.frozen

Dazpoet:
Mod no longer exists on SD based on #5256 

Freezing it incase it should ever return


techman83:
Thanks @Dazpoet

PR #5233, opened 2017-02-03
For new versions of SVT

linuxgurugamer:
New files, new directory layout

PR #5267, opened 2017-02-13
Updated files for SVT

linuxgurugamer:
Closs #5233


linuxgurugamer:
@politas  & @Dazpoet 
Please take a look, I took a different path for the regex, it seems to work when I test using a local netkan.exe.  Given the regex, I really would like someone else to double-check me on this

PR #5321, opened 2017-03-19
Add RealSolarSystem Visual Enhancements.

zilti:


benjwgarner:
According to the console log, the issue is that it can't find any releases on GitHub.

`1115 [1] WARN CKAN.NetKAN.Transformers.GithubTransformer (null) - No releases found for PhineasFreak/RSSVE`

It appears that the [releases on GitHub](https://github.com/PhineasFreak/RSSVE/releases) do not have a zip file uploaded for them, only the default "source code" archives automatically generated from the repository. @zilti To fix this, you need to edit the releases and add upload an additional zip file (which in this case will contain the same contents), then get the check to run again.

@politas I've seen this problem come up a few times before. Do you think it would be worth the effort to modify NetKAN so that it looks in the default source zip on GitHub if the release doesn't contain an additional zip file?

politas:
Well, in this case, they're all marked as pre-releases, so that wouldn't make a difference for this particular mod.

Olympic1:
#rebuild

PhineasFreak:
I have been getting false "bug reports" about Scatterer not working. **[RSSVE requires specific mod versions that are documented in the RSSVE wiki](https://github.com/PhineasFreak/RSSVE/wiki/Installation)** and not all mod versions will work with it.

Thank you for your time.

Issue #4131, opened 2016-05-23
Unlicense-licensed mods


plague006:
When CKAN 1.18 is released we will gain the `Unlicense`. The following mods currently use that license (not necessarily extensive):

- [x] ~~[ContractConfigurator-SCANsat](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/ContractConfigurator-SCANsat.netkan)~~ "            DO WHAT THE F YOU WANT TO PUBLIC LICENSE"
- [x] [ControlLock](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/ControlLock.netkan)
- [x] ~~[CoolRockets](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/CoolRockets.netkan)~~ "This work was released in the public domain but dtobi"
- [x] ~~[DangerAlerts](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/DangerAlerts.netkan)~~ MIT
- [x] [EasyVesselSwitch](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/EasyVesselSwitch.netkan)
- [x] [ExceptionDetector](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/ExceptionDetector.netkan)
- [x] ~~[FacilityReset](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/FacilityReset.netkan)~~ "Source and icon art are released as Public Domain."
- [x] ~~[HarderSolarSystem-1x](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/HarderSolarSystem-1x.netkan)~~ "This is public domain. Do whatever you wish with what I have released."
- [x] ~~[HarderSolarSystem-2x](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/HarderSolarSystem-2x.netkan)~~ "This is public domain. Do whatever you wish with what I have released."
- [x] ~~[HarderSolarSystem-3x](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/HarderSolarSystem-3x.netkan)~~ "This is public domain. Do whatever you wish with what I have released."
- [x] ~~[HarderSolarSystem-4x](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/HarderSolarSystem-4x.netkan)~~ "This is public domain. Do whatever you wish with what I have released."
[...]

PR #5322, opened 2017-03-19
Adding "1869 A Steampunk Mod"

zilti:


benjwgarner:
The check is failing with this error:

`NetKAN/1869.netkan: line terminator characters must be escaped inside string literals: u'\n\nStill in beta,...`

You might be able to escape the new line characters, but it's better to just leave it all on one line. CKAN will wrap the text when it displays it. Since the text display area is resizable and not the same size for everyone, it's better to just keep it all on one line.

benjwgarner:
Since support for the Unlicense was not part of the original spec (see #4131), you need to change the spec_version to 1.18.

`#   Failed test '1869 - spec_version v1.18+ required for license 'Unlicense''`

zilti:
Fixed.

linuxgurugamer:
The format for the spec version is wrong.  it should be as follows:

 "spec_version": "v1.18",

zilti:
Fixed. Maybe stuff like this should be added to the manual.

PR #5323, opened 2017-03-19
Adding The Maritime Pack mod

zilti:


benjwgarner:
According to the console log:

`4718 [1] ERROR CKAN.RelationshipResolver (null) - Dependency on 1869 found but it is not listed in the index, or not available for your version of KSP.`
`Module 1869 required but it is not listed in the index, or not available for your version of KSP.`

@zilti The check failed because 1869 is listed as a dependency for TheMaritimePack, but it is not yet indexed in CKAN. You need to fix the [line splitting problem](https://github.com/KSP-CKAN/NetKAN/pull/5322) in 1869 and wait for it to be indexed by CKAN. Once it has been picked up, TheMaritimePack should (unless there are any other problems) pass the checks.

linuxgurugamer:
#rebuild

zilti:
So there's still something wrong with the file?

politas:
A Corrupted Download, apparently. Lets try it again.

politas:
My local testing is successful. Merging to see if the bot will handle it properly

politas:
TheMaritimePack-019.ckan is in the repo now!

PR #5341, opened 2017-03-24
Path Change - USAFOrion

TiktaalikDreaming:
For reasons I've yet to determine, this mod's plugin fails to work with the existing folder structure.  Fixing the mod required moving it's install location.  Which, from a netkan point of view is a giant pain in the posterior.


TiktaalikDreaming:
Thanky

PR #5353, opened 2017-03-29
Create StationKeeping.netkan

swilbur:
A mod for precisely putting satellites in synchronous orbits.
http://forum.kerbalspaceprogram.com/index.php?/topic/158348-122-stationkeeping-precise-synchronous-orbits/

### DIFF
--- a/NetKAN/StationKeeping.netkan
+++ b/NetKAN/StationKeeping.netkan
@@ -1,0 +1,8 @@
+{
+    "spec_version" : 1,
+    "identifier"   : "StationKeeping",
+    "$kref"        : "#/ckan/github/swilbur/StationKeeping",
+    "license"      : "CC-BY-4.0",
+    "ksp_version"  : "1.2.2",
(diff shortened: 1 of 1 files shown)


NetKAN/StationKeeping.netkan
```diff
@@ -0,0 +1,8 @@
+{
+    "spec_version" : 1,
+    "identifier"   : "StationKeeping",
+    "$kref"        : "#/ckan/github/swilbur/StationKeeping",
+    "license"      : "CC-BY-4.0",
```
politas: Please add an abstract describing the mod, such as:

    "abstract": "Adds a button in the Tracking Station to set a ship's orbit. If you are with 1% (configurable) of your target orbit, you can simply push the button to set the orbit's semi-major axis to exactly what you want, so it won't drift out of alignment as you time warp.",

Olympic1: pinging @swilbur

PR #5354, opened 2017-03-30
Create PoodsOPMVO.netkan

Poodmund:
.netkan file for Pood's Outer Planets Mod - Visual Overhaul.

I did notice that the Outer Planets Mod is currently listed as KSP Max Version 1.2.1 and my mod states a requirement of 1.2.2. Outer Planets Mod works fine and is compatible with KSP 1.2.2 and should be updated to reflect this. I will raise a separate issue.

### DIFF
--- a/NetKAN/PoodsOPMVO.netkan
+++ b/NetKAN/PoodsOPMVO.netkan
@@ -13,7 +13,7 @@
 		{ "name"     : "Kopernicus", "min_version" : "2:release-1.2.2-5" },
 		{ "name"     : "OuterPlanetsMod", "min_version" : "1:2.1" },
 		{ "name"     : "EnvironmentalVisualEnhancements", "min_version" : "2:EVE-1.2-2" },
-		{ "name"     : "scatterer", "min_version" : "2:v0.0300" }
+		{ "name"     : "Scatterer", "min_version" : "2:v0.0300" }
 	],
 	"install"        : [
 		{ "find"     : "PoodsOPMVO", "install_to" : "GameData" }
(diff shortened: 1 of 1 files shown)


Poodmund:
I think this is all fine to go except for the Outer Planets Mod being specified for KSP 1.2.1.

politas:
#rebuild

Poodmund:
It looks like the Outer Planets Mod PR is still failing as the CommunityTerrainTexturePack is not being verified as compatible with 1.2.2 before the OPM verification pass. Doh!

politas:
Yep. Chasing the rabbit, here...

politas:
#rebuild

politas:
EVE Min version - you've set it as 2:EVE-1.2.2, but the actual version is 2:EVE-1.2**-**2

Poodmund:
@politas  it took me quite a while to mull over what you were saying here and what was wrong. I put mine version as redundancy if EVE gets updated and the metadata for this mod does not. Then the end users will be able to update the dependencies. 
Does CKAN not look at a min version reference as a greater than or equal to way?
Basically I've specified the current mod dependencies as min versions to ensure anyone using this mod is using the current latest builds of the dependencies. Is this not the way to go about it?

Poodmund:
Got there in the end. Was calling scatterer "scatterer" as the lower case 's' is used everywhere else for the mod file name wise but its listed here with an upper case 's'. Just UNIX things I guess.

Anyway, should be good to go. :D

politas:
Looks all good to me, now. Merging with many thanks for the persistence!

PR #5358, opened 2017-04-01
Outer Planets extend to 1.2.2

politas:
Should enable #5354 to pass checks


politas:
jenkins is complaining that CustomBarnKit is not available for 1.2.1

politas:
#rebuild

PR #5365, opened 2017-04-02
Adding HeatPump mod.

zilti:


politas:
Looks ok to me, assuming there are no conflicts with the GingerCorp Agency (I can't find any on a quick search). Merging with thanks!

PR #5380, opened 2017-04-06
Update SearchAndRescue

Olympic1:
New version


politas:
Version 9 to version 4? That'll need an epoch.

Olympic1:
I added the epoch the minute after: https://github.com/Olympic1/NetKAN/commit/0e04b1c0488f4a3a4acc4fb200c157320d92a55b

PR #5391, opened 2017-04-08
Use regex to find correct kRPC

### LINKED ISSUE #5387: kRPC 0.3.8.post1 NetKAN metadata
A small update was pushed to the kRPC repository but the changes were interpreted as a mod update from the NetKAN bot: https://github.com/KSP-CKAN/CKAN-meta/commit/313f07482a29e4da45c8a6c24ea447c5a8360a7b

This replaces the entire mod content with python source code files that have nothing to do with the actual mod (well, they are used to create the python kRPC server and client but they are not meant to be distributed).

Olympic1:
Only download the zip with the correct contents. In this case it should download `krpc-0.3.8.zip` and **not** `krpc-python-0.3.8.post1.zip`.

This regex code works, but can it be made simpler?

Fixes #5387

### DIFF
--- a/NetKAN/kRPC.netkan
+++ b/NetKAN/kRPC.netkan
@@ -1,7 +1,7 @@
 {
     "spec_version" : 1,
     "identifier"   : "kRPC",
-    "$kref"        : "#/ckan/github/krpc/krpc",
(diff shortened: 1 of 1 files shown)


Olympic1:
hmm, download field is correct but ckan downloads the latest version

PR #5413, opened 2017-04-18
Changed "UpgradesUIExtensions" plugin name to "UpgradesGUI"

gotmachine:
I changed my plugin name (I know, messy idea...) from "UpgradesUIExtensions" to "UpgradesGUI".


linuxgurugamer:
Would it be better to create a new .netkan file and freeze the old one?

PR #5414, opened 2017-04-18
SVE Change of packaging again

politas:
I hope that setting a conflict on a version range works. Don't know whether this PR will test properly.

### DIFF
--- a/NetKAN/StockVisualEnhancements.netkan
+++ b/NetKAN/StockVisualEnhancements.netkan
@@ -1,0 +1,38 @@
+{
+    "spec_version"   : "v1.4",
+    "$kref"          : "#/ckan/github/Galileo88/StockVisualEnhancements/asset_match/SVE.v.*.zip",
+    "$vref"          : "#/ckan/ksp-avc",
+    "identifier"     : "StockVisualEnhancements",
+    "name"           : "Stock Visual Enhancements",
+    "abstract"       : "A visual enhancement pack using EVE and Scatterer for the stock planets. ",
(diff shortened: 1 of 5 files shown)


NetKAN/StockVisualEnhancements.netkan
```diff
@@ -0,0 +1,38 @@
...
+    ],
+    "depends": [
+        { "name": "SVE-Textures" },
+        { "name": "Scatterer" },
+        { "name": "EnvironmentalVisualEnhancements" },
+        { "name": "ModuleManager"}
+    ],
+    "recommends": [
+        { "name": "DistantObject" },
+        { "name": "PlanetShine" },
```
Olympic1: remove comma

Olympic1:
Build successful, so I think the range is fine, will do some testing though. Other then that looks good.

Olympic1:
After some extensive testing it doesnt like the version range. If I deleted them, it works.

The download also includes a `SVE-Sunflare` config, so that would need to be added as a `provides` too.

politas:
It works without the ranges on the conflicts? But it depends on one of those! I don't think we can depend on a version range for a metapackage, either.

Olympic1:
I think it needs the scatterer configs/sunflares, but I'm not sure. It modifies the sunflare one.
@Galileo88 can you elaborate.

Galileo88:
Yes, it needs scatterer and its sun flare.

politas:
@Galileo88 , so the SVE_Scatterer folder is providing a sunflare configuration for Scatterer?

PR #5421, opened 2017-04-20
Added epoch to ETT

linuxgurugamer:
Looks like he sometimes forgets to add the "v" to the version number


politas:
And sometimes he puts dashes in the date format and sometimes not...

PR #5422, opened 2017-04-20
Reactivation of CollisionFX netkan file

Momo555:


Momo555:
Ok added the missing $vref and resources field with the homepage URL. Should I specify also a "name" and add the github link in the resources field?

Olympic1:
The name of the repository will be used for the mod name and the github link is added automatically.

PR #5435, opened 2017-04-26
Update SSRSS

Olympic1:
* Updated SSRSS
  * Added RSS and SSRSS-Rescale as dependencies
  * Added recommendations for the new options
  * Added suggestions for mod support
* Added `SSRSS-Rescale` as a 'fake' SigmaDimensions
* Added `SSRSS-StockIncl` to install stock inclinations for SSRSS
* Added `SSRSS-Clouds` to install Low and High Res clouds

---

This PR will fail if using `find` because CKAN doesn't like installing folders that only have folders. This is the current layout of the zip:
```
GameData
  | SSRSS
      | configs
        - files
      | visuals
          | more folders
            - files
```
It fails because it doesn't detect any files in **SSRSS** but only folders. If @Galileo88 could include/move a file in the **SSRSS** folder, it will detect a file and install the mod.

We should change [InstallsFilesValidator.cs](https://github.com/KSP-CKAN/CKAN/blob/master/Netkan/Validators/InstallsFilesValidator.cs) so it can detect folders with files too as a valid install.

### DIFF
--- a/NetKAN/SSRSS-Clouds-HR.netkan
+++ b/NetKAN/SSRSS-Clouds-HR.netkan
@@ -1,0 +1,27 @@
+﻿{
+    "spec_version": "v1.4",
+    "identifier": "SSRSS-Clouds-HR",
+    "$kref": "#/ckan/github/Galileo88/SSRSSContinued",
+    "$vref": "#/ckan/ksp-avc",
+    "name": "Stock Size Real Solar System - High Res Clouds",
+    "abstract": "Clouds for Stock Size Real Solar System.",
+    "license": "CC-BY-NC-ND",
+    "provides": [
+        "SSRSS-Clouds"
+    ],
+    "conflicts": [
+        { "name": "SSRSS-Clouds" }
+    ],
+    "depends":
+    [
+        { "name": "SSRSS-Cont" },
+        { "name": "EnvironmentalVisualEnhancements" }
+    ],
+    "install":
+    [
+        {
+            "find": "SSRSSClouds_HighRes",
+            "install_to": "GameData/SSRSS"
+        }
+    ]
+}
--- a/NetKAN/SSRSS-Clouds-HR.netkan
+++ b/NetKAN/SSRSS-Clouds-HR.netkan
@@ -7,10 +7,12 @@
     "abstract": "Clouds for Stock Size Real Solar System.",
     "license": "CC-BY-NC-ND",
     "provides": [
-        "SSRSS-Clouds"
+        "SSRSS-Clouds",
+        "EnvironmentalVisualEnhancements-Config"
     ],
     "conflicts": [
-        { "name": "SSRSS-Clouds" }
+        { "name": "SSRSS-Clouds" },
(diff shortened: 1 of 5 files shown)


politas:
Have we got SigmaBinary back, yet? Last I heard, we still needed author approval to re-add that.

Olympic1:
@Sigma88 wants to keep them off CKAN atm, so no we don't have it back. But SSRSS needs SigmaDimensions because it uses the mod to resize the solar system what is the purpose of SSRSS.

Sigma88:
I asked @Olympic1 to not add my mods to CKAN because I don't have much time to maintain my mods at the moment and I would not be able to sustain the greater "public" CKAN would expose my mods to

however, I can give you permission to install SD through CKAN only alongside SSRSS

I guess it is possible for you to make this (SD would need to be uninstalled when uninstalling SSRSS)

I'd also like if SD was not listed as a mod, but rather installed as a "module" of SSRSS so bug reports would go through SSRSS devs first and reach me only when the issue is actually related to SD

if all this is possible, you have my permission to install SD with SSRSS

if there are other mods that are currently blocked by SD being off CKAN let me know, I'll take a look and let you know for which one you can do the same

Sigma88:
pinging @Galileo88 to make sure he is aware of this

politas:
Ok, so given sigma88's permission, we sshould move SigmaDimensions to a "conflicts", and add an install stanza. 
As for the failing empty folder install, we can get around that by installing the subfolders into the SSRSS folder, which might be a better solution if all these mods are in a single file structure in the zip.

Sigma88:
I'm not sure what this means, but if you plan to install SigmaDimensions inside the SSRSS folder that probably won't work

Galileo88:
I'm not sure I understand what you mean when you say I have empty folders? I am pretty certain I laid everything out correctly for eventual CKAN compatibility. I will be glad to change anything so long as I understand what needs to happen :)

politas:
@Galileo88 , nothing you need to worry about. We've got an issue in Netkan where it doesn't count files in subfolders and sees an install target as empty when it contains only folders.

Olympic1:
@politas I solved it with using `SSRSS - Rescale`.
So if this way is ok by @Sigma88 , this can be merged.

Sigma88:
you have my blessing :)

politas:
Sneaky! I like the Dimensions solution! :-) 
Other than that, I may be off base, but I think SSRSS-Cont should be providing EnvironmentalVisualEnhancements-Config

Olympic1:
Totally overlooked that. Added it.

politas:
Cool. Merging!

politas:
Actually, now that I have slept on it (accidentally), I wonder  whether it might be the SSRSS-Clouds-*  that are providing "EnvironmentalVisualEnhancements-Config". @Galileo88, can you advise?

Olympic1:
Ok, I looked at every file and this should be the correct state of the EVE/Scatterer-configs.
@politas so this can be merged if nothing needs to be changed.

politas:
Looks good to me.merging!

PR #5442, opened 2017-04-27
SSRW 1.12 is a 1.2.x release

lamont-granquist:
There's one update to SSRW that works on 1.2.x

Set the upper bounds to 1.2.8 since 1.2.9 seems to be the
1.3.0 pre-release, dunno if you care, but 1.2.99 now seems
inappropriate...


politas:
Yeah, the 1.2.9 pre-release is a nightmare for us.

politas:
Looks all good to me. Merging with thanks!

PR #5443, opened 2017-04-29
CoyoteSpaceIndustries Cargo Freighter

linuxgurugamer:


dboi88:
Hey, this doesn't install properly if you've already got USI-Core installed (bundled with this mod) should it not have USI-Core listed as a dependency?
Next release will allso bundle CKK

PR #5446, opened 2017-04-30
Telemachus for KSP 1.2

goncalopp:
Hello,

This is my first CKAN metadata change, and I'm not very familiar with CKAN, so it's likely I botched this change :)

The latest [Telemachus versions](https://github.com/KSP-Telemachus/Telemachus/releases) seem to support KSP 1.2. Will this change on `NetKAN` also update existing metadata on `CKAN-meta`? Or will it only be applied to new versions?

### DIFF
--- a/NetKAN/Telemachus.netkan
+++ b/NetKAN/Telemachus.netkan
@@ -8,8 +8,8 @@
     "resources"    : {
         "homepage" : "http://forum.kerbalspaceprogram.com/index.php?/topic/22623-104-2015-07-04-telemachus"
     },
-    "ksp_version_min" : "1.0.4",
-    "ksp_version_max" : "1.2",
+    "ksp_version_min" : "1.2.0",
+    "ksp_version_max" : "1.2.2",
     "license"      : "restricted",
     "x_netkan_force_v": true
 }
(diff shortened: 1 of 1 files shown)


NetKAN/Telemachus.netkan
```diff
@@ -9,7 +9,7 @@
         "homepage" : "http://forum.kerbalspaceprogram.com/index.php?/topic/22623-104-2015-07-04-telemachus"
     },
     "ksp_version_min" : "1.0.4",
-    "ksp_version_max" : "1.0.5",
+    "ksp_version_max" : "1.2",
```
Olympic1: Could you set `ksp_version_min` to 1.2.0 and `ksp_version_max` to 1.2.2

Olympic1:
@goncalopp You did a good job on your first metadata change, but I requested some changes.

NetKAN/Telemachus.netkan
goncalopp: Done.
Regarding "1.2.2": is it CKAN's philosophy to only set metadata to actually tested versions?
I thought it would be likely this will keep working for 1.2.3, if there's such a thing, but I guess we can't know before it comes out

goncalopp:
Great, so this will start working as soon as it's reindexed :) Thanks for the clarification

Olympic1:
When our bot is checking on changes and updates, it will detect this change and apply it to the latest version of Telemachnus.

PR #5470, opened 2017-05-04
add ForScience KSP 1.2.x version

lamont-granquist:
this was updated for KSP 1.2.x but apparently nobody updated the
metadata.

also updated the license per the link on the top post in the
forum thread:

http://forum.kerbalspaceprogram.com/index.php?/topic/107803-12-forscience-v141-your-science-autopilot/&


politas:
Thanks for the heads-up. I think we might move this to automated discovery so we don't miss the next update.

PR #5471, opened 2017-05-04
For Science switch to automated discovery 

politas:
Closes #5470


politas:
Hmm. The Curse transformer is not replacing spaces in URLS with %20. That should be easy to fix.

politas:
#rebuild

Olympic1:
LGTM, good catch for the curse links

PR #5482, opened 2017-05-05
update targetron kref to spacedock

hashashin:


Olympic1:
LGTM, merging with my thanks

PR #5488, opened 2017-05-06
Add MKS Explainer

Gnurfos:
Untested. I abandonned the quest for a working netkan.exe after a relatively long time, but the json is quite simple so maybe it just works.


politas:
Looks fine to me. It's picking up the "MKS Explainer" name from Spacedock, so that's how it will show in the CKAN mod list.
Merging with many thanks!

PR #5517, opened 2017-05-23
Fix FAR

Olympic1:
Ref: https://github.com/KSP-CKAN/CKAN-meta/pull/1261

### DIFF
--- a/NetKAN/FerramAerospaceResearch.netkan
+++ b/NetKAN/FerramAerospaceResearch.netkan
@@ -32,5 +32,15 @@
             "optional"    : true,
             "description" : "FAR example craft"
         }
+    ],
+    "x_netkan_override": [
+        {
+            "version": "3:0.15.8",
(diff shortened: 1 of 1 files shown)


dbent:
D'oh, assumed this would be handled by the remote AVC file since it now contains the correct KSP versions, but the version file in the download is wrong (0.16.0.0) so it doesn't match the remote (0.15.8.0) and NetKAN refuses to use it.

dbent:
Also @techman83, shouldn't FAR have gone to staging instead of being automatically committed?

politas:
Yes, it looks like staging is not working properly. I don't think it ever has.

techman83:
Because @techman83 seemed to miss the convention of an underscore vs dash. So it works, but our metadata conventions are different to how I've coded the metadata parser.

Raised an issue KSP-CKAN/NetKAN-bot#60

Simple one to fix.

PR #5535, opened 2017-05-27
Update Kopernicus

Olympic1:
Kopernicus now uses an avc file.


dbent:
> Kopernicus now uses an avc file.

😀

PR #5536, opened 2017-05-27
Make OPM recommend, not require, CBK.

Starstrider42:
Custom Barn Kit is not compatible with some other mods and is not strictly required for gameplay, so it should be recommended, not required. CBK is needed only to increase the communication range of the tracking station in order to reach the outer planets. The same goal can be accomplished by using mods like [Extended Antenna Progression](https://github.com/LouisB3/KSP-ExtendedAntennaProgression), or just building a really overpowered relay satellite.

Pinging @CaptRobau, since the final decision should be his.


Olympic1:
Pinging @Galileo88 too, as he maintains an updated version of OPM.

Galileo88:
Yeah I can make that change

PR #5538, opened 2017-05-28
Add kerbalsimpit

phardy:
Adds my Kerbal Simpit mod.

I couldn't find any docs on what the `x_netkan_license_ok` field is expected to do, but all of the metanetkan entries I've checked have it, and my mod is licensed under simplified BSD so hopefully that's OK too.

### DIFF
--- a/NetKAN/KerbalSimpit.netkan
+++ b/NetKAN/KerbalSimpit.netkan
@@ -1,0 +1,5 @@
+{
+  "spec_version": "v1.16",
+  "identifier": "KerbalSimpit",
+  "$kref": "#/ckan/netkan/https://bitbucket.org/pjhardy/kerbalsimpit/downloads/KerbalSimpit.netkan"
+}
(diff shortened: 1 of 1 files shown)


phardy:
That's... unexpected validator output. I've been using this netkan file to generate CKAN files locally for a little while now, and the client seems to deal with that OK. Should I be setting ksp_version_strict to something else?

    Validating built/KerbalSimpit-1.0.0.36.ckan.. Failed! See below for error description.
    u'true' is not of type u'boolean'
    
    Failed validating u'type' in schema[u'properties'][u'ksp_version_strict']:
        {u'description': u'Optional enforcement of strict version checks (defaults to false)',
         u'type': u'boolean'}
    
    On instance[u'ksp_version_strict']:
        u'true'

phardy:
Oh. It should be a boolean not a string. 😢

Olympic1:
Looks good, merging

@phardy `x_netkan_license_ok` is used if a license is not in our database or if its written differently. Ex: CKAN uses `GPL-3.0` but author writes it as `GPLv3` then you need to use `x_netkan_license_ok`

phardy:
@Olympic1 Ah, cool. Thanks for the info, and the merge!

PR #5560, opened 2017-05-30
Add Sigma mods back

Olympic1:
I got permission from @Sigma88 to add his mods back to CKAN. I also added some non-sigma mods in this PR to satisfy the dependencies for his mods. Just waiting on the 1.3 releases.

* Unfroze
    * Human Colored Faces - **Unfroze to satisfy dep**
    * Sigma Binary
    * Sigma Dimensions
    * Sigma Eve Recolor
* Added
    * Kronometer
    * Revamped Stock Solar System - **Added to satisfy dep**
    * Stockalike Neptune - **Added to satisfy dep**
    * Science Def Renamer
    * Stockalike Solar System


Sigma88:
👍 

worth noting Human Colored Faces is not one of mine?

also, some of those you've added are in a semi-abandoned state (like Kerbin Recolor) and most likely don't work in the current version of the game

I don't think it matters, just thought I'd mention it

Olympic1:
> worth noting Human Colored Faces is not one of mine?

After I added your mods, I focused on the dependencies for SASS. So if there are mods in this PR that are not yours, its because they are needed as a dependency.

Sigma88:
nice!

Dazpoet:
#rebuild

politas:
#rebuild

politas:
@Olympic1 , any chance you could pull SigmaBinary-Core and SigmaDimensions out into a separate PR to get them sorted out quickly?

Olympic1:
This PR is mainly failing because of the `StockalikeSolarSystem` dependencies, they are not all updated for 1.3. We can merge this and I can update the deps for 1.3 later.

Sigma88:
idk what your policy is about installing outdated mods, but SASS provides all the changes needed to make the mod work in 1.3.0 even if the dependancies are outdated.

it actually requires those exact versions to be installed, if a version of those mods was to come up compatible with 1.3.0 it wouldn't probably be compatible with SASS

politas:
Oh, wow. That's really tricky to code in the CKAN spec.

Sigma88:
I don't expect you to.
if that's not possible I have zero issues with it.

just make sure that you don't add newer versions of those mods if they happen to be released, because I have no way of knowing if that would work

politas:
Are the specified versions included in the Zip file? We could install them as part of SASS and set a conflict instead of a depends, but it might mess up other mods' dependencies if they are more widely desired.

Sigma88:
no I don't redistribute other mods, I assume the best way to approach this would be to create definitions for those mods but specific to SASS

it's pointless to discuss this now since SASS is not up to date with KSP 1.3.1 and I don't give support for outdated KSP version, which measn adding it to CKAN would only be detrimental

as soon as I get a version working on 1.3.1 we can fix the metadata

politas:
Cool.

Sigma88:
another thing, SASS and GN both require a "simple" installation of all planet packs

which means only the basic stuff, without all the extra EVE/Scatterer/Clouds stuff which are usually stored on a different folder

I assume this can easily be setup in CKAN, since it's as simple as copying just the correct folder into GameData

but I haven't check to see if those mods are already split in modules or if all folders are provided by the same ckan module

Sigma88:
ok, SASS is released, I'll start whip up a set of metadata files and I'll pass it through you as soon as I get it done

Sigma88:
I can't remember where the GN discussion was, did we end up deciding what to do about that?

I was looking at how planet packs are installed and some of them would require a different installation for GN

I am willing to manage the metadata for those mods in case you want me to

PR #5563, opened 2017-05-31
replaced dependancy on InterstellarFuelSwitch by InterstellarFuelSwitch-…

sswelm:
…Core

in order to prevent mod conflict, reduced dependency requirement

### DIFF
--- a/NetKAN/KSPInterstellarExtended.netkan
+++ b/NetKAN/KSPInterstellarExtended.netkan
@@ -18,14 +18,15 @@
     ],
     "depends" : [
         { "name"        : "CommunityResourcePack", "min_version" : "0.6.5" },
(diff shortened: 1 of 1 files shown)


sswelm:
As expected, Fails on CommunityTechTree because there it is not updated to KSP 1.3 yet. I think this can be ignored since CTT is just a config mod

politas:
The min version for CTT will not work. It'll need to be 1:3.0.3

sswelm:
ok, lets try it, but what does 1:3.0.3 mean?

politas:
I means it has an epoch due to a historical discontinuity in version numbering systems. In this case, there was a accidental 2.4.0 that thought it came after the real 2.4 #4071 has the details.

PR #5565, opened 2017-05-31
Set DPAI to correct ksp version

Olympic1:
Fixes https://github.com/KSP-CKAN/CKAN-meta/pull/1270

### DIFF
--- a/NetKAN/DockingPortAlignmentIndicator.netkan
+++ b/NetKAN/DockingPortAlignmentIndicator.netkan
@@ -18,5 +18,11 @@
             "file": "GameData/NavyFish",
             "install_to": "GameData"
         }
+    ],
+    "x_netkan_override": [
+        {
+            "version": "6.6.0",
+            "override": { "ksp_version_min": "1.3.0" }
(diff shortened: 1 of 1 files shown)


politas:
What is making it set a min_version of 1.2.0? When I run on the existing netkan, I get a clean 1.3.0 ksp version, as per Spacedock. Was it wrong on Spacedock? I can't see an AVC .version file in the 6.6.0 download I have.

bfishman:
Ah, that explains it.  I uploaded 6.6.0 to spacedock originally with the wrong .version file in it. I noticed this within a minute, so re-uploaded the zip file with a fixed .version file, but the NETKAN scrape must have occurred by that point.

What's the easiest way to rectify this - should I delete 6.6.0 on spacedock and submit a new version?

Olympic1:
I you want to do that, go ahead. I dont have to remind myself to add the vref back in for the next release.

bfishman:
Alright, I'll publish 6.7.0 just to make everyone's life easier!

PR #5576, opened 2017-06-03
Update BurnTogether and VesselMover

Olympic1:
Remove ksp version
Override ksp version
Remove a duplicate `name`


Olympic1:
Weird, BurnTogether doesn't work without a ksp version even though it has a `vref`

politas:
The .version file in the zip is the old one for v0.0.7.2, while the specified remote AVC file is v0.0.7.1, so the Vref isn't matching the current version number at all. @PapaJoesSoup, are you planning on updating your .version files for BurnTogetherContinued?

PR #5577, opened 2017-06-03
TACLS + BackgroundResources

politas:
As advised by @JPLRepo in IRC


politas:
NetKAN failure can be ignored; it's complaining that it can't find the mod that's being added.

PR #5591, opened 2017-06-05
Fix IFS

Olympic1:
InterstellarFuelSwitch-Core now also install the `Resources` folder, so we need to filter that too.

Fixes #5587
Fixes #5588

PR #5592, opened 2017-06-05
reverted copying folder

sswelm:


Olympic1:
I had fixed the problem in #5591 but because you changed the folder structure, you should revert that commit too.

sswelm:
Oh crap, so it already worked?

PR #5647, opened 2017-06-21
Fix CryoTanks vref

blowfishpro:
I am not quite sure why this broke, but CryoTanks has its own .version
file and running NetKAN confirms that it is finding the correct one

For reference, the error is this:
```
AVC: Invalid path to remote GameData/KerbalAtomics/Versioning/KerbalAtomics.version, doesn't match any of: GameData/CryoTanks/Versioning/CryoTanks.version,
```


politas:
Looks good. Merging with thanks!

PR #5648, opened 2017-06-21
Fix NearFutureProps vref

blowfishpro:
again not quite sure why this broke, but NearFutureProps has its own .version
file and running NetKAN confirms that it is finding the correct one


politas:
LGTM. Merging with thanks!

PR #5650, opened 2017-06-22
Another small SETI mod

Y3mo:
With self-administered netkan file as usual.

Best regards,
Yemo


politas:
LGTM. Merging with thanks!

PR #5653, opened 2017-06-25
Update Animated Decouplers to 1.3.4 for KSP 1.3.0

sumghai:


Starwaster:
That's weird, I wonder why this should have been necessary? I looked over the version file and it looks correct to me. (though it's 6am and I'm not fully awake...)

sumghai:
The current version file in NetKAN has ksp_version tagged as 1.2.1, whereas my commit brings it up to 1.3.0.

PR #5654, opened 2017-06-27
And the next SETI administered mod

Y3mo:
Self administered netkan file as usual.

Thank you very much!
Yemo


politas:
Merging with thanks!

PR #5656, opened 2017-06-28
Final SETI addition to CKAN for the current project

Y3mo:
Self-administered netkan file as usual.

Thank you and best regards,
Yemo


politas:
Merging with thanks!

PR #5659, opened 2017-06-30
Add dynamic battery storage and fix metadata for dependent mods

blowfishpro:
Permanent fix for KSP-CKAN/CKAN-meta#1280


blowfishpro:
Actually Nertea just released an updated version of DynamicBatteryStorage with a .version file included.  So I think the original would work if rerun.  Not sure if it's worth adding 1.0.0 to CKAN, particularly since there was no KSP version change between them

politas:
#rebuild

politas:
Ok, I'll merge this and then fix it..

PR #5660, opened 2017-06-30
Add Community Trait Icons

cake-pie:
To facilitate release of next version of PortraitStats, which will require this as a dependency.

### DIFF
--- a/NetKAN/AirlockPlus.netkan
+++ b/NetKAN/AirlockPlus.netkan
@@ -1,0 +1,22 @@
+{
+    "spec_version" : "v1.2",
+    "identifier"   : "AirlockPlus",
+    "name"         : "AirlockPlus",
+    "abstract"     : "Allows the use of any airlock on a vessel in conjunction with any crew part on the vessel, without requiring manual crew transfer.",
+    "author"       : "cake-pie",
+    "license"      : "restricted",
+    "$kref"        : "#/ckan/github/cake-pie/AirlockPlus",
+    "$vref"        : "#/ckan/ksp-avc",
+    "resources": {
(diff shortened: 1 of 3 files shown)


cake-pie:
Question: should CTI be added as a dependency in PortraitStats.netkan now, or after the new version of Portrait Stats is released? (@DMagic1 is ready to release, just waiting on CTI to be available via CKAN).  
If I understand correctly, waiting until afterwards would mean that when the new PS becomes available, netkan-bot would generate metadata for it based on the existing netkan -- and CKAN users could end up installing it without the required dependency, until the metadata is updated. If that is indeed the case, let me know and I'll add the necessary changes to PortraitStats.netkan to this PR as well.

dbent:
```
Failed to download "https://ksp.sarbian.com/jenkins/job/ModuleManager/128/artifact/ModuleManager-2.8.1.zip" - error: Error: NameResolutionFailure
One or more files failed to download, stopped.
```
Failure seems transient, #rebuild.

> Question: should CTI be added as a dependency in PortraitStats.netkan now, or after the new version of Portrait Stats is released?

When the new version that actually requires it is released, since as you mentioned changes to the netkan will update the latest version. [Overrides](https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#x_netkan_override) can be used to stage changes. [Here](https://github.com/Apokee/HotSpot/blob/develop/HotSpot.netkan#L16) are some examples.

dbent:
#rebuild

cake-pie:
> Overrides can be used to stage changes

Thanks, that's exactly what I was looking to do

PR #5663, opened 2017-07-01
Add SmartTank

HebaruSan:
Automatically resizes ProceduralParts fuel tanks to fit a specified thrust to weight ratio

I'm the author. Should work just like the entry for Astrogator.


HebaruSan:
I don't think that build error is correct. The supported KSP versions are marked as 1.3.0 through 1.3.8, and it's only checking whether the ProceduralParts dependency is available for 1.3.8. Both mods **are** available for 1.3.0.

Please let me know if I need to make changes to account for that.

PR #5664, opened 2017-07-01
DynamicBatteryStorage to Spacedock kref

politas:
Ref #5659


blowfishpro:
It looks like it's not getting the version from AVC since it already has one from spacedock.  Is there any way to force it to use the AVC version?

PR #5679, opened 2017-07-09
Recommendation of DynamicBattery Storage

blackliner:
cleaned up and added DynamicBatteryStorage as a recommended mod, because i did some changes to support it

### DIFF
--- a/NetKAN/RealBattery.netkan
+++ b/NetKAN/RealBattery.netkan
@@ -4,13 +4,16 @@
     "$kref"        : "#/ckan/github/blackliner/RealBattery",
     "license"      : "MIT",
     "$vref"        : "#/ckan/ksp-avc",
-	"abstract" : "Realistic behaving electrical storage. Including state of the art Li-Ion recharchable batteries!",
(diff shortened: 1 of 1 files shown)


blackliner:
what's wrong?

Dazpoet:
I ran it through NetKAN locally with -debug flag set and it seems to get stuck on the parsing of the .version file. A guess would be that it is confused by the lack of a KSP_VERSION_MAX when a KSP_VERSION_MIN exists.

I don't recall exactly how the avc-parsing is done but I'm sure @zengei knows seeing as they wrote it.

blackliner:
I didn't change that part, that's funny...

blackliner:
I think i fixed it, how to run checks again?

techman83:
#rebuild

techman83:
LGTM, merging with thanks!

blackliner:
Awesome, thanks! It seems I forgot the PATCH variable for KSP_VERSION, but that's something the parser could handle IMHO...

techman83:
Patches welcome :slightly_smiling_face:

blackliner:
Okidoki ;-)

PR #5715, opened 2017-07-22
RN revert changes to 1.3

raidernick:
-Remove the changes to 1.3 because of it breaking ckan and my mod/the
game from loading.
@linuxgurugamer This needs to go hand in hand with removing those
metadata files I mentionned in the forum PM. Once all that is deleted I will quickly remove the releases for 1.3.


raidernick:
@linuxgurugamer changing my mods to use kspavc to version check from now on. This will hopefully prevent this from ever happening again in the future.

linuxgurugamer:
#rebuild

PR #5726, opened 2017-07-31
Update RealFuels to KSP 1.3

blowfishpro:


blowfishpro:
Failing because ` RealFuels-Engine-Configs` is not available for KSP 1.3 (provided by `RFStockalike` or `RealismOverhaul`

NathanKell:
LGTM. The jenkins fail is from no engine configs existing for 1.3, which is as expected.

PR #5731, opened 2017-07-31

themaster402:
Made scatterer a dependency as to avoid it being installed incorrectly.


politas:
That failing check appears invalid. Merging with thanks!

PR #5733, opened 2017-08-01
Indicate that RSS provides PlanetShine config

aw1621107:


aw1621107:
I think this may be broken, but I'm not sure how to fix it... Manually adding the "conflicts" and "provides" fields to the entry in registry.json makes RSS show up when choosing a config, but the install fails later, complaining that it can't find something that provides PlanetShine-Config

aw1621107:
Now that I took a second look, I think I was mistaken in my reasoning. This PR is incorrect. Sorry.

PR #5734, opened 2017-08-01
Indicate that RSSVE provides Scatterer sunflare/config

aw1621107:


aw1621107:
Just like #5733, I'd expect that this one is broken in some way, but unfortunately I'm not familiar enough with CKAN to determine what exactly is wrong...

politas:
Merging with thanks!

PR #5740, opened 2017-08-01
Defreeze CivilianPopulation

rleroy:


politas:
#rebuild

politas:
Looks like NetKAN doesn't like something about the zipfile

politas:
Tested locally, downloads and installs fine.

PR #5757, opened 2017-08-09
Update MusicMuteContinued supported KSP versions

nimobeeren:
Also updated the spec version to v1.16 because of `ksp_version_strict` and added repository URL.

### DIFF
--- a/NetKAN/MusicMuteContinued.netkan
+++ b/NetKAN/MusicMuteContinued.netkan
@@ -9,7 +9,7 @@
     "ksp_version_max"    : "1.3",
     "ksp_version_strict" : "false",
     "resources": {
-        "homepage"   : "http://forum.kerbalspaceprogram.com/index.php?/topic/107951-1"
+        "homepage"   : "http://forum.kerbalspaceprogram.com/index.php?/topic/107951-1",
         "repository" : "https://github.com/nimobeeren/MusicMute"
     },
     "install": [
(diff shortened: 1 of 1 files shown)


nimobeeren:
Not sure why it's not passing...

NetKAN/MusicMuteContinued.netkan
```diff
@@ -1,13 +1,16 @@
...
+    "spec_version"       : "v1.16",
+    "identifier"         : "MusicMuteContinued",
+    "$kref"              : "#/ckan/github/nimobeeren/MusicMute",
+    "name"               : "MusicMute",
+    "abstract"           : "A simple mod that toggles the game's music with a hotkey.",
+    "license"            : "GPL-3.0",
+    "ksp_version_min"    : "1.2",
+    "ksp_version_max"    : "1.3",
+    "ksp_version_strict" : "false",
     "resources": {
```
techman83: Wrapping `"false"` in quotes turns it into a string of text. The schema expects `true` or `false` in that field.

```
Validating built/MusicMuteContinued-1.1.3.ckan.. Failed! See below for error description.
u'false' is not of type u'boolean'

Failed validating u'type' in schema[u'properties'][u'ksp_version_strict']:
    {u'description': u'Optional enforcement of strict version checks (defaults to false)',
     u'type': u'boolean'}

On instance[u'ksp_version_strict']:
    u'false'
```

politas:
Merging with thanks!

PR #5759, opened 2017-08-09
Update SolarSailNavigator.netkan

bld:
I am the author of the mod. I have recompiled it to work with KSP 1.3.0.


politas:
This will disappear from the client until Persistent Thrust is available for 1.3

PR #5772, opened 2017-08-15
Update name from "Astronomer's Visual Pack"

themaster402:
To possibly fix the disappearance of the updated version of the same name, and prevent conflict. Name changed to "Astronomer's Visual Pack (deprecated)", can be changed at will.


politas:
This won't fix the issue. Need to change the .CKAN file in CKAN-meta.

PR #5777, opened 2017-08-18
Add Trim Indicator

formicant:
A minimalistic add-on displaying trim values inside the control gauges.


politas:

PR #5785, opened 2017-08-22
Crew R&R Removed conflicts with KCT, since they are compatible

linuxgurugamer:


linuxgurugamer:
!rebuild

linuxgurugamer:
#rebuild

linuxgurugamer:
!rebuild

linuxgurugamer:
Squashing because I only removed a conficts line.  It's erroring on MM not being available, but it is, I think this is an error becase the version file has a min & max, and it looks like the netkan is using the max only

PR #5797, opened 2017-08-27
netkan for AFLC and OPM-Galileo

Galileo88:
earlier pull requests kept showing up failed. I'm unsure why so I uploaded again, this time with success, I think.


politas:
All looks good. Merging with huge thanks!

PR #5805, opened 2017-08-30
Case sensitivity in dependency identifier

madlemur:
Case sensitivity with the identifier for the Scatterer mod was preventing AVP from installing. I changed it in my local registry, and AVP upgraded without issue. This should allow others to upgrade.


politas:

PR #5808, opened 2017-08-31
SVT name update

Galileo88:
Its no longer separated by OS. Works for all


politas:
Rather than change the name on SVT-Windows, it would be better to freeze both SVT-Windows and SVT-MacLinux, and reactivate SVE-Terrain

Galileo88:
Ok cool I'll do that

PR #5809, opened 2017-09-01
Add GPO from SpaceDock

politas:
Also extends CTTP KSP compatibility (texture pack only)
Closes #5657


politas:
CTTP update is included. It is supposed to test the change as a whole, but this doesn't always work.

PR #5819, opened 2017-09-03
Fix JPL-RSS version overrides format

HebaruSan:
The "v" in the ksp_version_min and _max fields confuses the NetKAN bot. Currently causing "JSON deserialization error" at http://status.ksp-ckan.org/, discovered by making that error more verbose in a local fork of netkan.exe.


politas:
Oooh, my bad. Thanks for catching that! Merging with thanks!

PR #5820, opened 2017-09-03
Fix name of REKT's AVC file

HebaruSan:
This mod's $vref says to look for rekt.version, but the mod ships RecoverableEmergencyKerbalTransport.version instead, currently causing "AVC: Invalid path to remote rekt.version, doesn't match ..." on http://status.ksp-ckan.org/

To confirm, see: https://github.com/steedcrugeon/REKT/blob/master/SHED/REKT/RecoverableEmergencyKerbalTransport.version

### DIFF
--- a/NetKAN/REKT.netkan
+++ b/NetKAN/REKT.netkan
@@ -11,7 +11,7 @@
       "filter": "RetractableLiftingSurface"
     }
   ],
-  "$vref": "#/ckan/ksp-avc/rekt.version",
+  "$vref": "#/ckan/ksp-avc/RecoverableEmergencyKerbalTransport.version",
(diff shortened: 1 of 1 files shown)


politas:
The install is wrong now, too:

`Showing 5 installed files:
- GameData/REKT-mk1A
- GameData/REKT-mk1A/AeroVersion-B.mbm
- GameData/REKT-mk1A/AeroVersion-B_NRM.mbm
- GameData/REKT-mk1A/escapePod3.mu
- GameData/REKT-mk1A/rektmk1A.cfg`

politas:
And there's a new dependency.

politas:
Thanks for the heads-up!

PR #5839, opened 2017-09-12
file/find confusion

HebaruSan:
This one currently says "Module contains no files to install" on the [status page](http://status.ksp-ckan.org/), but the download exists and has mod files in it. If I change `file` to `find`, it generates a valid `.ckan` file. And the Source directory is actually Sources.


linuxgurugamer:
Looks like the mod author changed the layout, thanks for catching this

PR #5840, opened 2017-09-12
file/find confusion part deux

HebaruSan:
[Status page](http://status.ksp-ckan.org/) error: "Module contains no files to install."

Cause: Uses `file` instead of `find`, and the `.netkan` file has a space where the download zip has an underscore

Generates a valid `.ckan` file after this change in my testing.

### DIFF
--- a/NetKAN/SolarisHypernautics.netkan
+++ b/NetKAN/SolarisHypernautics.netkan
@@ -1,5 +1,5 @@
 {
-    "spec_version": 1,
+    "spec_version": "v1.4",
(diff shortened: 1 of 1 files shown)


techman83:
You'll need to set `"spec_version": "v1.4"` to use find.
```
#   Failed test 'SolarisHypernautics - spec_version v1.4+ required for install with 'find''
#   at t/metadata.t line 139.
```

HebaruSan:
Good point, updated.

PR #5841, opened 2017-09-12
switch to find_regexp

HebaruSan:
This mod includes the version in the directory internal to the zip, so a regexp is needed to match.

E.g.:

```
  Length      Date    Time    Name
---------  ---------- -----   ----
     1636  2016-06-11 08:23   North Kerbin Weaponry v0.83/effects/Explosion_150.mu
```

Current [status page](http://status.ksp-ckan.org/) error: "Module contains no files to install."

### DIFF
--- a/NetKAN/NorthKerbinDynamics.netkan
+++ b/NetKAN/NorthKerbinDynamics.netkan
@@ -10,7 +10,7 @@
     ],
     "install": [
         {
-            "file"       : "North Kerbin Weaponry",
+            "find_regexp": "North Kerbin Weaponry.*",
             "install_to" : "GameData"
         }
     ]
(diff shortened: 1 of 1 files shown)


Olympic1:
@HebaruSan for some reason it only installs the flag folder

HebaruSan:
Good catch, I missed that! I'll take a look and update (hopefully with a solution, possibly by closing this PR).

HebaruSan:
See https://github.com/KSP-CKAN/CKAN/pull/2120 for why it only installs the flags folder (it's the shortest directory entry that matches the regexp). I think this is a CKAN bug, and this PR will work if that other one is merged into CKAN.

HebaruSan:
The CKAN fix is merged. Can you please re-trigger the tests for this PR?

linuxgurugamer:
#rebuild

politas:
#rebuild

PR #5848, opened 2017-09-12
Remove absent install path

HebaruSan:
There's no Converter path of any kind in this zip. There's a YongeTech_TechTreesPlugin folder, and there's ETT in another folder in the same archive (which has its own `.netkan` file sharing this SpaceDock link).

Current [status page](http://status.ksp-ckan.org/) error: "Could not find GameData/YongeTech_TechTreeConverter entry in zipfile to install"

### DIFF
--- a/NetKAN/YongeTechTreesPlugin-Revived.netkan
+++ b/NetKAN/YongeTechTreesPlugin-Revived.netkan
@@ -19,10 +19,6 @@
         {
             "find"       : "GameData/YongeTech_TechTreesPlugin",
             "install_to" : "GameData"
-        },
-                {
-            "find"       : "GameData/YongeTech_TechTreeConverter",
-            "install_to" : "GameData"
         }
     ],
     "suggests":[


linuxgurugamer:
I don't understand this, the netkan file seems to be installing from Spacedock/mod/199, which is the ETT.
Rather than pull the mod from the ETT file, it would be better to actually get a clean release of the mod

HebaruSan:
I don't think there is one, since the author is inactive. The ETT zip is probably packaging your own recompile of these DLLs for 1.3.0 (you released on 2017-07-03, timestamps in ETT's zip are 2017-07-04):

https://forum.kerbalspaceprogram.com/index.php?/topic/112886-104-yongetech-tech-trees-plugin-v13/&do=findComment&comment=3110062

Are you suggesting I make a new zip for this?

HebaruSan:
This was addressed in #5868. Closing.

PR #5885, opened 2017-09-27
Fix repo URL for SpudMoon

HebaruSan:
This mod currently has the wrong Github repo in its $kref, causing "Could not find Spud entry in zipfile to install" on the status page.


Olympic1:
Oops. Thx for the fix

PR #5901, opened 2017-10-02
Update folder path

HebaruSan:
Latest download of this mod changed the ordering of the words in the directory name.


politas:
Tests fine locally. Merging with thanks!

PR #5905, opened 2017-10-05
Retire two obsolete Asclepius supplemental mods

HebaruSan:
Asclepius-clouds and ContractConfigurator-Asclepius were parts of the Asclepius mod that are no longer included in the download. The clouds portion is replaced by configs that are included in the main mod, and the contract part is replaced by the FarOut contract pack which is separately included in CKAN. Since these are both obsolete, I propose renaming them both to `.frozen` status.

Author's confirmation:
- https://forum.kerbalspaceprogram.com/index.php?/topic/113949-13-asclepius-a-kopernicus-planet-ssto-playground/&do=findComment&comment=3186904


politas:
Does Asclepius now require EVE? If so, that should be added as a dependency to Asclepius.netkan.

HebaruSan:
It's an optional dependency. If EVE is installed, then Asclepius has configs that will be used, but Asclepius will still work without EVE.

PR #5920, opened 2017-10-09
Switch WindowShine from DropBox to SpaceDock

HebaruSan:
The latest version of this mod is on SpaceDock, and the older DropBox download seems to have been removed, as the bot now errors out on this mod. The internal structure of the zip file changed slightly as well.

https://forum.kerbalspaceprogram.com/index.php?/topic/110080-13-windowshine-v15-toroidal-tank-update-2-october-17/


politas:
Looks good to me. Very glad they've switched to Spacedock. Those Dropbox downloads are a pain for us!

PR #5929, opened 2017-10-11
Update GPP netkan

Galileo88:

PR #5931, opened 2017-10-11
Match Scatterer compatility to Forum thread

politas:
Should allow #5929 to pass Jenkins


Sigma88:
I noticed that [Scatterer_config](https://github.com/politas/NetKAN/blob/master/NetKAN/Scatterer-config.netkan) doesn't have "Scatterer" as a requirement

is this intended?

it feels like it should

PR #5932, opened 2017-10-11
Update WarpDrive.netkan

Real-Gecko:
WarpDrive mod conflicts with KSPInterstellarExtended


politas:
Thanks for catching that! Merging with thanks!

PR #5934, opened 2017-10-11
Removed Firespitter dependency for RecycledParts

linuxgurugamer:


linuxgurugamer:
Only change was removing a dependency on Firespitter.  Other mods haven't yet updated, so it's failing because of that

PR #5941, opened 2017-10-12
Update DescentMode NetKAN file

infoman:
Hello. Please update my netkan file to use KSP-AVC version data instead of hardcoded one. I pushed a prerelease version and it seems to generate a correct .ckan file, and will make it a full release when .netkan is updated, to not break existing 1.2.2 installs

### DIFF
--- a/NetKAN/DescentMode.netkan
+++ b/NetKAN/DescentMode.netkan
@@ -15,5 +15,13 @@
     ],
     "conflicts" : [
         { "name" : "RealismOverhaul" }
+    ],
+    "x_netkan_override" : [
+        {
+            "version"  : "v0.2.0",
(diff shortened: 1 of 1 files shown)


politas:
Hi @Infoman!. The only issue with this appears to be that your current release will be re-created to have no ksp_version at all. You can avoid that by adding the following to the end:

    "x_netkan_override": [
        {
            "version": "v0.2.0",
            "override": {
                "ksp_version" : "1.2.2"
            }
        }
    ]

infoman:
Thanks @politas, i updated my branch to include your suggestion.

politas:
Looks all good now! Merging with thanks!

PR #5948, opened 2017-10-13
updated dependancies KSPIE

sswelm:


Olympic1:
Merging as this only fails for deps that aren't updated yet

sswelm:
oh, what dependencies were they, perhaps I should drop them for the time beeing

HebaruSan:
TweakScale is the one that made it error out.

PR #5942, opened 2017-10-12
Add Kessler Syndrome and PAWS - correct license for EarnYourStripes

severedsolo:
I've been a bit lax getting my mods onto the CKAN - so thought I'd correct it. Also the license you had for EarnYourStripes was incorrect it's actually MIT.

PR #5949, opened 2017-10-13
Add Dependency on ModuleManager to PAWS

severedsolo:


severedsolo:
As discussed in #5942

PR #5954, opened 2017-10-15
Stock-alike Solar System metadata

Sigma88:
@Olympic1  @politas 

I think something along the lines of this is the best solution for SASS

can you please take a look and see if everything is fine?


HebaruSan:
Direct `download` links in the netkan will need to be updated manually with each release.
Could these use `$kref` instead? E.g.:

```
    "$kref": "#/ckan/github/GregroxMun/Duna-Restoration-Project",
```

That way CKAN's automated indexing bot would take care of checking for new releases and updating the metadata for them, including the versions.

Sigma88:
@HebaruSan I did this on purpose because I plan to update manually the metadata

I cannot guarantee that new released versions of those mods will be compatible, but I can guarantee the setup I have tested personally will always work

Olympic1:
Every `SASS-...` file needs a:
* `"version"` - To see if there is a new version and file generation
* `"name"` - To show in the GUI
* `"abstract"` - To show in the GUI (not sure about this one)

Also doesn't OPM depend on `CustomBarnKit` and `CommunityTerrainTexturePack`?

---

@HebaruSan most of the dependencies for SASS are not/will not be updated for 1.3, thats why its done this way. This way the deps are only installed with SASS or none at all.

Sigma88:
thanks @Olympic1 I'll add those

Sigma88:
I've added CTTP manually as everything else, and added all the parameters you said.

hopefully it should not fail this time

Olympic1:
LGTM.

If you find it better to host your own metadata you can do the same as Roverdude is doing.

His repo: https://github.com/BobPalmer/CKAN
On NetKAN: https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/UKS.netkan

Sigma88:
would that be more risky? could an error on my side cause problems to CKAN outside of the issues in installing SASS?

Olympic1:
The only problem this causes is that if you push a change on there, it doesn't get checked by Jenkins, while on here it does. You may choose what is easier and best for you.

PR #5962, opened 2017-10-17
Updated KDEX information

mwerle:
masTerTorch has handed the baton for the KDEX mod to me ([see forum post](https://forum.kerbalspaceprogram.com/index.php?showtopic=50039)); as such I've taken the liberty to update the KDEX CKAN information as well.

I've taken the KDEX.frozen file as a basis; Please double-check and if ok, merge.


jenyaza01:
https://ci.ksp-ckan.org/job/NetKAN/6321/console , scroll down...

- 'KDEX' matches an existing frozen identifier: NetKAN/KDEX.frozen

There already a file named KDEX. Here is it
https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/KDEX.frozen

edit it with same data that you add in your new file, than rename it from .frosen to .netkan. 
Good luck

mwerle:
Ah didn't realise the actual netkan doesn't override the frozen file.

Olympic1:
There is no `.version` file included in the download, so every new release will always be the latest KSP version.

This is not breaking anything, just a heads up.

PR #5968, opened 2017-10-19
Exception detector version overrides

MoreRobustThanYou:
I put it through JSONLint. I also made it so:

version 1 supports 1.0-1.0.5
2 supports 1.1-1.1.3
3 supports 1.3+


politas:
Ok, this looks good, but it won't go back and grab the old versions, unfortunately. Merging with thanks, and I'll do a CKAN-meta PR to fix up versions 1 and 2

PR #5969, opened 2017-10-19
Makes EVE Configs NOT required

MoreRobustThanYou:
so you can install a cloud pack not on CKAN (or make your own) without manually removing it


politas:
What will EVE do if it has no configuration files?

MoreRobustThanYou:
Nothing, I think.

politas:
OK, then this is probably a good idea. Thanks for the suggestion!

PR #5966, opened 2017-10-18
Add DoubleDouble from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of SamBelanger, to add [DoubleDouble](https://spacedock.info/mod/1553/DoubleDouble) to CKAN.

Mod details:
    name = /mod/1553/DoubleDouble
    author = SamBelanger
    abstract = This mod that adds binary systems!
    license = MIT
    Homepage = 20 oct 2017 ONLY
    description = **Hello guys this mod that adds binary systems!**

 

**README**

1. Download the mods Kopernicus (1.3.0-6/1.3.1-2 RECOMMEND!!), ModularFlightIntegrator (Integrate to Kopernicus!), ModuleManager (Integrate to Kopernicus!) and BetterTimeWarp

2. Put the mods files in the GameData folder

3. Start the game (KSP)

4. Have fun :)

**And if you have a suggestion or find bugs or errors, contact us (pikmin-123@hotmail.com or make a private message in the forum of KSP or otherwise) Thank you very much!**

Please direct questions about this pull request to [SamBelanger](https://spacedock.info/profile/SamBelanger).

PR #5972, opened 2017-10-20
Add DoubleDouble and SamBelangerFlags

politas:
Closes #5966


politas:
This may fail due to the dependency not yet existing.

SamBelanger:
:D

SamBelanger:
It really doesn't matter, it can happen that it fails!

PR #5974, opened 2017-10-20
Add MandatoryRCS part pack

gotmachine:


politas:
Looks perfect to me. Merging with many thanks!

PR #6009, opened 2017-10-31
Update and rename ActionGroupManager.frozen to ActionGroupManager.netkan

Real-Gecko:
I'm wondering if this changes are enough to make ActionGroupManager Continued available to CKAN from SpaceDock.


HebaruSan:
Looks like @linuxgurugamer asked the author whether it was OK to add this mod, and he didn't receive a "Yes." It's probably not a good idea to merge this as long as that's the case. Please let us know whether that situation changes.

https://forum.kerbalspaceprogram.com/index.php?/topic/161919-131-en-es-de-action-group-manager-continued/&do=findComment&comment=3172862

PR #6014, opened 2017-11-01
Fix DavinciCombatKit install

politas:
Saw this was broken on NetKAN status page


politas:
Damn. We can't rename the GameData directory? Well, that makes this mod basically impossible to fix.

PR #6022, opened 2017-11-06
Update DiverseKerbalHeads.netkan

Real-Gecko:
Diverse Kerbal Heads now has new maintainer and is modified to be compatible with [Texture Replacer Replaced](https://forum.kerbalspaceprogram.com/index.php?/topic/167229-131-trr-diverse-kerbal-heads-v11-11052017/). I've contacted the author and he granted permissions for mod to be indexed by CKAN:
screen

### DIFF
--- a/NetKAN/DiverseKerbalHeads.netkan
+++ b/NetKAN/DiverseKerbalHeads.netkan
@@ -1,29 +1,18 @@
 {
-    "spec_version": 1,
+    "spec_version": 1.4,
     "identifier": "DiverseKerbalHeads",
-    "name": "Diverse Kerbal Heads",
-    "abstract": "Texture pack to provide racial and gender diversity to kerbonauts. Requires TextureReplacer.",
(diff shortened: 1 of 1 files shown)


HebaruSan:
Can the dependency remain TextureReplacer? TextureReplacerReplaced has a "provides" directive allowing it to substitute for TextureReplacer, so that would support both 1.3.0 and 1.3.1. But depending on TextureReplacerReplaced directly would only support 1.3.1.

Real-Gecko:
Nope, it won't work. TRR has different folder layout and also recommends keeping mod textures in separate folder. Loading additional textures is done by .cfg file
```
TextureReplacerReplaced
{
	Folders
	{
		Heads = DiverseKerbalHeads/Heads/
	}
}
```

HebaruSan:
Gotcha; well, it was worth asking. I suppose the old 1.0 version will cover 1.3.0 anyway.

Real-Gecko:
Thanks!

PR #6025, opened 2017-11-07
Update DMTanks-SphericalTanks.netkan

Bezzier:
change max supported version to 1.3.1


Bezzier:
So how did I mess up this time?

HebaruSan:
> So how did I mess up this time?

You didn't; the validation checks are having trouble due to a change in GitHub's APIs.
I tested this locally and it looks fine...

PR #6030, opened 2017-11-08
"KourageousTourists"

linuxgurugamer:


techman83:
#rebuild

HebaruSan:
The downloads are done as links embedded in the release descriptions rather than as part of the Downloads list:

image

I don't think netkan can handle that. @linuxgurugamer, are you in contact with this author to explain the problem?

linuxgurugamer:
Yes, I am, will contact him asap

PR #6035, opened 2017-11-13
Move SamBelangerFlags to Spacedock

politas:
@SamBelanger, I really don't want to force people installing Optional Atmospheres to download the nearly 1GB that is DoubleDouble just to get the flags. If you want to have a SamBelanger Flags that is simply v1.0 and doesn't change, you could make a zip file available anywhere and we could point to that. We don't need a GitHub/Spacedock release to scrape for unchanging mods.

### DIFF
--- a/NetKAN/SamBelangerFlags.netkan
+++ b/NetKAN/SamBelangerFlags.netkan
@@ -4,7 +4,6 @@
     "name"           : "SamBelanger Flags",
     "$kref"          : "#/ckan/spacedock/1525",
     "abstract"       : "Flags used in SamBelanger's mods",
-    "x_netkan_epoch" : 1,
     "license"        : "MIT",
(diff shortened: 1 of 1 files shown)


politas:
Not really blocked, just waiting for a response.

HebaruSan:
Is this still needed now that Curse is fixed?

SamBelanger:
I give you permission to do it! ;)

SamBelanger:
@politas @HebaruSan can I put a github for her? :/

SamBelanger:
But is done! https://github.com/SamBelanger/SamBelangerFlags

politas:
Great! Thanks, SamBelanger!

SamBelanger:
@politas You welcome ;)

PR #6037, opened 2017-11-15
Added HeatControl as recommendation to KSPIE

sswelm:

PR #6038, opened 2017-11-15
Update HeatControl's license

HebaruSan:
This mod's code is CC-BY-NC-SA-4.0, but its models are All Rights Reserved.

(That is the reason for #6037.)

### DIFF
--- a/NetKAN/HeatControl.netkan
+++ b/NetKAN/HeatControl.netkan
@@ -4,7 +4,7 @@
     "$kref"          : "#/ckan/spacedock/537",
     "$vref"          : "#/ckan/ksp-avc",
     "release_status" : "stable",
-    "license"        : "CC-BY-NC-SA-4.0",
+    "license"        : ["CC-BY-NC-SA-4.0", "restricted"],
     "depends" : [
         { "name" : "ModuleManager" }
     ],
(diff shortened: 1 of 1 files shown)


HebaruSan:
Waiting for fix in https://github.com/KSP-CKAN/xKAN-meta_testing/pull/28 .

dbent:
Multiple licenses in the array mean the mod is licensed under all and the user can elect which license to use, for example in the case of dual-licensed software. The schema does not currently provide a mechanism to specify *which* files are licensed under a particular license and doing so has always been relatively low priority.

(Although for anyone interested the basic idea would be to allow `license` to be an object where the keys are a regex which matches on installed file paths and the values would be a string or array of strings indicating which license(s) those files are licensed under. We'd then need a check in NetKAN which would ensure that every file that would be installed has a license specified and refuse to generate otherwise if we want to be super strict. Alternatively, treat unspecified files as `restricted` by default and have NetKAN and CKAN issue a warning instead.)

The rule for mods with multiple licenses for different files is to choose the most restrictive, which generally goes: public domain -> permissive -> copyleft -> restricted. So in this case, `restricted` should be the one and only license.

HebaruSan:
Gotcha.
Do we also need to purge the old versions from archive.org? It may be distributing All Rights Reserved content because of this metadata snafu.

dbent:
Probably, @techman83 would be the guy to talk to about that.

techman83:
Doesn't look like they can be purged easily (from the interface at least). I've made them "dark", which makes them only visible to the account owner (which is CKAN's archive.org account.)

PR #6039, opened 2017-11-16
Fix ksp_version_min for OPM 2.1

HebaruSan:
See https://github.com/KSP-CKAN/CKAN-meta/pull/1306; this pull request sets up a version override in the NetKAN to fix the problem of OPM 2.1 trying to install on the incompatible 1.1.x KSP releases.


politas:
Looks good. Merging

PR #6061, opened 2017-11-26
Create KKtoSD.netkan

Sigma88:
you might also want to add this as "suggested" to KK and/or SD


HebaruSan:
Can you make the abstract a bit more verbose? Users might not know what KK or SD refer to or what the advantage is to loading groups of one into the other.

Sigma88:
thanks for adding the suggestions 👍

HebaruSan:
Abstract looks a lot clearer now! Merging with thanks...

PR #6063, opened 2017-11-28
Create Sigma-KerbalDescriptions.netkan

Sigma88:


HebaruSan:
Is the Settings.cfg file going to be changed by the plugin during gameplay? That might confuse CKAN when the time comes to update. If so, we can add a filter to leave that file out during installation, assuming the plugin can auto-create it.

Sigma88:
if the cfg is missing it will be automatically generated or if it is corrupted (someone goes inside and deletes stuff) my mod will modify it by re-adding the necessary components

but generally speaking the cfg provided with the archive works fine so my mod has no reason to modify it

which also means it will be removed by CKAN when uninstalling the mod

Sigma88:
Let me know if you need more info about this mod

HebaruSan:
I'm flummoxed about what to do with the settings.cfg. I'd like a more experienced team member to weigh in, rather than merging it myself and potentially causing problems for users down the line.

Sigma88:
all my mods have a Settings.cfg, and except for "Sigma Replacements" which works differently, all the others have the settings node installed through CKAN

that's also the way the mod is intended to be installed. and the content of the cfg is generally modified only through MM patches which means CKAN will uninstall it without any problems

Olympic1:
Normally CKAN doesn't complain about a settings file that was modified. I don't recall to have seen an issue about this.

Sigma88:
even if that was the case, the chances of that file being modified are slim to none. so there should not be any issues there anyways

HebaruSan:
Thanks, @Olympic1, and sorry for the delay, @Sigma88, I just didn't know the right thing to do. I'll keep in mind for the future that config files are OK to index.

Sigma88:
@HebaruSan don't worry

better safe than sorry :D

PR #6066, opened 2017-11-30
Filter empty PluginData folder in DistantObject-default (fixes #6065)

### LINKED ISSUE #6065: Directory ownership conflict between DistantObject and DistantObject-default
Current master CKAN, KSP 1.3.1, Ubuntu. Installing DistantObject, the install works fine:

```
$ mono ckan.exe install DistantObject --ksp steam
1) DistantObject-RealSolarSystem (Distant Object Enhancement Real Solar System config)
2) DistantObject-default (Distant Object Enhancement default config)
Enter a number between 1 and 2 (To cancel press "c" or "n".): 
2
About to install...

 * Distant Object Enhancement v1.9.1(cached)
 * Distant Object Enhancement default config v1.9.1(cached)

Continue? [Y/n] 


Installing mod "DistantObject v1.9.1"
Installing mod "DistantObject-default v1.9.1"
Updating registry
Commiting filesystem changes
Rescanning GameData
Done!
```

However, when removing, there's a "failure in locating file" error for GameData/DistantObject/PluginData:

```
$ mono ckan.exe remove DistantObject DistantObject-default --ksp steam
About to remove:

 * Distant Object Enhancement v1.9.1
 * Distant Object Enhancement default config v1.9.1

Continue? [Y/n] 
Removing DistantObject...
Removing DistantObject-default...
3783 [1] ERROR CKAN.ModuleInstaller (null) - Failure in locating file /home/user/.steam/steam/SteamApps/common/Kerbal Space Program/GameData/DistantObject/PluginData : Could not find file "/home/user/.steam/steam/SteamApps/common/Kerbal Space Program/GameData/DistantObject/PluginData"
Done!
```

It's not even an exception, just a raw print to screen, which messes up the UI from KSP-CKAN/CKAN#2177. And probably gets ignored on GUI, which is strange if we consider it a serious error condition. Here's the code:

https://github.com/KSP-CKAN/CKAN/blob/786fe07bb6e7117ace9774d0e224664bb57bc32b/Core/ModuleInstaller.cs#L896

Relevant excerpt from `registry.json`; the directory and its only contained file are listed as owned by different mods:

```json
        "installed_files": {
                "GameData/DistantObject/PluginData": "DistantObject-default",
                "GameData/DistantObject/PluginData/Settings.cfg": "DistantObject",
```

I think this means that when DistantObject is removed, it removes Settings.cfg, and then notices that the directory is empty, so it removes that too (because CKAN tries to be a good neighbor nowadays). However, then it tries to remove the mod that was supposed to own that directory, and it's not there anymore.

As for the fix, at a glance, multiple things look wrong here. Shouldn't the Settings.cfg file be in the `-default` suffix mod, since that's a config package? And shouldn't the package that doesn't contain the Settings.cfg also not have anything to do with PluginData?

HebaruSan:
DistantObject-default currently installs an empty GameData/DistantObject/PluginData folder, which causes conflicts with DistantObject at uninstallation (see #6065). This pull request removes the empty folder.

### DIFF
--- a/NetKAN/DistantObject-default.netkan
+++ b/NetKAN/DistantObject-default.netkan
@@ -14,7 +14,7 @@
     "install": [{
         "file": "GameData/DistantObject",
         "install_to": "GameData",
-        "filter": ["DistantObject.dll", "Flare", "Icons", "Settings.cfg", "DistantObject.version"]
+        "filter": ["DistantObject.dll", "Flare", "Icons", "Settings.cfg", "DistantObject.version", "PluginData"]
     }],
     "provides": ["DistantObject-config"],
     "conflicts": [


Olympic1:
Wauw, that this issue wasn't noticed for so long. Merging

PR #6067, opened 2017-11-30
Move GPP's conflicting recommends to suggests

HebaruSan:
GPP currently recommends a number of mods that conflict with one another.

Cloud packs:
- GPPCloudsHighRes
- GPPCloudsLowRes

Rescales:
- Rescale-25
- Rescale-32
- Rescale-64
- Rescale-10
- Rescale-10625

Currently if you install GPP from the command line, it installs Rescale x2.5, which should probably be something the user chooses rather than the default. And GUI defaults to installing them all:

image

This pull request moves these mods from recommends to suggests, so the user can choose which ones they want, if any.

Pinging @Galileo88 for FYI/feedback.

### DIFF
--- a/NetKAN/GPP.netkan
+++ b/NetKAN/GPP.netkan
@@ -28,7 +28,7 @@
         {
             "find": "Optional Mods/GPP_FinalFrontier/GameData/Nereid",
             "install_to": "GameData",
-            "comment": "This correctly installs Final Fronteir for GPP"
+            "comment": "This correctly installs Final Frontier for GPP"
         },
         {
             "find": "Optional Mods/LoadingScreen/GameData/LoadingScreenManager",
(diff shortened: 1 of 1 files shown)


HebaruSan:
@Sigma88, should some of your recently indexed mods be added as dependencies here?

Sigma88:
not necessarily dependancies, but I think probably "Recommends" would be the best choice for:

- SigmaReplacements-Heads
- SigmaReplacements-Suits
- SigmaReplacements-SkyBox
- SigmaReplacements-Navigation

idk if *Sigma Binary* would be better as recommended or suggested

anyways, I know @Galileo88 was waiting for KerbalDescriptions to be merged into CKAN so he could make a PR for GPP so I assume he will fix all those in his PR

HebaruSan:
Cool, I've added those. I was initially just going to fix the conflicting recommendations, but with this other stuff going on, we might as well tee up the changes we know about, and when he's ready @Galileo88 can stop by and look them over rather than starting his own metadata from scratch.

HebaruSan:
We're already indexing the new version of some of the other GPP related mods. I think it'll be better to merge now rather than waiting further, since more changes can always be made...

Issue #6057, opened 2017-11-25
Merge Nehemiah's mods?


mwerle:
Original issue #5557 

After more thinking and also requesting information from the users, as well as checking the download stats, it seems that having split these mods apart is far more bother than worth.

Download Info:
> KEES_0.3_beta_2.zip (1.70 MiB) - downloaded 68 times. Last updated on 2017-07-27
> KeminiResearch_0.2_beta_2.zip (0.87 MiB) - downloaded 43 times. Last updated on 2017-07-27
> KerbalLifeScience_0.2_beta_2.zip (3.90 MiB) - downloaded 43 times. Last updated on 2017-07-27
> NehemiahIncSciencePack_0.2_beta_2.zip (14.03 MiB) - downloaded 4,434 times. Last updated on 2017-07-27
> OrbitalMaterialScience_0.7_beta_2.zip (11.18 MiB) - downloaded 50 times. Last updated on 2017-07-27

As such, I'm wondering whether it's possible to replace the various mods with one new one "Nehemiah Engineering Orbital Science" (NEOS).  I would remove the version files from all the sub-directories.

Thoughts?

If we go ahead with this change, we'd need to:
* Mark the old mods as deprecated or replaced-by or something.
* Mark the new mod as conflicting with the old mods

I'm presuming that's possible?

HebaruSan:
This was done in #6071.

PR #6071, opened 2017-12-02
NehemiahEngineeringOrbitalScience

mwerle:
#6057 - nobody commented so I created this patch.
It makes the old files "frozen" and creates a new "all-in-one" which conflicts with the old files.


mwerle:
Ok, now I'm stumped as there are releases which should match the kref:

Running NetKAN for NetKAN/NehemiahEngineeringOrbitalScience.netkan
1423 [1] WARN CKAN.NetKAN.Transformers.GithubTransformer (null) - No releases found for mwerle/OrbitalMaterialScience
1585 [1] ERROR CKAN.CkanModule (null) - NehemiahEngineeringOrbitalScience missing required field download
1587 [1] FATAL CKAN.NetKAN.Program (null) - NehemiahEngineeringOrbitalScience missing required field download

HebaruSan:
The format of the asset_match filter is a regular expression rather than a file wildcard format.

This metadata should be fine now, but the validation is messed up until KSP-CKAN/CKAN#2203 gets merged. Hopefully that won't be more than a day or two, but it depends on when folks are available to double check the code.

What is the upgrade experience going to be like for your users? Will they be able to figure out what happened and get everything working again?

Have you considered keeping NehemiahInc-Complete active and using it instead?

mwerle:
Re NehemiahInc-Complete:
Well that's why i was trying to get a discussion going in the issue i raised to try to figure out the best approach. 
I was hoping there'd be a mechanism for tagging a nod as "replaced by" or similar. 

Although I do want a tiny bit of effort on behalf of the users anyway instead of just a ckan update to try to get then to read the changelog as the save format has changed slightly
I don't suppose there's a mechanism for popping up a notification on update? On Debian any major changes on update force a press-enter-to-continue dialog.

HebaruSan:
- "Replaced by" is in progress at KSP-CKAN/CKAN#1888. I think we're hoping to include it in the next release if it's ready, but currently it's not available. Until then, I believe your users would have to figure out that they need to uninstall the old mods and install the new one.
- Popups at upgrade have been suggested in KSP-CKAN/CKAN#837 and KSP-CKAN/CKAN#1488 but not implemented. I'm a pessimist about getting users to put in effort at upgrade, though; usually the desire is to get the newest version and play it right away, and if a message pops up they'll just look for the OK button to make it go away as quickly as possible.
- You're probably right about the backslash. Feel free to push more commits if you want to tweak the string further, or just click the pencil icon in the file to edit.

As for #6057, I didn't know the history of these mods beyond what I read in the previous issues/PRs, and I was not clear on the reasons for separating them or merging them, so I opted not to put my foot in my mouth. I also don't know whether there's been another case of mod merger like this before.

HebaruSan:
#rebuild

mwerle:
1. I saw the "replaced-by" - very encouraging :) 

1. Popups - well, yes, most people will "click-through"; but it does give me as a mod-publisher an "out" if people break their saves without backing up after having to read the warning.  Currently it's "too easy" to upgrade using CKAN if a mod update has some breaking changes. (I do it myself - just click on update all without reading changelogs etc). So I think it would be a -very- good addition, although it should be a single popup with a merged set of messages instead of one per mod.

1. Eh, couldn't care less as long as it works :)

HebaruSan:
(The full depth and breadth of my thoughts on popups would be off-topic here, so I wrote [a GitHub gist](https://gist.github.com/HebaruSan/b93971894b25918b8da01f1ba5505d0b) about it instead of derailing this issue further.)

HebaruSan:
Conflicts have these effects:

- If you try to install two or more mods that conflict with one another, an error appears and changes will be aborted
- Likewise if you try to install something that conflicts with any of your current mods
- The relationship resolver for dependencies will avoid selecting conflicting mods, for example if your mod `recommends` multiple conflicting mods, installing it via the command line will only choose one of them

They do _not_ prompt to uninstall currently installed mods. Hence the desire for `replaced_by`.

Do you want to proceed with this change now, or possibly wait till `replaced_by` is available? I'm not sure how urgent these changes are.

mwerle:
> Do you want to proceed with this change now, or possibly wait till replaced_by is available? I'm not sure how urgent these changes are.

I think we may as well go ahead; it's not urgent per-se but the new release is out. This way people can get it via CKAN as well (if they can figure it out) and when "replaced_by" becomes available we can update the metadata again to encourage the remaining users onto the new release.

HebaruSan:
Alright, merging with thanks. Good luck!

mwerle:
Thankee muchly :)

PR #6079, opened 2017-12-06
Update required for release IXSWarpShipOS v0.4.0.8

bartblommaerts:
- changed suggested SpaceDock to SpaceDockRedeployed
- moved Alcubierre Drive from depends to recommends
- added WarpDrive to recommends
- added suggests CommunityTechTree
- added depends ModuleManager

I did not find how to define an OR.
Either "AlcubierreStandalone" or "WarpDrive" can be installed, not both because that would create namespace issues.


HebaruSan:
We could do an OR by requiring an abstract name that isn't a real mod, like maybe "WarpDriveEngine" or something, and then marking mods with `"provides": [ "WarpDriveEngine" ]`. Do you want to try that? Will your mod crash if neither of them is installed?

HebaruSan:
I'm not sure why this closed; that wasn't intentional. I was just trying to resolve the conflicts. I'll look into it...

HebaruSan:
Please see my explanation in #6080; I made some errors in trying to resolve this PR's conflicts that GitHub won't allow me to fix. I'd like to continue the discussion of OR/`provides` over there if that's OK with you.

PR #6080, opened 2017-12-06
Update required for IXSWarpShipOS release 0.4.0.8

HebaruSan:
- changed suggested SpaceDock to SpaceDockRedeployed
- moved Alcubierre Drive from depends to recommends
- added WarpDrive to recommends
- added suggests CommunityTechTree
- added depends ModuleManager

@bartblommaerts, I apologize, I appear to have messed up #6079 beyond my ability to repair. There was a conflict with master because your branch was out of date, and in the process of resolving the conflicts, I erroneously committed instead of `rebase --continue`ing, then pushed to your branch thinking the changes were ready. But git decided to push unchanged master instead, after which your branch had no changes; GitHub decided this meant it should auto-close your PR, and without an open PR, I can no longer push changes to your branch.

The good news is that I do still have the changes and was able to resolve the conflicts properly and push them to my own branch. So while I can't restore your PR, I can create this new one to replace it.

### DIFF
--- a/NetKAN/IXSWarpshipOS.netkan
+++ b/NetKAN/IXSWarpshipOS.netkan
@@ -2,7 +2,7 @@
     "license": "CC-BY-NC-SA-4.0",
     "identifier": "IXSWarpshipOS",
     "$kref": "#/ckan/spacedock/589",
-    "spec_version": "v1.4",
+    "spec_version": "v1.16",
     "depends": [
         { "name": "CommunityResourcePack" },
         { "name": "ModuleManager" }
--- a/NetKAN/IXSWarpshipOS.netkan
+++ b/NetKAN/IXSWarpshipOS.netkan
@@ -23,7 +23,12 @@
         {
             "find"       : "IXSWarpShipOS",
             "install_to" : "GameData",
-            "filter"     : ".DS_Store"
+            "filter"     : [ ".DS_Store", "Example Craft", "old Spacedocks.zip" ]
+        },
+        {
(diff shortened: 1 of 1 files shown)


HebaruSan:
@bartblommaerts, you mentioned that installing WarpDrive and AlcubierreStandalone would result in namespace issues. Is that the case for those two mods in general, or just when IXSWarpShipOS is installed? If there is a conflict there, I'd like to update the metadata, because currently CKAN allows those to mods to be installed together.

bartblommaerts:
Hi HebaruSan,

I'm not 100% certain, but while i was looking into separating the warpdrive
module from the parts and looking at the code of both warpdrive mods, i
noticed some functions being named the same. Also, i think that internally,
both use "WarpDrive" as a module name, i'll check again. And i also have
not yet gone over all the source of the Interstellar WarpDrive mod, just
the files relating to the drive itself.

I like your idea of working with the dependency on a 'provides' though. But
let me check the 2 mods again and see if my fear is an actual reality.

Grtz

Denko

HebaruSan:
OK, I made a few changes; the zip file is not installed anymore and the craft goes into Ships\VAB.

I see MODULE[AlcubierreDrive] in one mod and MODULE[USI_ModuleWarpEngine] in the other, so based on that my guess is that they're compatible. They might have similar or the same class names in the DLLs, but as long as they're in different namespaces, that should be fine.

bartblommaerts:
oh ok, i didn't get what the namespaces were for other then identifying the
dll with in MM.

Thanks for the changes to the netKan file. most appreciated!

PR #6086, opened 2017-12-10
Updated RecycledParts

linuxgurugamer:
Added netkans for individual mods in RecycledParts


linuxgurugamer:
This is adding a number of netkans to pull out individual mods from the RecycledMods.  It's complex enough that I would appreciate some review of them

HebaruSan:
Looks like the individual parts mods "provides" themselves, i.e. the value in "provides" is the same as the `identifier`. I've seen that confuse the client when that name is a dependency, it prompts the user to choose between two copies of the same mod. See #6068. I think that's because all mods already inherently provide themselves, and I believe there's even code that adds `identifier` to `provides` automatically when the registry is loaded. Recommend removing those lines.

politas:
Conflicting with themselves is no problem, that's a common trick to stop multiple providers being installed, but any "provides" for themselves should be removed.

Mecripp:
on the AtomicAge the Radiators don't glow if you change
 MODULE
	{
      		name = ModuleAnimateHeat
     		ThermalAnim = RadiatorRadialHeat
	}

MODULE
 {
        name = FXModuleAnimateThrottle
       animationName = RadiatorRadialHeat
 
responseSpeed = 0.001
dependOnEngineState = True
dependOnThrottle = True
}

you can get the glow back

linuxgurugamer:
I'm closing this, cause I deleted my local repo, I've recreated them all and removed the provides, that will be a new  PR soon

PR #6121, opened 2018-01-02
North Kerbin Dynamics changed path again

HebaruSan:
Latest download uses just "NKD"


HebaruSan:
[Inquired on the forum about the CommunityAmmunitionLibrary dependency](https://forum.kerbalspaceprogram.com/index.php?/topic/118667-wip131-north-kerbin-dynamics-nuclear-bombs-and-heavy-ordnance-pack-and-other-things-v084b/&do=findComment&comment=3257915).

PR #6126, opened 2018-01-02
Borrow updated issue template from CKAN repo

HebaruSan:
There have been a few issues filed for NetKAN that required follow-up questions for clarification. This pull request copies the issue template from CKAN to NetKAN, with modifications, to encourage issue reports to include relevant info in the first draft.


HebaruSan:
#rebuild

PR #6145, opened 2018-01-07
Recycled Parts, individual install files

linuxgurugamer:


politas:
Each mod will need a distinct `"name"` if they are to show up as separate mods in the CKAN client.

linuxgurugamer:
When I had names, I was told it would be a conflict.
Do I need to have a different name from the "provides"?

linuxgurugamer:
@politas 
I fixed it, please check again, thanks

Issue #6154, opened 2018-01-09
Version mismatches between NetKAN and forum threads


HebaruSan:
#6148 gave me nightmares, so I wrote [a script to check versions hard-coded in NetKAN against their forum threads](https://github.com/HebaruSan/KSP_Tools/blob/master/ckan_find_bad_versions); output below. Some of these are false positives (ScottManleyHeadPack is 1.3.1 compatible but the title just isn't updated), but others probably cause CKAN to install incompatible mods.

| Netkan | ksp_version | Forum thread | Links |
| --- | --- | --- | ---- |
| df95e48ca0e1108e66b20b988278ff654cb6b48f MechJeb2.netkan | 1.3 | 1.3.1 | [netkan](https://github.com/KSP-CKAN/NetKAN/tree/master/NetKAN/MechJeb2.netkan) [forum](http://forum.kerbalspaceprogram.com/index.php?/topic/154834-*) |
| SDHI-StrobeOMatic.netkan | 1.3.0 | 1.3 | [netkan](https://github.com/KSP-CKAN/NetKAN/tree/master/NetKAN/SDHI-StrobeOMatic.netkan) [forum](http://forum.kerbalspaceprogram.com/index.php?/topic/137804-11-sdhi-strobe-o-matic-warning-rotator-lights-v10-23-april-2016/) |
| SDHI-ServiceModuleSystem.netkan | 1.3.0 | 1.3 | [netkan](https://github.com/KSP-CKAN/NetKAN/tree/master/NetKAN/SDHI-ServiceModuleSystem.netkan) [forum](http://forum.kerbalspaceprogram.com/index.php?/topic/48073-121-sdhi-service-module-system-v321-20-november-2016/) |
| NewHorizons.netkan | 1.3.0 | 1.3 | [netkan](https://github.com/KSP-CKAN/NetKAN/tree/master/NetKAN/NewHorizons.netkan) [forum](http://forum.kerbalspaceprogram.com/index.php?/topic/102776-nh/) |
| LaunchPadLamp.netkan | 1.3.0 | 1.3 | [netkan](https://github.com/KSP-CKAN/NetKAN/tree/master/NetKAN/LaunchPadLamp.netkan) [forum](https://forum.kerbalspaceprogram.com/index.php?/topic/164677-13-launch-pad-lamp/) |
[...]

PR #6161, opened 2018-01-10
Make UnchartedLands version-independent

HebaruSan:
UnchartedLands is marked as 1.3.1 compatible on its forum page, but we have it as 1.3.0. According to discussion on its forum thread, it works with both. It has no KSP-AVC file.

Discovered during investigation of #6154.


Olympic1:
Planet packs should never be set to `any` as a new version of Kopernicus can break them.

HebaruSan:
Thanks, I wasn't sure about that. Changed the PR to mark it for "1.3" instead.

PR #6170, opened 2018-01-15
Fix CLS revert

Sniper296:
As much as a trailing newline is nice, I don't think it was the intended change by @linuxgurugamer in #6127.


HebaruSan:
Thanks for the fix!

PR #6180, opened 2018-01-16
host GPP netkans on Galileo's repo

Sigma88:


HebaruSan:
Good idea.
One suggestion on the textures: You could make it auto-update by using the same `$kref` as GPP but with `asset_match` set to pick out the right zip file. You have this set up for Rescale, and I think it would work the same way here.

Sigma88:
I'll leave that to @galileo88 now that he has easy access to the files

Thanks for the suggestion!

PR #6187, opened 2018-01-19
Update SCANsat netkan for Overdrive incompatibility

DMagic1:
Overdrive changes the stock UI in a way that breaks SCANsat and is therefore incompatible with it.

### DIFF
--- a/NetKAN/SCANsat.netkan
+++ b/NetKAN/SCANsat.netkan
@@ -17,7 +17,7 @@
     ],
     "conflicts": [
         { "name": "ContractConfigurator-ScanSatLite" },
-		{ "name": "Overdrive" }
+	{ "name": "Overdrive" }
     ],
     "resources": {
         "repository": "https://github.com/S-C-A-N/SCANsat"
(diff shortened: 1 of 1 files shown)


HebaruSan:
Thanks for the update!
Should we check other mods for Overdrive-related conflicts?

DMagic1:
It's possible that other mods might have a similar problem, I think the specific issue for SCANsat is that Overdrive replaces some, but not all of the stock toggle UI elements, and SCANsat tries to access those on loading. SCANsat could be made to be more fault tolerant of this specific issue, but it would still break the UI is some way.

Other mods that use the new Unity UI and grab their UI elements from the stock skin could have similar problems.

The real solution is for Overdrive to not remove stock elements without replacing them with something else.

linuxgurugamer:
I am in the process of taking Overdrive over, I'll be looking into this.
Any details you can provide would be useful, since I don't use it normally

DMagic1:
Can I get this reversed (my branch is all out of sync and I don't want to fuss with getting it back in working order to make a new pull right now), the two work fine together now with SCANsat 18.3.

HebaruSan:
@DMagic1, done in #6223. Thanks for the heads-up!

PR #5975, opened 2017-10-21
Freeze retired mod

HebaruSan:
This mod's content has been merged into BDArmory Continued, and it is no longer supported.

https://forum.kerbalspaceprogram.com/index.php?/topic/125381-discontinued-cal-community-ammunition-library-for-bdarmory/

PR #5864, opened 2017-09-17
Remove missing mod

HebaruSan:
I believe @Ippo343 has deleted [this repo](https://github.com/Ippo343/rt-config). It was a config file for Remote Tech, with a minimum version of KSP 0.25, and only one release in CKAN-meta, so there's a good chance it was out of date or no longer needed.

Current error on the [status page](http://status.ksp-ckan.org/): "The remote server returned an error: (404) Not Found."

PR #6192, opened 2018-01-20
Re-add deleted netkans as frozen

HebaruSan:
## Problems

Early in my CKAN career, I was eager to eliminate all errors from http://status.ksp-ckan.org/, several of which were due to mods that had been removed from their host sites. In three instances I addressed this by deleting the .netkan files, which I found out in #5975 is a no-no; we should rename them to .frozen instead.

| Pull request | File |
| --- | --- |
| #5842 | AxisSystem.netkan |
| #5843 | RejuvenatedDuna.netkan |
| #5864 | rt-config-by-ippo.netkan |

@Olympic1, next time please tell the clueless new person about freezing. :wink: 

## Changes

These files are now retrieved from the repo archive and re-added as .frozen as they should have been.

### DIFF
--- a/NetKAN/AxisSystem.frozen
+++ b/NetKAN/AxisSystem.frozen
@@ -1,0 +1,10 @@
+{
+    "spec_version": "v1.4",
+    "identifier":   "AxisSystem",
+    "$kref":        "#/ckan/spacedock/1438",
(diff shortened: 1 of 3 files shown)


Olympic1:
Sorry I didn't mention that to you

PR #6194, opened 2018-01-21
Update NetKAN files for Pood's Skyboxes

Poodmund:
Switching from TextureReplacer to Sigma Replacements: Skybox as a dependency and now listing Distant Object Enhancement as a suggested mod.

### DIFF
--- a/NetKAN/PoodsCalmNebulaSkybox.netkan
+++ b/NetKAN/PoodsCalmNebulaSkybox.netkan
@@ -1,27 +1,27 @@
 {
     "spec_version": 1,
     "name": "Pood's Calm Nebula Skybox",
-    "abstract": "Skybox that presents the Kerbol system within a sparse Nebula. Features a subtle Galaxy star-line and patches of Pink/Blue Nebula proto-Star fields. To the polar South a very energetic solar nursery can be seen.",
+    "abstract": "Skybox that presents the Kerbal Solar System within a sparse Nebula. Features a subtle Galaxy star-line and patches of Pink/Blue Nebula proto-Star fields. To the polar South a very energetic solar nursery can be seen.",
     "author": "Poodmund",
     "identifier": "PoodsCalmNebulaSkybox",
     "$kref": "#/ckan/spacedock/924",
     "license": "CC-BY-NC-ND-4.0",
     "resources": {
-        "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/96851-11-texturereplacer-2413-442016/&do=findComment&comment=2388097"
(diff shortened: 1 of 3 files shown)


HebaruSan:
Looks like the download folder is still TextureReplacer. What was the intent of this change?

```json
    "install": [
        {
            "file": "GameData/PoodsSkyboxes",
            "install_to": "GameData"
        }
    ],
```

Poodmund:
I haven't uploaded the new distributions to Spacedock yet as I wanted the CKAN metadata to be correct for it automatically updated.

I'm changing dependencies from one mod to another. I am not sure how CKAN handles this for the end user when updating through Spacedoc without the new metadata being correct.

HebaruSan:
Uploading the new files first generally works best. If we committed this pull request as-is, it would mess up the metadata for the older versions.

Poodmund:
But to counter that, uploading the new versions first also causes issues as CKAN automatically pulls that there has been an update from Spacedock, the end user then updates the mod but due to the dependency change, it would no longer work for them in-game.

Its a chicken and egg situation I expect.

HebaruSan:
In this case, if you've changed the install paths, then when the netkan bot tries to pull the new version, it will error out and not index it. That gives us time for metadata updates.

Poodmund:
Ah so in future if I need to do anything like this again, I should wait for the SpaceDuck pull request to be made, then quickly post in the PR to hold fire as I need to make further changes to the metadata.

HebaruSan:
Posting quickly in the PR isn't a factor; I haven't applied any sort of hold. It's just that the netkan bot would look for the old path in the zip, and when it's not found, it would fail to index that new version. That would start happening as soon as the new version went up, and it would keep happening until the metadata was fixed.

Poodmund:
@HebaruSan the new distributions have been uploaded to SpaceDock for all 3 listings and are ready to go.

HebaruSan:
#rebuild

Poodmund:
Thank you ever so much for your assistance in getting this resolved. :smile:

HebaruSan:
As techman83 says, "LGTM" :+1: .
As politas says, merging with thanks...

Poodmund:
@HebaruSan I am an complete fool. I forgot to add a Module Manager dependency on to the listings as now it is required. Do I have to open another pull request?

HebaruSan:
Yes, a pull request would be the best way to handle that, and yes, it probably has to be a new one, I haven't found a way to re-open a merged PR on GitHub.

Don't worry about it, we all miss things like this from time to time.

PR #6215, opened 2018-01-31
Added NetKAN file for USI Kontainers (without reactors)

pamidur:
Made it possible to install Kontainers only. Without reactors.


linuxgurugamer:
This netkan is pulling files from Roverdude's repos on Github.
Does @pamidur have permission from him to do so?

pamidur:
Ok, I've got puzzled with the license.

PR #6219, opened 2018-02-03
Create BladeTweaks.netkan

blade1981m:


HebaruSan:
Thanks for the contribution!

The MechJeb.cfg and Squad.cfg files both set bulkheadProfiles for mk2DockingPort, to different values. Is that intentional?

blade1981m:
No, that is not. Not sure how I managed that. I will make a change here real quick. Thanks for catching that.

blade1981m:
Ok, github repo is updated but there are no changes to the netkan file. Do I need to do anything else regarding this pull request?

HebaruSan:
Nope, you're all set. Thanks again for submitting this!

PR #6232, opened 2018-02-11
Add Toolbar dependency to ScienceAlert

HebaruSan:
Toolbar is required for ScienceAlert. I think this was added to the thread after it was added to CKAN (and it's still not mentioned on SpaceDock).


linuxgurugamer:
I posted a message in his thread about the spacedock issue

PR #6238, opened 2018-02-14
Create DockingCameraKURS

DennyTX:
edits of name and description  when old mod back to service

### DIFF
--- a/NetKAN/DockingCameraKURS
+++ b/NetKAN/DockingCameraKURS
@@ -1,0 +1,42 @@
+{
+    "spec_version": "v1.4",
+    "identifier": "DockingcameraKURSstyle",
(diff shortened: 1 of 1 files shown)


HebaruSan:
The DockingcameraKURSstyle module has already picked up the 1.3.3.1 version and has a $kref from this same SpaceDock module:

https://github.com/KSP-CKAN/CKAN-meta/blob/master/DockingcameraKURSstyle/DockingcameraKURSstyle-1-1.3.3.1.ckan

That existing netkan should be sufficient here, and adding another one would create confusion.

HebaruSan:
The "abstract" field is pulled from "Short description" on SpaceDock. Try editing the mod in SpaceDock and changing that field.

PR #6259, opened 2018-02-20
Create dawnguard.netkan

subdeviant:
dawn guard solar sail


HebaruSan:
We have a request for this already at #6258, but it needs some changes before merging, and we need SpaceDock to come back up to finish that. Would it be OK with you if we follow through on that rather than starting this one?

subdeviant:
hi i really have no idea what i am doing. it is up and running on spacedock now and that is where i have decided to host until i start to understand github better. sorry if i am spamming requests. it's the first time i have tried anything like this.

PR #6266, opened 2018-02-20
Create DockRotate.netkan

peteletroll:
New mod - depends on ModuleManager, which is included in the release .zip

### DIFF
--- a/NetKAN/DockRotate.netkan
+++ b/NetKAN/DockRotate.netkan
@@ -1,0 +1,13 @@
+{
+	"spec_version": 1,
+	"identifier": "DockRotate",
+	"$kref": "#/ckan/github/peteletroll/DockRotate",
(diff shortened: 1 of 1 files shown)


HebaruSan:
This will install two copies of Module Manager, one via the dependency, and one from your ZIP. We try to avoid installing bundled dependencies. Any chance you could move ModuleManager.3.0.1.dll into the parent directory, so it would still be there for manual installers but ignored for CKAN?

peteletroll:
Yes, no problem. It has to be in GameData, right?

HebaruSan:
Yes, that's my understanding. Thanks!

peteletroll:
New version released.

HebaruSan:
#rebuild

HebaruSan:
Looks good. Thanks for the submission!

PR #6285, opened 2018-02-24
Update GPO.netkan on behalf of Gameslinx

DeltaDizzy:


HebaruSan:
Hey @DeltaDizzy, there were lots and lots of problems with your changes, so they couldn't be used as-is. I've tried to clean them up a bit, but it would probably be best if you tried to summarize what you wanted to accomplish (typically that's what the text portion of a pull request is for).

Regarding this mod's bundled mods, CKAN prefers to use proper dependencies for those. I'll look into adding true netkan entries for:

- Sigma LoadingScreens
- KS3P
- INSTANTIATOR

... and then we can list them as true dependencies of GPO. If their authors aren't comfortable being indexed independently in CKAN, then installing them from the GPO package would be an even worse idea.

Galileo88:
@DeltaDizzy Sigma loading screens is not a dependency and I'm fairly certain Gameslinx bundled it, which is what the mod is intended for.

HebaruSan:
Yup, SLS is bundled in GPO, but bundling doesn't play nice with other mods that share the bundled dependency.

Galileo88:
I responded to this in the sigma issue you opened.

Galileo88:
And believe me, we give Gameslinx enough crap about bundling mods for those reasons lol

HebaruSan:
OK, KS3P and INSTANTIATOR are now indexed as of #6286, and added as proper dependencies of GPO in this pull request. LoadingScreens is not; if it's moved into the main GPO folder in a future download as @Galileo88 suggested in Sigma88/Sigma-LoadingScreens#12, then the bot will take care of it automatically.

PR #6296, opened 2018-02-28
Force GPP hash recalc

HebaruSan:
GPP had a new release, then replaced the same download URL with a ZIP with an updated KSP-AVC file.

The metadata hash doesn't match the current download, which causes errors at installation, but the netkan bot refuses to re-index the download:

image

This must mean the original file is still in the bot's cache. This pull request makes whitespace-only changes to the affected netkans, so their CKAN-meta entries will be regenerated by Jenkins.

### DIFF
--- a/NetKAN/GPP.netkan
+++ b/NetKAN/GPP.netkan
@@ -1,6 +1,6 @@
 {
-	"spec_version": "v1.4",
-	"identifier": "GPP",
-	"$kref": "#/ckan/netkan/https://raw.githubusercontent.com/Galileo88/Galileos-Planet-Pack/master/CKAN/GPP.netkan",
-	"x_netkan_license_ok": true
+    "spec_version": "v1.4",
+    "identifier":   "GPP",
+    "$kref":        "#/ckan/netkan/https://raw.githubusercontent.com/Galileo88/Galileos-Planet-Pack/master/CKAN/GPP.netkan",
+    "x_netkan_license_ok": true
 }
--- a/NetKAN/GPPCloudsHighRes.netkan
+++ b/NetKAN/GPPCloudsHighRes.netkan
@@ -1,6 +1,6 @@
 {
-	"spec_version": "v1.4",
-	"identifier": "GPPCloudsHighRes",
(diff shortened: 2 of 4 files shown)


HebaruSan:
OK, the new hash looks correct. If the bot overwrites it with the old value, then that confirms that it has the old file in its cache, and we can figure out how to fix that at that point. Either way this should give users a window of having it work, and I think techman83 is probably sleeping now.

HebaruSan:
Hmm, that did nothing. So the post-commit reindexer must share the same cache as the bot. Troubling.

@techman83 , when you see this, can you please delete the bot's cache entry beginning with 5BE7E32D? That should make the bot download the new file. And we should figure out how to handle this case in general now that we check hashes.

Galileo88:
I can just release another version which should resolve all this.

linuxgurugamer:
Note that is what the Epoch is for, to correct for mistakes made with version numbers, etc

This was still a problem, I added an epoch to gpp and that seems to have fixed the problem

HebaruSan:
No, the epoch didn't fix it. The hashes are still wrong in the new file.

```
    "download_hash": {
        "sha1": "95BB6B4ECDCAC1B7F0A62E352AE0B5787720893D",
        "sha256": "66AF7E65B0EAB598C97DCC222B8047DCD675F02D8DF614D01001C2D6D88920BB"
    },
```

HebaruSan:
(The above hashes are now corrected as of KSP-CKAN/CKAN-meta#1326)

techman83:
@HebaruSan yes the Index + the Webhook endpoint share a cache. It does purge the cached objects over 30 days old, but that doesn't cover this scenario. I've manually purged those items.

An Epoch won't fix this as the indexer only uses the hash at the start to resolve the file. I've pondered different ways to work around this, but in practice it's not super common.

HebaruSan:
Thanks @techman83. I think I have a start on a long term solution in KSP-CKAN/CKAN#2337; the GitHub API has fields for a file's timestamp, which should tell us when a file has been updated.

PR #6298, opened 2018-02-28
Gpp epocch

linuxgurugamer:


HebaruSan:
This isn't the problem, and it's still broken. Please hold off on this.

HebaruSan:
@linuxgurugamer , can you please roll this back? We don't need an epoch on this, and the extra CKAN-meta file is going to mess up the fix that I'm working on.

Galileo88:
would it be easier for me to just create a new release and upgrade the version? I'm certain that a fix for this will type of behavior will happen regardless of if I update again?

HebaruSan:
Yes, incrementing the version would fix the problem and be extremely easy for the CKAN team. It's something we'll need to figure out regardless, but maybe not with a horde of GPP fanboys screaming at us. ;)

linuxgurugamer:
Yes, please do so asap
Thanks

Galileo88:
Yeah that's what I thinking. 

GPP fanboys lol

linuxgurugamer:
It's working now

Galileo88:
We haven’t released anything yet

HebaruSan:
It's working now, _temporarily_, due to KSP-CKAN/CKAN-meta#1326. A new release would fix it permanently.

HebaruSan:
The re-break has happened: https://github.com/KSP-CKAN/CKAN-meta/compare/b2d7b4c1dc...d9f25d54b2

Will re-apply the fix momentarily.

I think we can update NetKAN to use info from the GitHub API to detect and handle same-file updates. Investigation continues...

Galileo88:
I just released a new version

PR #6305, opened 2018-03-04
BetterLookingOceans netkan

Galileo88:


HebaruSan:
The title and GitHub repo are "BetterLookingOceans", but you have "BetterLookinOceans" without the 'g' as the module identifiers and netkan file names here. Is that intentional?

Galileo88:
No lol I’ll fix it

HebaruSan:
Looks good now, thanks for adding this.

PR #6310, opened 2018-03-08
FlareReplacer netkan

Galileo88:


HebaruSan:
Could these have more differentiated abstracts, so users have some basis for choosing one of them?

Galileo88:
I mean, not really. They are sun flares and describing them will be even more confusing as some have the same characteristics as others. They can’t be categorized by color either because they are all white or greyscale, and the color is changed by a cfg in the download. There is an image of each flare inside each zip. If the user doesn’t like it, I don’t think it’s that big of an inconvenience to download another flare at that’s less than 2mb in dl size. Do you have any suggestions? I figured doing it by name would suffice. What would be really cool is if CKAN could add a small thumbnail in a tab next to “Versions”. That’s a long shot, I’m sure, but could be helpful.

HebaruSan:
Can you add a link to the forum thread via:

```json
    "resources": { "homepage": "url here" },
```

Then maybe folks could browse the images.

Galileo88:
I can do that. I think what I will do also is compile all the flares into a single zip. This will give them all of the flares, and from there, they can swap them out without having do download them again, instead they will be cached for them., maybe thats not the best way to do it though.
There is now a grand total of 15 flares for people to choose from... I think individual zips will get out of hand...

Galileo88:
Thinking about it, I don't think there is a way to to get these on CKAN without spamming the UI with all these entries. With 15 now, its ridiculous. I don't think I'll add these after all, unless you guys dont have an issue with a ton of flares on CKAN?

HebaruSan:
Personally it doesn't bother me as long as there's some way for people to tell them apart. :)
But if you want to limit it, maybe take a vote on your forum thread and list the top three?

PR #6318, opened 2018-03-10
Update DockRotate.netkan

peteletroll:
updated to ksp 1.4.0


HebaruSan:
Have you considered shipping a .version file instead? Then CKAN could automatically determine the compatible game versions based on the download.

peteletroll:
Will do. Should I remove the ksp_version from the .netkan file then?

HebaruSan:
Yes, once a .version file is available, we would replace the game version with a flag that tells it to check for a .version file. You can see an example at #6319. Though to be clear, your 1.4.0 update _is_ committed now, so you don't have to worry about incorrect versioning in the meantime.

PR #6324, opened 2018-03-13
Updated AlphabeticalKerbals to KSP version 1.4

cecilkorik:
Addon has been tested and doesn't need to be modified, this is just marking it as compatible now that testing is complete.


HebaruSan:
#rebuild

HebaruSan:
#rebuild

PR #6331, opened 2018-03-15
move Scatterer SVE

Galileo88:
Scatterer is no longer required


HebaruSan:
FYI, Scatterer won't appear in recommendations lists: KSP-CKAN/CKAN#2268

Are you sure about this change given that?

Galileo88:
yeah, I'm sure. Scatterer is currently too broken to use in 1.4.1, so I provide a cheap alternate in SVE. Even once Scatterer updates and fixes, Ill likely leave it up to the user to download it.

PR #6335, opened 2018-03-15
Update FuelTanksPlus.netkan

NecroBones:
Remove IFS as a dependency, and instead make it a suggestion.

### DIFF
--- a/NetKAN/FuelTanksPlus.netkan
+++ b/NetKAN/FuelTanksPlus.netkan
@@ -5,11 +5,11 @@
     "license": "CC-BY-NC-SA",
     "spec_version": 1,
     "depends": [
-        { "name": "ModuleManager" },
-        { "name": "InterstellarFuelSwitch-Core"}
+        { "name": "ModuleManager" }
     ],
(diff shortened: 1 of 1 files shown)


HebaruSan:
The forum thread still says IFS is required, which makes me think this change may only apply to a new version. If I merge this now, is it going to mess up an older version?

NecroBones:
That's a good point. This change is valid for 2.0.0 of FTP, going forward. Is there a way to update the netkan to leave it required for previous versions?

NecroBones:
I've updated the forum thread to indicate that the requirement is for pre-2.0.0 versions.

HebaruSan:
The previous versions are effectively frozen in time once a newer version is out, so one way of doing it is to release the new version and then update the netkan after that. The downside of this is that the new version temporarily has the old unupdated metadata. That's probably what I would recommend here, as having an unnecessary dependency for a few hours isn't a big problem.

There are other options (internal .ckan files, version-specific overrides as Sarbian uses), but they're more complicated.

PR #6343, opened 2018-03-16
Add MSP-3000 Material Science Pod

HebaruSan:
This is a part mod that adds a materials science bay with the shape and mass of a mystery goo.

Fixes https://forum.kerbalspaceprogram.com/index.php?/topic/154922-ckan-the-comprehensive-kerbal-archive-network-v1240-bruce/&do=findComment&comment=3324094

### DIFF
--- a/NetKAN/MSP3000.netkan
+++ b/NetKAN/MSP3000.netkan
@@ -1,0 +1,13 @@
+{
+    "spec_version": "v1.4",
+    "identifier":   "MSP3000",
+    "name":         "MSP-3000 Material Science Pod",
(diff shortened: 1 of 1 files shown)


HebaruSan:
@TykoTek, should this mod be limited to any particular game versions? I believe the way we have it now, it will be available on any game version where ModuleManager is also available.

TykoTek:
Should work with any version. It’s just a part config and a couple of models. 

Unless Squad changes how the science or animation modules work it’ll be fine 

It’s been tested with 1.2.x, 1.3.x and 1.4.x

Thanks for your help!

HebaruSan:
Sounds good, thanks for the reply. Merging...

PR #6355, opened 2018-03-18
Update DockRotate.netkan

peteletroll:
Added .version file


HebaruSan:
Looks good! Merging...

PR #6359, opened 2018-03-18
RN FASA update

raidernick:
-Update fasa to use avc version file now, also fixed repo link and
changed TR to RP.

### DIFF
--- a/NetKAN/DockRotate.netkan
+++ b/NetKAN/DockRotate.netkan
@@ -2,8 +2,8 @@
 	"spec_version": 1,
 	"identifier": "DockRotate",
 	"$kref": "#/ckan/github/peteletroll/DockRotate",
+	"$vref": "#/ckan/ksp-avc",
 	"license": "MIT",
-	"ksp_version": "1.4.1",
 	"resources": {
 		"homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/170484-dockrotate-rotation-control-on-docking-ports/"
 	},
--- a/NetKAN/FASA.netkan
+++ b/NetKAN/FASA.netkan
@@ -6,25 +6,28 @@
     "abstract"      : "Re-release of FASA by Raidernick under the new management of Nathankell team.",
     "comment"       : "No Comment",
     "resources": {
-        "homepage"  : "http://forum.kerbalspaceprogram.com/index.php?/topic/22888-105-fasa-544"
+        "homepage"  : "https://github.com/KSP-RO/FASA-RO"
     },
     "$kref"         : "#/ckan/github/KSP-RO/FASA-RO",
     "license"       : "CC-BY-SA",
-    "ksp_version_min"   : "1.2.0",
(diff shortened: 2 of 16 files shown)


raidernick:
It says this passed but can someone double check the syntax and make sure the file has no errors please.

HebaruSan:
The syntax looks fine, and I don't think you need to change the epoch.

However, as it is there are no game version constraints, so the current version would be considered compatible with all versions of KSP. Is that intentional?

Also it would be nice to leave the "homepage" property set to the forum thread. A $kref pointing at GitHub will pull in the repo page as the "repository" property automatically:

```
Resources:
- homepage: https://github.com/KSP-RO/FASA-RO
- repository: https://github.com/KSP-RO/FASA-RO
```

raidernick:
The forum thread is dead and contains a version of the mod that breaks ksp. I don't want to link to a thread that will cause people to break their game if they decide to download something from there.

The game version constraints are in the .version file same as with all my other mods.

HebaruSan:
Ah, OK, never mind the forum thread then.

I just checked [the download](https://github.com/KSP-RO/FASA-RO/releases/download/v7.2/FASA_v7.2.zip), and there is no .version file in it. That's probably why it's not picking up the game versions. If this isn't compatible with all game versions, then we should resolve that before merging this pull request.

raidernick:
It is compatible with all versions since they changed the file structure when mods used to be dae files, that was somewhere back before .18. And no it does not contain the version file yet because I was waiting for this to merge before releasing a new version. If I release that new version first, then you merge this, that version won't be indexed. I'll then need to release another new version AGAIN just to get it indexed, so i'll end up with 2 identical versions.

HebaruSan:
I'm not following that. Why wouldn't the new version be indexed?

raidernick:
I mean if I make a new release right now with a new version file in it, then after that you merge this. That previous release will not be indexed, the same way a mod that is marked as compatible with a certain version of ksp before that version comes out won't be indexed until you make a NEW release after that version of ksp drops.

It's kind of hard to explain, but all of my mods are marked as compatible with up to version 1.3.99 of ksp, so you'd assume that when 1.3.1 dropped for instance, that they would automatically be indexed for that version, but they aren't and only will show up if I make a new release again and that new release is indexed. This is the same situation. That's why I wanted to have this merged right before I made the release.

HebaruSan:
OK, we'll do it your way. This will mess up the metadata for v7.2, but hopefully that won't matter.

raidernick:
Why would it mess up the 7.2 metadata? It shouldn't reindex that release again right? Anyway I'm going to make a 7.3 release right now with an updated version file.

HebaruSan:
The latest release is always re-indexed after a netkan is changed:
https://github.com/KSP-CKAN/CKAN-meta/commit/40f0632a3af0e1260094aa7fcdc2e206a27d0bfa

techman83:
That'd be the webhooks that fire off when commits are submitted to the NetKAN repo.

PR #6375, opened 2018-03-22
Add OhScrap!

severedsolo:


HebaruSan:
Thanks for the contribution, and nice metadata :+1:! Question about this in the forum thread:

> - I've designed this with Kerbal Construction Time in mind, so would recommend that mod (make sure you use the latest dev version)
> - If you plan to actually re-use your parts (and have them fail), you'll probably want StageRecovery too (see Kerbal Construction Time link)

You can represent this in the netkan with optional dependencies. It might look roughly like this:

```json
    "recommends": [
        { "name": "KerbalConstructionTime" }
    ],
    "suggests": [
        { "name": "StageRecovery" }
    ]
```

The difference between `recommends` and `suggests` is whether the checkbox is checked by default (the user still gets to choose either way). Any interest in that? If not, we can merge this as-is, but I wanted to point out the option just in case. Thanks again.

severedsolo:
Oh yeah, I always forget CKAN has that feature. Added them both as recommendations (and fixed an error in the installation path to boot)

HebaruSan:
LGTM! Merging...

HebaruSan:
@techman83 , the status page says "3 days ago" for all mods (and this mod hasn't been indexed). Can you please check on the bot?

techman83:
@HebaruSan the git index had become corrupted, so they've been cleaned out.

HebaruSan:
Thanks! Folks were getting antsy on the forum thread, and that just unleashed [100+ new commits](https://github.com/KSP-CKAN/CKAN-meta/compare/6ce96e87cd...72d174217d)! Looks good now.

PR #6377, opened 2018-03-22
Move nightingale mods to use mod-owned .netkan files.

jrossignol:
Most of the old-style KSP forum links broken and I needed to fix them.  I figured this was a good opportunity to move everything in the netkans into my own repos to make this type of change easier in future.

### DIFF
--- a/NetKAN/ContractConfigurator-AnomalySurveyor.netkan
+++ b/NetKAN/ContractConfigurator-AnomalySurveyor.netkan
@@ -1,30 +1,6 @@
 {
-    "spec_version" : "v1.4",
-    "identifier"   : "ContractConfigurator-AnomalySurveyor",
-    "$kref"        : "#/ckan/github/jrossignol/ContractPack-AnomalySurveyor",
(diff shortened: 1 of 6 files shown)


jrossignol:
Hmm....  will need someone to help me out here, everything looks good as far as I can see, but got this failure:
```
2841 [1] FATAL CKAN.NetKAN.Program (null) - Could not find WaypointManager entry in zipfile to install
```

HebaruSan:
That means it's looking for a folder in the ZIP called WaypointManager and can't find it. I think that's because the `$kref` is pointing to the wrong mod:

```json
    "$kref": "#/ckan/github/jrossignol/Strategia",
```

jrossignol:
Got it, thanks for catching that.  Fixed by https://github.com/jrossignol/WaypointManager/commit/6afec181912bec40abc6231554f933220b6d7b6c, so this can be retried.

HebaruSan:
#rebuild

HebaruSan:
LGTM, merging...

PR #6386, opened 2018-03-24
Make CRP a dependency

Olympic1:


HebaruSan:
@techman83, it looks like the web hooks aren't running anymore either. This problem seems to have started 3 days after the bot freeze-up, but maybe they're still related?

techman83:
@HebaruSan it is possible that a webhook gets missed. The way they work means that all comments get sent to the hooks and it'll wake it up if the original one was missed. Part of the way the Jenkins plugin works.

PR #6388, opened 2018-03-24
Kerbalism: move CRP to dependency

politas:
Requested by N70 on IRC:
><N70> hey
><N70> anyone on
><N70> can someone add CommunityResourcePack as a requirement for Kerbalism now


politas:
I guess because it's identical, there's no conflict.

PR #6402, opened 2018-03-27
Fix spacetux

HebaruSan:
This mod currently suffers from "Could not find spacetux entry in zipfile to install"

This pull request updates the license and install stanza.


HebaruSan:
#rebuild

PR #6429, opened 2018-04-01
Scatterer fix

Dazpoet:
Hoping this works better than #6428 since it include the files that one halted on. I recall the bot handling these things better now so I no longer need to stage PRs like this?

closes #6428


Dazpoet:
Ugh how I hate JSON after not having touched it in forever but I think it works now :)

PR #6442, opened 2018-04-03
Aviation Lights v4.0.0+ now supports KSP 1.4.x - Upping version number.

Rohaq:
I'm honestly a little unsure as to how this affects versions previous to v4.0.0 of the mod, which should still be marked as 1.3 compatible only.


HebaruSan:
There's a .version file in this download, so we might not need to hard code it. Let's see what the test script says.

Rohaq:
Fair enough; MOARdV doesn't do CKAN configs, so hopefully there's a better way forward to keep the version checks up to date!

PR #6451, opened 2018-04-04
Kerbalism

Olympic1:
Adding path for version file to index new release.


HebaruSan:
Looks like this change will break the next "backport" of this mod, if there is one, because the current backport includes only a .version file that doesn't match this pattern:

image

And in the meantime, anyone with KSP-AVC will get errors due to the redundant conflicting version files.

PR #6463, opened 2018-04-08
added jrbuddasKerbalEngineer

CorvetteCole:
This is a fork of the original kerbal engineer redux which is no longer being maintained

### DIFF
--- a/NetKAN/jrbuddasKerbalEngineer.netkan
+++ b/NetKAN/jrbuddasKerbalEngineer.netkan
@@ -1,0 +1,26 @@
+{
+    "spec_version" : 1,
+    "identifier"   : "jrbuddasKerbalEngineerRedux",
+	"name": "Kerbal Engineer Redux",
+    "abstract": "Reveals important statistics about your ship and its orbit during building and flight",
+    "author": "CYBUTEK & jrbudda",
+    "license": "GPL-3.0",
+    "release_status": "testing",
+    "resources": {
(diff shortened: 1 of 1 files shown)


HebaruSan:
@jrbudda, do you want your fork on CKAN? A third party submitted an entry for it here, but it's not clear whether you want that level of exposure yet.

HebaruSan:
I guess we might as well clean this up in the meantime...

@CorvetteCole , this should be using `$kref` instead of `download` and `$vref` instead of hard coded version numbers. See the current file for an example:

https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/KerbalEngineerRedux.netkan

CorvetteCole:
@HebaruSan I think I fixed the syntax things, it is basically the same file as the original redux at this point, just with a new link and auther

jrbudda:
I keep hoping @CYBUTEK will pull my changes and put out an update, I don't *reallly* wan't to become the new 'face' of KER.

I'm not familiar with the inner workings of CKAN, tho I do use it. Would this be a separate listing from the original? or just update the version/download link so folks will get an update notification?

jrbudda:
If it's easy enough to switch to my fork as a 'continuation' of the existing indexed mod then it's also easy to switch it back if cybutek becomes active again (or to someone else if I can't do it anymore). 

In that case let's go ahead and update it for ckan that way.  Do I need to do anything on my end?

PR #6467, opened 2018-04-09
New maintainer for KER

HebaruSan:
@jrbudda has begun maintaining a fork of KER until the original author has time for updates. This pull request updates KER to look at that fork.

Replaces #6463, which attempted to add it as a separate mod.

### DIFF
--- a/NetKAN/KerbalEngineerRedux.netkan
+++ b/NetKAN/KerbalEngineerRedux.netkan
@@ -1,23 +1,17 @@
 {
-    "spec_version" : 1,
-    "identifier" : "KerbalEngineerRedux",
-    "$kref" : "#/ckan/github/CYBUTEK/KerbalEngineer",
-    "$vref" : "#/ckan/ksp-avc",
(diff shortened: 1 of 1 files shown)


HebaruSan:
@politas, @techman83, @Olympic1, is this a good idea? Redirecting KER seems like a big deal due to how widely used it is, and I'm not sure I should be doing that by myself.

Dazpoet:
Switching it back later when the original author comes back might require and epoch bump if they don't follow the current maintainers versioning but that'll still provide an upgradepath so should be fine. Iirc back when we did things like these pretty often, in the wild west days when much was done by hand, we always ensured that users didn't get stuck without a way to upgrade correctly and it usually worked out.

A problem might arise if Cybutek come back and revert some of the changes made to KER in their absence and leave people with the choice to upgrade into a version of the mod which lacks functionallity they presume to still be there.

jrbudda:
So yes? no?

PR #6471, opened 2018-04-10
Get newer DynamicBatteryStorage

HebaruSan:
DynamicBatteryStorage is one of those mods that exists only as a bundled dependency in multiple other mods. Currently it's sourced from KerbalAtomics' download, which hasn't been updated to 1.4. Now it's switched to NearFutureElectrical, which is updated to 1.4. It also will hopefully pull from the right KSP-AVC file now.

Fixes KSP-CKAN/CKAN#2404.


Dazpoet:
Nice, I was wondering about the rationale of this just a few hours ago. It seems NFE and KA depend on this but since KA also depends on NFE-Core it won't work without both NFEs dll and this mod. However NFE does not depend on KA so it would seem logical to drag Battery from NFE instead.

Oh well all is fine now :)

PR #6474, opened 2018-04-12
Trust version file of bundled mods

HebaruSan:
In KSP-CKAN/CKAN#2406, we added the ability to extract the versions of bundled mods separately from the containing mod. I thought there would be several of these, but a script to find shared $krefs found only two:

- CryoTanks (bundled in KerbalAtomics)
- DynamicBatteryStorage (bundled in NearFutureElectrical)

This pull request updates both of them to take advantage of the new setting to get the correct version numbers.


HebaruSan:
#rebuild

PR #6481, opened 2018-04-15
Update ForScience.netkan

numshub:
Update to version 1.5.2

### DIFF
--- a/NetKAN/ForScience.netkan
+++ b/NetKAN/ForScience.netkan
@@ -1,5 +1,5 @@
 {
-    "spec_version": "v1.5",
+    "spec_version": "v1.5.2",
     "identifier": "ForScience",
     "name": "ForScience!",
     "abstract": "ForScience is autopilot for running, collecting and resetting experiments.",


Dazpoet:
Your change is trying to change the spec_version, meaning the version of the ckan.spec we use to interperet the .netkan file. Is the mod not updating? It seems to be hosted through a kref and should update itself in due time from curse.

HebaruSan:
Running netkan.exe locally, I get `ForScience-v1.5.0.ckan`, which is not the latest version:

https://www.curseforge.com/kerbal/ksp-mods/for-science/files

image

It's not picking up v1.5.2 for some reason. At a glance, maybe something to do with the .zip extension?

HebaruSan:
v1.5.2 is missing from the Curse API listing:

https://api.cfwidget.com/kerbal/ksp-mods/for-science

@citricsquid doesn't seem to have a repo for the widget anymore where an issue could be filed, but maybe he'll react to this notification...

citricsquid:
Hi! Sorry for the inconvenience, please visit the [widget homepage](https://www.cfwidget.com/) for further information, specifically:

> Due to changes made to the availability of CurseForge data to third party services we are no longer able to provide fresh data. All data served by this service is accurate as of March 17th 2018. I am working with the CurseForge team to determine the long term status of this service, whether that's shutting down part of (or all of) it, or changes that will allow the service to fetch fresh data again. This page will be updated with any news. Please reach out to me via email (sryan@twitch.tv) with any questions.

HebaruSan:
Thanks @citricsquid, that's a shame but important for us to understand.

Do you have any links regarding "changes made to the availability of CurseForge data to third party services"? I assume that any new effort to scrape this data from curseforge's HTML would be similarly frowned upon.

citricsquid:
unfortunately I don't have any more information at this time. The full back story is that I was notified by someone that their project information was out of date so I investigated and discovered that all attempts to update projects had been failing for a couple of weeks. I reached out to a member of the CurseForge team who informed me that they have made the decision to block scraping of CurseForge, which includes my service.

I have explained that `cfwidget.com` delivers many millions of requests per month and I made mention of this project specifically as one of the many users, however my service was only ever intended as a way to *increase* the success of CurseForge and if it is doing the opposite then I will shut it down -- providing a notice period, during which time it will continue to return stale data as it is now.

I haven't heard back from the team yet. There is the potential that the service will be given permission to continue to scrape CurseForge but at present I have no idea of the likelihood of that. I'll provide you with an update as soon as I know more about the next steps.

HebaruSan:
This pull request's change is invalid and doesn't fix the problem, and it's not a problem that can be fixed in netkan metadata anyway. Closing.

PR #6484, opened 2018-04-16
Updated Perl file, now JSON validator should ignore it

Rohaq:
Assuming I did everything right on the xKAN-meta_testing side, anyway 🤞


HebaruSan:
Isn't it satisfying to see a fix working like that? :+1: 

```
Skipping file 'bin/firespitter-downloader': Not in the NetKAN directory.
```

Rohaq:
I never had a doubt, really 😉

PR #6500, opened 2018-04-22
Add DarkSideTechnologyContinued.netkan

willwill2will54:
The mod is on SpaceDock and Github, but since the author did not use the integrated function in SpaceDock, and apparently does not create 'releases' on GitHub, I decided to create this stopgap until one of those things change.


HebaruSan:
Hmm, it's already indexed (sort of):
https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/DarkSideTechnology.netkan

HebaruSan:
#rebuild

HebaruSan:
#rebuild

HebaruSan:
This is for 1.4.2, dependency RetractableLiftingSurface is for 1.4.1.

willwill2will54:
Perhaps it would be best to reevaluate that label.

PR #6425, opened 2018-04-01
Add K.E.T.O from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of Dankgum, to add [K.E.T.O](https://spacedock.info/mod/1781/K.E.T.O) to CKAN.

Mod details:
    name = /mod/1781/K.E.T.O
    author = Dankgum
    abstract = Adds super-high res clouds to Kerbin, and adds other nice features to planets.
    license = CC - NC - SA 3.0
    Homepage = https://forum.kerbalspaceprogram.com/index.php?/topic/173295-keto-kerbal-extreme-texture-overhaul/
    description = Adds ultra high res clouds to Kerbin and makes the other planets look nice.

**Dependencies**
============
* DiRT (included)
* Scatterer
* EnvironmentalVisualEnhancements
* Kopernicus
* ModuleManager
===========
**Installation Instructions**
============
1. Install dependencies
2. Extract contents of the 'KETO' zip into your gamedata folder
3. ???
4. Profit
============
**Recommended Mods**
============
* PlanetShine
* DistantObjectEnhancement
* StockVisualTerrain
============
**WARNING: A STRONG COMPUTER IS REQUIRED TO RUN THIS MOD!**

Please direct questions about this pull request to [Dankgum](https://spacedock.info/profile/Dankgum).

PR #6512, opened 2018-04-27
Add DiRT

HebaruSan:
Drop-in Replacement Textures is a simplified texture replacement mod.

It's a dependency for #6425. A new release just came out with fixes to the .version file that makes it possible to list on CKAN.


HebaruSan:
#rebuild

PR #6515, opened 2018-04-28
RN change dependency to TU

raidernick:
-Change reflection plugin to Textures Unlimited, this will be the new
dependency from now on


raidernick:
It says the textures unlimited dependency is missing for the 1.3.1 version of KSP and the build failed. This is not true TU DOES have several version for 1.3.1. I think there is an issue with the CKAN-META repo that needs to be fixed here.

HebaruSan:
That looks better. Merging...

PR #6522, opened 2018-05-01
Update KerbalKonstructs.netkan

willwill2will54:
The SpaceDock release 10 days behind the GitHub one, and counting. The GitHub release appears to be released and is perhaps a better $kref for this particular mod. This has been observed in the past, with version 1.3.9.6 of the mod releasing 9 days later on SpaceDock than on GitHub, as one example. CKAN would improve as a service were this switch to be made.


HebaruSan:
Looks reasonable to me; this mod already has a $kref, so it doesn't lose much by moving off of SD.
Merging with thanks...

PR #6527, opened 2018-05-04
limit my mods to 1.3.1

Sigma88:


HebaruSan:
I took a look at SigmaBinary. In the download for 1.6.11 there's a .version file, but it's missing in 1.7.0. That's probably why there's no game version limitation. Can you add a .version file to that download?

Removing the $kref would prevent any further releases from being detected.

Sigma88:
No I removed the KSP-AVC version file because it was yet another thing to update and I don't have the time or patience to update every single thing anymore.

This is also why I would like ckan to only provide that one version for those players that are still on 1.3.1 but not showing the newer versions

Galileo88:
Any ETA on when this will be merged? 

Not to rush or anything, it’s just that GPP has an update in the wings and it depends on this PR getting merged :)

HebaruSan:
I would like to help resolve whatever the specific issue is here, but I don't (personally, just my opinion) see this as merge-able as-is due to the potential harm in hard coding the versions and downloads. Users will complain that new releases aren't being indexed. Indexing them will require hand crafted pull requests for each mod each time such a release comes out. This is the reason why the $kref property exists, to automate the things that can be automated.

Is the only motivation here that specific versions are showing up for KSP 1.4 and shouldn't be? If so there are ways to achieve that without causing other problems. We can even set up version-specific overrides to set KSP compatibility based on some simple comparison logic (see #6528 for an example).

I'll be happy to start a successor pull request once I'm sure I understand the requirements.

HebaruSan:
So no automatic indexing of future updates, ever?
If that's the case it would be simpler to freeze these netkans so they aren't processed at all anymore.

Sigma88:
if you prefere freezing them it's fine for me

HebaruSan:
OK, I will start a pull request for that.
May I inquire as to why you wish to block further updates?

HebaruSan:
Further follow-up question:
This PR has KKtoSD being hard coded to version v1.0.4, but the download is for v0.1.4, and there's a later version already indexed at v0.1.5. Do you need v0.1.5 to be deleted?

Also @Galileo88, why do you need this for GPP? It's just disabling future updates, not adding or removing any current versions. How does that help GPP?

HebaruSan:
I also see that there are some versions indexed now with an extra ".ckan" in the version number, e.g. SigmaDimensions-0.10.0.ckan.ckan. Is that related to this somehow?

Sigma88:
I don't like the attitude of squad and I have lost all interest in making mods for KSP

I'm only updating my mods for galileo at this point

HebaruSan:
OK, I believe the latest versions of your mods should now be as per your preference. Try a refresh in CKAN to make sure.

Sigma88:
yes, it seems they are correct now, thanks

PR #6528, opened 2018-05-04
Create CraftManager.netkan

willwill2will54:
I have reached out to the developer and gained permission. However, he asked that I try to also register a older release of the mod, which is compatible with ksp 1.3.
Should I add something to this file, or create CraftManagerLegacy.netkan? The developer does not use KSP-AVC, though I have advised them to.


HebaruSan:
I added a version override so if he releases more 1.0.X versions they'll automatically be marked as compatible with KSP 1.3.
We'll have to create the .ckan file for 1.0.1 manually, but that's not too difficult.

willwill2will54:
Thanks

PR #6546, opened 2018-05-16
Use version file for DecouplerShroud

nav0t:
I've added a version file to my mod. The "$vref" line will get the ksp version from that file, correct? And does this affect the current release, or just new ones?


HebaruSan:
Yes indeed, $vref does it. You can see exactly what the new metadata will look like if you click on the pull request checks and scroll down:

https://ci.ksp-ckan.org/job/NetKAN/7412/console

Version 0.4.4 would be updated automatically once this is committed, and any future releases would also be affected. Is that OK?

nav0t:
Thanks for the info, everything looks good!

PR #6556, opened 2018-05-20
Ballutes

linuxgurugamer:


linuxgurugamer:
#rebuild

linuxgurugamer:
@HebaruSan Thanks for removing that extra vref.  But it didn't help.
I rebuilt the zip, reuploaded it, and it is still failing, says it can't find the path in the zip, but it is there, and there ARE files in the directory, so I'm stumped.  Any help would be appreciated

linuxgurugamer:
#rebuild

linuxgurugamer:
#release

linuxgurugamer:
#rebuild

linuxgurugamer:
Strange.  I changed the kref to use github and it worked.  Same file

HebaruSan:
It's down to `find` vs `file` in the install stanza. `find` matches subdirectories while `file` only matches full paths, so `file` would need to be set to `GameData/KermangeddonIndustries` to install this mod.

The commit history in this PR is confusing, though. It looks like my commit changed it back to `file`, but I definitely did not do that. Either way, the commits containing `find` have a green checkmark, and the ones with `file` have a red X.

PR #6584, opened 2018-06-05
Adds OSX version of Advanced Fly By Wire

willwill2will54:


HebaruSan:
@linuxgurugamer, these are your mods; should they be indexed?

willwill2will54:
@HebaruSan, is this level of consultation typical for this kind of PR? One of these were already indexed, and the other two are the same mod, just for other platforms.
I don't want to come off as hostile or anything, I am just new to the CKAN platform and am eager learn it's ways.

HebaruSan:
Yes, when a third party requests to index someone else's mod, it's standard practice to try to contact the author.

In this specific case, @linuxgurugamer is very well familiar with how to index his own mods (and typically does it several times per week), so if these aren't indexed yet, there may be a reason for that. It at least makes sense to ask.

willwill2will54:
Well. I found the 'reason for that'. The windows version was already there, and the OSX version was added less than a week ago. I fixed for these new developments, and changed the existing windows config over to the use the provide feature.

willwill2will54:
As a side note, the new system is not really any more consise or simple than the old, but I feel it is more graceful, and covers the eventuality of a 4th major OS taking the world by storm, somehow.
I shall be unreachable for the next hour

linuxgurugamer:
In general, you should always ask permission BEFORE adding anything to CKAN if its not yours.
@HebaruSan I dont have any objection.  Ill leave it to you to validate and approve this.
Im just concerned about existing installs being borked by the change in what is being provided

linuxgurugamer:
There was a very good reasoning n I didnt change the name, so this is changing it.

willwill2will54:
I just tried it out on my own repo. It left previous installs unbeborked.

willwill2will54:
The only problem I can think of is if someone tries to install the OSX version, before the metadata for the one they already have installed is updated. Will the bot update the metadata for mods whose .netkan changed, but have no new release?

HebaruSan:
OK, I think this looks fine. Merging...

PR #6587, opened 2018-06-10
Add KerbalAlarmClock to "recommends" and move UniversalStorage to "suggests".

mwerle:


mwerle:
Incidentally, any reason why only the "abstract" is displayed in CKAN, and not the "description"? Other mods show extended descriptions in the GUI.

HebaruSan:
Looks good, merging...

As for the abstract, that's just how the GUI is coded. Those other mods probably have long abstracts.

https://github.com/KSP-CKAN/CKAN/blob/871678aa88f28b1a5ae3f64af149a1c766ae070e/GUI/MainModInfo.cs#L144

mwerle:
Thanks for the quick merge!  And I left an improvement suggestion for CKAN.

PR #6597, opened 2018-06-14
Add KOOSE

TiktaalikDreaming:
Forgot to tag as CKAN when creating, now SD isn't triggering the scripted thingamy thing.


HebaruSan:
Looks good. Thanks for the submission!

PR #6608, opened 2018-06-28
Add PhotonSailor

HebaruSan:
As per https://forum.kerbalspaceprogram.com/index.php?/topic/154922-ckan-the-comprehensive-kerbal-archive-network-v1251-broglio/&do=findComment&comment=3406789

### DIFF
--- a/NetKAN/PhotonSailor.netkan
+++ b/NetKAN/PhotonSailor.netkan
@@ -5,7 +5,8 @@
     "$kref":        "#/ckan/curse/296653",
     "license":      "CC-BY-NC-3.0",
     "resources": {
-        "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/176144-*"
+        "homepage":   "https://forum.kerbalspaceprogram.com/index.php?/topic/176144-*",
+        "repository": "https://github.com/sswelm/KSP-Interstellar-Extended"
     },
     "install": [ {
         "find":       "PhotonSail",
--- a/NetKAN/PhotonSailor.netkan
+++ b/NetKAN/PhotonSailor.netkan
@@ -1,8 +1,12 @@
 {
-    "identifier": "PhotonSailor",
-    "name":       "Photon Sailor",
-    "$kref":      "#/ckan/curse/296653",
-    "license":    "CC-BY-NC-3.0",
+    "spec_version": "v1.4",
+    "identifier":   "PhotonSailor",
+    "name":         "Photon Sailor",
+    "$kref":        "#/ckan/curse/296653",
+    "license":      "CC-BY-NC-3.0",
+    "resources": {
+        "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/176144-*"
+    },
     "install": [ {
         "find":       "PhotonSail",
         "install_to": "GameData"
--- a/NetKAN/PhotonSailor.netkan
+++ b/NetKAN/PhotonSailor.netkan
@@ -1,0 +1,15 @@
+{
+    "identifier": "PhotonSailor",
+    "name":       "Photon Sailor",
+    "$kref":      "#/ckan/curse/296653",
+    "license":    "CC-BY-NC-3.0",
+    "install": [ {
+        "find":       "PhotonSail",
+        "install_to": "GameData"
+    } ],
+    "recommends": [
+        { "name": "PersistentRotation"      },
+        { "name": "KSPInterstellarExtended" },
+        { "name": "SolarSailNavigator"      }
+    ]
+}
--- a/NetKAN/PhotonSailor.netkan
+++ b/NetKAN/PhotonSailor.netkan
@@ -2,7 +2,7 @@
     "spec_version": "v1.4",
     "identifier":   "PhotonSailor",
     "name":         "Photon Sailor",
-    "$kref":        "#/ckan/curse/296653",
+    "$kref":        "#/ckan/spacedock/1895",
     "license":      "CC-BY-NC-3.0",
     "resources": {
         "homepage":   "https://forum.kerbalspaceprogram.com/index.php?/topic/176144-*",


HebaruSan:
#rebuild

HebaruSan:
#rebuild

HebaruSan:
The Curse API is acting up; it can't find [this project](https://kerbal.curseforge.com/projects/photonsail) whereas [others are fine](https://api.cfwidget.com/project/239249):

https://api.cfwidget.com/project/296653

```json
{"error":"not_found","title":"Project Not Found","message":"Project cannot be found in CurseForge database."}
```

Maybe it needs more time to update its database?
Tagging @citricsquid in case there's a known outage.

citricsquid:
Hey! This is, technically, not an issue. Below are 2 emails I sent last week in response to a similar question.

TLDR: you're trying to use a legacy URL for a non-legacy project. The system has many legacy URLs in the cache, and continues to support them, but for new projects you must use the *current* valid CurseForge URL to access the project data through the widget.

Email 1:

> The widget pulls the data from CurseForge using the project slug, you can determine if a URL is valid based on replace "api.cfwidget.com" with "www.curseforge.com". 
> 
> The reason that the last 2 examples you've provided have results available via the API is because the widget used to be able to support legacy URLs (which what you've provided are), unfortunately that's no longer possible, but we've kept all old URLs cached to prevent those existing URLs from breaking so you'll see inconsistent results with cached legacy URLs working, and uncached legacy URLs not working. Sorry for the inconvenience there.
> 
> To use the widget API you should visit the CurseForge page for a project and then replace "www.curseforge.com" with "api.cfwidget.com". If you're on a project page (e.g: https://minecraft.curseforge.com/projects/forever-stranded-lost-souls) then click "View on CurseForge" (on the right in "About This Project") to be taken to the CurseForge page.
> 
> Due to the fact that this system has existed for many years now and CurseForge has evolved a lot in those years, this system has to account for a lot of legacy things like old URLs. You'll quite often see legacy URLs (e.g: projects/name, e.g: All Of The Mods 3 will redirect to api.cfwidget.com/projects/all-of-the-mods-3) but you should ignore them: only ever try and access a project's API (or widget) using the current www.curseforge.com path and let the system deal with the legacy handling :)

Email 2:

> The reason for this is the legacy URLs. CurseForge projects have been accessible in... 6 different ways over the last few years:
> 
> 1. curse.com (shut down)
> 2. mods.curse.com (shut down)
> 3. curseforge.com (old version) (shut down)
> 4. game.curseforge.com (access blocked)
> 5. curseforge.com (current version) (access blocked)
> 6. private curseforge service 
> 
> My service has supported each of these URLs, and continues to support them for projects that have already been requested even though they are no longer usable to access CurseForge. This is because there are lots of places that these widgets have been embedded with old URLs -- this service is, primarily, intended to deliver widgets, ensuring the best widget service is the priority. This means that my service must keep all of these old URLs cached, and my service determines which is the canonical URL for a project based on which URL was requested first.
> 
> The first time someone requested "All the Mods 3" was last year, using the `projects` URL when the `projects` URL was supported, so the system now believes that `projects/all-the-mods-3` to be the canonical URL for "All the Mods 3".
> 
> When making API requests (to any API, not just this API) it is customary to follow redirects, as your code should not care about where the data resides or the API's implementation details.
> 
> 1. Make a request to the intended URL (e.g: https://api.cfwidget.com/minecraft/modpacks/all-the- mods-3)
> 2. Follow any redirects (default behaviour for most libraries/languages, and if not default, it will be configurable)
> 3. Retrieve the JSON data from the response
> 
> For example: https://repl.it/repls/StupidSalmonThing
> 
> That said, as there is a standardised canonical URL for accessing projects now (`www.curseforge.com/{game}/{mod}/{project}`) I will potentially update the widget system to always redirect to that URL (instead of the first URL it saw when a project was requested) in future, however I am wary of breaking any existing implementations that do not follow redirects so that may not happen for months, or even years.
> 
> So, to summarise:
> 
> 1. Make a request to my API
> 2. Follow the redirects
> 3. Retrieve the content from the body of the final location
> 
> My service will always support legacy URLs.
> 
> Let me know if you need any further information or clarity.

---

Let me know if you need any other information.

HebaruSan:
Hmm, I tried:

- https://api.cfwidget.com/project/photonsail
- https://api.cfwidget.com/projects/photonsail
- https://api.cfwidget.com/kerbal/project/photonsail
- https://api.cfwidget.com/kerbal/projects/photonsail

... and they all return `not_found`. What's the current valid CurseForge URL?

citricsquid:
You must be on a `www.curseforge.com` page, then replace `www.curseforge.com` with `api.cfwidget.com`. This is the only URL that will work for non-legacy projects.

For the project you're talking about here:

1. Visit the project page: https://kerbal.curseforge.com/projects/photonsail
2. Click "View on CurseForge"
3. Replace `www.curseforge.com` with `api.cfwidget.com`

`https://www.curseforge.com/kerbal/ksp-mods/photonsail` becomes `https://api.cfwidget.com/kerbal/ksp-mods/photonsail`

e.g: https://api.cfwidget.com/kerbal/ksp-mods/photonsail

HebaruSan:
Thanks. (Though I would protest that the page I was looking at does say "CurseForge" twice in its header and once in the URL, hence thinking it was a valid CurseForge URL.)

citricsquid:
Yeah, it's confusing :( I tried to ensure that the documentation is as clear as possible ("Visit the www.curseforge.com page for a project...") but given the expectation that when a website says "CurseForge" that it *is* CurseForge people will inevitably be tripped up.

HebaruSan:
@sswelm, we'll need to make some core changes to CKAN to add this mod. This might take a week or two; I have a first pass at the change working locally, but it'll need to be inspected and tested by other team members. If it's important to add it sooner, we could achieve a faster turnaround time if the download was hosted on SpaceDock or GitHub.

Another note, it's detecting the version number of the backport as 1.3.1, because Curse doesn't track an explicit mod version number, so we try to parse the file name, which is "Photon Sailor 1.0 backport to KSP 1.3.1", and we look for a string of numbers and dots at the end, which in this case finds "1.3.1". You might wish to rename it to just "Photon Sailor 1.0" to address that. Since Curse does track _game_ versions explicitly, that info wouldn't actually be lost by such a change.

sswelm:
instead of the Curse download location, why not use the Spacedock download location at https://spacedock.info/mod/1895/Photon%20Sailor

HebaruSan:
Nice, I didn't know it was on SpaceDock! That solved all of the problems we were having, including the version number thing. Merging...

PR #6613, opened 2018-07-01
Force v for FullAutoStrut

HebaruSan:
This mod has some versions with a 'v' and some without, and it's causing confusion.

image

image

This pull request adds the `x_netkan_force_v` property to try to work around this. I haven't used this property before, so I'm not sure it will work, and will be checking the validation output to find out.


HebaruSan:
Looks like that worked.

PR #6615, opened 2018-07-01
added TexturesUnlimited as a recommended plugin

sswelm:


HebaruSan:
The TextureUnlimited.cfg file uses ModuleManager syntax, so I added it as a dependency.
Also told it to look at your .version file.

sswelm:
TextureUnlimited is technically an optional visual feature, the mod will still work without TextureUnlimited, so Module manager is not really a hard dependency. But keep it as it is because I intend to make use of MM script in the future.

PR #6617, opened 2018-07-03
KKL

craidler:
Kerbal Kustomized Ltd - KSP mod to fulfill some gameplay tweaks

### DIFF
--- a/NetKAN/KerbalKustomzLtd-1.0.netkan
+++ b/NetKAN/KerbalKustomzLtd-1.0.netkan
@@ -1,0 +1,28 @@
+{
+    "spec_version"   : "v1.2",
+    "name"           : "Kerbal Kustomz Ltd (KKL)",
+    "abstract"       : "KSP mod to fulfill some gameplay tweaks",
+    "identifier"     : "KKL",
+    "download"       : "https://github.com/craidler/KerbalKustomzLtd/archive/KKL-1.0.zip",
+    "license"        : "LGPL-2.1",
+    "version"        : "1.0",
+    "release_status" : "stable",
+    "ksp_version_min": "1.4.0",
+    "resources" : {
+        "repository"   : "https://github.com/craidler/KerbalKustomzLtd"
+    },
+    "tags" : [
+        "resources"
+    ],
+    "depends" : [
+        { "name" : "ModuleManager", "min_version" : "2.3.5" },
+        { "name" : "MechJeb2" }
+    ],
+    "recommends" : [
+        { "name" : "Mk1CabinHatch" },
+        { "name" : "InterstellarFuelSwitch" },
+        { "name" : "NearFutureSolar" },
+        { "name" : "NearFutureConstruction" },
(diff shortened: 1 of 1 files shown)


HebaruSan:
#rebuild

HebaruSan:
#rebuild

HebaruSan:
Hi @craidler, thanks for the submission!

A netkan file can be set up to automatically index new releases as they come out, but the way this is set up right now you would have to do a pull request in this repo for each release. There are a couple of decent options:

- Meta-netkan: We could make it reference the netkan file on your own repo, so you could update the metadata as needed without pull requests.

- GitHub kref: We could set it up to get metadata and releases from your repo. This would require adding official releases at https://github.com/craidler/KerbalKustomzLtd/releases, but a CKAN bot would take care of the rest.

Do either of those sound appealing to you?

craidler:
Hi thanks, yeah sounds great. What can I do to automate the process? I can do the releases thingy. Do you have a link to a How To set that up?

HebaruSan:
For that you would replace the download property with:

```json
    "$kref": "#/ckan/github/craidler/KerbalKustomzLtd",
```

Then it will check the list of releases and index the latest one automatically. This also causes it to auto-populate several fields from the repo itself, such as name, author (missing currently), and abstract.

For maximum maintainability, you can add a .version file and this as well:

```json
    "$vref": "#/ckan/ksp-avc",
```

But that's probably overkill for a few MM configs.

craidler:
Yeah, dont need to overkill, but reduce overhead. So I have replaced the download property. How will CKAN now keep the versions, dependencies and other properties up to date? I want to include some crafts and stuff in future. So some structures will change.

HebaruSan:
The properties will be copied from the netkan file to the index entry for the latest version when the bot runs (about every 3-4 hours), except for the `download` and `version` properties, which will be set based on the contents of the release from https://github.com/craidler/KerbalKustomzLtd/releases . So if you need to add dependencies or craft files, you would update the netkan file after you create the new release.

craidler:
Did I do stuff right? I have added a pre release tag in my repo there. Want to do some testing on the plugin part, before pushing a v1.1 public. Maybe I can get some feedback on what changes the repo needs?

HebaruSan:
CKAN doesn't index pre-releases, to allow authors a way to create a test version without having all CKAN users install it, so we'll need a true release for the pull request to work.

And you'll probably want to prepare a proper download ZIP for installation, minus the source code. Right now all that's available is an archive of the whole repo, which isn't needed just to install the mod. Some examples here: https://github.com/HebaruSan/Astrogator/releases

craidler:
Aight, thank you very much! I'll work on it.

HebaruSan:
#rebuild

HebaruSan:
Nice! Glad it could be of service. :grin: 

This pull request looks good now, too. GitHub wasn't recognizing that latest commit for a while, so I put off merging this, but that problem seems to have sorted itself out. Merging now...

PR #6618, opened 2018-07-06
Added recommended mods to KSPIE

sswelm:


HebaruSan:
#rebuild

HebaruSan:
Updated CTT to support its official range of 1.4.2-1.4.9 in #6619. Now InterstellarFuelSwitch-Core is the problem, but that one seems to be officially for 1.4.3.

These changes are fine, though. There are times when we want a missing dependency to block things, but this isn't one of them. Merging...

PR #6626, opened 2018-07-10
Add Tundra Technologies

politas:
Closes KSP-CKAN/CKAN#2473


politas:
Or should it be a recommend, rather than a depends?

politas:
Thanks for fixing that, @HebaruSan! I've clearly been ignoring metadata for too long!

PR #6631, opened 2018-07-14
update PAPIPluginContinued

bssthu:


HebaruSan:
Why are you changing the override section? This version already has those values, and we probably don't want to remove the old ones:

https://github.com/KSP-CKAN/CKAN-meta/blob/master/PAPIPluginContinued/PAPIPluginContinued-v0.4.9.ckan

bssthu:
In this case, I shouldn't change that section.

PR #6632, opened 2018-07-14
Add KSP_GPWS addon to CKAN

bssthu:


bssthu:
will fix "Invalid Identifier" error

PR #6633, opened 2018-07-14
Add KSPGPWS addon to CKAN

bssthu:


HebaruSan:
We've got an old placeholder for this mod here:

https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/GPWS.frozen

I'll make a few changes to account for that...

HebaruSan:

PR #6644, opened 2018-07-21
Couple of fixes for Tundra Exploration

blowfishpro:
* It depends on Kerbal Reusability Expansion, so include that dependency
* Allow installation of example craft

This was done in talks with the author, @damonvv

### DIFF
--- a/NetKAN/TundraExploration.netkan
+++ b/NetKAN/TundraExploration.netkan
@@ -20,6 +20,12 @@
             "find"       : "TundraExploration",
             "install_to" : "GameData",
             "filter"     : "Thumbs.db"
+        },
+        {
+            "find": "Ships/VAB",
+            "install_to": "Ships",
+            "optional": true,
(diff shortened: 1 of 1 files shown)


blowfishpro:
Any advice on the failure?  It seems to be looking for installed ships in `<ksp_root>/Ships/Ships`

blowfishpro:
@HebaruSan can you force a reindex for the most recent version or is there a different process this should go through?

HebaruSan:
That happens automatically after a merge:

https://github.com/KSP-CKAN/CKAN-meta/commit/cea4a688736f39aed102206622df9363a78a0f9c

blowfishpro:
Oh good.  I was looking at status.ksp-ckan.org but I guess it doesn't update the last indexed timestamp unless there's actually a new version

HebaruSan:
Oh, that page only reflects the status of the bot; pull requests are treated separately. It is confusing!

PR #6647, opened 2018-07-28
Update KerbalLaunchVehicles.netkan

efalder413:
Mod author has moved Github accounts, updating for 1.4


HebaruSan:
Thanks for the update!

PR #6648, opened 2018-07-31
Added full package recommendation to Core

sswelm:


HebaruSan:
What problem is being solved here? There was probably a reason these were split out in the first place, which would be undermined by this change.

PR #6652, opened 2018-08-02
Create PreciseEditor.netkan

jfrouleau:
Adding a new mod called Precise Editor.

### DIFF
--- a/NetKAN/PreciseEditor.netkan
+++ b/NetKAN/PreciseEditor.netkan
@@ -1,0 +1,17 @@
+{
+    "spec_version"   : 1,
+    "identifier"     : "PreciseEditor",
+	"name"           : "Precise Editor",
+	"abstract"       : "Editor mod to set a part's position and angle values with great precision. Also show CoM, CoL and CoT coordinates.",
+	"author"         : "jfrouleau",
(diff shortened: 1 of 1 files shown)


HebaruSan:
Here's the current error message:

```
#   Failed test 'PreciseEditor has a $kref/$vref and a download field, this is likely incorrect.'
#   at t/metadata.t line 63.
```

Usually you want a `$kref` so CKAN can determine the `download` property automatically. However, that will also raise an error because of the way the releases are currently set up; you want the download to be listed under Assets rather than linked in the description, see this example:

https://github.com/Kopernicus/Kopernicus/releases/tag/release-1.4.5-2

So I'd recommend:

- Edit the release to list the download under Assets
- Remove the `download` property from the .netkan

HebaruSan:
Looks good now! Just a few more sanity checks to go...

The `license` field says it's GPL 3, but the LICENSE.md file in the download looks like the MIT license. Should that be updated as well?

jfrouleau:
I'm not too familiar with licences but I think the one I'm looking for is GPL 3.0. I will update my mod and get back to you.

jfrouleau:
Okay I made a new release of the mod with the updated license to GPL 3.0. It's ready now. :-)

HebaruSan:
Cool, looks good, and thanks for the submission! Merging...

PR #6653, opened 2018-08-03
CommunityPartsTitles

yalov:


yalov:
What is wrong there?

HebaruSan:
Looks like it was caused by a [byte order mark](https://en.wikipedia.org/wiki/Byte_order_mark):

```
NetKAN/CommunityPartsTitles.netkan:1:0: Warning: JSON document was prefixed by a BOM (Byte Order Mark): '\xef\xbb\xbf'
   |  At line 1, column 0, offset 0 (AT-START)
```

Removing the character before the open brace seems to have fixed it.

PR #6658, opened 2018-08-04
Add RemoteTechRedevAntennas.netkan

KSP-TaxiService:
Hi, I would like to add a new part package of RemoteTech Redev (2) to the CKAN list for third-party modders to refer without getting into license issues.

Thank you.

### DIFF
--- a/NetKAN/RemoteTechRedevAntennas.netkan
+++ b/NetKAN/RemoteTechRedevAntennas.netkan
@@ -10,7 +10,7 @@
     ],
     "install": [
         {
-            "file": "GameData/RemoteTech2/RemoteTech-Antennas",
+            "file": "GameData/RemoteTech",
             "install_to": "GameData"
         }
     ],
--- a/NetKAN/RemoteTechRedevAntennas.netkan
+++ b/NetKAN/RemoteTechRedevAntennas.netkan
@@ -10,7 +10,7 @@
     ],
     "install": [
         {
-            "file": "GameData/RemoteTech/RemoteTech-Antennas",
+            "file": "GameData/RemoteTech2/RemoteTech-Antennas",
             "install_to": "GameData"
         }
     ],
--- a/NetKAN/RemoteTechRedevAntennas.netkan
+++ b/NetKAN/RemoteTechRedevAntennas.netkan
@@ -3,8 +3,6 @@
     "identifier"   : "RemoteTechRedevAntennas",
     "$kref"        : "#/ckan/spacedock/1929",
     "license"      : "GPL-3.0",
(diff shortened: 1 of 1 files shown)


HebaruSan:
It's currently not installing to the same path as in the ZIP file; the RemoteTech folder is removed, e.g.:

- GameData/RemoteTech-Antennas/Parts/GigaDish1/model004.dds

Is that OK?

yalov:
@HebaruSan 
CKAN uses `identifier` for "conflicts", "depends", etc, or also uses folder name?

Will third-party fail, if path would be `GameData/RemoteTech/RemoteTech-Antennas/`?
 
```
    "conflicts": [
        { "name": "RemoteTech" }
    ],
    "depends": [
	{ "name": "RemoteTechRedevAntennas" }
    ],
```

@KSP-TaxiService can you place it somewhere different from GameData/RemoteTech/, for preventing confusion?
`GameData/RemoteTech2/RemoteTech-Antennas/` or `GameData/RemoteTech-Antennas/`

HebaruSan:
Either path would be fine for CKAN. I'm only asking because some mods require specific paths, and if this is one of them, then it might not work as currently configured.

KSP-TaxiService:
Hi, I didn't realise this naming conflict. I made a renaming and sent new commit.

Please help to restart the failed Jenkins build as I forgot to upload the new SpaceDock release before this commit.

HebaruSan:
Sorry, I wasn't clear, there was no name conflict relating to identifiers or folders or anything like that; please disregard @yalov's remarks, as they're not relevant. You can put the ZIP back the way it was.

The only problem is that it might be installing incorrectly. Inside the ZIP, you have this path prefix:

- GameData/RemoteTech/RemoteTech-Antennas

And so presumably someone installing manually would install to that same folder. But with the current metadata, CKAN would install it to this path instead:

- GameData/RemoteTech-Antennas

As you can see, part of the path has been stripped out (the RemoteTech folder level), so if there are any parts of the mod that rely on the path, they would break. This is because of how the install stanza in the metadata is set up, not because of any inherent limitation of CKAN or a naming conflict; it's no problem to install to that full path if it's necessary. To fix it, you would make this change:

```json
    "install": [
        {
            "file": "GameData/RemoteTech",
            "install_to": "GameData"
        }
    ],

```

KSP-TaxiService:
Hi, 

Thanks for clarifications on this. I pushed a new commit and can see the correct directory `GameData/RemoteTech/RemoteTech-Antennas/*`

I think all is set to be ready.

HebaruSan:
Yup looks good! Thanks for your patience, merging...

PR #6660, opened 2018-08-06
add CommNetAntennasExtension

yalov:
After passing thorough the netkan,  in CKAN - "File/install from .ckan" do nothing

### DIFF
--- a/NetKAN/CommNetAntennasExtension.netkan
+++ b/NetKAN/CommNetAntennasExtension.netkan
@@ -1,0 +1,25 @@
+{
+    "spec_version" : "v1.4",
+    "identifier"   : "CommNetAntennasExtension",
+    "$kref"        : "#/ckan/github/yalov/CommNetAntennasExtension",
+    "name"         : "CommNet Antennas Extension",
+    "author"       : "flart",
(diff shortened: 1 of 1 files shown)


HebaruSan:
RemoteTechRedevAntennas is only available for 1.3.0 (because that's what's set on SpaceDock), so this mod would have that same limitation. @KSP-TaxiService, should we make that mod available in more versions?

KSP-TaxiService:
I tried to mark the same release file on SpaceDock for both 1.3 and 1.4 and ended up in going with lowest common denominator. I guess I will have to bump version by tiny and put up for 1.4 later.

HebaruSan:
Another option is to include a KSP-AVC .version file in the download. CKAN can consume such files to support the exact version range you wish, for example:

https://github.com/HebaruSan/Astrogator/blob/master/Astrogator.version

HebaruSan:
#rebuild

HebaruSan:
OK, I added an override to expand RemoteTechRedevAntennas's compatible game versions in #6663. We can still go the .version file route down the line, but for now this should work.

PR #6662, opened 2018-08-07
add Survey Transponder

efalder413:
https://forum.kerbalspaceprogram.com/index.php?/topic/111834-14-survey-transponder-052a/


HebaruSan:
@efalder413, if we add this, would you be willing to help us keep it up to date? Since taniwha uses his own server that we can't query for new versions, we would have to update this file manually every time he releases a new version of this mod (as we do for all of his mods). Dropping us a note when that happens would be appreciated.

efalder413:
I certainly can!

HebaruSan:
Thanks!

PR #6693, opened 2018-08-28
Use RealHeat's .version file

### LINKED ISSUE #6692: What's the right way to maintain RealHeat (and other mods)
Background
----------

**CKAN Version:** 1.25.3

**KSP Version:**  1.4.5

**Operating System:** Windows

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?** yes

**Affected mods and mod versions:** RealHeat

Problem
-------

**What steps did you take in CKAN?**
View the RealHeat mod 

**What did you expect to happen?**
it would show a min ksp version of 1.4.5

**What happened instead?**
the min version is 1.4.3

I'm asking this because i'm trying to figure out what the _right_ wait to properly maintain this on ckan is.  https://github.com/KSP-CKAN/NetKAN/pull/6573 said that RealHeat does not have a .version file, however there is one here: https://github.com/KSP-RO/RealHeat/blob/master/RealHeat/RealHeat.version

What is the preferred method for maintaining a mod on ckan?  I've seen at least 2 approaches, having a .netkan file somewhere in the project, or having a .version file.  Is it also possible to just add the .netkan file as a PR to NetKAN? Is the .version file only used if you're using AVC?  I'm trying to understand what I need to do to make this a bit more seamless when I updated the versions of mods.

HebaruSan:
RealHeat's latest release has a version file, so this pull request removes the hard coding of `ksp_version` in favor of a `$vref`.

@rsparkyc, tagging you in case GitHub doesn't send you a notification automatically.

Fixes #6692.

### DIFF
--- a/NetKAN/RealHeat.netkan
+++ b/NetKAN/RealHeat.netkan
@@ -2,7 +2,7 @@
     "spec_version" : "v1.4",
     "identifier"   : "RealHeat",
     "$kref"        : "#/ckan/github/KSP-RO/RealHeat",
-    "$vref"        : "ksp-avc",
+    "$vref"        : "#/ckan/ksp-avc",
     "name"         : "RealHeat",
     "abstract"     : "Improves the realism of KSP's thermodynamics.",
     "license"      : "CC-BY-SA",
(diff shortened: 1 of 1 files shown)


HebaruSan:
To sum up, if a .netkan file has this:

```json
    "$vref": "#/ckan/ksp-avc",
```

... then the `ksp_version` / `ksp_version_min` / `ksp_version_max` properties will be set based on the .version file contained in the download.

Thanks for helping us automate this!

PR #6694, opened 2018-08-28
RN update rss textures version

raidernick:
-update version to support ksp 1.4.x

### DIFF
--- a/NetKAN/RSSTextures2048.netkan
+++ b/NetKAN/RSSTextures2048.netkan
@@ -7,7 +7,7 @@
     "license"         : "CC-BY-NC-SA",
     "release_status"  : "stable",
     "ksp_version_min" : "1.3.0",
-    "ksp_version_max" : "1.3.999",
+    "ksp_version_max" : "1.4.999",
     "resources": {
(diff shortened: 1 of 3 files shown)


HebaruSan:
Thanks for the updates! Merging...

raidernick:
@HebaruSan is this some sort of new change to how versioning works where you don't need to come up with some large version number to cover all potential game versions? Does it need to be done in the .version files included with the mods themselves also that don't set the version number in the actual netkan file?

HebaruSan:
It's not new; the code that translates "1.4" into [1.4.0, 1.5.0) is at least 2 years old:

https://github.com/HebaruSan/CKAN/blob/547976009d479b0ca5a24e51375eaa186866cd52/Core/Versioning/KspVersion.cs#L261-L265

As for .version files, I would defer to the author and/or maintainer of KSP-AVC, as [the spec](http://ksp.cybutek.net/kspavc/Documents/README.htm) doesn't say whether the `PATCH` property is required. But in general, CKAN's `ksp_version` properties are related to but not exactly the same as KSP-AVC's `KSP_VERSION` properties.

PR #6695, opened 2018-08-28
update KerbalRenamer to use .version file for versioning

rsparkyc:
this is my first NetKAN PR, so please let me know if I did something wrong, but this should be similar to what was just don for RealHeat here https://github.com/KSP-CKAN/NetKAN/pull/6693


HebaruSan:
Looks good! You can click "Show all checks" and then "Details" to see the validation log.

https://ci.ksp-ckan.org/job/NetKAN/7620/

Merging...

PR #6696, opened 2018-08-29
Use .version info for KSCSwitcher

rsparkyc:


HebaruSan:
This download doesn't have a .version file, unfortunately, so this change drops all compatibility info. Looks like the best we could do would be to update the hard coded version.

rsparkyc:
Ahh, the repo does, but it needs to be in the download.  I can fix that

rsparkyc:
This should now have the .version file
https://github.com/KSP-RO/KSCSwitcher/releases/tag/1.4.5.1

HebaruSan:
#rebuild

PR #6738, opened 2018-10-07
Update EvaFollower.netkan

Artemonim:
The correct homepage for the mod is set.


HebaruSan:
Looks good, keep 'em coming!

PR #6748, opened 2018-10-08
Update CommunityTechTree.netkan

Artemonim:
Suggest CTT supported mods.


HebaruSan:
Should this be "supports" as well?

Artemonim:
> Should this be "supports" as well?

My fall.

PR #6751, opened 2018-10-08
Update JX2Antenna.netkan

Artemonim:
Updated homepage info and added compatible mods as suggested.


HebaruSan:
Let's not set the homepage here. It will be pulled from SpaceDock as needed, which allows the author to control it more directly.

PR #6753, opened 2018-10-08
Update GPP.netkan

Artemonim:
Added JX2Antenna as suggestions since given this mod as an example on JX2Antenna homepage.


HebaruSan:
Try a PR here instead, so the author can retain control:
https://github.com/Galileo88/Galileos-Planet-Pack/blob/master/CKAN/GPP.netkan

PR #6754, opened 2018-10-08
Update OuterPlanetsMod.netkan

Artemonim:
Added JX2Antenna as suggestions since given this mod as an example on JX2Antenna homepage.

### DIFF
--- a/NetKAN/OuterPlanetsMod.netkan
+++ b/NetKAN/OuterPlanetsMod.netkan
@@ -18,7 +18,8 @@
         { "name": "CustomBarnKit" }
     ],
     "suggests": [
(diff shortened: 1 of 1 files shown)


Artemonim:
I can't understand what's wrong :(

HebaruSan:
This is the error message:

> OuterPlanetsMod 2:2.2.1 requires CommunityTerrainTexturePack but it is not listed in the index, or not available for your version of KSP.

That's happening because the script is trying to install on KSP 1.4.5, but CTTP only goes up to 1.4.2. This is one of the errors that we sometimes override if it looks like it will be OK.

PR #6758, opened 2018-10-08
Update KSTS.netkan

Artemonim:
Add conflict and suggests sections


HebaruSan:
I believe this would override the author's ability to control those fields.
Please submit a PR here instead:
https://github.com/mmoench/KSTS/blob/master/KSTS.netkan

PR #6759, opened 2018-10-08
Revert "Update SpaceXLegs.netkan"

Artemonim:
Reverts KSP-CKAN/NetKAN#6757

Suggests is supports. My fall  ¯\\\_(ツ)_/¯


HebaruSan:
We don't need to revert the previous change, we can just change suggests to supports in a new change.

HebaruSan:
Replaced by #6760.

PR #6768, opened 2018-10-09
Update ColorCodedCans.netkan

Artemonim:
Add supports section

### DIFF
--- a/NetKAN/ColorCodedCans.netkan
+++ b/NetKAN/ColorCodedCans.netkan
@@ -12,5 +12,12 @@
     ],
     "recommends": [
         { "name": "FuelTanksPlus" }
+    ],
+    "supports": [
+        { "name": "ModularRocketSystem" },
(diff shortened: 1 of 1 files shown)


HebaruSan:
Hey @Artemonim, I just wanted to say thanks for the flurry of updates!

This might be taking advantage, but there's an under-used "tagging" feature that needs a lot of metadata added to be useful. See:

- KSP-CKAN/CKAN#1021
- KSP-CKAN/CKAN#2034
- KSP-CKAN/CKAN#2155
- https://github.com/KSP-CKAN/CKAN/wiki/Suggested-Tags

If you were interested, we would also be grateful for any pull requests along those lines.

Artemonim:
I just wanted to choose mods for my game and stuck here for a week :D

Maybe I'll try it, but I can't understand where CKAN users can see that tags and how they use it now. This could be the basis of an excellent mod recommendation system by the way.

PR #6770, opened 2018-10-09
Update SurfaceExperimentPack.netkan

Artemonim:
Update homepage link


HebaruSan:
Forum thread says:

> Need a lander to carry your SEP down to the surface? Like a fine wine, SEP pairs excellently with Bluedog Design Bureau - which includes a stockalike version of the LEM lander!

That sounds like a "supports" clause.

PR #6775, opened 2018-10-13
Update OuterPlanetsMod.netkan

Artemonim:
Sync info with homepage and add tags section

### DIFF
--- a/NetKAN/OuterPlanetsMod.netkan
+++ b/NetKAN/OuterPlanetsMod.netkan
@@ -38,8 +38,7 @@
         { "name": "DeadlyReentry" }
     ],
     "tags": [
-        "other",
-        "environment"
+        "planet-pack"
     ],
     "x_netkan_epoch": "2",
     "$vref": "#/ckan/ksp-avc",
--- a/NetKAN/OuterPlanetsMod.netkan
+++ b/NetKAN/OuterPlanetsMod.netkan
@@ -37,7 +37,7 @@
         { "name": "SCANsat" },
         { "name": "DeadlyReentry" }
     ],
-    tags : [
+    "tags": [
         "Other",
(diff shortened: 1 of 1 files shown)


HebaruSan:
"other" and "environment" doesn't scream OPM to me. We should probably have a tag for planet packs. Can you think of a better one than "planet-pack"?

Artemonim:
I think that "environment" is good for all mods that change world of KSP such as OPM, Kerbin-Side, etc

HebaruSan:
`environment` sounds too broad, it could mean something like enhancements to grass and trees. If I was looking for planet packs in CKAN, I wouldn't necessarily expand a list called "environment". `locations` is better, but that could mean things like new launch sites around Kerbin.

How about `solar-system`?

Maybe we should do a poll on the forum, or invite more folks to post here.

Artemonim:
English is not my native language so I'm not good at finding semantic synonyms.
I think that CKAN can use tags with different width of meanings like Netflix and other recommendation system. So, there may be a tag cloud in the form of subcluds like `environment`\\`solar-systems`\\`planets`\\`plants`

Poodmund:
In future updates we will plan to host our own NetKAN metadata file remotely so these changes are a moot point going forward for future updates. Suggestions on the metadata are welcome however.

Artemonim:
I believe that sections "recommendations" and "suggests" is enough.
The only thing that needs to be done is to set up a two-way recommendations for mods. Now we have only one-way recommendation system.

HebaruSan:
Well, I asked for feedback on the forum, and surprisingly we got only one response, an endorsement of `planet-pack`. I guess most folks are waiting to complain after we pick something they don't like.

I'm going to revise the wiki a bit and go with that.

PR #6776, opened 2018-10-14
Update GAP.netkan

inigmatus:
need to update recommended mods.


HebaruSan:
Thanks! This will also update the other file automatically.

PR #6781, opened 2018-10-17
Update AutomatedAerialRefueling.netkan

Artemonim:
Removed unnecessary dependencies


HebaruSan:
Confirmed the forum thread says:

> DOWNLOAD: SpaceDock (Also in CKAN, but ckan will download Aircraft Carrier Accessories also, we're trying to fix it.)

How do you even find these?? :+1:

PR #6789, opened 2018-10-24
OPM - Remote hosting of NetKAN metadata

Poodmund:
Update the NetKAN metadata to point to remotely hosted NetKAN metadata.

### DIFF
--- a/NetKAN/OuterPlanetsMod.netkan
+++ b/NetKAN/OuterPlanetsMod.netkan
@@ -1,50 +1,7 @@
 {
     "spec_version": "v1.4",
-    "install": [
-        {
-            "find": "OPM",
-            "install_to": "GameData"
-        }
-    ],
-    "resources": {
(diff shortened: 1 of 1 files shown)


Poodmund:
We are planning to manage our own metadata going forward as we have with GPP. I am not sure if the changes made here, to the NetKAN repo data, is correct as I just copied it from the local GPP one. Please advise if necessary.

HebaruSan:
You'll want to keep this part in your remote netkan.

```json
    "x_netkan_epoch": "2",
```

Without it, new versions will be indexed as being older than previous versions.

Poodmund:
As so: https://github.com/Galileo88/Outer-Planets-Mod/blob/master/CKAN/OPM.netkan ?

HebaruSan:
#rebuild

Poodmund:
Ah I see. Thanks for the assistance and the knowledge bomb. 😄

PR #6790, opened 2018-10-24
Changes Kerbalism to use spacedock

steamp0rt:
because i'm more likely to update the spacedock lol


HebaruSan:
Fair enough! Looks like all the properties are set properly. Merging...

PR #6791, opened 2018-10-25
CTTP - Remote hosting of NetKAN metadata

Poodmund:
Update the NetKAN metadata to point to remotely hosted CTTP NetKAN metadata.

### DIFF
--- a/NetKAN/CommunityTerrainTexturePack.netkan
+++ b/NetKAN/CommunityTerrainTexturePack.netkan
@@ -1,19 +1,5 @@
 {
     "spec_version": "v1.4",
-    "identifier": "CommunityTerrainTexturePack",
-    "name": "Community Terrain Texture Pack",
-    "abstract": "High-quality textures for use by planet creators.",
-    "license": "CC-BY-NC-SA-4.0",
-    "$kref": "#/ckan/github/Galileo88/Community-Terrain-Texture-Pack",
-    "$vref": "#/ckan/ksp-avc",
-    "x_netkan_epoch": "1",
-    "resources": {
-        "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/165873-ksp-13-community-terrain-textures-pack-10-26-sept-2017/"
-    },
-    "install": [
-        {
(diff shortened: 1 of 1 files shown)


HebaruSan:
It seems to be confused by

```json
    "ksp_version": "any",
```

... but that's explicitly allowed according to the spec, so we may have some code fixing to do here. Please stand by.

HebaruSan:
OK, I've submitted a fix for that. It might take a couple of days for another CKAN dev to review and merge it, so in the meantime if you wish you can workaround this issue by removing that `"ksp_version"` property from your netkan.

Poodmund:
I don't mind waiting. That was pretty much the whole reason for the metadata update as CTTP is just a folder with a bunch of textures in and will work on any version of KSP; we don't want to have to keep updating it going forward. 

Thanks for your work.

HebaruSan:
OK, waiting is fine with me, but just for the record removing that property would not change any behavior because ["any" is the default](https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md):

> If no KSP target version is included, a default of "any" is assumed.

The current difficulty is just a weird parsing quirk.

Poodmund:
Oh okay that is good to know. Well I'll leave it as is anyhow as then it documents the edge case nicely.

HebaruSan:
#rebuild

HebaruSan:
New error:

```
Validating built/CommunityTerrainTexturePack-1-1.0.4.ckan.. Failed! See below for error description.
None is not of type u'string'

Failed validating u'type' in schema[u'properties'][u'ksp_version']:
    {u'description': u'A version of KSP',
     u'pattern': u'^(any|[0-9]+\\.[0-9]+(\\.[0-9]+)?)$',
     u'type': u'string'}

On instance[u'ksp_version']:
    None
```

The fix worked fine (since netkan was able to generate the .ckan file), but apparently the bot's extra validation checks don't like the output. Investigating...

HebaruSan:
#rebuild

HebaruSan:
w00t

PR #6795, opened 2018-10-26
EarlyBird: Delete bak file, restore vref, use ksp-avc kref

HebaruSan:
#6794 added a file called `EarlyBird.netkan.bak`. This is now deleted.

It also added `EarlyBird.netkan` with hard coded download and version info. Now this is replaced with a kref using this mod's public .version file (added in KSP-CKAN/CKAN#2517), plus a vref.

The abstract had a bunch of BOM characters embedded in it. Now they're gone.


linuxgurugamer:
Thanks, sorry about the .bak.  Could you add the following to the .gitignore for the directory:

*~
*.bak

Would avoid these issues in the future

HebaruSan:
No problem. Sure, *~ is already in the ignore, added bak in 56d07e3ba94edf1944bf5ed404ea771ab27edc20.

PR #6797, opened 2018-10-27
Update KSPIE folder path

HebaruSan:
The latest download doesn't include GameData.


HebaruSan:
The InterstellarHybridRocketry folder is missing as well. The author said something implying that this is kind of a beta release, but it's been uploaded to SpaceDock as normal. Not sure about the best way to proceed here...

HebaruSan:
#rebuild

HebaruSan:
This isn't needed anymore. GameData is back.

PR #6807, opened 2018-10-31
Update m2X Netkan

SuicidalInsanity:
Updates M2X dependency following switch from IFS to B9PS


HebaruSan:
Thanks for the update. Merging...

PR #6808, opened 2018-10-31
ScienceLabInfo abstract

yalov:
"—" was showed as "???" in the abstract


HebaruSan:
If we delete that field, then it'll pull from the description of your GitHub repo:

image


What do you think of that option?

yalov:
ok, I didn't know. The field can be deleted

PR #6811, opened 2018-11-01
New FAR fork

dkavolis:
Replaced the restricted assets with different ones so there are no licensing issues. Would like to get this on CKAN so everyone can use one version and not many different recompiles. 

If I was providing backports, how would that work with CKAN? Would I need to create multiple releases one for each KSP version?

Edit: build fails because ModularFlightIntegrator is not found for KSP 1.5 but a version from KSP 1.4.5 works.

### DIFF
--- a/NetKAN/FerramAerospaceResearch.netkan
+++ b/NetKAN/FerramAerospaceResearch.netkan
@@ -1,18 +1,18 @@
 {
     "spec_version": 1,
     "identifier": "FerramAerospaceResearch",
-    "$kref": "#/ckan/github/dkavolis/Ferram-Aerospace-Research",
+    "$kref": "#/ckan/github/ferram4/Ferram-Aerospace-Research",
     "$vref": "#/ckan/ksp-avc",
     "x_netkan_epoch": "3",
     "x_netkan_staging": true,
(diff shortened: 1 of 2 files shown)


HebaruSan:
I'm a bit wary of replacing the existing entry; if Ferram does come back, he might receive problem reports from people having difficulties with CKAN version sequencing and have to skip other people's version numbers with his releases. He's already not the biggest fan of CKAN, so it would be nice to avoid antagonizing him in that way.

What would you say about creating a new entry called `FerramAerospaceResearchContinued` (or `Redux` or any of the other suffixes people use for continued mods)? That way we could keep it separate from the original.

dkavolis:
Yeah, no problem. Besides changing the .netkan file name, what else would need to be changed?

HebaruSan:
Just the file name and the `identifier` property.

HebaruSan:
There are two key factors with backports:

1. Select a versioning scheme that "makes room" for the backports properly; for example, you might designate a 1.4.X series of your mod as KSP 1.4-compatible, and a 1.5.X series as KSP 1.5-compatible.
2. Upload one release at a time, and allow 3-5 hours in between uploads to give the CKAN bot time to index each one. The bot only checks the most recently uploaded release each time it runs.

HebaruSan:
Looks OK now, thanks for the submission. Merging...

dkavolis:
Maybe the name should also be changed so that users can see it's a different version.

HebaruSan:
Oops, good point. Added "Continued", should be in the repo momentarily.

PR #6812, opened 2018-11-01
Update M3X netkan

SuicidalInsanity:
Updates M3X dependency from IFS to B9PS and adds M3X example craft.


HebaruSan:
Thanks for the update!

PR #6818, opened 2018-11-05
Add SciFiVE

HebaruSan:
This mod is an EVE texture pack and was requested by a user on its thread.

As per https://forum.kerbalspaceprogram.com/index.php?/topic/151190-14sci-fi-visual-enchancements-v-14-high-performance-alt-to-sve/&do=findComment&comment=3481223

I forgot to add `provides` and `resources`...


HebaruSan:
Dropbox got rid of permalinks to specific files. Sad.

HebaruSan:
Should be ready now, but author hasn't given me an OK yet.

PR #6822, opened 2018-11-06
moved several dependancies to suggest mods

sswelm:


sswelm:
I guess I should reupload with InterstellarHybridRocketry

HebaruSan:
#rebuild

PR #6827, opened 2018-11-08
SpaceDock catch-up 4

HebaruSan:
KSP-SpaceDock/SpaceDock#188 remains unfixed, so here's another round of manually indexed mods that requested CKAN support on SpaceDock. Previous was #6788.


HebaruSan:
#rebuild

PR #6836, opened 2018-11-11
Add KXAPI to NetKAN

Sujimichi:
Adding new mod KXAPI (KerbalX API), a dependency mod for other mods to use to interact with KerbalX.com

KXAPI 1.0.x versions will be compatible with KSP 1.3.x
KXAPI 1.1.x versions will be compatible with KSP 1.4.x & 1.5.x

I've set ksp_version_min to 1.4 (as the 1.1.x series will work with 1.5 as well), but in the x_netkan_override section I've set ksp_version to 1.3 for versions less than 1.1.0.  Not sure if that is correct syntax (using ksp_version_min and ksp_version in the override).

KXAPI 1.0.0 has been released. 1.1.0 will be released once NetKAN has picked up the 1.0.0 release.


HebaruSan:
Looks good to me. Thanks for the submission!

PR #6838, opened 2018-11-12
Realistic Atmospheres

linuxgurugamer:
Requested by OhioBob


HebaruSan:
We already have an entry called that:

https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/RealisticAtmospheres.netkan

HebaruSan:
Confirmed that's the same mod (both by OhioBob). I'll try to de-index this duplicate before it causes too much confusion...

Issue #6856, opened 2018-11-20
ksp-ckan.org domain expired?


HebaruSan:
Problem
-------

**What steps did you take?**
Go to https://ci.ksp-ckan.org/job/NetKAN/7841/

Or http://status.ksp-ckan.org

**What did you expect to happen?**
Page to load with outcome of pull request test scripts or status

**What happened instead?**

```
This site can’t be reached ci.ksp-ckan.org’s server IP address could not be found.
Try:

Checking the connection
Checking the proxy, firewall, and DNS configuration
Running Windows Network Diagnostics
ERR_NAME_NOT_RESOLVED
```

`wget` says 

> wget: unable to resolve host address 'ci.ksp-ckan.org'

@politas @techman83 @Olympic1 @dbent, I think the domain name has expired.

dbent:
Whelp, it's gone now.

```
> whois ksp-ckan.org
Domain Name: KSP-CKAN.ORG
Registry Domain ID: D174543473-LROR
Registrar WHOIS Server: whois.namesilo.com
Registrar URL: www.namesilo.com
Updated Date: 2018-11-21T01:27:42Z
Creation Date: 2014-11-20T15:19:35Z
Registry Expiry Date: 2019-11-20T15:19:35Z
Registrar Registration Expiration Date:
Registrar: Namesilo, LLC
Registrar IANA ID: 1479
Registrar Abuse Contact Email: abuse@namesilo.com
Registrar Abuse Contact Phone: +1.4805240066
Reseller:
Domain Status: clientHold https://icann.org/epp#clientHold
Domain Status: clientTransferProhibited https://icann.org/epp#clientTransferProhibited
Domain Status: autoRenewPeriod https://icann.org/epp#autoRenewPeriod
Registrant Organization: See PrivacyGuardian.org
Registrant State/Province: AZ
Registrant Country: US
Name Server: NS-1526.AWSDNS-62.ORG
Name Server: NS-1563.AWSDNS-03.CO.UK
Name Server: NS-666.AWSDNS-19.NET
Name Server: NS-383.AWSDNS-47.COM
DNSSEC: unsigned
URL of the ICANN Whois Inaccuracy Complaint Form https://www.icann.org/wicf/)
>>> Last update of WHOIS database: 2018-11-21T04:08:01Z <<<

For more information on Whois status codes, please visit https://icann.org/epp

[...]

PR #6858, opened 2018-11-21
Create KerbalChangelog.netkan

BenjaminCronin:
If possible I'd like this mod to be available for KSP v.1.4.x AND 1.5.x. I take it that `ksp_version` is the max KSP version that the mod is available for, is this the case?

### DIFF
--- a/NetKAN/KerbalChangelog.netkan
+++ b/NetKAN/KerbalChangelog.netkan
@@ -3,8 +3,8 @@
     "identifier"   : "KerbalChangelog",
     "$kref"        : "#/ckan/github/BenjaminCronin/KerbalChangelog",
     "license"      : "MIT",
-	"$vref" : "#/ckan/ksp-avc"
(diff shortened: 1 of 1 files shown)


HebaruSan:
No, `ksp_version` allows an exact match only. `ksp_version_min` and `ksp_version_max` can be used to support a range.

HebaruSan:
And a more flexible approach is to put a KSP-AVC .version file in your download, which can then be read into CKAN metadata with the $vref property.

BenjaminCronin:
@HebaruSan I think that should do the trick, but I'm not sure why the build is failing...

HebaruSan:
Well there was a syntax error, but I would expect it to still fail until the download contains a .version file. I don't see one in the latest ZIP.

HebaruSan:
Looks good now. (The usual validation scripts aren't available due to #6856, but I ran netkan.exe locally and it looks fine.)

PR #6859, opened 2018-11-22
Add Kerbal Changelog as a dependency of Oh Scrap

severedsolo:


severedsolo:
I assume this won't effect old versions? Only new ones?

HebaruSan:
Changing the netkan affects the current version plus future versions.
To change older versions, edit them here:
https://github.com/KSP-CKAN/CKAN-meta/tree/master/OhScrap

severedsolo:
That's fine, that's what I want. I don't want old versions to be touched as Changelog isn't compatible with 1.3.x.

PR #6861, opened 2018-11-23
Create Science-Full-Transmit.netkan

mylesmarkevich:
My first KSP mod


mylesmarkevich:
Hello, is there something I'm missing? I was hoping to have my mod added to CKAN...

HebaruSan:
There are some problems on the CKAN side of things that are preventing the [pull request validation script output](https://ci.ksp-ckan.space/job/NetKAN/7849/) from being accessed, see #6856. I'd like to get that resolved so we can confirm that this is installing correctly.

mylesmarkevich:
Is there anything I can do to help? Or is there some other way I should submit my netkan? I see that other commits are still being made.

HebaruSan:
#rebuild

HebaruSan:
1.5.9 isn't a valid KSP version, so this would never show up as compatible with any KSP install.
What would you like to use instead?

mylesmarkevich:
You're right, it looks like I did that incorrectly. I'm now using ksp_version_min and ksp_version_max instead. Sorry about that.

HebaruSan:
Looks good now! Sorry for the delay, merging...

PR #6871, opened 2018-11-30
Create kunedo.netkan

Kunedo83:


HebaruSan:
The identifier should probably be FanRadiator rather than your user name, in case you decide to make another mod at some point, and to match the `name` property. Any objections?

Kunedo83:
> The identifier should probably be FanRadiator rather than your user name, in case you decide to make another mod at some point, and to match the `name` property. Any objections?

ok changing indentifier to fanradiator , thx for warning :)

Kunedo83:
Is it possible to install it in gamedata/kunedo/fanradiator? for next mods install in same folder.

HebaruSan:
If you structure the ZIP file that way, yes. Currently you would have GameData/Kunedo/Parts/fanradiator as it is in the ZIP.

Kunedo83:
okey, 5 minutes im changing

PR #6873, opened 2018-12-02
Update manually maintained mods

HebaruSan:
These mods have out of date metadata.

- CTTStockRebalance was re-uploaded by someone else and marked as compatible with KSP 1.5.1
- Kethane is hosted on taniwha.org and requires manual maintenance of everything
- SDHI-ServiceModuleSystem is on GitHub but has no version file, so it was listed as compatible with the wrong game version

Now they're fixed.


avalanchas:
Is it encouraged to add .version files with a PR to repos that don't have one?

PR #6876, opened 2018-12-05
Update TestFlight to 1.3.1 Release

pap1723:


HebaruSan:
What are you trying to change here? The title doesn't fit as this already has a $kref and $vref so it would already be on the latest release, and it looks like some important things may have gotten broken (such as the TestFlightConfig dependency).

pap1723:
Yep, you are right, the existing one works well. Thanks!

PR #6881, opened 2018-12-09
RP-1.netkan for Release v.1.00 for KSP 1.3.1

pap1723:


pap1723:
The old RP-0.netkan needs to be removed as it is giving the wrong dependencies

pap1723:
I will re-submit

PR #6884, opened 2018-12-10
Fix installation of RSSVE

DrRaphello:
Currently, scatterer configs for stock are not installed and RSSVE depends on them. This makes installation of RSSVE broken.

### DIFF
--- a/NetKAN/RSSVE-HR.netkan
+++ b/NetKAN/RSSVE-HR.netkan
@@ -12,21 +12,19 @@
     },
     "provides": [
         "EnvironmentalVisualEnhancements-Config",
-        "RSSVE-Textures",
-        "Scatterer-config",
-        "Scatterer-sunflare"
+        "RSSVE-Textures"
     ],
     "conflicts": [
         { "name": "EnvironmentalVisualEnhancements-Config" },
-        { "name": "RSSVE-Textures" },
-        { "name": "Scatterer-config" },
-        { "name": "Scatterer-sunflare" }
+        { "name": "RSSVE-Textures" }
     ],
     "depends": [
         { "name": "ModuleManager" },
         { "name": "EnvironmentalVisualEnhancements" },
         { "name": "RealSolarSystem" },
-        { "name": "Scatterer", "version": "2:v0.0320b" }
+        { "name": "Scatterer", "version": "2:v0.0320b" },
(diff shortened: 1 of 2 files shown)


DrRaphello:
BTW, this is done after discussion with Phineas Freak, RSSVE developer

HebaruSan:
Thanks for the correction.
(Checking the scatterer cfg files in this mod, they use the ModuleManager `@` syntax to edit existing values, which confirms that they depend on Scatterer's configs since that wouldn't work otherwise.)
Merging...

DrRaphello:
@HebaruSan thanks for the merge! 
There seems to be a problem with scatterer-config though - when I try to install RSSVE-HR through CKAN, I get error message "Scatterer-config requires a version 2:v0.0320b. However an incompatibile version, 2:v0.0324, is in the resolver"
What does this mean and how can I fix it?

HebaruSan:
Looks like the relationship resolver is pulling in the latest Scatterer-config while processing the Scatterer dependency, [since that relationship is version agnostic](https://github.com/KSP-CKAN/CKAN-meta/blob/1cc3d5099be447e99ddd7465ea29ec2ff37a63c6/Scatterer/Scatterer-2-v0.0320b.ckan#L17-L24), and then later it sees that RSSVE needs a different version and quits. There's even a `TODO` in the code regarding this case:

https://github.com/KSP-CKAN/CKAN/blob/84db54385d0fa7b1b884bf85f6c2df64bfb52d9b/Core/Relationships/RelationshipResolver.cs#L353

Does RSSVE absolutely require Scatterer-config 2:v0.0320b, or could any version of Scatterer-config suffice? Removing that dependency would be the most obvious way to fix this.

DrRaphello:
The config files are different between the versions and I would not install dll from one version and then configs from another.
Would it be better to make the relationship you've linked to version locked?

DrRaphello:
I was able to perform the install by installing the config and sunflare v0.0320b manually, then scatterer 0.0320b and then RSSVE. This is however not user friendly enough, I'm afraid.

DrRaphello:
Well, there is no issue for sunflare for sure. Config though has a significant difference: "useAlternateShaderSQRT = True" is removed and nearClipPlane is 10 times smaller.

No other way?

HebaruSan:
What problems does that cause?

PR #6890, opened 2018-12-20
Switch Chaka to platform neutral version of AnimatedDecouplers

HebaruSan:
This mod is out for KSP 1.4.3, but one of its dependencies only goes up to KSP 1.1:

image

This is because AnimatedDecouplers-x86 is now just AnimatedDecouplers. This pull request makes that change.


HebaruSan:
Apparently it's also bundling an incompatible KJR. Not much we can do about that.

PR #6892, opened 2018-12-22
Fix kerbalism vref

HebaruSan:
Latest download has no GameData folder.


HebaruSan:
Dependencies aren't marked compatible with latest game version yet. Should be fine.

PR #6897, opened 2018-12-25
Delete rt-config-by-ippo.frozen

Ippo343:
That config hasn't been updated in ages
and it's not even compatible with remote tech anymore.


HebaruSan:
.frozen files are ignored by the bot, and I thought deleting files permanently was not the preferred way to retire a mod.

Ippo343:
Sorry about that, I didn't know. What's the preferred way to do it?

PR #6932, opened 2019-01-09
Fix KJR to provide for different game versions

raidernick:
-Update this to act like FAR does for the different versions for game
versions by adding provides.

Before this is merged do any changes need to be made to the meta repo
for this to work? I'd rather not release new versions of the mods again
for no reason just to get this to propogate.


raidernick:
nvm on this i screwed it up

PR #6949, opened 2019-01-13
Create Insight

Kunedo83:


Kunedo83:
ouch,xD thx man!

HebaruSan:
There's already an `InsightLander` indexed. https://spacedock.info/mod/1899
Do you want to add something distinctive to make it easier for users to tell which is which?

PR #6952, opened 2019-01-16
vref for SEP

HebaruSan:
Latest version of this mod supposedly has a .version file (CobaltWolf/Surface-Experiment-Pack#40 got merged, but I haven't checked it myself yet). This PR updates the netkan to use it.


HebaruSan:
The validation error is just about KAS compatibility, but this mod goes back to KSP 1.4. N/A.

PR #6955, opened 2019-01-17
Source Nertea dependencies from independent repositories

HebaruSan:
## Motivation

Up until recently, some of Nertea's mods only existed as bundled pieces of other mods, and their netkans had to choose one download as the source. This created a sort of artificial dependency situation, where (for example) KerbalAtomics might not be installable until NearFutureElectrical updated, or vice versa, even though they're technically independent, because they depend on bundled mods that we source from one or the other.

See KSP-CKAN/CKAN#2404 for an example.

## Changes

Nertea asked on a Discord about opportunities to improve how his mods work with CKAN, and I mentioned this. He has graciously decided to provide standalone releases for these dependencies.

Now the bundled dependencies are sourced from their own independent repositories.

| Mod | Change |
| --- | --- |
| CryoTanks | Change kref from KerbalAtomics SD to CryoTanks GH |
| DeployableEngines | Change kref from KerbalAtomics SD to DeployableEngines GH |
| DynamicBatteryStorage | Change kref from NearFutureElectrical SD to DynamicBatteryStorage GH |
| NearFutureProps | Remove old override that could mess up a future version |

This will mean that if the dependencies are released first, KerbalAtomics and NearFutureElectrical will be installable immediately upon release.

(As of the initial submission of this PR, some of the repos may not yet have any releases. This is all still being finalized, so we will rebuild this PR as needed.)

### DIFF
--- a/NetKAN/CryoTanks.netkan
+++ b/NetKAN/CryoTanks.netkan
@@ -5,7 +5,6 @@
     "name":         "Cryogenic Tanks",
     "$kref":        "#/ckan/github/ChrisAdderley/CryoTanks",
     "$vref":        "#/ckan/ksp-avc/CryoTanks.version",
-    "x_netkan_trust_version_file": true,
     "license":      "CC-BY-NC-SA-4.0",
     "depends": [
         { "name": "ModuleManager"         },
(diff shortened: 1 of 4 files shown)


HebaruSan:
The generated files look fine locally, this validation error isn't a showstopper.

PR #6956, opened 2019-01-18
Fix mk4 version file and deps

HebaruSan:
MarkIVSpaceplaneSystem's .version file got renamed, and some of its dependencies haven't been included in the download for quite a while.


HebaruSan:
#rebuild

PR #6968, opened 2019-01-22
Add Custom Asteroids OPM config.

Starstrider42:
This PR registers a new Custom Asteroids config with CKAN. The proper NetKAN file can be found at https://github.com/Starstrider42/Custom-Asteroids-Extras/blob/master/CustomAsteroids-Pops-OPM-Outer.netkan.

### DIFF
--- a/NetKAN/CustomAsteroids-Pops-OPM-Outer.netkan
+++ b/NetKAN/CustomAsteroids-Pops-OPM-Outer.netkan
@@ -1,0 +1,5 @@
+{
+    "spec_version": "v1.2",
+    "identifier": "CustomAsteroids-Pops-OPM-Outer",
+    "$kref": "#/ckan/netkan/https://raw.githubusercontent.com/Starstrider42/Custom-Asteroids-Extras/master/CustomAsteroids-Pops-OPM-Outer.netkan"
+}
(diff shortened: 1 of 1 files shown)


HebaruSan:
Relevant part of the validation error:

> u'license' is a required property

HebaruSan:
#rebuild

Starstrider42:
Hmm...
> The following inconsistencies were found:
> * CustomAsteroids-Pops-Stock-Outer v1.6.0 conflicts with OuterPlanetsMod 2:2.2.2
> Install canceled. Your files have been returned to their initial state.

I think it must have installed `Pops-OPM-Outer` and `CustomAsteroids` first, then installed `Pops-Stock-Outer` as a recommended mod of `CustomAsteroids`, then tried to install OPM. Is this a bug (it didn't even suggest `Pops-Stock-Outer` when I was testing with the CKAN GUI), or should I remove those recommendations?

HebaruSan:
Oh, that would explain it. I wonder if it would help if the OPM dependency was listed first. Then the recommendation should just get filtered out.

Starstrider42:
That fixed it. Though I wonder why the behavior was different compared to "File -> Install from .ckan".

HebaruSan:
GUI has completely different handling of recommendations than CmdLine, so it can prompt you to choose yes/no on each. One of my longer-term TODO items is to unify all of that.

PR #6973, opened 2019-01-24
Fix SciFiVE install

HebaruSan:
This mod had a folder missing, now it's back and we install it.

https://forum.kerbalspaceprogram.com/index.php?/topic/151190-16sci-fi-visual-enchancements-v-15-high-performance-alt-to-sve/&do=findComment&comment=3527711

ckan compat add 1.4


HebaruSan:
#rebuild

PR #6984, opened 2019-01-31
Create ProgressiveColonizationSystem.netkan

SteveBenz:
This is my first mod submission, so be gentle but thorough please.

### DIFF
--- a/NetKAN/ProgressiveColonizationSystem.netkan
+++ b/NetKAN/ProgressiveColonizationSystem.netkan
@@ -1,0 +1,22 @@
+{
+    "spec_version": 1,
+    "identifier": "ProgressiveColonizationSystem",
+    "$kref": "#/ckan/github/SteveBenz/ProgressiveColonizationSystem",
+    "license": "MIT",
+    "ksp_version_min": "1.5.1",
+    "ksp_version_max": "1.6.99",
+    "version": "0.0.1",
+    "resources": {
+        "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/181852-15x~16x-progressivecolonizationsystem-life-support-colonization-for-more-fun-in-late-game/"
+    },
+    "author": "NermNermNerm",
+    "abstract": "A colonization and production system that encourages building a variety of ship types - space stations, rovers, transports and, of course, landed bases.  It also creates a progression system outside of the science grind.  It includes a very simple life-support system, and so should not be used in mods with life support systems.",
+    "tags":  [ "colonization", "production" ],
+
+    "depends": [
+        { "name": "ModuleManager" },
+        { "name": "StationPartsExpansionRedux" },
+        { "name": "PlanetaryBaseInc" },
+        { "name": "B9_Aerospace_HX" }
+    ]
+}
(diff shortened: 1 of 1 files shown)


HebaruSan:
This is the current error:

> 2109 [1] FATAL CKAN.NetKAN.Program (null) - /tmp/CdFileMgr/eef002df-6ad6-4f.tmp is not a valid ZIP file: Error in step EntryHeader for GameData/ProgressiveColonizationSystem/ModuleManager.cfg: Exception during test - 'Central header and local header file name mismatch'

Which program did you use to generate the ZIP file?

SteveBenz:
I used .Net's System.IO.Compression.ZipFile.CreateFromDirectory from a Powershell Script.

SteveBenz:
I didn't notice that second commit on my chain before...  Dunno what's up with that, I did as the instructions suggested and made the pull request through the editor.  Perhaps I should man up and fork the repo myself.

HebaruSan:
Also the `PlanetaryBaseInc` dependency doesn't exist, so it will never be possible to install this. Is that supposed to be `KerbalPlanetaryBaseSystems`?

SteveBenz:
You're probably right about the dependency - I'll update the PR this evening.  Maybe I can make the world a better place by creating a check in NetKan for that.  But meanwhile, here's an interesting read:

https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/mitigation-ziparchiveentry-fullname-path-separator

My PowerShell script probably ends up using .Net 4.5

SteveBenz:
I've got my doubts about that failing check.  I used 7zip, which is the top zip program out there (at least in Google's estimation) and it has a linux download.  No joy.  I unpacked the 7zip'ed file on my Raspberry pi; no problems there.  I packaged it up again on the PI and lo, I have a thing that works...  But not really a sensible solution going forward.  What program do folks typically use to Zip up mods for CKAN?

HebaruSan:
Impossible to say for sure because we don't ask that question of people when their files are valid, which is almost all the time, but I'm sure there's a lot of variety in 1600+ working mods. My guess is there's a lot of WinZip and Windows Explorer's "Send to compressed folder", and on Unix I know the `zip` utility does fine.

Can I see your 7z file?

SteveBenz:
Releases 0.0.1-3 were the ones that failed the check:

https://github.com/SteveBenz/ProgressiveColonizationSystem/releases

HebaruSan:
So 0.0.2 and 0.0.3 were created with 7zip?

SteveBenz:
I'm definitely still not happy with my pipeline, but I'm good with shipping 0.0.4.

I've cloned the NetKan sources and can run it locally and see the same thing as the website does, so I can iterate locally.  What I'd love to be able to do is create a pull request that makes the messaging coming out of NetKan provide a clear explanation of what the problem is and prescriptive guidance on how to fix it.  When I get some time, I'll test WinZip.  Given the name, you'd think it'd be least-likely to work, but you're right that there are so many mods that don't have this problem, there must be something in common use that does it correctly.

HebaruSan:
In this case Netkan is just passing along an error that it gets from the SharpZipLib package's validation checks (see KSP-CKAN/CKAN#2287 and KSP-CKAN/CKAN#2288), so that pull request would probably have to be made there:

- https://github.com/icsharpcode/SharpZipLib

But I don't think it has an active maintainer at the moment, so you might not get a quick reaction.

Merging...

SteveBenz:
FWIW, my release build now uses the .Net 4.7 Zip classes to build up the ZIP file and NetKan is happy with it.  It's a shame I couldn't find a way to convince Powershell to use the later .Net version and had to build up a C# application just to do it.  But whatever.  It's automated and it works.

PR #6986, opened 2019-01-31
Update ProbesBeforeCrew.netkan

Silven88:
ckan compat add 1.5


Silven88:
My mod affects this ScienceParamModifier dependency by changing values of a file within that mod. It cannot be affected with ModuleManager, so how will CKAN ensure my edits are included?

HebaruSan:
(Continued via forum PM; these changes can and should be done via ModuleManager.)

Silven88:
Thanks for your help HebaruSan!

PR #6992, opened 2019-02-03
MechJeb2 compatibility and staging

HebaruSan:
MechJeb's game version is hard coded in the netkan, which has caused the latest version to be out of date. Now it's updated from 1.5 to 1.6.

#4794 made the point that staging might be a good way to avoid this, since it would allow us to check the game versions before they go live. So now staging is enabled.

### DIFF
--- a/NetKAN/MechJeb2.netkan
+++ b/NetKAN/MechJeb2.netkan
@@ -14,7 +14,8 @@
         "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/154834-*",
         "repository": "https://github.com/MuMech/MechJeb2"
     },
-    "ksp_version": "1.5",
+    "ksp_version": "1.6",
+    "x_netkan_staging": true,
     "install": [
         {
             "find": "MechJeb2",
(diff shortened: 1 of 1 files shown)


HebaruSan:
@techman83 , I think this broke the bot. The web hook never pushed a commit for it, and the status page says the last update was MastodonPatch 2 hours ago, which is immediately before MechJeb2 alphabetically.

techman83:
@HebaruSan , I don't see anything particularly ominous. But I'll do a debug run and see what happens.

HebaruSan:
There's a 4 hour gap here:

image

I think the bot made it up to MastodonPatch and then either froze or aborted at MechJeb2 on the last pass.

techman83:
I forgot part of my sentence, particularly ominous in the logs :slightly_smiling_face: - Debug will reveal all. Probably something with the staged branch.

techman83:
Well debug didn't reveal anything useful. So that likely means it's something Git related. It hadn't created a branch, which is how the staging works. I'll wait for the next run to complete.

PR #6997, opened 2019-02-05
Fix Moar Filter Extension Configs

HebaruSan:
This module's folder path changed, and the license on SD says "Creative Commons Zero v1.0 Universal" but is "unknown" in the netkan.


linuxgurugamer:
Thanks, I was going to do that this evening.
however, it now has a .version file, so it needs to be updated with a vref. 
I'll check this evening to see if it's been upated

PR #6999, opened 2019-02-06
Fix folder name

leatherneck6017:
Added the "t" to Extension


HebaruSan:
That's what's in the ZIP:

image

Closing because this would break the netkan.

leatherneck6017:
I updated the zip on Spacedock this morning.

HebaruSan:
Ahh OK, good to know.

leatherneck6017:
Thank you.

PR #7005, opened 2019-02-09
Obey UniversalStorage2's version file

### LINKED ISSUE #7004: Universal Storage II compatibility data
According to the [forum thread](https://forum.kerbalspaceprogram.com/index.php?/topic/177385-universal-storage-ii/), the current version (1.6.0.9) is listed as being compatible with KSP 1.4.x upwards, whereas CKAN only has 1.6.0 compatibility noted. The same might also be true for the intermediate 1.5.x.y mod versions.

HebaruSan:
Apparently this mod has a version file now, but its netkan is only set up to use the version from SpaceDock. Now it has a vref.

Fixes #7004.

### DIFF
--- a/NetKAN/UniversalStorage2.netkan
+++ b/NetKAN/UniversalStorage2.netkan
@@ -2,6 +2,7 @@
     "spec_version": "v1.4",
     "identifier":   "UniversalStorage2",
     "$kref":        "#/ckan/spacedock/1933",
+    "$vref":        "#/ckan/ksp-avc",
     "license":      "restricted",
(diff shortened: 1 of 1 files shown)


HebaruSan:
Note that the version file has the wrong module version (1.5.1.9 instead of 1.6.0.9) but right game versions, reported at https://bitbucket.org/uvs2/universalstorage2/issues/18/bug-out-of-date-version-field-in-ksp-avc. This is OK for the current version and hopefully future versions will be fixed.

PR #7018, opened 2019-02-22
Add Flight Tracker, Make EarnYourStripes a dependency.

severedsolo:


HebaruSan:
#rebuild

HebaruSan:
#rebuild

HebaruSan:
The validation script isn't running (no clue why), so I tried this locally. These mods are for different game versions, which might cause hiccups when users try to install:

FlightTracker:

```json
    "ksp_version_min": "1.6.0",
    "ksp_version_max": "1.6.99",
```

EarnYourStripes:

```json
    "ksp_version_min": "1.5.0",
    "ksp_version_max": "1.5.99",
```

Users would have to mark both 1.5 and 1.6 as compatible to install. Is that OK?

https://github.com/KSP-CKAN/CKAN/wiki/User-guide#choosing-compatible-game-versions

severedsolo:
Is it possible to specify a version of EarnYourStripes that FlightTracker must be installed for and leave previous ones alone - that would be the ideal. The current "in the wild" release works standalone, but 2.0 will need Flight Tracker.

HebaruSan:
Yes, we can do that pretty easily by merging this pull request after 2.0 is out.

severedsolo:
Right, leave it with me. I'll release in a minute and give you a ping once it's been done.

severedsolo:
@HebaruSan - Earn Your Stripes 2.0 is now out. I assume you need to wait for the bot to catch up.

HebaruSan:
In this case it's better to merge as quickly as possible. A script runs at merge to (re-)index affected mods based on the latest netkan.

PR #7021, opened 2019-02-23
[x] Science! Continued

Flupster:
Taking over for Z-Key Aerospace due to inactivity

### DIFF
--- a/NetKAN/xScience.netkan
+++ b/NetKAN/xScience.netkan
@@ -1,26 +1,17 @@
 {
     "spec_version": 1,
     "identifier": "xScience",
-    "$kref": "#/ckan/spacedock/234",
+    "$kref": "#/ckan/github/Flupster/KSP-X-Science",
     "license": "CC-BY-NC-SA-4.0",
     "author": ["Z-Key Aerospace", "Brodrick"],
-    "abstract": "The Science Report and Checklist for KSP.  [x] Science! keeps track of the science experiments you have completed and recovered to Kerbal Space Center. It also tracks science you have gathered which is held on vehicles and kerbals.  No matter what game scene you are in, what you current vehicle is, [x] Science! always reports on all science held anywhere.",
-    "resources": {
+    "abstract": "The Science Report and Checklist for KSP.  [x] Science! keeps track of the science experiments you have completed and recovered to Kerbal Space Center. It also tracks science you have gathered which is $    "resources": {
         "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/107661-ksp-105-x-science",
-        "repository": "https://github.com/thewebbooth/KSP-X-Science"
+        "repository": "https://github.com/Flupster/KSP-X-Science"
     },
     "install": [
         {
-            "file": "GameData/[x] Science!",
+            "file": "[x] Science!",
             "install_to": "GameData"
-        }
-    ],
-    "x_netkan_override": [
-        {
-            "version": "4.14",
-            "override": {
-                "download": "https://themoose.co.uk/ksp/x-science-v4.14.zip"
-            }
         }
     ],
     "x_netkan_force_v": true
--- a/NetKAN/xScience.netkan
+++ b/NetKAN/xScience.netkan
@@ -4,7 +4,8 @@
     "$kref": "#/ckan/github/Flupster/KSP-X-Science",
     "license": "CC-BY-NC-SA-4.0",
     "author": ["Z-Key Aerospace", "Brodrick"],
-    "abstract": "The Science Report and Checklist for KSP.  [x] Science! keeps track of the science experiments you have completed and recovered to Kerbal Space Center. It also tracks science you have gathered which is $    "resources": {
+    "abstract": "The Science Report and Checklist for KSP.  [x] Science! keeps track of the science experiments you have completed and recovered to Kerbal Space Center. It also tracks science you have gathered which is held on vehicles and kerbals.  No matter what game scene you are in, what you current vehicle is, [x] Science! always reports on all science held anywhere.",
+    "resources": {
         "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/107661-ksp-105-x-science",
         "repository": "https://github.com/Flupster/KSP-X-Science"
     },
--- a/NetKAN/xScienceContinued.netkan
+++ b/NetKAN/xScienceContinued.netkan
@@ -7,7 +7,7 @@
     "author": ["Z-Key Aerospace", "Brodrick", "Flupster"],
     "abstract": "The Science Report and Checklist for KSP.  [x] Science! keeps track of the science experiments you have completed and recovered to Kerbal Space Center. It also tracks science you have gathered which is held on vehicles and kerbals.  No matter what game scene you are in, what you current vehicle is, [x] Science! always reports on all science held anywhere.",
     "resources": {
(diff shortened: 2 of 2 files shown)


NetKAN/xScience.netkan
```diff
@@ -1,27 +1,18 @@
 {
     "spec_version": 1,
     "identifier": "xScience",
-    "$kref": "#/ckan/spacedock/234",
+    "$kref": "#/ckan/github/Flupster/KSP-X-Science",
     "license": "CC-BY-NC-SA-4.0",
     "author": ["Z-Key Aerospace", "Brodrick"],
-    "abstract": "The Science Report and Checklist for KSP.  [x] Science! keeps track of the science experiments you have completed and recovered to Kerbal Space Center. It also tracks science you have gathered which is held on vehicles and kerbals.  No matter what game scene you are in, what you current vehicle is, [x] Science! always reports on all science held anywhere.",
-    "resources": {
+    "abstract": "The Science Report and Checklist for KSP.  [x] Science! keeps track of the science experiments you have completed and recovered to Kerbal Space Center. It also tracks science you have gathered which is $    "resources": {
```
DasSkelett: You deleted some text and a new line here by accident

Flupster: I'm stupid! fixed, thanks for noticing

DasSkelett:
Do you want to add yourself as an author?

Also, the name will differ to the original -> `"name": "KSP X Science",` (updated) vs `"name": "[x] Science!",` (original).

Also, it will conflict with the odd usage of the 'v' in the version strings of the previous versions:
grafik
I think this can be resolved by removing it from the old .ckans?

Flupster:
> Do you want to add yourself as an author?
> 
> Also, the name will differ to the original -> `"name": "KSP X Science",` (updated) vs `"name": "[x] Science!",` (original).
> Edit: This is because CKAN pulls the name from the GitHub repo name.
> If you add a explicit `"name": "[x] Science!",` to the netkan, it should override this.
> One more edit: I think this also is the reason, why CKAN can't read the .version file.
> 
> Third, it will conflict with the odd usage of the 'v' in the version strings of the previous versions, CKAN thinks the older versions are newer ones (if you have 1.4 marked ad compatible):
> grafik
> I think this can be resolved by removing it from the old .ckans?

I decided to just create from scratch, this means if the original developer comes back he can merge my PR and [x] Science! can live on, for now [x] Science! Continued (this) will just be here so people using ckan can grab the mod magically :)

linuxgurugamer:
it will conflict with the odd usage of the 'v' in the version strings of the previous versions, CKAN thinks the older versions are newer ones (if you have 1.4 marked ad compatible):
> grafik
> I think this can be resolved by removing it from the old .ckans?

Would be better to do an epoch rather than removing the old ones

DasSkelett:
@Flupster now decided to make it a separate mod, so this issue is resolved.

Also I think it now would make sense to create a new forum thread, @Flupster, so you can communicate with the users outside the comment section of the original thread.

HebaruSan:
#rebuild

NetKAN/xScienceContinued.netkan
```diff
@@ -0,0 +1,20 @@
...
+    "spec_version": 1,
+    "identifier": "xScienceContinued",
+    "name": "[x] Science! Continued",
+    "$kref": "#/ckan/github/Flupster/KSP-X-Science",
+    "license": "CC-BY-NC-SA-4.0",
+    "author": ["Z-Key Aerospace", "Brodrick", "Flupster"],
+    "abstract": "The Science Report and Checklist for KSP.  [x] Science! keeps track of the science experiments you have completed and recovered to Kerbal Space Center. It also tracks science you have gathered which is held on vehicles and kerbals.  No matter what game scene you are in, what you current vehicle is, [x] Science! always reports on all science held anywhere.",
+    "resources": {
+        "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/107661-ksp-105-x-science",
+        "repository": "https://github.com/Flupster/KSP-X-Science"
```
Tekaoh: Should `repository:` be removed in order to rely on the conversion from $kref instead?
```suggestion
```

NetKAN/xScienceContinued.netkan
```diff
@@ -0,0 +1,20 @@
+{
+    "spec_version": 1,
+    "identifier": "xScienceContinued",
+    "name": "[x] Science! Continued",
+    "$kref": "#/ckan/github/Flupster/KSP-X-Science",
+    "license": "CC-BY-NC-SA-4.0",
+    "author": ["Z-Key Aerospace", "Brodrick", "Flupster"],
+    "abstract": "The Science Report and Checklist for KSP.  [x] Science! keeps track of the science experiments you have completed and recovered to Kerbal Space Center. It also tracks science you have gathered which is held on vehicles and kerbals.  No matter what game scene you are in, what you current vehicle is, [x] Science! always reports on all science held anywhere.",
+    "resources": {
+        "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/107661-ksp-105-x-science",
```
Tekaoh: Suggestion: Wildcard URL.
Better suggestion: @Flupster Make a new forum thread for your continuation and link it instead.
```suggestion
        "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/107661-*",
```

HebaruSan:
Two mods suggest or recommend xScience. Should this mod have `"provides": [ "xScience" ],` to pull it in in those cases?

Tekaoh:
> Two mods suggest or recommend xScience. Should this mod have `"provides": [ "xScience" ],` to pull it in in those cases?

One of those mods is mine. I'll likely just make a PR to suggest Flupster's continuation instead anyways.

Tekaoh:
> Two mods suggest or recommend xScience. Should this mod have `"provides": [ "xScience" ],` to pull it in in those cases?

Alternately, if this PR is able to be merged, I can take a stab at adding `replaced_by:` to the old xScience for the sake of testing that out.

DasSkelett:
Does a `replaced_by` relationship imply a `conflicts` (@HebaruSan / @politas) ?

politas:
@DasSkelett No, any conflicts or provides would need to specified separately.

Flupster:
Anything needed to be done by me before this gets merged?

HebaruSan:
> Anything needed to be done by me before this gets merged?

Is there going to be a new forum thread? That was suggested above and sounds like a good idea.

Flupster:
> > Anything needed to be done by me before this gets merged?
> 
> Is there going to be a new forum thread? That was suggested above and sounds like a good idea.

I don't see a point, there are no additional features or changes other than recompiling for newer versions of KSP, I'll continue to monitor the older thread but seeing as there are no additional features there's no need to make a new thread.

linuxgurugamer:
@Flupster The problem with not creating a new thread is that you don't have control of the OP, which can cause issues down the road (I'm running into this with the OSE Workshop)

Tekaoh:
@Flupster One example is that you would be able to update the version number in the thread title. It would also just make it clear that you're continuing it as another mod and no longer just updating the same mod.

Flupster:
I've created a thread and changed the homepage in the netkan meta.

Tekaoh:
Hm, I wonder if a new thread takes a few minutes before being available to anyone else? I was on my way over there to give you kudos but got a 404...

Flupster:
I'm still below 5 posts, it's in the process of being approved 😅

Tekaoh:
Oh hahaha that's hilarious

PR #7024, opened 2019-02-25
[xScience] - Point to a different repository for now

Tekaoh:
This mod hasn't been updated in it's present location for many months now. Another user has been updating the mod to work with newer versions of KSP. I propose that CKAN should make his updates available for now, at least until such time as the maintainer returns or someone else wants to take over.

### DIFF
--- a/NetKAN/xScience.netkan
+++ b/NetKAN/xScience.netkan
@@ -1,26 +1,19 @@
 {
     "spec_version": 1,
     "identifier": "xScience",
-    "$kref": "#/ckan/spacedock/234",
+    "$kref": "#/ckan/github/Flupster/KSP-X-Science",
+    "ksp_version_min": "1.4.0",
+    "ksp_version_max": "1.6.9",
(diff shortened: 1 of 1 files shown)


HebaruSan:
Redundant with #7021. Closing because this one hard codes the game versions.

Tekaoh:
@HebaruSan Sorry about that, I just noticed Flupster already created a pull request the other day. I have a question though, if you don't mind. Your comment mentioned that the pull request is closed because of the hard coded game versions. However, I've made previous pull requests with hard coded game versions that you accepted. For future reference, what's the policy on that? I was under the impression that that was acceptable practice.

Thanks!

HebaruSan:
It's not a policy, just a tie-breaker when there are two options for doing something. The other PR is getting the versions from the .version file, which is much better for maintenance.

Tekaoh:
Oh ok, thanks for the info. Sorry about the redundant PR. I should have double checked before submitting, but it had already been so many months with no action that I didn't think to, hah. @HebaruSan

Issue #7027, opened 2019-02-26
Testing required


politas:
@KSP-CKAN/wranglers, there's a new pre-release available that includes the long-delayed "replaced_by" addition to the spec. (Huge thanks to @HebaruSan for finishing up the code that had me dithering about implementation specifics!)

We really need some of the old entries in CKAN-Meta modified to use the new field so we can test out the new client.

HebaruSan:
For reference, this was the original feature request: KSP-CKAN/CKAN#904

One possible test case is xScience &rarr; xScienceContinued in #7021.

HebaruSan:
<details><summary>Expand for the (very long) list of identifiers in NetKAN that might be applicable...</summary>

- AirPark → AirParkContinued
- AnalogControl → AnalogControlContinued
- AnyRes → AnyResContinued
- AudioMuffler → AudioMufflerRedux
- B9-PWings → B9-PWings-Fork
- BahaEPL → BahaEPLRedrilled
- BAM → BAMCont
- BargainRockets → BargainRocketsCont
- BDArmory → BDArmoryContinued
- BetterScienceLabs → BetterScienceLabsContinued
- BetterTimeWarp → BetterTimeWarpCont
- BOSS → BOSSContinued
- BurnTogether → BurnTogetherContinued
- CactEyeTelescopes → CactEyeTelescopesContinued
- CareerManager → CareerManagerContinued
- ChampagneBottle → ChampagneBottleRedux
- CivilianPopulations → CivilianPopulationsRevived
- CollisionFX → CollisionFXUpdated
- ContractFilter → ContractFilterContinued
- CrewQueue → CrewQueueTwo
- DangerAlerts → DangerAlertsContinued
- DangIt → DangItContinued
- EditorExtensions → EditorExtensionsRedux
- EngineIgnitor → EngineIgnitorReignited
- EVAEnhancements → EVAEnhancementsContinued
- EvaFuel → EvaFuelCont
- EVAHandrailsPack → EVAHandrailsPackContinued
- ExceptionDetector → ExceptionDetectorUpdated
- ExtensiveEngRpt → ExtensiveEngRptContinued
- FarFromKerbin → FarFromKerbinCont
- FASALaunchClamps → FASALaunchClampsContinued
- FerramAerospaceResearch → FerramAerospaceResearchContinued
- FlagRotate → FlagRotateCont
- FMRS → FMRSContinued
- ForScience → ForScienceContinued
[...]

PR #7048, opened 2019-03-06
Add replaced_by to OSE Workshop branches

HebaruSan:
OSEWorkShopCont goes up to version 1.2.5 on KSP 1.4.2.
OSEWorkShopReworked is a new fork with versions up to 1.2.9 on KSP 1.6.1.

This pull request marks the former as `replaced_by` the latter.
Part of #7027.

ckan compat add 1.4 1.5 1.6

### DIFF
--- a/NetKAN/OSEWorkShopCont.netkan
+++ b/NetKAN/OSEWorkShopCont.netkan
@@ -1,5 +1,5 @@
 {
-  "spec_version": "v1.4",
+  "spec_version": "v1.26",
   "install": [
     {
       "install_to": "GameData",
--- a/NetKAN/OSEWorkShopCont.netkan
+++ b/NetKAN/OSEWorkShopCont.netkan
@@ -16,5 +16,6 @@
   "$kref": "#/ckan/spacedock/1448",
   "x_via": "Automated Linuxgurugamer CKAN script",
   "identifier": "OSEWorkShopCont",
-  "license": "CC-BY-NC-SA-4.0"
+  "license": "CC-BY-NC-SA-4.0",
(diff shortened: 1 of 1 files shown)


HebaruSan:
```
5847 [1] FATAL CKAN.NetKAN.Program (null) - Could not cast or convert from System.String to CKAN.ModuleRelationshipDescriptor.
```

I forgot it has to be a full `{ "name": ... }` clause rather than just an identifier...

HebaruSan:
Worked that time, but it shouldn't because the `spec_version` isn't updated to 1.26.
Fixing in KSP-CKAN/xKAN-meta_testing#46.

HebaruSan:
#rebuild

techman83:
We both triggered a rebuild :laughing:

HebaruSan:
That looks better, but it didn't show up as replaceable in the installed module list.
I think we can make that happen...

HebaruSan:
#rebuild

linuxgurugamer:
You ARE aware that the Reworked version only starts on 1.6.1?  I'm not "replacing" it on any earlier version of KSP, so why put the "replaced" in?

DasSkelett:
I think you got the `replaced_by` property wrong. If a mod is no longer maintained and another author takes over the development (i.e. you in most cases), we stick a `replaced_by` on the old mod, so a user who has installed the old mod can select "replace" and gets the new mod.
Kinda like an update, but with replacement.

linuxgurugamer:
Then I think you have a LOT of "replaced_by"'s to add :-)
Thanks for the explanation

PR #7050, opened 2019-03-06
Adds indexing for Restock and Restock+

ChrisAdderley:
Two new mods, pulling their metadata from our repo.


HebaruSan:
Do you want to update the homepage before we merge this? That way the early users wouldn't see "tbd".

ChrisAdderley:
Ok, changed this a bit actually. Want to be able to tweak the netkans without pushing a release to our production, so this targets master for collecting them

HebaruSan:

PR #7056, opened 2019-03-09
update CommunityPartsTitles 0.4.0

yalov:


yalov:
waiting confirmation about the KAX/L

yalov:
how to suggest a mod, starting only with some version, so until it's getting there, the mod isn't even suggested?

`{ "name": "Ringworld", "version": ">3.0.0" }`

HebaruSan:
Try what `IR-Model-Rework-Core.netkan` does:

```json
{ "name": "InfernalRobotics", "version_min": "0.19.4" }
```

yalov:
ready to go

HebaruSan:
Sorry, I got that wrong, it's actually `min_version`, and the metadata I quoted needs to be cleaned up:

https://github.com/KSP-CKAN/CKAN/blob/6b4cd143996f029b94162ec89a2e0cb86f9a7c3f/Core/Types/CkanModule.cs#L37-L42

PR #7062, opened 2019-03-11
DMagic Orbital Science recommendation changed from US1 to US2

DMagic1:
Some weird SCANsat netkan change seems to have tagged along with this update for some reason, I'm not sure how to get rid of it. It seems to be white space changes and a block for OverDrive, the UI replacement mod. I seem to remember that mod causing issues with SCANsat at some point, but I'm not sure if that is still the case.


DMagic1:
Ok, ignore all that stuff about the SCANsat netkan file, I see that the conflict with Overdrive was fixed a long time ago, so I took it back out.

PR #7064, opened 2019-03-12
Populate known staging reasons

HebaruSan:
As of KSP-CKAN/NetKAN-bot#88 we now have the option to set a property `x_netkan_staging_reason` in a netkan to tell metadata maintainers why a module has staging enabled. This property is used as the body of a staged pull request when present.

In several cases, the `comment` property had this information; these are now updated to use `x_netkan_staging_reason` instead.

In other cases the reason for staging could be figured out by checking the pull requests where it was enabled. These now have `x_netkan_staging_reason` added.

RoverWheelSounds now has its game versions updated to match its forum thread.

PlanetWiki-OPM had its reason for staging removed in #4988, so it is now no longer staged.

ckan compat add 1.1

### DIFF
--- a/NetKAN/ContractConfigurator-HistoricMissions.netkan
+++ b/NetKAN/ContractConfigurator-HistoricMissions.netkan
@@ -1,14 +1,14 @@
 {
     "spec_version": "v1.4",
-    "license": "CC-BY-NC-SA-4.0",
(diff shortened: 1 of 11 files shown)


HebaruSan:
#rebuild

PR #7067, opened 2019-03-14
Migrate Pood's Skyboxes to use "any_of" dependency feature

Poodmund:


HebaruSan:
The validation scripts need some updating. Please stand by.

techman83:
#rebuild

HebaruSan:
#rebuild

HebaruSan:
@Poodmund, this will hide the latest version of these mods for anyone not using CKAN 1.26. Since there's only a pre-release now, shall we plan to merge this after the full release?

Poodmund:
@HebaruSan 	yes that is fine to wait for 1.26 before the merge.

Poodmund:
Presuming that 1.25.4 is effectively the above discussed 1.26 in this case, thanks for all the work you guys have done to get this all to function properly. Its greatly appreciated.

DasSkelett:
Can you elaborate on the first part? I don't quite understand.
But thanks for the second part :D

PR #7080, opened 2019-03-19
Update Near Future Propulsion netkans

ChrisAdderley:
Few items in this pull request
1. Move netkan files into the mod's own repository for cleaner control
2. Remove an invalid/old mod - this mod shouldn't be installable separately as it's already part of the base install (not sure on procedure here, renaming it to `.frozen`?
3. Add a new separately installable package

The linked netkans (appreciate a quick once over if possible):

- https://github.com/ChrisAdderley/NearFuturePropulsion/blob/master/CKAN/NearFuturePropulsion-XenonHETs.netkan

- https://github.com/ChrisAdderley/NearFuturePropulsion/blob/master/CKAN/NearFuturePropulsion.netkan


HebaruSan:
The validation script doesn't like the new identifier:

```
#   Failed test 'NearFuturePropulsion-XenonHETs.netkan identifier should match filename'
#   at t/metadata.t line 23.
#          got: 'NearFuturePropulsionXenonHETs'
#     expected: 'NearFuturePropulsion-XenonHETs'
# Looks like you failed 1 test of 13420.
```

DasSkelett:
```
Running NetKAN for NetKAN/NearFuturePropulsion-XenonHETs.netkan 13227 [1] FATAL CKAN.NetKAN.Program (null) - Too many .version files located: GameData/CommunityResourcePack/CRP.version, GameData/B9PartSwitch/B9PartSwitch.version, GameData/NearFuturePropulsion/Versioning/NearFuturePropulsion.version
```

HebaruSan:
Right, I believe this would tell it which AVC file to use:

```json
    "$vref": "#/ckan/ksp-avc/NearFuturePropulsion.version",
```

ChrisAdderley:
Interesting - I'm curious as to why it worked before then. That package is the same as before and the `vref` line was unchanged.

HebaruSan:
That's the newly created one.

HebaruSan:
OK, I think the exact difference is due to this two-pass logic; first it checks to see if there are any *installable* AVC files (meaning that they match the "install" directives of the metadata), and if not, it checks for any at all:

https://github.com/KSP-CKAN/CKAN/blob/4f3b3d7cdf9c65afd17283fd0631bd1749aa3f3a/Netkan/Services/ModuleService.cs#L92-L102

So for the main NFP, it finds exactly one AVC file in that first pass and uses it. For XenonHET, it finds none in the first pass, then multiple in the second pass, and errors out.

ChrisAdderley:
Ok, so in summary, https://github.com/ChrisAdderley/NearFuturePropulsion/blob/master/CKAN/NearFuturePropulsion-XenonHETs.netkan#L8

Must be changed from 
` "$vref": "#/ckan/ksp-avc/",`
to
` "$vref": "#/ckan/ksp-avc/NearFuturePropulsion.version",`

HebaruSan:
Yes, that should do it.

ChrisAdderley:
And it passed! Great!

HebaruSan:
Yup, seems to be in order. Thanks for the fixes, merging...

PR #7083, opened 2019-03-20
NF Spacecraft self-hosted NetKANs

ChrisAdderley:
Similar to latest PR (these will roll out for all mods slowly, moves

- Moves netkans to the mod's own repository for closer control
- Creates new entry for a previously unlisted Extras package
- Updates some dependencies and metadata


HebaruSan:
LGTM, merging...

PR #7089, opened 2019-03-22
Add 'OPM-Reconfig' Custom Asteroids config

Starstrider42:
This PR adds the OPM-Reconfig.cfg asteroid pack by @HSJasperism as a separate download for Custom Asteroids. It also changes the name of the existing OPM-Outer to clarify the difference between the two.

### DIFF
--- a/NetKAN/CustomAsteroids-Pops-OPM-Reconfig.netkan
+++ b/NetKAN/CustomAsteroids-Pops-OPM-Reconfig.netkan
@@ -1,8 +1,8 @@
 {
     "spec_version": "v1.4",
-    "identifier": "CustomAsteroids-Pops-OPM-Outer",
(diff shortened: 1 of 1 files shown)


HebaruSan:
Regarding the `max_version` values, do you actually want the dependencies to be considered incompatible if OPM releases a 2:3.0 version or if CustomAsteroids releases a v2.0 version? If not, it would be safer to leave out those properties.

Starstrider42:
Well, yes, since a major release generally breaks things (and I can vouch that Custom Asteroids 2.0, if it ever exists, will break existing configs). Allowing compatibility only with versions 2 and 1, respectively, strikes me as much safer than assuming an arbitrary future version will work until proven otherwise.

HebaruSan:
OK, that makes sense to me, and everything else looks fine. Thanks for the update! Merging...

PR #7095, opened 2019-03-28
Adds KerbalChangelog to reccomends of Kerbalism

steamp0rt:


DasSkelett:
You added it to the dependencies, not the recommends.

steamp0rt:
Fixed

PR #7098, opened 2019-03-30
Spacedock catch-up 13

DasSkelett:
KSP-SpaceDock/SpaceDock#188 is not fixed.
These are the mods added since #7066.

@HebaruSan is on holiday, so I tried my best.


HebaruSan:
@DasSkelett, RedOnion has a capitalized `Install` property instead of `install`. Any reason?

DasSkelett:
Well....... no 🤷‍♂️

PR #7100, opened 2019-03-31
Feature/neos

mwerle:
Mark old mods by Nehemiah as "replaced-by" NehemiahEngineeringOrbitalScience and add some additional metadata.
(Also replaces all space-indentation with tab-indentation to reduce character count in files)


mwerle:
Part of #7027

PR #7111, opened 2019-04-08
Revert "Add B9PartSwitch to Kerbalism deps"

steamp0rt:
Reverts KSP-CKAN/NetKAN#7109

### DIFF
--- a/NetKAN/Kerbalism.netkan
+++ b/NetKAN/Kerbalism.netkan
@@ -14,9 +14,6 @@
 		},
 		{
 			"name": "CommunityResourcePack"
-		},
-		{
-			"name": "B9PartSwitch"
 		}
 	],
 	"recommends": [{


DasSkelett:
Context: https://github.com/KSP-CKAN/CKAN-meta/pull/1435 and 

As soon as I merge this, the bot reruns and generates the 2.2.0 .ckan.
If I delete it afterwards, it will generate it in 4 hours again.

Can you delete the version from spacedock? This prevents the bot from pulling it.

steamp0rt:
Is there even a delete button on spacedock?

DasSkelett:
I think another option is to freeze the 2.2.0.ckan. Not sure if this works though.

DasSkelett:
Huh, no idea.

DasSkelett:
I'll try to freeze it.

DasSkelett:
image
Worked!
So what about the dependency? For what version(s) do you want it added or removed?

steamp0rt:
all of em. I added em too early -- SirMortimer thought it'd be needed due to a bug in the Configure module, but he found a different way to do things

PR #7118, opened 2019-04-12
AlphabeticalKerbals-0.3 update compatibility to 1.7

cecilkorik:
Smoke tested addon with 1.7 and all seems well. Updated netkan file with new version info.


HebaruSan:
Thanks for the update!

PR #7119, opened 2019-04-13
Add support for taniwha's "Custom Bulkhead Profiles" mod.

mwerle:


mwerle:
Permission given by mod author - https://forum.kerbalspaceprogram.com/index.php?/topic/181645-161-custom-bulkhead-profiles-010/&do=findComment&comment=3581993

HebaruSan:
Yeah, it's the other way around, the Netkan's values take precedence. And if we hard code the version, then when the next release came out we'd generate incorrect metadata using the old values.

How likely is it that a future release will require a game version greater than 1.6? We might want to enable staging so someone can double check the compatibility before indexing new releases.

mwerle:
>Yeah, it's the other way around, the Netkan's values take precedence. And if we hard code the version, then when the next release came out we'd generate incorrect metadata using the old values.

Ah.. my memory is playing tricks on me. On re-reading the Netkan specs, they are perfectly clear : " fields will be auto-filled **if not already present**". (my emphasis). For some reason I believed the remote information would override the Netkan-supplied fields - and I only re-read the pertinent bits of the spec this morning while writing the .netkan! Nevermind.

> How likely is it that a future release will require a game version greater than 1.6? We might want to enable staging so someone can double check the compatibility before indexing new releases.

It's not my mod and I haven't done an in-depth audit of the code so I can't really answer this. I'd say "low" though based on the functionality of the mod.

HebaruSan:
OK, I turned on staging as a precaution. It's easy enough to deal with.

PR #7121, opened 2019-04-13
Make `Install` property of RedOnion lower case

DasSkelett:
Thanks @HebaruSan for pointing it out.

### DIFF
--- a/NetKAN/RedOnion.netkan
+++ b/NetKAN/RedOnion.netkan
@@ -3,7 +3,7 @@
     "identifier":   "RedOnion",
     "$kref":        "#/ckan/spacedock/2116",
     "license":      "MIT",
-    "Install": [
+    "install": [
         {
             "find": "Licenses",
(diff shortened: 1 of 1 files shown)


HebaruSan:
I'm trying to figure out whether the bundled ZIP files need to be unzipped, but I don't see any installation instructions. Any idea?

DasSkelett:
No, see [here](https://github.com/evandisoft/RedOnion/issues/28)

DasSkelett:
Out of interest, would there be a way for CKAN to unzip folders (besides the root folder)?

HebaruSan:
Do you mean unzip nested ZIPs? No, that's not implemented as far as I know.

If you mean extract regular folders from the main ZIP, you're already doing that here with Plugins.

DasSkelett:
First one, extract nested zips inside the root zip. Good to know, thank you.

PR #7122, opened 2019-04-13
Splits Kerbalism into config/base

steamp0rt:
Splits Kerbalism into 2 packages -- the core (DLL + parts + icons), and the config (settings.cfg, System, Profile, etc)

the config is basically the settings and all the mod support, module manager configs, the profiles, et cetera

Why? Because the RO compatability w/ Kerbalism will likely change configs/profiles, and therefore this is a good way to avoid file conflicts.


NetKAN/Kerbalism.netkan
```diff
@@ -65,5 +65,22 @@
 	],
 	"conflicts": [{
 		"name": "UKS"
+	}],
+	"x_netkan_override": [{
+		"version": ">2.2",
+		"override": {
+			"depends": [{
```
steamp0rt: Will this add on or fully override?

steamp0rt: Actually, should x_netkan_override even be needed?

steamp0rt:
This is ready

steamp0rt:
whoops wrong button

DasSkelett:
@HebaruSan can you review this?

HebaruSan:
@steamp0rt, would you mind writing a description of the change and why/what you're trying to accomplish? I find that extremely helpful when reviewing code or metadata.

steamp0rt:
@HebaruSan this simple splits Kerbalism into 3 packages -- the core (DLL + parts + icons), the profile (Default.cfg, etc), and the config (settings.cfg, System, etc)

Why? Because the RO compatability w/ Kerbalism will likely change configs/profiles this, and therefore this is a good way to avoid file conflicts.

HebaruSan:
@steamp0rt, thanks, that's a good start. Some follow-up questions:

- What's a profile?
- What's a config?

steamp0rt:
the config is basically the settings and all the mod support, module manager configs, et cetera.

the profile is just a single config that's used to configure Kerbalism's features, and can't be changed by Module Manager (and it's always required to be in the GameData/Kerbalism/Profiles folder)

HebaruSan:
Thanks, that helps a lot.

So what are some examples of profiles and configs that might be created? You mentioned RO might have both, and obviously there's the Default, presumably for stock. Are there other examples of when a profile and/or config would be created?

In particular, I'm curious when you might override just the config, or just the profile, since the only examples we have so far provide both. Might it make sense to merge the profile and config, to simplify the packaging?

Standecco:
I'm the one to blame for these changes. I'm configuring Kerbalism to work for RO and the first thing that occurred to me is that, as steamp0rt said, it wouldn't have been possible to do it properly without separate configs.
To answer your questions, the profile decides kerbals' resource consumptions, fuel cells, scrubbers and processes in general; configs also needed to be different for RO because they patch stock parts, which RO also does.
Regarding merging config and profile, I think it would work just as well. There aren't really that many use cases for only one of them, unless for SMURFF or similar mods, but they will work just as well with a unified config.

steamp0rt:
@HebaruSan done

Standecco:
@HebaruSan any news? Why isn't this getting merged? It's been quite a long time

steamp0rt:
yeah this should be ready

steamp0rt:
Bump

Standecco:
Could you at least respond?

HebaruSan:
Thanks for the response and sorry for my unresponsiveness on this. It somehow grew into a big deal in my mind and I couldn't muster the effort.

Looks like we've done this one before, in #3967, and un-did it in #5010. What happened there? Did the mod actually change in a way that a profile module was no longer needed and then change back, or was that a misunderstanding?

steamp0rt:
> starting from Kerbalism version 1.1.5 for KSP 1.2.1 (to be released in the following days), multiple profiles can cohexist in the same user installation.
> 
> As a consequence of this, the package Kerbalism-Profile-Default is now obsolete and has no reason to keep existing. I don't know how to proceed about this, but I doubt that is as simple as deleting the metadata associated.
> 
> In addition, the package Kerbalism has a special provision: files with extension .cfg in the Profile/ subfolder are deleted post-installation. This is not necessary anymore, for the same reason specified above.

The reason for undoing this is due to the fact multiple configs cannot co-exist (such as the ROKerbalism configs).

HebaruSan:
Thanks, I think that answers my last question on this. Let's do this...

PR #7123, opened 2019-04-13
 Fix author and homepage to avoid confusion

henrybauer:
I've been maintaining this mod for 3 years, but never updated the NetKAN file.  Apparently this is causing confusion. :)

See https://forum.kerbalspaceprogram.com/index.php?/topic/137489-16-gpospeedfuelpump-continued-v1818/&do=findComment&comment=3582515


henrybauer:
I've asked for the wrong CKAN-meta files to be updated in https://github.com/KSP-CKAN/CKAN-meta/pull/1440

DasSkelett:
Do you want to fix already existing ckans too? You could add that to #1440.

PR #7127, opened 2019-04-15
Fix Tundra's relationships and licenses

### LINKED ISSUE #7125: Tundra Exploration 1.4 for KSP 1.7
Hi there,

For the new version of Tundra Exploration some dependencies have been changed:

The current status for the following required mods are:
- Kerbal Reusability Expansion 
- B9 Part Switch
- Animated Decouplers
- Module Manager

Recommended mods are:
- Near Future Solar
- Near Future Props
- Smokescreen
- FMRS

Not everything is on CKAN, so if it can be as close to the ones listen on the forum would be nice.
[Forum link](https://forum.kerbalspaceprogram.com/index.php?/topic/166915-17x-tundra-exploration-v14-april-13th-spacex-falcon-9-dragon-v2-and-starship/)

For licensing, the current licensing on all CKAN METAs are wrong. Tundra Exploration uses CC BY-NC-ND 4.0.

For TundraTechnologies (which is included within the spacedock download) only needs Tundra Exploration as required mod. Same licensing applies.

I am not home to CKAN nor have I ever used it before, so if you need more info please ask.

HebaruSan:
These mods' author has requested several changes:

- Fix the licenses (we had the wrong one)
- Different depends and recommends

Fixes #7125.

### DIFF
--- a/NetKAN/TundraExploration.netkan
+++ b/NetKAN/TundraExploration.netkan
@@ -1,30 +1,31 @@
 {
-    "$kref": "#/ckan/spacedock/223",
-    "$vref": "#/ckan/ksp-avc/TundraExploration.version",
     "spec_version": "v1.4",
-    "identifier": "TundraExploration",
-    "license": "CC-BY-NC-SA-4.0",
+    "identifier":   "TundraExploration",
+    "$kref":        "#/ckan/spacedock/223",
+    "$vref":        "#/ckan/ksp-avc/TundraExploration.version",
+    "license":      "CC-BY-NC-ND-4.0",
     "depends": [
-        { "name": "NearFutureSolar-Core" },
-        { "name": "B9PartSwitch" },
-        { "name": "TundraTechnologies" },
-        { "name": "ModuleManager" },
-        { "name": "SpaceXLegs" }
+        { "name": "SpaceXLegs"         },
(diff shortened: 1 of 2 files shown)


HebaruSan:
@damonvv, I think these are the changes you requested. They should be available to users about 4 hours after merging. I'll update the licenses of the older versions separately since the files are in a different place.

damonvv:
@HebaruSan thanks! I just noticed in Tundra Exploration it said it required Tundra Technologies and that should still be there. I forgot that was a thing as well. Rest looks ok so thanks for that!

HebaruSan:
@damonvv hm, but Technologies requires Exploration as well. Is it really two mods that require each other? That's not what I'd expect because then it's never possible to install one without the other; they might as well be one mod.

damonvv:
It is one mod. Both folders are in the same .zip file. There's no separate download for Tundra Technologies.

HebaruSan:
But there are separate folders under GameData, and separate .version files. So that makes it look like two separate mods in the same download.

Reading KSP-CKAN/CKAN#2473 and #6626, it looks like originally we only had Exploration indexed, and then Technologies was indexed later.

But it looks like they're both just parts packs, with no functional dependency in either direction. Does either mod actually break if the other isn't installed?

PR #7130, opened 2019-04-15
Update SDHI-ServiceModuleSystem.netkan's game version

HebaruSan:
Latest version of this mod is for KSP 1.7, which we track in the netkan because there's no version file.

ckan compat add 1.6


HebaruSan:
#rebuild

PR #7136, opened 2019-04-16
SpaceDock catch-up 14

HebaruSan:
KSP-SpaceDock/SpaceDock#188 is not fixed.
These are the mods added since #7098.

ckan compat add 1.6


HebaruSan:
#rebuild

PR #7142, opened 2019-04-18
add KVASS

yalov:


HebaruSan:
The description on the repo says:

> if you vessel name isn't started with the Simulation pattern, then you need "to plan" your vessel before a launch. For the planning time management Kerbal Alarm Clock is used.

Should KAC be a dependency? It sounds like it's needed for this mod to work.

yalov:
There is an option to disable a planning time and to leave only the simulation, so I don't want to make KAC a dependency on a CKAN, but I mixed up recommends and suggests: it should enable the KAC by default, so be "recommends"

HebaruSan:

PR #7144, opened 2019-04-19
Portrait Stats explicit install directive

cake-pie:
Portrait Stats bundles CommunityTraitIcons for manual installs. If I understand correctly, should not include the bundled dependency when doing CKAN install so it can be obtained and updated separately.

Will this fix the existing ckans in CKAN-meta automatically or should I PR corresponding changes there as well?

### DIFF
--- a/NetKAN/PortraitStats.netkan
+++ b/NetKAN/PortraitStats.netkan
@@ -4,6 +4,12 @@
     "$kref":        "#/ckan/spacedock/133",
     "$vref":        "#/ckan/ksp-avc",
     "license":      "MIT",
+    "install": [
+        {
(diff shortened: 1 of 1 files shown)


HebaruSan:
When there's no `install` property, the default is

```json
    {
        "find": "<identifier>",
        "install_to": "GameData"
    }
```

https://github.com/KSP-CKAN/CKAN/blob/5331bf9344f6729a6b96f52bea45bdec245a3bb3/Core/Types/ModuleInstallDescriptor.cs#L109-L120

So this change would have no effect and is unnecessary.

cake-pie:
Thanks for clarification.

PR #7147, opened 2019-04-20
Update PKS to KSP 1.7

SteveBenz:
If there's a way to make it so that CKAN pulls the max KSP version out of the GitHub release, please point me at it.


HebaruSan:
To automate this, you can put a `ProgressiveColonizationSystem.version` file in your download containing the relevant version numbers for that download, format spec:
http://ksp.cybutek.net/kspavc/Documents/README.htm

Then we would add this to your netkan to tell the bot to use the .version file:

```json
    "$vref": "#/ckan/ksp-avc",
```

HebaruSan:
Thanks for the update.

SteveBenz:
Thanks!  I'll get that done for next time.

PR #7151, opened 2019-04-22
Update KerbalJointReinforcement.netkan

pap1723:
Maintenance Release for 1.3.x to 1.7.x from the RO Team now


HebaruSan:
The RO team's KJR lives here (see #6899):

https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/KerbalJointReinforcementContinued.netkan

PR #7157, opened 2019-04-23
CryoEngines and related extras

ChrisAdderley:
This should be getting to the end of these. Moves to self hosting and adds a previously un-indexed mod.

### DIFF
--- a/NetKAN/CryoEngines-LFO.netkan
+++ b/NetKAN/CryoEngines-LFO.netkan
@@ -1,16 +1,5 @@
 {
     "spec_version": "v1.4",
     "identifier": "CryoEngines-LFO",
-    "$kref": "#/ckan/spacedock/709",
-    "$vref": "#/ckan/ksp-avc/GameData/CryoEngines/Versioning/CryoEngines.version",
-    "name": "Cryogenic Engines Extras: LFO Patch",
-    "abstract": "Optional patch to make the engines burn KSP's normal liquid fuel and oxidizer mixture instead of the custom liquid hydrogen and oxidizer mixture that they use by default.",
-    "license": "CC-BY-NC-SA-4.0",
-    "x_netkan_epoch": 1,
-    "depends": [
-        { "name" : "CryoEngines" }
-    ],
-    "install": [
-        { "find" : "CryoEnginesLFO", "install_to" : "GameData" }
-    ]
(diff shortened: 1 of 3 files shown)


NetKAN/CryoEngines-LFO.netkan
```diff
@@ -1,16 +1,5 @@
 {
     "spec_version": "v1.4",
     "identifier": "CryoEngines-LFO",
-    "$kref": "#/ckan/spacedock/709",
-    "$vref": "#/ckan/ksp-avc/GameData/CryoEngines/Versioning/CryoEngines.version",
-    "name": "Cryogenic Engines Extras: LFO Patch",
-    "abstract": "Optional patch to make the engines burn KSP's normal liquid fuel and oxidizer mixture instead of the custom liquid hydrogen and oxidizer mixture that they use by default.",
-    "license": "CC-BY-NC-SA-4.0",
-    "x_netkan_epoch": 1,
```
HebaruSan: This `x_netkan_epoch` property was lost in the transition; it needs to be preserved, or the bot will treat newer versions as old versions.

ChrisAdderley: So it's an override on version numbers? eg 
`1.0.` with `x_netkan_epoch": 1` would count as newer than say `14.0` with `x_netkan_epoch": 0`

HebaruSan: Yes. In this case it was added in #1911 because the 1.81 release was followed by 1.9, and 9 < 81.

ChrisAdderley: Mm ok. This is exactly why I couldn't support CKAN before validating and managing all this meta myself. Presumably there would have been a typo in a package somewhere having version as `1.81` versus `1.8.1` but to my knowledge this mod has never exceeded version 0.6.6 so... 🤷‍♂️ 

I restored the field in the host repo.

HebaruSan: Sorry, typo on my part; the relevant versions were 0.1.81 and 0.1.9. I got confused by where one number stopped and the next began.

image

HebaruSan:
#rebuild

NetKAN/CryoEngines-LFO.netkan
ChrisAdderley: Ah yeah, same issue I guess - likely a borked upload to I guess, KerbalStuff back then.

HebaruSan:
Looks good now. Thanks for another update!

PR #7161, opened 2019-04-24
Force v for SpecialMissionProjects versions

DasSkelett:
The mod versions are out of order due to irregular usage of the 'v' in the version strings.
grafik
The abstract is now set manually because the generation from spacedock did some weird stuff in previous versions. The new one is copied from the first section of the spacedock description.

I'll open another PR in ckan-meta to fix the .ckan files.

### DIFF
--- a/NetKAN/SpecialMissionProjects.netkan
+++ b/NetKAN/SpecialMissionProjects.netkan
@@ -9,7 +9,7 @@
   "$kref": "#/ckan/spacedock/1238",
   "x_via": "Automated Linuxgurugamer CKAN script",
   "identifier": "SpecialMissionProjects",
-  "abstract": "This mod is made up with tweaked parts to be more useful such as adding more fuel in tanks, more powerful batterys, self refueling parts that are faster and use alot less resources than the stock parts, new types of fuel. In SMP you get new ships with each update that use SMP parts.",
+  "abstract": "This mod is made up with tweaked parts to be more useful such as adding more fuel in tanks, more powerful batteries, self refueling parts that are faster and use alot less resources than the stock parts, new types of fuel. In SMP you get new ships with each update that use SMP parts.",
   "license": "unrestricted",
   "x_netkan_version_edit": "^V?(?<version>.+)$"
 }
--- a/NetKAN/SpecialMissionProjects.netkan
+++ b/NetKAN/SpecialMissionProjects.netkan
@@ -9,5 +9,7 @@
   "$kref": "#/ckan/spacedock/1238",
   "x_via": "Automated Linuxgurugamer CKAN script",
   "identifier": "SpecialMissionProjects",
-  "license": "unrestricted"
+  "abstract": This mod is made up with tweaked parts to be more useful such as adding more fuel in tanks, more powerful batterys, self refueling parts that are faster and use alot less resources than the stock parts, new types of fuel. In SMP you get new ships with each update that use SMP parts.
+  "license": "unrestricted",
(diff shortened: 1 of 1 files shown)


HebaruSan:
But... it's a *capital* V.

image

DasSkelett:
Does this affect sorting?

HebaruSan:
Yes, but I think you'll end up with `vV`:

image

https://github.com/KSP-CKAN/NetKAN/blob/65af7ceffe6b98ad06dbe1853ea966927202c800/NetKAN/xScience.netkan#L26

DasSkelett:
Oh, okay...
Having a "x_netkan_force_capital_v" would be nice, it seems :smile: 
So let the double v persist, or remove the big Vs again?

HebaruSan:
I don't know if there is a good solution. You might be able to strip the 'V' with another property:

```json
    "x_netkan_version_edit": "^ModuleManager-(?<version>\\d+\\.\\d+\\.\\d+)\\.zip$",
```

Possibly something like

```json
    "x_netkan_version_edit": "^V?(?<version>.+)$",
```

DasSkelett:
Sounds good. So I will reverse https://github.com/KSP-CKAN/CKAN-meta/pull/1450 to remove all 'V's, and change `"x_netkan_force_v": true` to `"x_netkan_version_edit": "^V?(?<version>.+)$"` in this PR.

DasSkelett:
Tried it out:
```
-rwxrwxrwx 1 root root    1373 Apr 25 19:43 SpecialMissionProjects-10.0.ckan*
-rwxrwxrwx 1 root root    1371 Apr 25 19:43 SpecialMissionProjects-4.2.ckan*
-rwxrwxrwx 1 root root    1373 Apr 25 19:43 SpecialMissionProjects-5.0.ckan*
-rwxrwxrwx 1 root root    1372 Apr 25 19:43 SpecialMissionProjects-6.0.ckan*
-rwxrwxrwx 1 root root    1372 Apr 25 19:43 SpecialMissionProjects-6.1.ckan*
-rwxrwxrwx 1 root root    1372 Apr 25 19:43 SpecialMissionProjects-6.2.ckan*
-rwxrwxrwx 1 root root    1371 Apr 25 19:43 SpecialMissionProjects-7.0.ckan*
-rwxrwxrwx 1 root root    1372 Apr 25 19:43 SpecialMissionProjects-8.0.ckan*
-rwxrwxrwx 1 root root    1371 Apr 25 19:43 SpecialMissionProjects-9.0.ckan*
```
Works!
I love the --releases option!

HebaruSan:
Don't run netkan as root! :fearful:

DasSkelett:
Didn't actually tell it to do so. There is some weird owner thing going on in my whole "git" directory, every file in there belongs to root. I think this is due to it being a symlink to another hdd, where my Windows Home folder is located. It seems perma-mounted drives always belong to root. I have to figure out a solution to this one time, a simple `chown` doesn't seem to work...

HebaruSan:
If you mount that drive in `/etc/fstab`, try adding `uid=1000,gid=1000` (replace 1000 with the desired user and group IDs), as in:

```
UUID=XXXXXXXXXXXXXXXX /media/Windows ntfs defaults,ro,uid=1000,gid=1000,umask=0022,noatime,user,nofail,nobootwait 0 2
```

DasSkelett:
Woaaah, this worked :tada:
Those hours of googling, should've asked you first :smile:

PR #7163, opened 2019-04-25
Add support for Semi-Saturable Reaction Wheels

pap1723:


HebaruSan:
We should update the current listing unless both forks are planned to be maintained in parallel:

https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/SemiSaturatableRW.netkan

HebaruSan:
Also "GNU" isn't in the list of licenses. If you just delete that line, we'll get the proper license directly from the repo automatically.

HebaruSan:
FYI you can easily search for indexed mods on the status page:

http://status.ksp-ckan.space/

pap1723:
I never knew that page existed. Thank you! I will resubmit. Sorry all of
the recent ones have required more work on your end.

PhineasFreak:
The new SSRW version is only valid for KSP 1.4+. Older KSP versions should be served by the original SSRW distribution.

HebaruSan:
I took the liberty of migrating your changes to the existing netkan.
@PhineasFreak, don't worry, the old versions won't be affected. They'll still be there.

HebaruSan:
#rebuild

PR #7169, opened 2019-04-26
Kurs

linuxgurugamer:
using github for now

ckan compat add 1.6


HebaruSan:
It needs an `abstract`. Might as well add it to the GH repo:

abstract

HebaruSan:
#rebuild

linuxgurugamer:
I've done that, and the output is very strange.  First, it has this error:

DockingCamKURS 1.3.4 requires ModuleManager but it is not listed in the index, or not available for your version of KSP.

and second, even though the .version in the file it downloaded from Github has the following:
`  "KSP_VERSION_MIN": {
    "MAJOR": 1,
    "MINOR": 4,
    "PATCH": 5
  },
  "KSP_VERSION_MAX": {
    "MAJOR": 1,
    "MINOR": 6,
    "PATCH": 1
  }`

it still seemed to think the min version was 1.7.0.  Is it somehow picking uup the old version?

HebaruSan:
The "URL" property overrides the contents of the AVC file, and this one says 1.7.0:

https://raw.githubusercontent.com/linuxgurugamer/DockingCam/master/DockingCamera.version

PR #7178, opened 2019-05-02
Update ProceduralFairings-ForEverything.netkan

pap1723:
New Release, bring the vref to a local version .netkan file


HebaruSan:
#rebuild

HebaruSan:
Looks good, thanks for the update.

PR #7179, opened 2019-05-02
Update ProceduralFairings.netkan

pap1723:
New maintainer and moving the .netkan version locally

### DIFF
--- a/NetKAN/ProceduralFairings.netkan
+++ b/NetKAN/ProceduralFairings.netkan
@@ -1,16 +1,6 @@
 {
-    "spec_version": 1,
+    "spec_version": "v1.4",
     "identifier": "ProceduralFairings",
-    "$kref": "#/ckan/github/rsparkyc/ProceduralFairings",
-    "$vref": "#/ckan/ksp-avc",
-    "x_netkan_epoch": 1,
-    "name": "Procedural Fairings",
-    "abstract": "Fairings that automatically reshape for any attached payload",
-    "license": "CC-BY",
-    "resources" : {
-        "homepage" : "http://forum.kerbalspaceprogram.com/index.php?/topic/161733-13-1"
-    },
-    "suggests": [
-        { "name" : "PFTextures" }
-    ]
+    "$kref": "#/ckan/netkan/https://raw.githubusercontent.com/KSP-RO/ProceduralFairings/v1.6.1/ProceduralFairings.netkan",
+    "x_netkan_license_ok": true
 }
--- a/NetKAN/ProceduralFairings.netkan
+++ b/NetKAN/ProceduralFairings.netkan
@@ -1,7 +1,7 @@
 {
     "spec_version": "v1.4",
     "identifier": "ProceduralFairings",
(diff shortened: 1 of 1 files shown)


NetKAN/ProceduralFairings.netkan
```diff
@@ -1,16 +1,6 @@
 {
-    "spec_version": 1,
+    "spec_version": "v1.4",
     "identifier": "ProceduralFairings",
-    "$kref": "#/ckan/github/rsparkyc/ProceduralFairings",
-    "$vref": "#/ckan/ksp-avc",
-    "x_netkan_epoch": 1,
```
HebaruSan: It's not safe to drop `x_netkan_epoch` properties. Newer releases will be treated as older.

pap1723:
OK, would you like me to add it back? I looked at the specs and couldn't
figure out what it actually does. Can you shed some light on this?

HebaruSan:
It sets the part of the version before the colon:

https://github.com/KSP-CKAN/CKAN-meta/blob/52dbdadc71cc585a44aa13c42326c5e359f91659/ProceduralFairings/ProceduralFairings-1-1.5.0.5.ckan#L12

This creates a total override of the version ordering; any version with an epoch is considered newer than any version without an epoch, and any version with a higher epoch is considered newer than any version with a lower epoch. Once `x_netkan_epoch` is added to a netkan, it has to stay forever (or be increased).

pap1723:
Thank you. I will be back at my desk soon and will update

NetKAN/ProceduralFairings.netkan
```diff
@@ -1,16 +1,7 @@
...
-    "name": "Procedural Fairings",
-    "abstract": "Fairings that automatically reshape for any attached payload",
-    "license": "CC-BY",
-    "resources" : {
-        "homepage" : "http://forum.kerbalspaceprogram.com/index.php?/topic/161733-13-1"
-    },
-    "suggests": [
-        { "name" : "PFTextures" }
-    ]
+    "$kref": "#/ckan/netkan/https://raw.githubusercontent.com/KSP-RO/ProceduralFairings/v1.6.1/ProceduralFairings.netkan",
```
HebaruSan: This URL probably shouldn't have v1.6.1 in it. When you release a new version, either the URL would no longer make sense or you would have to update it with a PR here, somewhat defeating the purpose of a metanetkan.

NetKAN/ProceduralFairings.netkan
HebaruSan: Are you sure you want the epoch in the NetKAN repo instead of in your metanetkan? Either will work, but putting it in your metanetkan would give more control.

pap1723:
Yes, leaving the epoch there is good, then we won't accidentally remove it
again. :)

NetKAN/ProceduralFairings.netkan
HebaruSan: https://raw.githubusercontent.com/KSP-RO/ProceduralFairings/ProceduralFairings.netkan isn't a valid URL. 404.

pap1723:
Sorry about that

HebaruSan:
Looks good now, thanks for the update.

PR #7182, opened 2019-05-06
add unBlur

cake-pie:
Lets end users fix blurry UI textures in other mods via config, without having to modify or rename installed texture files.

### DIFF
--- a/NetKAN/unBlur.netkan
+++ b/NetKAN/unBlur.netkan
@@ -1,0 +1,24 @@
+{
+    "spec_version" : "v1.16",
+    "identifier"   : "unBlur",
+    "name"         : "unBlur",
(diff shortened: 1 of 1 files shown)


HebaruSan:
A folder for this mod could be useful if you decide to add more files, such as a .version file or some default .cfgs. Might also help with ModuleManager `:NEEDS[]` clauses. Are you wedded to installing in the root of GameData?

cake-pie:
No foreseeable plans to add more files, chose to install in root precisely because there is only the one dll. Any technical hurdle to changing the install arrangement in future if the need does arise? If not I'm inclined to just leave things as-is.

HebaruSan:
No problem with changing it in the future if needed. I just wanted to bring it up since it's not the usual practice (I think only ModuleManager installs directly to GameData).

PR #7183, opened 2019-05-06
Changed location of folder in compressed file

sswelm:


sswelm:
I think CKAN didn't retrieve the latest files in which the folder structure has been changed

DasSkelett:
It's failing because of this:
```
PhotonSailor 1.4.8.4 requires ModuleManager but it is not listed in the index, or not available for your version of KSP.
```

Also I think the netkan would work if you just have 
```
"find":       "PhotonSail",
"instal_to":  "GameData"
```
so, as it is right now. Netkan searches through the folder structure to find the top-most directory with this name.

HebaruSan:
Meanwhile the bot updated it without problems. :)

https://github.com/KSP-CKAN/CKAN-meta/commit/dc1ed2649507692e7950d66902423e239cf69287

I don't think this needs to be changed. As @DasSkelett explained, `find` searches the whole ZIP. (A `file` property on the other hand only searches the root, but that's not what we have here.)

PR #7188, opened 2019-05-12
Add replacement mod, delete old mod

droric:
I am replacing my LaunchPadLamp mod with TrackingLights.


HebaruSan:
That's a neat idea for a mod. Thanks for the submission!

PR #7189, opened 2019-05-14
update

meirumeiru:
update needed for next release


HebaruSan:
The conflict relationship does not need to be updated; since KJRContinued `provides` KJR, it is already considered a conflict:

image

HebaruSan:
Thanks for the update!

PR #7190, opened 2019-05-14
Update Science-Full-Transmit.netkan

Tekaoh:
Bumping up the max compatible KSP version and putting the repo in a Github organization rather than my account directly as part of my new strategy for organizing my stuff and in case my username ever changes.


HebaruSan:
Thanks for the update!

PR #7200, opened 2019-05-21
Add Harmony dependency for future AirlockPlus versions

### LINKED ISSUE #7131: Index Harmony
I mentioned this earlier in https://github.com/KSP-CKAN/NetKAN/pull/6792#issuecomment-482929482 
> Wouldn't it be better to have [Harmony](https://github.com/pardeike/Harmony) -- which TiltEm has as a dependency -- indexed separately?  It's conceivable that other mods down the line might use it as well.

This is now actually becoming an issue because I also found myself using it as a dependency for a WIP mod ( [forum](http://forum.kerbalspaceprogram.com/index.php?showtopic=183776) | [github](https://github.com/cake-pie/BetterSandboxSpecializations) ).

Harmony is a patching library for .NET in general, and not a KSP mod in the usual sense, so I don't think it's appropriate to ask them to put a *.version file somewhere. On the other hand we probably don't have to worry about its compatibility with different versions of KSP. And any KSP mods that use it as a dependency will probably be fine with using the latest version (unless something gets deprecated or there is some kind of regression.)

Is there anything clever that can be done on CKAN side of things to grab the version number from the git tag of the release, or the dll assembly info?

cake-pie:
Resolves #7131 -- will wait no further, moving ahead with staging necessary changes for next release.

~Testing `AirlockPlus.netkan` locally with `netkan.exe` generated `AirlockPlus-v0.0.10.ckan` for the existing version but it contained the new `Harmony-DONTNET` line, shouldn't `x_netkan_override` have removed that, or am I missing something?~ Fixed. Thanks @DasSkelett!

### DIFF
--- a/NetKAN/AirlockPlus.netkan
+++ b/NetKAN/AirlockPlus.netkan
@@ -11,8 +11,17 @@
         "homepage": "http://forum.kerbalspaceprogram.com/index.php?showtopic=160268"
     },
     "depends": [
+        { "name": "Harmony-DOTNET" },
         { "name": "ModuleManager" }
     ],
+    "x_netkan_override": [{
+        "version": "<=v0.0.10",
+        "override": {
+            "depends": [
+                { "name": "ModuleManager" }
+            ]
+        }
+    }],
     "recommends": [
         { "name": "CommunityTraitIcons" }
     ],
--- a/NetKAN/Harmony-DOTNET.netkan
+++ b/NetKAN/Harmony-DOTNET.netkan
@@ -1,0 +1,20 @@
+{
+    "spec_version" : "v1.16",
+    "identifier"   : "Harmony-DOTNET",
+    "name"         : "Harmony",
+    "abstract"     : "A library for patching, replacing and decorating .NET and Mono methods during runtime.",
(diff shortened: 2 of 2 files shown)


NetKAN/AirlockPlus.netkan
```diff
@@ -11,8 +11,17 @@
         "homepage": "http://forum.kerbalspaceprogram.com/index.php?showtopic=160268"
     },
     "depends": [
+        { "name": "Harmony-DOTNET" },
         { "name": "ModuleManager" }
     ],
+    "x_netkan_override": [{
+        "version": "<=0.0.10.0",
```
DasSkelett: You need to change this to
```
"version": "<=v0.0.10",
```
That means, include the leading 'v' and cut the trailing '.0'.

cake-pie:
Can this get merged please, my release of AirlockPlus has already been [delayed for almost 2 weeks](https://github.com/KSP-CKAN/NetKAN/issues/7131#issuecomment-491340090) and I am fast running out of patience here.

HebaruSan:
This is done in a different way in #7220.

PR #7201, opened 2019-05-23
Add AirlockPlus-Localization

HebaruSan:
AirlockPlus ships its localization assets in a separate ZIP. This pull request indexes it.

Came up in discussion of KSP-CKAN/CKAN#2760. Note that this pull request will **not** set the `localizations` property because it's not part of the spec/schema yet.

### DIFF
--- a/NetKAN/AirlockPlus-Localization.netkan
+++ b/NetKAN/AirlockPlus-Localization.netkan
@@ -1,5 +1,5 @@
 {
-    "spec_version" : "v1.2",
+    "spec_version" : "v1.4",
     "identifier"   : "AirlockPlus-Localization",
     "name"         : "AirlockPlus Localization",
     "abstract"     : "Localizations for AirlockPlus",
--- a/NetKAN/AirlockPlus-Localization.netkan
+++ b/NetKAN/AirlockPlus-Localization.netkan
@@ -1,0 +1,20 @@
+{
+    "spec_version" : "v1.2",
+    "identifier"   : "AirlockPlus-Localization",
+    "name"         : "AirlockPlus Localization",
+    "abstract"     : "Localizations for AirlockPlus",
+    "author"       : "cake-pie",
+    "license"      : "restricted",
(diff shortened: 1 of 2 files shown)


NetKAN/AirlockPlus-Localization.netkan
```diff
@@ -0,0 +1,20 @@
+{
+    "spec_version" : "v1.4",
+    "identifier"   : "AirlockPlus-Localization",
+    "name"         : "AirlockPlus Localization",
+    "abstract"     : "Localizations for AirlockPlus",
+    "author"       : "cake-pie",
+    "license"      : "restricted",
```
cake-pie: Localizations are actually CC-BY-SA 4.0 (see [TRANSLATING.md](https://github.com/cake-pie/AirlockPlus/blob/42e57bbcbbcdb1c24bf6c6a50ad1d4b74e959d07/Localization/TRANSLATING.md))

Might get a bit weird because your bots could then automatically submit this to archive.org but won't do so for the actual mod itself.

cake-pie:
While we're here, #7200 please?

HebaruSan:
Please don't derail other issues. :)

cake-pie:
Please break up into 8 separate modules so that end user can select individual language(s) to to be installed, rather than be forced to install all 8.

cake-pie:
Maybe you forgot, but I'm on the opt-out list, and I haven't given my consent, since I wasn't done reviewing the PR yet.

HebaruSan:
Ahh, I did forget, sorry. And no, I'm not going to spam the module list unnecessarily for this.

HebaruSan:
Do you want me to actually revert the commit and start a new PR (looks like it could be a bit messy)? Or would you be willing to grant consent after the fact if no issues are found?

HebaruSan:
OK, I removed it.

https://github.com/KSP-CKAN/NetKAN/commit/8456e9ae2dd7ac4c15bf09cfa48255da782c0726

PR #7202, opened 2019-05-23
Add AirlockPlus-Localization

HebaruSan:
Re-doing #7201 because it was merged without permission from author on opt-out list.

AirlockPlus ships its localization assets in a separate ZIP. This pull request indexes it.

Came up in discussion of KSP-CKAN/CKAN#2760. Note that this pull request will **not** set the `localizations` property because it's not part of the spec/schema yet.


HebaruSan:
Never mind.

PR #7204, opened 2019-05-24
Move to self-hosted for last NF mods

ChrisAdderley:
Changes the last 3 NF CKAN packages to self-hosted versions.

### DIFF
--- a/NetKAN/NearFutureConstruction.netkan
+++ b/NetKAN/NearFutureConstruction.netkan
@@ -1,28 +1,5 @@
 {
-    "spec_version"        : "v1.4",
-    "identifier"          : "NearFutureConstruction",
-    "$kref"               : "#/ckan/spacedock/563",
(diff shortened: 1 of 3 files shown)


HebaruSan:
Just making sure, did you mean to remove NearFutureSolar's ModuleManager dependency? The forum thread still says,

> Dependencies (Required and Bundled)
> Module Manager

ChrisAdderley:
It's not actually required anymore so it could be removed from the forum post/distribution. I've re-added it though, just as it's more work to go and get all those out than it is to update the CKAN file ;).

HebaruSan:
#rebuild

HebaruSan:
Looks good, thanks for the updates.
One side note, you might want to update the licenses on SpaceDock, since these mods are still listed under the more permissive licenses there.

ChrisAdderley:
I'll do that... sometime. I would hope that people understand that generally the forum thread and the zips themselves are the most authoritative.

PR #7205, opened 2019-05-25
Water Launch Sites - recommend MakingHistory-DLC

yalov:
\-\-\-
Edited by @DasSkelett:
See https://github.com/KSP-CKAN/CKAN/pull/2326
\-\-\-

### DIFF
--- a/NetKAN/WaterLaunchSites.netkan
+++ b/NetKAN/WaterLaunchSites.netkan
@@ -20,5 +20,9 @@
     "depends": [
         { "name": "KerbalKonstructs" },
         { "name": "ModuleManager" }
-    ]
+    ], 
+
+    "recommends": [
+    	{ "name": "MakingHistory-DLC" }
+	]
 }


DasSkelett:
This will need some changes in the client before it works.
Current output if I try to install it via commandline:
```
4246 [1] INFO CKAN.RelationshipResolver (null) - MakingHistory-DLC is recommended/suggested but it is not listed in the index,
or not available for your version of KSP.
```

yalov:
it's ok, you can't buy it via CKAN. It is suppose to be informational

DasSkelett:
Right now you can only see it in the recommendations tab. If that's enough for you, we don't need to change anything.
If you want it to show up in the recommendation screen, as a simple text entry or something, it would need some work.

yalov:
dependency tab is ok

HebaruSan:
Looks OK (with the already noted caveat that it won't do anything), merging...

PR #7206, opened 2019-05-26
Update supported version range for BetterLoadSaveGame

jefftimlin:
I've switched to use a version range since my mod works with all versions in this range.


HebaruSan:
Nice, thanks for the update.

FYI, there's a file in there that probably isn't intended:

```
- GameData/BetterLoadSaveGame/.gitignore
```

This isn't a problem for CKAN, but I wanted to let you know since if you decide to remove it, things will go more smoothly if you do that before we merge this PR.

jefftimlin:
Thanks for letting me know. I've removed the .gitignore file now.

HebaruSan:
Looks good, merging with thanks!

PR #7208, opened 2019-05-26
Outer Planets Mod - Change metadata external reference URL

Poodmund:
We've just transferred ownership of the Git repo so the external URL reference just needs updating. :D


HebaruSan:
Nice of Galileo to confirm the transfer in the thread's OP. Thanks for the update!

PR #7209, opened 2019-05-27
Add TelemachusReborn

DasSkelett:
See https://github.com/TeleIO/Telemachus-1/issues/15

Telemachus Reborn is a maintained fork of Telemachus - a mod to access telemetry data from your web browser.

@DanGSun is going to include an AVC .version file in the future.
For the current release, a `x_netkan_override` is used to set the KSP version compatibility.

I had to do some trickery with the `x_netkan_version_edit` regex, because the latest release contains the release name in the Git(Hub) tag:
https://github.com/KSP-CKAN/NetKAN/blob/b72748d63fc7656e5c3ed346be3fe911a4f6a43d/NetKAN/TelemachusReborn.netkan#L11
@HebaruSan, any suggestions to improve the that, or is it okay like this?

### DIFF
--- a/NetKAN/TelemachusReborn.netkan
+++ b/NetKAN/TelemachusReborn.netkan
@@ -21,7 +21,7 @@
     ],
     "recommends" : [
         {
-            "name": "RasterPropMonitor",
+            "name": "RasterPropMonitor"
         }
     ],
     "install" : [
--- a/NetKAN/TelemachusReborn.netkan
+++ b/NetKAN/TelemachusReborn.netkan
@@ -1,0 +1,42 @@
+{
+    "spec_version"   : 1,
+    "identifier"     : "TelemachusReborn",
+    "$kref"          : "#/ckan/github/TeleIO/Telemachus-1",
+    "$vref"          : "#/ckan/ksp-avc",
+    "name"           : "Telemachus Reborn",
+    "abstract"       : "Telemachus is a universal Web Telemetry Mod for Kerbal Space Program. It allows you to access any aspects of game from browser.",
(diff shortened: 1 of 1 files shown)


HebaruSan:
Nice, I tried indexing this in January but there was something wrong with the ZIP file.

I think the thread is here:  https://forum.kerbalspaceprogram.com/index.php?/topic/179887-*

The version edit looks OK to me.

DasSkelett:
Thanks. Yeah, the second to last release zip is broken, but the current one works again

DanGSun:
> Nice, I tried indexing this in January but there was something wrong with the ZIP file.
> 
> I think the thread is here: https://forum.kerbalspaceprogram.com/index.php?/topic/179887-*
> 
> The version edit looks OK to me.
@HebaruSan, there were problems with automated zip-archive creation (we have automated building pipeline from Houston on circleci to Telemachus itself on AppVeyor. And the problem was on AppVeyor, so technically, the latest zip was created manually. If sb know how to create openable on MacOS and Linux archive on AppVeyor side, I would be very grateful.

HebaruSan:
@DanGSun, I don't know anything about AppVeyor. For my own mods, I just use the Linux command line utility `zip` to create archives.

The issue that I saw with that ZIP was that the file paths used backslashes instead of forward slashes as the ZIP spec says they should. This came up before in #6984, where that author explained, "I used .Net's System.IO.Compression.ZipFile.CreateFromDirectory from a Powershell Script." So if AppVeyor is using that method, then it's possible they suffer from the same issue.

PR #7212, opened 2019-05-28
Add missing InnerLock install stanza

HebaruSan:
See https://forum.kerbalspaceprogram.com/index.php?/topic/154922-ckan-the-comprehensive-kerbal-archive-network-v1260-baikonur/&do=findComment&comment=3606503

This mod has another folder that should be installed. This pull request adds it.

### DIFF
--- a/NetKAN/InnerLock.netkan
+++ b/NetKAN/InnerLock.netkan
@@ -2,6 +2,7 @@
     "spec_version": "v1.4",
     "identifier":   "InnerLock",
     "$kref":        "#/ckan/spacedock/2113",
+    "$vref":        "#/ckan/ksp-avc",
     "license":      "MIT",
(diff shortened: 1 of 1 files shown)


HebaruSan:
Ooh, and it's got a version file. vref incoming...

whale2:
Sorry for breaking in - how could I avoid confusing CKAN with future InnerLock releases or other mod releases that are similarly structured?

HebaruSan:
Hi @whale2, it's mainly when new folders are added that we run into problems. There are a few options for that:

1. Put new files in an existing folder when possible
2. [Open an issue](https://github.com/KSP-CKAN/NetKAN/issues/new) to tell us when new folders are planned to be added
3. Don't worry about it and let us fix it after the fact

PR #7217, opened 2019-05-31
Update NMSG.netkan

Tekaoh:
This mod is currently listed in CKAN as having a max version of 1.4.1. It actually works fine on 1.7.1 as well, so I propose overriding the max version to make it available and easy to install for anyone using an up-to-date game.

At the same time, I propose recommending xScienceContinued instead of xScience. The original mod is no longer maintained and has been continued for newer KSP versions by @Flupster. (Thanks @Flupster!)

If there's a better way to do it, or if I should fork the mod and maintain it myself, I'm open to that as well.

### DIFF
--- a/NetKAN/NMSG.netkan
+++ b/NetKAN/NMSG.netkan
@@ -6,7 +6,7 @@
         { "name": "ModuleManager", "min_version": "2.6.0" }
     ],
     "recommends": [
-        { "name": "xScience" },
+        { "name": "xScienceContinued" },
         { "name": "ContractConfigurator" }
     ],
     "license": "MIT",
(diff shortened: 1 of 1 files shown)


Tekaoh:
If there's a better way to do it, or if I should fork the mod and maintain it a little more properly myself, I'm open to that as well.

HebaruSan:
On its face this seems reasonable, especially since it's just a ModuleManager patch. I'd even wonder whether making it compatible with all versions would be better. However, I think we generally try to keep the compatibility info reflective of what the mod author supports, so users can get support if needed (and so mod authors aren't overwhelmed by false bug reports caused by using the wrong game version).

@maculator, is this OK with you?

Tekaoh:
We'll give it some more time just in case, but I'm not sure if he's around these days. No forum or ksp-related github activity in over a year.

HebaruSan:
> if I should fork the mod and maintain it myself, I'm open to that as well.

Would you mind doing that? Then we'd have an active maintainer for this.

PR #7219, opened 2019-06-01
Add Kopernicus-Backport

HebaruSan:
See https://forum.kerbalspaceprogram.com/index.php?/topic/154922-ckan-the-comprehensive-kerbal-archive-network-v1260-baikonur/&do=findComment&comment=3609503

Kopernicus releases additional versions for older game versions on a separate repo.

https://github.com/Kopernicus/Kopernicus-Backport/releases

Currently they're all missing from CKAN, so anyone using Kopernicus on an older version will have an out of date copy of the mod. This pull request adds a netkan that checks the backport repo and indexes its releases under the Kopernicus identifier. A `x_netkan_version_edit` property is used to replace "backport" in the version string with "release", to prevent this from causing problems with the version sort.

### DIFF
--- a/NetKAN/Kopernicus-Backport.netkan
+++ b/NetKAN/Kopernicus-Backport.netkan
@@ -1,0 +1,42 @@
+{
+    "spec_version"  : "v1.4",
+    "identifier"    : "Kopernicus",
+    "name"          : "Kopernicus Planetary System Modifier",
+    "abstract"      : "Allows users to replace the planetary system used by the game.",
+    "$kref"         : "#/ckan/github/Kopernicus/Kopernicus-Backport",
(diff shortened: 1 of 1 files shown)


HebaruSan:
This test is not applicable in this case; we need the identifier to be Kopernicus but the filename can't be. Propose skipping this and merging anyway.

```
#   Failed test 'Kopernicus-Backport.netkan identifier should match filename'
#   at t/metadata.t line 23.
#          got: 'Kopernicus'
#     expected: 'Kopernicus-Backport'
# Looks like you failed 1 test of 13414.
t/metadata.t .. 
Dubious, test returned 1 (wstat 256, 0x100)
Failed 1/13414 subtests 
```

DasSkelett:
Problem is this makes other PRs fail to, see #7221 and #7220.

HebaruSan:
Yup, sorry about that. Working on it in #7222.

PR #7220, opened 2019-06-01
Use Harmony bundled with AirlockPlus

cake-pie:
Since #7131 and #7200 are clearly going nowhere fast due to continued inertia, this is intended to provide some kind of workable interim arrangement so as to stop holding up my releases.

For now I will be bundling Harmony for manual installs in a manner similar to what I laid out in https://github.com/KSP-CKAN/NetKAN/issues/7131#issuecomment-484625693 which follows the generally established convention for bundling Module Manager.

In this compromise approach, CKAN can take that bundled dll and put it in the mod folder instead of GameData to prevent collision between multiple mods that bundle Harmony in this way. This is effectively the same as the "suggestion" in #7131 that all mods that require Harmony to include their "own" copy of it.

For reasons already discussed in #7131 this is not my preferred approach; I cannot guarantee that this won't run into problems down the line. #7200 is still the more robust and recommended solution.

Syntax check requested before merge.  
For sample refer to [prerelease version](https://github.com/cake-pie/AirlockPlus/releases/tag/v0.0.10.9).

### DIFF
--- a/NetKAN/AirlockPlus.netkan
+++ b/NetKAN/AirlockPlus.netkan
@@ -1,5 +1,5 @@
 {
-    "spec_version" : "v1.2",
+    "spec_version" : "v1.16",
     "identifier"   : "AirlockPlus",
     "name"         : "AirlockPlus",
     "abstract"     : "Allows the use of any airlock on a vessel in conjunction with any crew part on the vessel, without requiring manual crew transfer.",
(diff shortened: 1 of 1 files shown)


DasSkelett:
#rebuild

DasSkelett:
#rebuild

DasSkelett:
@HebaruSan why does the rebuild not work here?

HebaruSan:
> it needs to be rebased, doesn't it?

Yes.

DasSkelett:
Thanks.
@cake-pie, can you rebase this branch on the current master?
The tests are failing due to an unrelated problem that got fixed by now.

DasSkelett:
Thank you. Running netkan locally works fine, the Harmony dll is installed correctly:
installed files

The validation fails because it doesn't check the pre-release, and the last full release had no Harmony. But the regex looks good to me.

@HebaruSan any objections? Else I would merge this now.

HebaruSan:
@DasSkelett no objections. Also I'm not in charge of anything, you don't need my approval. :)

DasSkelett:
> @DasSkelett no objections. Also I'm not in charge of anything, you don't need my approval. :)

Okay, just wanna ask if you spotted something that won't work, better know it before than fixing it afterwards :)

DasSkelett:
@cake-pie this is merged, feel free to finally release the new version.

Issue #7101, opened 2019-03-31
Discussion: Best Practices for managing mod NetKAN metadata?


mwerle:
G'day,
Raising as issue since nobody seems to be active on IRC and there's a note to raise a Github issue instead.. please let me know if this type of discussion would be better on the Forum Thread instead.

In general it's fairly straightforward, but wondering about some Best Practices for managing the CKAN metadata for mods.

----

What are the pros/cons about hosting NetKAN files myself for the mods I manage (ie, Metanetkan's) versus sending pull-requests against the master NetKAN's in the NetKAN repo?

Seems that if I host them myself, I can easily manage any minor changes and updates without having to wait for a NetKAN developer to review and accept pull requests, and in the event that I totally stuff up or go AWOL, the master NetKAN can be modified to remove link to the Metanetkan to fix things.

----

Both Curse and Spacedock have a mechanism to mark a released Mod as being compatible with a new version of KSP.

So wondering whether it's possible to tag a mod as being compatible with a new version of KSP without having to make a new release which would force a user to re-download a mod just for an updated .version file.

----

Thanks!

DasSkelett:
It's better as a issue here, else we would spam the forum thread with messages and all the users watching it.
I don't have much experience with metadata yet, neither do I maintain any KSP mod, so I can't speak out of a modder's side of view

DasSkelett:
Sorry misclick :/

politas:
It is possible to specify exceptions to the default scraping of KSP version compatibility in the Netkan file for the latest release, and in the .ckan files for old releases.The format is a little finicky, so hosting the Netkan file here is probably better from that regard unless you want to invest a fair bit of time understanding the .netkan structure.

The main reason for self-hosted .netkan files is when you are making fairly substantial changes to the file structure of your mod and need to adjust the install instructions for each release.

HebaruSan:
> So wondering whether it's possible to tag a mod as being compatible with a new version of KSP without having to make a new release which would force a user to re-download a mod just for an updated .version file.

Yes, this is one function of the "URL" property in the .version file (in addition to checking for new releases). If Netkan is able to find a valid .version file at the specified URL, and if both "VERSION" properties are the same, then the game compatibility data from the online file will override the data from the file in the download. I have done this with my mods and it works; this commit:

[...]

PR #7221, opened 2019-06-01
Remove KJR provides from KJRNext

HebaruSan:
## Problems

#7101 indexed KerbalJointReinforcementNext with a `provides` property containing `KerbalJointReinforcement`, because this module is a successor module and might be preferred by some users who, say, install a craft module recommending KJR. This has caused two problems so far:

- In #7203 a RealismOverhaul team member reported that KJRNext providing KJR causes issues for them:
  > Also I would like to point out KJR NEXT should NOT be able to provide for KJR. RO/RSS does not support next at all. It has known issues with the RO/RSS suite of mods, in fact both RO and KJR Continued conflicts with next yet ckan seems to be allowing people to install it with these mods.
- In #7210 we discovered that it's very difficult, if even possible, for IRNext to conflict with old KJR and KJRContinued without also conflicting with KJRNext. Since KJRNext provides KJR, any `conflicts` relationship with KJR implicitly includes KJRNext as well, and the `min_version` and `max_version` properties of such a conflict relationship are only applied to KJR, not KJRNext.

## Changes

Now KJRNext no longer provides KJR. This should help the RO folks with whatever conflicts they were having, and also allow us to have IRNext conflict with KJR and KJRContinued while recommending KJRNext.

(Note that older versions will be updated in a separate pull request in CKAN-meta.)

### DIFF
--- a/NetKAN/KerbalJointReinforcementNext.netkan
+++ b/NetKAN/KerbalJointReinforcementNext.netkan
@@ -8,9 +8,6 @@
     "$vref"          : "#/ckan/ksp-avc",
     "license"        : "GPL-3.0",
     "release_status" : "stable",
-    "provides"       : [
-        "KerbalJointReinforcement"
-    ],
     "conflicts"      : [
         { "name": "KerbalJointReinforcement" }
     ],


HebaruSan:
#rebuild

HebaruSan:
@DasSkelett , do you think this is the right way forward?

PR #7228, opened 2019-06-05
Move Kerbalism configuration files into GameData/KerbalismConfig

steamp0rt:


steamp0rt:
Will require 3.0 be pushed to SpaceDock before it compiles, BUT 3.0 will not install properly on ckan without this. HM.

HebaruSan:
Is it going to be two separate files like on GitHub? If so we'll need to make a few additional changes.
https://github.com/Kerbalism/Kerbalism/releases/tag/v3.0-beta-1

steamp0rt:
it's gonna be on spacedock as one file with two folders.

HebaruSan:
OK, good luck!

PR #7230, opened 2019-06-06
Fork NMSG and update my own

Tekaoh:
As suggested in [#7217](https://github.com/KSP-CKAN/NetKAN/pull/7217), I have forked NMSG and am resubmitting it for any version.

Simultaneously, I am updating my other (very similar) mod to install to any version since that seemed like a good idea and really shouldn't cause any trouble.

Closes #7217.


HebaruSan:
Thanks! Merging...

PR #7232, opened 2019-06-06
NMSG replaced_by Science-Full-Reward

Tekaoh:
Suggest the new fork as an upgrade path.  I'm not sure if this is required or even recommended, but I figured I'll propose it and see what happens.


HebaruSan:
Sure, why not?

PR #7238, opened 2019-06-12
Add Kerbalism RO config

Standecco:


HebaruSan:
Looks good. Thanks for the submission, merging...

PR #7259, opened 2019-06-17
Freeze RSSVE

HebaruSan:
This mod's (locked) forum thread says:

> I. CKAN installation
> 
> Installation of RSSVE via CKAN is not currently supported.

> CKAN is not currently supported due to various issues with the pre-release installation procedures.

**All** of its releases are marked as pre-releases, and the one that we do have indexed has been deleted. This looks like an attempt to prevent indexing.

Now it's frozen.


HebaruSan:
As of 29 days ago they've got a non-pre-release, and another 17 days ago.

https://github.com/KSP-RO/RSSVE/releases

HebaruSan:
Hi @Theysen, do you happen to remember what led to those CKAN disclaimers being added for RSSVE on the forum? It may be something we can fix, but I'm not sure what "pre-release installation procedures" means there.

Theysen:
Hi, 

so I've been wanting to get it up on CKAN again since it literally only needs EVE and acts as a config provider and whatever scatterer is marked compatible with the player's game version above 1.8. 

The only two recommendations are DOE and Planetshine and should be ticked as "install" automatically. 
I've been looking into it lately and trying to restructure the .meta but work got in my way.

HebaruSan:
Sounds good, submitted KSP-CKAN/NetKAN#8595 to start doing that, then noticed the version file is very out of date and submitted KSP-RO/RSSVE#18 for that.

Issue #7268, opened 2019-06-26
[Bug] 404 errors for Curseforge hosted mods


DasSkelett:
Problem
-------
**Describe the bug**
See status.ksp-ckan.space, (all?) mods hosted on Curseforge are failing with `The remote server returned an error: (404) Not Found.`.
Get the same error inflating locally.

The JSON returned by the API URL (e.g. https://api.cfwidget.com/project/283006) looks fine though.

**NetKAN stacktrace, useful for example API URL (mod DiscordRP):**<details><summary>Click to expand</summary>
```
516 [1] INFO CKAN.NetKAN.Sources.Curse.CurseApi (null) - Calling https://api.cfwidget.com/project/283006
516 [1] DEBUG CKAN.Net (null) - About to download https://api.cfwidget.com/project/283006
2424 [1] DEBUG CKAN.NetKAN.Processors.Inflator (null) - Deleting downloads for failed inflation
2425 [1] FATAL CKAN.NetKAN.Program (null) - The remote server returned an error: (404) Not Found.
2430 [1] FATAL CKAN.NetKAN.Program (null) -   at System.Net.HttpWebRequest.GetResponseFromData (System.Net.WebResponseStream stream, System.Threading.CancellationTokencancellationToken) [0x00146] in <e8eb3d7a311640f484845e45cbec8973>:0 
at System.Net.HttpWebRequest.RunWithTimeoutWorker[T] (System.Threading.Tasks.Task`1[TResult] workerTask, System.Int32 timeout, System.Action abort, System.Func`1[TResult] aborted, System.Threading.CancellationTokenSource cts) [0x000f8] in <e8eb3d7a311640f484845e45cbec8973>:0 
at System.Net.HttpWebRequest.GetResponse () [0x00016] in <e8eb3d7a311640f484845e45cbec8973>:0 
at CKAN.NetKAN.Sources.Curse.CurseApi.ResolveRedirect (System.Uri url) [0x00025] in <fd6408a50cda4dc0acff3f19b62dcdb8>:0 
at CKAN.NetKAN.Sources.Curse.CurseFile.GetDownloadUrl () [0x00049] in <fd6408a50cda4dc0acff3f19b62dcdb8>:0 
at CKAN.NetKAN.Sources.Curse.CurseFile.GetFileVersion () [0x0000f] in <fd6408a50cda4dc0acff3f19b62dcdb8>:0 
[...]

PR #7281, opened 2019-07-01
Freeze Curse modules

HebaruSan:
As per #7268, the Curse API doesn't work anymore. The download URLs it returns are invalid and as such we can no longer auto-index Curse-hosted mods.

Only a few mods are still indexed from Curse. Most are very outdated and unlikely to ever be updated and/or have a -Continued successor mod already indexed on another host. These are now frozen.

Four others are indexed via metanetkans owned by @ihsoft, whom I have notified of the problem so he can take action if desired.

DiscordRP is a problem. It's only indexed on Curse *and* relatively recent and active and has no -Continued successor. I guess I'll ask on the forum thread? https://forum.kerbalspaceprogram.com/index.php?/topic/168196-*


HebaruSan:
#rebuild

HebaruSan:
#rebuild

HebaruSan:
Weird error.

```
Installing mod "HumanColoredFaces 1:1"
Unhandled Exception:
CKAN.FileNotFoundKraken: Trying to install HumanColoredFaces 1:1, but it's not downloaded or download is corrupted
  at CKAN.ModuleInstaller.Install (CKAN.CkanModule module, System.Boolean autoInstalled, System.String filename) [0x00073] in <df853a3d54734416bbf20fdf2e543a94>:0 
  at CKAN.ModuleInstaller.InstallList (System.Collections.Generic.ICollection`1[T] modules, CKAN.RelationshipResolverOptions options, CKAN.IDownloader downloader, System.Boolean ConfirmPrompt) [0x001b8] in <df853a3d54734416bbf20fdf2e543a94>:0 
  at CKAN.ModuleInstaller.InstallList (System.Collections.Generic.List`1[T] modules, CKAN.RelationshipResolverOptions options, CKAN.IDownloader downloader) [0x0004c] in <df853a3d54734416bbf20fdf2e543a94>:0
```

This is *after* netkan.exe runs, so it should be cached already.

techman83:
Yeah, I'm not 100% sure what is happening. But maybe ckan is hiting some problem with extracting the zip that NetKAN isn't? It's been a while since I looked at that part of the codebase.

I ran the job with -xe and everything looks to be pointing in the right places.
https://ci.ksp-ckan.space/job/NetKAN/8502/console

HebaruSan:
The file is caching with a weird name:

image

So `FileIdentifier.IdentifyFile` is not able to identify it:

https://github.com/HebaruSan/CKAN/blob/3250f57aeab7432a18b593bc9352d7b3044c893f/Netkan/Services/CachingHttpService.cs#L41-L66

The link for convenient investigation:

https://spacedock.info/mod/622/Human%20Colored%20Faces%20-%20RSS/SASS%20Kerbal%20Texturepack/download/1

Confirmed that something weird is going on with that file:

```bash
$ file Nuke_Refueled-v0.10-eta.zip
Nuke_Refueled-v0.10-eta.zip: Zip archive data, at least v1.0 to extract

$ file TacLifeSupport_V0.13.13.zip
TacLifeSupport_V0.13.13.zip: Zip archive data, at least v2.0 to extract

$ file Human_Colored_Faces_-_RSS_SASS_Kerbal_Texturepack-1.zip
Human_Colored_Faces_-_RSS_SASS_Kerbal_Texturepack-1.zip: data
```

Maybe it's a 7z?

HebaruSan:
Forget it, we'll just freeze it. We can figure out the file format problems if there's a request to re-index this mod.

techman83:
It's definitely a zip file, it's however missing its [magic number](https://en.wikipedia.org/wiki/File_format#Magic_number). Which is valid for a zip file, but quite unusual and CSharpZipLib isn't super forgiving. Maybe appending a .zip might help, but its been 4 years since I looked at the Zip code in CKAN.

PR #7328, opened 2019-07-29
Updating TweakScale to cope with more than one folder on the package.

Lisias:
See https://forum.kerbalspaceprogram.com/index.php?/topic/179030-14-tweakscale-under-lisias-management-2431-2019-0725/&do=findComment&comment=3642572 and subsequent; the default install stanza was finding `Extras/TweakScale` instead.

### DIFF
--- a/NetKAN/TweakScale.netkan
+++ b/NetKAN/TweakScale.netkan
@@ -13,12 +13,12 @@
         "pellinor",
         "LisiasT"
     ],
-    "install" : {
+    "install" : [
         {
             "file" : "GameData/TweakScale",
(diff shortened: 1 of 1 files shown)


HebaruSan:
@Lisias, are any previous versions affected by this problem? Only the latest one will be fixed automatically by this PR, but we can apply the fix to older versions if needed.

Lisias:
No. This stunt was applied to TweakScale 2.4.3.0 and later only. But only 2.4.3.1 reached CKAN, so just know I was bitten by it.

Thank you.

PR #7334, opened 2019-08-03
Add netkan files for new TestFlight configs.

Starstrider42:
This PR adds `.netkan` files for the major TestFlight config packs in Starstrider42/TestFlight-Configs. 

Local testing shows that while the dependency metadata allows/forbids the correct combinations of mods, the conflict between `TestFlightConfig-Stock-S42` and the original `TestFlightConfigStock` is not handled consistently -- the latter sometimes displays on the relationship tab as "not indexed" (despite being blocked correctly), and CKAN won't let you switch between the two config packs without uninstalling TestFlight first. 😲 I'm not sure if that's a bug in CKAN's dependency resolution or in the `.ckan` files.

### DIFF
--- a/NetKAN/TestFlightConfig-KerbalAtomics.netkan
+++ b/NetKAN/TestFlightConfig-KerbalAtomics.netkan
@@ -2,7 +2,7 @@
     "spec_version": "v1.4",
     "identifier": "TestFlightConfig-KerbalAtomics",
     "name": "TestFlight Config for Kerbal Atomics",
-    "abstract": "This config pack adds TestFlight support for Kerbal Atomics parts.",
+    "abstract": "This config pack adds TestFlight support for Kerbal Atomics parts. It does not affect stock parts; install a stock TestFlight config if desired.",
     "$kref": "#/ckan/github/Starstrider42/TestFlight-Configs",
     "release_status": "testing",
     "license": "MIT",
--- a/NetKAN/TestFlightConfig-KerbalAtomics.netkan
+++ b/NetKAN/TestFlightConfig-KerbalAtomics.netkan
@@ -1,0 +1,30 @@
+{
+    "spec_version": "v1.4",
+    "identifier": "TestFlightConfig-KerbalAtomics",
+    "name": "TestFlight Config for Kerbal Atomics",
+    "abstract": "This config pack adds TestFlight support for Kerbal Atomics parts.",
+    "$kref": "#/ckan/github/Starstrider42/TestFlight-Configs",
+    "release_status": "testing",
+    "resources": {
+        "bugtracker": "https://github.com/Starstrider42/TestFlight-Configs/issues"
+    },
+    "provides": [
+        "TestFlightConfig"
+    ],
+    "depends": [
+        { "name": "TestFlightConfigLibrary" },
+        { "name": "ModuleManager" },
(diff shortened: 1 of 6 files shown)


DasSkelett:
I understand this right, that `TFC-KerbalAtomics`, `TFC-NFPropulsion` and `TFC-NFSpacecraft` should be installed **on top** of another testflight config, like `TestFlightConfigStock`, `TestFlightConfig-Stock-S42` or the one of RO, yes?

Starstrider42:
They're designed to be mixed and matched at will; any of the config packs will work with only `Generic` as a dependency. While I can't imagine somebody wanting one of the mod-specific configs but not stock, I don't see any reason to enforce it.

Note that `TestFlightConfigStock` does exclude other configs (probably because the developer assumed people play either stock or RO), and RO is unlikely to work well with stockalike part mods like Kerbal Atomics or Near Future. But that seems like something that should be handled in the primary mods' configs, not the `TestFlight` packs.

DasSkelett:
> While I can't imagine somebody wanting one of the mod-specific configs but not stock, I don't see any reason to enforce it.

I do. Because if not, users might just install one of the mod specific extensions and assume this also includes configs for the stock parts, but it doesn't.
I think we should remove the `provides: ["TestFlightConfig"]` for those three extension configs, because
1) they don't provide a "full" config, only extensions that need any of the both stock configs to be installed, but which isn't enforced or indicated right now.
2) This should solve the siimavs stock config conflicting with them, whilst keeping the conflicts between `TestFlightConfigStock` <-> `TestFlightConfig-Stock-S42` <-> `TestFlightConfigRO` / `RO`, to ensure they don't get installed side by side.

Starstrider42:
I'm sorry, I don't quite follow what you mean by "full" config. A TF config pack adds TF modules to certain parts, no more, no less. Can't you just as well argue that a config for the stock parts isn't a full config because it doesn't support parts from mods (either these, or something that hasn't been covered at all yet, like SpaceY)?

If you want to restrict `provides: ["TestFlightConfig"]` to only config packs that support stock or total conversion mods (which seems to be what you're saying), is there some other mechanism to link single-mod configs to the `TestFlight` mod? I'd rather that config authors not be expected to add a `suggests` block to `TestFlight` itself; something that centralized would almost certainly get forgotten.

DasSkelett:
> is there some other mechanism to link single-mod configs to the `TestFlight` mod?

Currently not. There's always been the wish for [conditional recommends](https://github.com/KSP-CKAN/CKAN/issues/538), but unfortunately they are not a thing yet.

So I understand your considerations. I would go with your suggestion and change the `conflicts` of the original `TestFlightConfigStock` to `TestFlightConfig-Stock-S42, RealismOverhaul`, the counterpart for what you did here in `TestFlightConfig-Stock-S42.netkan`.
This will allow the parallel installation of the mod-specific configs to the "base"/stock game config, as you wanted.

To point out that these three configs do not modify the stock parts (some might think this is the case), could you add a sentence like
**`This config does not modify stock parts. To use the TestFlight functionality for those, you have to install one of the TestFlight configs for stock`** to the abstract of these three? (You are free to word it better than me)

This should sort that out too.

DasSkelett:
Thanks! Let's try if it works out.

PR #7335, opened 2019-08-04
Epoch AJEExtendedConfigs

DasSkelett:
Another interesting versioning system...
versions

My plan is to epoch all new versions (including the latest) twice,
and this weird `202` version once. Should get it sorted.
Also put a `x_netkan_force_v` in, so future capital `V` get converted to lower ones.


DasSkelett:
Worked :tada:
new versions

PR #7339, opened 2019-08-07
Added Urgent Contracts mod

GarwelGarwel:
Fixed BOM in the .netkan file, should be ok now


DasSkelett:
The missing upper version compatibility limit is intended?
Just asking.

GarwelGarwel:
This is going to be fixed in the next release.

DasSkelett:
Okay. Otherwise looks good, thanks for the submission!

PR #7344, opened 2019-08-13
Epoch for lost 'v' prefix

HebaruSan:
image


DasSkelett:
Always interesting with how many different versioning systems for one mod an author can come up with.

linuxgurugamer:
Yup.  That's why I have standard ways to define versions, the only time I have problems is when I do something manually.

I know it's a pipe dream, but I really wish there was a standardized method for version numbers for KSP

PR #7361, opened 2019-08-24
Add SpacePlane Corrections from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of Avera9eJoe, to add [SpacePlane Corrections](https://spacedock.info/mod/1194/SpacePlane%20Corrections) to CKAN.

Please direct questions about this pull request to [Avera9eJoe](https://spacedock.info/profile/Avera9eJoe).

Mod details:
    name = /mod/1194/SpacePlane%20Corrections
    author = Avera9eJoe
    abstract = SPC replaces the stock plane parts with revamped textures, and integrates with FStextureSwitch to let the player customize which textures and colors to use.
    license = MIT
    Homepage = http://forum.kerbalspaceprogram.com/index.php?/topic/106489-12-spc-v014-spaceplane-corrections-fstextureswitch-integration-19-october-16/
    description =
Changelog is on the forum website.

Album of more changes: [http://imgur.com/a/KjyV5](http://imgur.com/a/KjyV5)

v0.13 Changelog
 * Revamped standard elevon "shuttle" variant textures
 * Renamed from SPPCC to SPC
 * Removed depreciated custom "black" textures
 * Removed depreciated shuttle and fuel tank textures

v0.12 Changelog
 * Replaced "Black" Nasa Stripe option with Default stripes, revamped Nasa tanks
 * Replaced/revamped wing textures for better seams 
 * Removed smallhardpoint

v0.11 Changelog
 * Added Fire Spitter FStextureSwitch2 integration! Part widget to switch textures!
 * Black, mk2Gray, and mk3Gray color variants for all pertinent spaceplane parts

v0.10 Changelog
 * Added black trim to small hardpoint
 * Added black cap to front of all new landing gear
 * Added black trim to AV-R8 winglet, large tailfin
 * Darkened lines on AV-R8 winglet
 * Added optional download for mk1Cockpit with recolored windows

v0.08 Changelog
 * Matched standard canard and swept wing
 * Shoutout to InsanePlumber for helping with optimized mipmaps ONCE AGAIN! Thanks man.

v0.07 Changelog
 * Optimized images
 * Merged TextureReplacer and copy-replace downloads

v0.06 Changelog
 * improved BIG-S delta
 * re-added advanced canard

v0.05 Changelog
 * Brightened flaps
 * Added "slit" indents into the SLS tank black stripes

v0.04 Changelog
 * darkened from v0.03

release v0.03 Changelog
 * Added ablative paneling to all flaps
[...]

PR #7057, opened 2019-03-09
Update Spectra.netkan

HebaruSan:
This mod has an indexing error currently due to the changing of some folders:

```
Module contains no files matching: file="Step 2 - Spectra/GameData/TextureReplacerReplaced", file="Step 2 - Spectra/GameData/Spectra", file="Step 4 - Extras/PlanetShine/Config", file="Step 4 - Extras/PlanetShine/Icons"
```

In addition, some dependencies are out of date.

ckan compat add 1.4

PR #7362, opened 2019-08-24
Delete SPPCC.frozen

Avera9eJoe:
This file is a now defunct version of Spaceplane Correcetions, SPPCC stands for Space Plane Color Corrections, but I abbreviated the name to Spaceplane Corrections a number of years ago and recreated the page on SpaceDock/CKAN. This file is safe to delete so as to not confuse users.


HebaruSan:
Users will never see this file, but it makes sense to remove if we're indexing this mod in #7361.

HebaruSan:
BTW @Avera9eJoe, I just caught up on your chat with @allista in the SpaceDock IRC; in case you're curious how CKAN figured out how to install Spectra, it's mostly through manual inspection: #7057

Avera9eJoe:
Hm, HebaruSan the reason I requested this deleted was because users *can* see it, or at the very least I was seeing it when I searched "Spaceplane Color Corrections" on CKAN. I'm fairly certain it is this exact file after doing some searching through files in my local CKAN database. I could be wrong though and it might be something different. I'm not experienced at all when it comes to this sort of thing!

HebaruSan:
Nope, this file has/had no effect on CKAN. It skips files with a .frozen extension.
You may have been seeing this one:
https://github.com/KSP-CKAN/CKAN-meta/tree/master/SPPCC

Avera9eJoe:
Ahhh yes that would be it. The filename was SPPCC. How would I go about requesting that to be deleted? Thank you for all the help!

HebaruSan:
You can click the file, click the pencil icon, change the extension to "frozen", and then use the "commit changes" fields at the bottom to start a pull request.

PR #7374, opened 2019-09-03
Kerbalism metadata update

SirMortimer:
* New download location (Github)
* Updated the meta data: homepage, authors
* Updated list of recommended / conflicting mods

### DIFF
--- a/NetKAN/Kerbalism-Config-Default.netkan
+++ b/NetKAN/Kerbalism-Config-Default.netkan
@@ -13,5 +13,9 @@
 	}],
 	"conflicts": [{
 		"name": "Kerbalism-Config"
+	}],
+	"install": [{
+		"find": "KerbalismConfig",
+		"install_to": "GameData"
 	}]
 }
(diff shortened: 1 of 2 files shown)


SirMortimer:
Yay!

Nice build system you got there, btw :)

HebaruSan:
> Nice build system you got there, btw :)

Thanks, we like it, too. :tada: 

A few questions / observations:

- The old abstract was more helpful in explaining what the mod does. For GitHub we pull it from the short description field of the repo (between the tabs and the tags), or we can override it in the netkan file. Mind if I set up an override?
- Why not have the homepage link point to the forum thread?

SirMortimer:
> Mind if I set up an override?

Not at all, please go ahead.

> Why not have the homepage link point to the forum thread?

Because the forum post is not very practical for us to keep updated. Kerbalism is currently looked after by 4-5 different people. We set up permissions in github for anyone on the team to update the README.md (or the Wiki), but the OP in the forum can only be updated by one. It should be the same content anyway, but the github version is very likely to be the most recent.

HebaruSan:
> the OP in the forum can only be updated by one

Ahh, I see. The nice thing about a forum thread is that you can have discussions, but I can see the limits on editing the OP being a dealbreaker. Just wanted to make sure this was a considered change and not a typo. :+1:

PR #7401, opened 2019-09-23
Community Part Titles, separating Extras

yalov:
does it ok?

### DIFF
--- a/NetKAN/CommunityPartsTitles.netkan
+++ b/NetKAN/CommunityPartsTitles.netkan
@@ -10,13 +10,16 @@
         "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/174189-*"
     },
     "recommends": [
+            { "name": "CommunityPartsTitlesExtrasCategory" },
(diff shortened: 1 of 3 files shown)


HebaruSan:
Would a user ever want to install `CommunityPartsTitlesExtrasNoCCKDup` without the other two?

yalov:
yes I had had this patch before started to make CPT

yalov:
Fore example: https://forum.kerbalspaceprogram.com/index.php?/topic/149840-discussion-community-category-kit/&do=findComment&comment=2954690

HebaruSan:
OK, that makes sense...

yalov:
Does bot run once per day?

DasSkelett:
It *should* run about once every 2-3 hours now, but the new bot isn't running perfectly yet and the outage of Spacedock messed it up somehow.

techman83:
Well the old bot would have had the same problem with the spacedock outage :wink:  - But the inflator seems to have gotten stuck, so I've given it a kick.

PR #7436, opened 2019-10-18
Freeze OpenTree

DasSkelett:
Deleted from SpaceDock: https://spacedock.info/mod/1623

Confirmed discontinuation in the forum: https://forum.kerbalspaceprogram.com/index.php?/topic/110365-131-opentree-v25-2512018/&do=findComment&comment=3456854


DasSkelett:
All versions up to (including) 2.2 have their download property already set to Archive.org in CKAN-meta, will do it for the remaining 3 versions too.

PR #7438, opened 2019-10-19
Override Firespitter 7.13's max KSP version to 1.7

HebaruSan:
@sarbian reports on Discord that Firespitter 7.13 is not built for KSP 1.8 (which includes a Unity upgrade and therefore mostly breaks compatibility across the board), but this commit marked the max KSP version as 1.8.9 back in August:

https://github.com/snjo/Firespitter/commit/d6ec90eb76f09a03a1858db85f745a1ad789018f

If we can confirm this, these overrides will set it back to 1.7.

### DIFF
--- a/NetKAN/Firespitter.netkan
+++ b/NetKAN/Firespitter.netkan
@@ -7,6 +7,12 @@
     "license"      : "restricted",
     "$vref"        : "#/ckan/ksp-avc",
     "x_netkan_force_v" : true,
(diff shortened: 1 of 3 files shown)


HebaruSan:
See MuMech/MechJeb2#1188
Also check https://github.com/snjo/Firespitter/releases, "Firespitter 7.13.0 for KSP 1.7.x"

PR #7442, opened 2019-10-22
Update Author for TestFlight

siimav:
Project itself is under the KSP-RO group which means that multiple people have write access to that repo. Recently people have wrongly thought that I'm the author of that mod. I find it better to give credits to the actual author of the mod who has done most of the work.


DasSkelett:
We can also do a list of authors, if that suits better.

siimav:
Right, I guess adding KSP-RO group as the second author is actually better. 
Amended my last commit with that change.

PR #7467, opened 2019-10-31
Updating netKAN to cope with the new Redist scheme.

Lisias:
New scheme for Redist.dll distribution is needed to avoid spamming the KSP.log with AddOn Binder errors.


Lisias:
Applied a new pull request after updating SpaceDock

PR #7477, opened 2019-11-04
ProjectManager 1.8 demarcation

todi:
ProjectManager v2.3 is a recompile for 1.8 and does not work with previous ksp versions. This can be expected to be the case for all following versions of the mod.


HebaruSan:
Thanks for reporting this, too!

PR #7479, opened 2019-11-05
Create RestockRigidLegs.netkan

Poodmund:
Add RestockRigidLegs to the NetKAN repository.


Poodmund:
Would the URL to the .netkan be better as:
`https://raw.githubusercontent.com/PorktoberRevolution/ReStocked/master/CKAN/ReStockRigidLegs.netkan` ?

If so, would it also be better to then change ReStock and ReStockPlus to this style of URL rather than those that they currently have?

HebaruSan:
The problem wasn't the form of the URL, it's that it was a 404.

HebaruSan:
Thanks for the update!

PR #7488, opened 2019-11-09
Set Unlicense for applicable mods

HebaruSan:
As noted in #4131, several mods are released under the Unlicense and should be marked that way. Now they are (and their `spec_version` is updated to `v1.18`).

In some cases, the Unlicense in fact was not used, so those are skipped or updated to their actual license. A few mods are switched from a `download` property to a `$kref`, have their `homepage` updated, etc.

Note that the Unlicense is just a specific version of public domain (it's one of the built-in choices on GitHub). Technically a mod released into the public domain and a mod using the Unlicense have the same terms and conditions. Since CKAN supports both `public-domain` and `Unlicense` as license values, I have opted to set any mod that specifically mentions "unlicense" as `Unlicense` and leave those simply stating "public domain" as `public-domain`.

This is a partial fix for #4131. Some of the affected mods have no netkans and will need a separate PR in CKAN-meta.

ckan compat add 1.0

### DIFF
--- a/NetKAN/ContractConfigurator-SCANsat.netkan
+++ b/NetKAN/ContractConfigurator-SCANsat.netkan
@@ -22,7 +22,7 @@
     ],
     "install": [
         {
-            "find":       "CC_Contracts",
+            "find":       "ContractPacks",
             "install_to": "GameData"
         }
     ],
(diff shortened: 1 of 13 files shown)


HebaruSan:
#rebuild

HebaruSan:
#rebuild

PR #7496, opened 2019-11-14
Add ROCapsules and ROTanks to NetKan

pap1723:


HebaruSan:
```
1961 [1] WARN CKAN.NetKAN.Transformers.GithubTransformer (null) - No releases found for KSP-RO/ROCapsules
2033 [1] FATAL CKAN.NetKAN.Program (null) - ROCapsules missing required field download
```

pap1723:
Should be working correctly now.

HebaruSan:
#rebuild

HebaruSan:
Yup, looks good.

HebaruSan:
... and the last thing I wanted to check: Yes, there is a precedent for a mod named "ROSomething", instead of "RealismOverhaulSomething", so we won't be creating any confusion on that score here:

- https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/ROEngines.netkan

PR #7497, opened 2019-11-14
Update Procedural Parts to use the new (and correct) Repo

pap1723:


HebaruSan:
@jrodrigv just asked us to switch it in #7485.
Can you guys come to an agreement regarding which is correct?

HebaruSan:
```
4051 [1] FATAL CKAN.NetKAN.Program (null) - No files found in GameData/ProceduralParts to install!
```

pap1723:
Should be working correctly now.

HebaruSan:
#rebuild

HebaruSan:
Yup, looks good.

jrodrigv:
I don't mind to have pap's repo tracked on CKAN. The only reason I asked to track mine is because his doesn't provide a release for KSP 1.8.X 
@pap1723 Would mind to review my PR an to do a new release asap?

HebaruSan:
It's worth noting that if we merge this, we will have this in the metadata:

| PP version | KSP versions |
| --- | --- |
| v1.3.18 | 1.3.1–1.7 |
| v1.3.20 | 1.8 |
| v1.4 | 1.6.1–1.7 |

That's probably not ideal.

HebaruSan:
(clicked the wrong button)

todi:
It would be great if we could find a solution for this. CKAN currently offers PP v1.3.18 for installation in 1.7.3, which has a bug that makes it unusable on that KSP version.

HebaruSan:
@pap1723 , would you be willing to renumber your v1.4.0 release as v1.3.19? As noted above, there are already two v1.3.X versions for KSP 1.8, and having the compatibility jump backwards like that would be confusing for users and messy for CKAN. Luckily there's a gap in the numbering that could be used to address this.

DRVeyl:
As there's likely to be more than 1 release to support 1.7, can we address by:

1.3.19: Renumbering of the currently-proposed 1.4
1.3.19.1: The next 1.7 release
1.3.19.2+: ALL subsequent 1.7 releases

And either continue 1.8 releases with 1.3.22+, or increase the major or minor version number entirely (eg to 1.4+ or 2.0+)

HebaruSan:
@DRVeyl, that should work. Thanks for the reply. It's not ideal by any means but we are kind of painted into a corner here.

HebaruSan:
#rebuild

pap1723:
#rebuild

HebaruSan:
Doesn't look like that override is needed, the .version file would set that compatibility already.

HebaruSan:
The `author` list doesn't seem to cover the current fork. Any preference regarding new entries?

```json
    "author": [
        "Starwaster",
        "Tidal-Stream",
        "jrodrigv"
    ],
```

HebaruSan:
Added "KSP-RO" as the newest author, plus the earlier ones that are mentioned in the acknowledgments. We can update it later if you like.

PR #7506, opened 2019-11-18
Update PKS for 1.8 and to use version number supplied in the deployment

SteveBenz:
The version number is in the ZIP file, in GameData\ProgressiveColonizationSystem\ProgressiveColonizationSystem.version

https://github.com/SteveBenz/ProgressiveColonizationSystem/blob/master/src/ProgressiveColonizationSystem/GameData/ProgressiveColonizationSystem/ProgressiveColonizationSystem.version

### DIFF
--- a/NetKAN/ProgressiveColonizationSystem.netkan
+++ b/NetKAN/ProgressiveColonizationSystem.netkan
@@ -1,5 +1,5 @@
 {
-    "spec_version": 2,
+    "spec_version": 1,
     "identifier": "ProgressiveColonizationSystem",
     "$kref": "#/ckan/github/SteveBenz/ProgressiveColonizationSystem",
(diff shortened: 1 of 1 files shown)


DasSkelett:

SteveBenz:
NetKan doesn't seem to be recognizing the latest release... Mostly.  The version number displayed in my CKAN is v1.0.6.  If I install it, I get the v1.0.6 package.  The release is out there, and it should be v2.0.0.  Funny thing, though, the CKAN client says that the max supported KSP version is 1.8, and the only place it could have gotten that idea would be from the v2.0.0 package itself.  Any idea what's up?

DasSkelett:
It's appearing for me:
grafik
Are you on a KSP 1.7 instance?

PR #6877, opened 2018-12-07
Add PolandBall from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of whadk, to add [PolandBall](https://spacedock.info/mod/2024/PolandBall) to CKAN.

Mod details:
    name = /mod/2024/PolandBall
    author = whadk
    abstract = This mod adds polandballs
    license = CC-BY-SA 2.0
    Homepage = nope
    description = This is your mod listing! You can edit it as much as you like before you make it public.

To edit **this** text, you can click on the "**Edit this Mod**" button up there.

By the way, you have a lot of flexibility here. You can embed YouTube videos or screenshots. Be creative.

You can check out the SpaceDock [markdown documentation](/markdown) for tips.

Thanks for hosting your mod on SpaceDock!

Please direct questions about this pull request to [whadk](https://spacedock.info/profile/whadk).

PR #7507, opened 2019-11-18
Delete polandball.netkan

SteveBenz:
Looks like a bug in the spacedock integration caused it to fail to notice the differently cased version.


SteveBenz:
Looks like a bug in the spacedock integration caused it to fail to notice the differently cased version. (There's a newer version of this exact same file named with an uppercase "B" in the filename.)

DasSkelett:
See #6877, it's a different mod actually. Since this mod here hasn't been updated in a long time, it was decided that the similar name/identifier wouldn't be a problem.

SteveBenz:
Hm..  Well, I noticed it because I'm on a Windows machine, and Windows is a case-insensitive filesystem, so Git For Windows can't support both versions of the capitalization.

If you really think supporting both the older and the newer version of the mod is a good idea, I think the name should be split somehow.

SteveBenz:
To be clear, this is what you see when you clone the repo on a Windows box:

E:\repos\NetKAN>git status
On branch master
Your branch is ahead of 'origin/master' by 449 commits.
  (use "git push" to publish your local commits)

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   NetKAN/PolandBall.netkan

no changes added to commit (use "git add" and/or "git commit -a")

DasSkelett:
I blame that oversight on the lack of Windows-using team members :P

You're right, only differencing mods with capitalization, even if all OSes would use proper case-sensitive file systems, is not best practice.
We should rename the newer one. Any suggestions?

SteveBenz:
Not my mod - I'd be as lost as you.  But it sounds like this PR should be closed as not the right fix.

DasSkelett:
Should be fixed now, thanks for letting us know.

PR #7511, opened 2019-11-18
Make ChattererExtended indefinitely compatible

HebaruSan:
See KSP-CKAN/CKAN-meta#1537, this mod relies on Chatterer to perform its version-specific functions, and in theory it should be fine to support it on any version. This PR creates an override for that, so we can close that other PR while we decide whether this is a good idea.

Watch here to see if the author responds:

https://forum.kerbalspaceprogram.com/index.php?/topic/160545-142-chatterer-extended-more-kerbalized-chatter-for-chatterer-v06/&do=findComment&comment=3698839

### DIFF
--- a/NetKAN/ChattererExtended.netkan
+++ b/NetKAN/ChattererExtended.netkan
@@ -1,15 +1,18 @@
 {
-  "spec_version": "v1.4",
-  "install": [
-    {
-      "install_to": "GameData",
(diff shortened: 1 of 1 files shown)


HebaruSan:
I'll try a forum PM...

HebaruSan:
image

PR #7516, opened 2019-11-20
Repossess abandoned metanetkans

HebaruSan:
## Problems

See KerboKatz/Netkan#11 and Y3mo/SETI-NetKAN#2; these modules have metanetkans with bad or outdated metadata that users have noticed and reported. We have reached out to the authors to try to address these problems in their copies of the files, but so far neither has responded.

## Changes

Now the contents of these metanetkans are pulled back into their main netkans, with the fixes from the above pull requests applied.

@Y3mo or @KerboKatz, if you want to have these files back, let us know and we can collaborate on the problems users have noted and how to address them. Otherwise we are happy to maintain them going forward.


HebaruSan:
Inflated these locally with a token and they worked fine. We may need to set a token in Jenkins.

PR #7529, opened 2019-11-28
Bump Historian-expanded epoch

HebaruSan:
image

https://forum.kerbalspaceprogram.com/index.php?/topic/138848-181-historian-expanded/&do=findComment&comment=3710890


DasSkelett:
Oh well, I definitely didn't spot that, despite looking at the version history on GitHub...
I think this confirms we should _not_ put `x_netkan_allow_out_of_order` for this mod :)

PR #7545, opened 2019-12-10
Tag third group of mods

HebaruSan:
Follow-up to #7544, next most downloaded group plus random similar mods.


HebaruSan:
> 1739 [1] FATAL CKAN.NetKAN.Program (null) - libcurl

Too many in one PR again, I guess.

techman83:
Credits seem ok. Do we set api tokens? Might be worth generating a set of read only ones for the testing infrastructure.

HebaruSan:
Yeah, a token would be nice. For now I'm just going to split it up.

PR #7548, opened 2019-12-10
Updated Kerbalism mod conflicts

SirMortimer:
Kerablism mod conflicts depend on what Kerbalism is configured to do. This moves the mod conflict definitions where they belong.

See also https://github.com/KSP-CKAN/NetKAN/pull/7546


HebaruSan:
Nice! When you said "config" I thought it might be something the user would manipulate at runtime, this is a lot better.

PR #7550, opened 2019-12-10
Tag sixth group of mods

HebaruSan:
Successor to #7549, 35 more.


HebaruSan:
> 34368 [1] FATAL CKAN.NetKAN.Program (null) - Disk full. Path /tmp/CdFileMgr/10dca4b6-d268-4c.tmp

I guess putting all the biggest remaining mods in one PR was not a good idea.
Crossing fingers that adding tags won't break anything...

PR #7569, opened 2019-12-14
Removed deplyment of  seperate folder for InterstellarHybridRocketry

sswelm:


HebaruSan:
What's the background here? Was this incorrect? Is it changing? Etc.

sswelm:
It has become obsolete, there now is much better support for aluminum propulsion integrated in KSPIE

HebaruSan:
OK, good luck! It may be a day or two before this takes effect as the NetKAN bot is currently experiencing some downtime.

PR #7575, opened 2019-12-16
Remove unnecessary author arrays and tag some modules

HebaruSan:
I wanted to check the outcome of KSP-CKAN/CKAN#2942, but my search of CKAN-meta's commits was mildly obstructed by mods with author arrays with only a single element. Now they're converted to strings. The ones that didn't have tags now do, and a few more modules are also tagged to fill up the PR, which makes this a successor to #7570.


HebaruSan:
#rebuild

PR #7577, opened 2019-12-20
Tag twenty-fifth group of mods

HebaruSan:
Successor to #7576.

ckan compat add 1.7


HebaruSan:
Command to print untagged modules sorted by number of downloads, from NetKAN/NetKAN folder:

```
wget -q -O - https://github.com/KSP-CKAN/CKAN-meta/raw/master/download_counts.json | jq -rMS 'to_entries | map([.value, .key] | @tsv) | .[]' | sort -n | while read LINE ; do ID=$(echo $LINE|cut -d' ' -f 2); if ! grep -q '"tags"' $ID.netkan ; then echo "$LINE" ; fi; done
```

HebaruSan:
#rebuild

PR #7604, opened 2019-12-26
Add TelemachusForAll

CraigCottingham:


CraigCottingham:
I fixed the syntax error which failed the NetKAN build check, but I don't see a way to rerun the check.

HebaruSan:
#rebuild

HebaruSan:
Thanks for the submission! It may be a while before it appears in CKAN because indexing bot appears to be stopped, but once that's resolved your mod should be OK.

PR #7608, opened 2019-12-31
New RPM fork

HebaruSan:
## Problem

~~We're creating RasterPropMonitor-Legacy as part of a plan to fix an issue with how RPM is organized. The problem is this folder:~~

- `JSI/RPMPodPatches/BasicMFD`

This folder contains the multi-function display prop. But props should be in `JSI/RasterPropMonitor/Library/Props`.

The RasterPropMonitor-Core module is supposed to include the plugin and props, while RasterPropMonitor installs the IVAs. But RasterPropMonitor-Core only installs the `JSI/RasterPropMonitor` directory, so the MFD prop is in RasterPropMonitor instead.

The B9 and Deltaglider modules use the current paths of the MFD files directly, so the files cannot simply be moved without breaking these mods.

## Changes

Now the `JSI/RPMPodPatches/BasicMFD` folder is installed by RasterPropMonitor-Core and excluded from RasterPropMonitor.

ckan compat add 1.8

### DIFF
--- a/NetKAN/RasterPropMonitor-Core.netkan
+++ b/NetKAN/RasterPropMonitor-Core.netkan
@@ -2,28 +2,30 @@
     "spec_version"   : "v1.4",
     "identifier"     : "RasterPropMonitor-Core",
     "name"           : "RasterPropMonitor Core",
-    "abstract"       : "RasterPropMonitor is a plugin and toolkit that provides functional props within your IVA.",
+    "abstract"       : "Plugin and props for use in IVAs",
     "comment"        : "This package contains the plugin, and the standard parts library.",
-    "author"         : [ "MOARdV", "Mihara" ],
-    "$kref"          : "#/ckan/github/Mihara/RasterPropMonitor",
+    "author"         : [ "Mihara", "MOARdV", "JonnyOThan" ],
+    "$kref"          : "#/ckan/github/JonnyOThan/RasterPropMonitor",
     "$vref"          : "#/ckan/ksp-avc",
     "x_netkan_epoch" : "1",
     "license"        : "GPL-3.0",
+    "resources": {
+        "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/190737-*",
+        "repository": "https://github.com/JonnyOThan/RasterPropMonitor/releases"
+    },
     "tags": [
         "plugin",
         "crewed"
     ],
-    "depends" : [
-        { "name" : "ModuleManager" }
+    "depends": [
+        { "name": "ModuleManager" }
     ],
-    "install" : [
-        {
-            "find"       : "JSI/RasterPropMonitor",
-            "install_to" : "GameData/JSI"
-        }
-    ],
-    "resources": {
-        "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/105821-14x-rasterpropmonitor-still-putting-the-a-in-iva-v0300-14-march-2018/",
-        "repository": "https://github.com/Mihara/RasterPropMonitor"
-    }
+    "install": [ {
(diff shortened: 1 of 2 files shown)


HebaruSan:
#rebuild

HebaruSan:
#rebuild

JonnyOThan:
Ack..if we can’t update B9’s dependencies directly then I wonder if these legacy files should just be included in RasterPropMonitor-core for perpetuity.  I’ll still move the prop definitions but can leave the deprecated txt files where they are.  Saves one extra confusing package in the database.  Thoughts?

NetKAN/RasterPropMonitor-Core.netkan
```diff
@@ -18,11 +18,13 @@
...
-        { "name": "ModuleManager"            },
-        { "name": "RasterPropMonitor-Legacy" }
+        { "name": "ModuleManager" }
     ],
     "install": [ {
         "find"      : "JSI/RasterPropMonitor",
         "install_to": "GameData/JSI"
+    }, {
+        "find":       "JSI/RPMPodPatches/BasicMFD",
+        "install_to": "GameData/JSI/RPMPodPatches"
```
JonnyOThan: Is it possible to leave a comment here explaining why this is included?

Do you also need to remove this folder from the main RasterPropMonitor package?

HebaruSan: > Is it possible to leave a comment here explaining why this is included?

Good idea, added.

> Do you also need to remove this folder from the main RasterPropMonitor package?

Yes, that's what the `"filter"` property does, `BasicMFD` will be excluded when installing RasterPropMonitor:

https://github.com/KSP-CKAN/NetKAN/blob/e091ca5cf06c16a5a1fc93e1a3225824aa32567b/NetKAN/RasterPropMonitor.netkan#L46

JonnyOThan: Oops, missed that one. 
Ok, looks good to me.  Ship it!

HebaruSan:
#rebuild

HebaruSan:
```
Oh no! We tried to overwrite a file owned by another mod!

If this problem re-occurs, then it maybe a packaging bug.
Please report it at:

https://github.com/KSP-CKAN/NetKAN/issues/new

Please including the following information in your report:

File           : GameData/JSI/LICENSE.md
Installing Mod : RasterPropMonitor 1:v0.31.2
Owning Mod     : JSIPartUtilities
CKAN Version   : v1.26.7+2a7ae1da9807

Your GameData has been returned to its original state.
```

JSIPartUtilities is a dependency of VesselViewerContinued, which is recommended by ALCORIVAPatch. I'll remove that from this PR.

JonnyOThan:
> ```
> Oh no! We tried to overwrite a file owned by another mod!
> Please try a `ckan update` and try again.
> 
> If this problem re-occurs, then it maybe a packaging bug.
> Please report it at:
> 
> https://github.com/KSP-CKAN/NetKAN/issues/new
> 
> Please including the following information in your report:
> 
> File           : GameData/JSI/LICENSE.md
> Installing Mod : RasterPropMonitor 1:v0.31.2
> Owning Mod     : JSIPartUtilities
> CKAN Version   : v1.26.7+2a7ae1da9807
> 
> Your GameData has been returned to its original state.
> ```
> 
> 
> JSIPartUtilities is a dependency of VesselViewerContinued, which is recommended by ALCORIVAPatch. I'll remove that from this PR.

Ack!  I had JSIPartUtilities installed as well, and included this file by accident.  What’s the best remedy?

HebaruSan:
> included this file by accident. What’s the best remedy?

We can filter the extra files out. Is it LICENSE.md, README.md, and the Agencies folder?

PR #7610, opened 2020-01-03
Update RPM depends for ChakaMonkeyExplorationSystems

HebaruSan:
Spiritual successor to KSP-CKAN/CKAN-meta#1738.

ChakaMonkeyExplorationSystems currently depends on RasterPropMonitor-Core, however it uses RasterPropMonitorBasicMFD internally which is only provided by the full RasterPropMonitor module. Now it depends on that. Will update the historical versions afterwards.

ckan compat add 1.7


HebaruSan:
#rebuild

PR #7611, opened 2020-01-04
Ksts

linuxgurugamer:
new maintainer


HebaruSan:
Please remember not to delete tags when adopting new mods.

https://github.com/KSP-CKAN/CKAN/wiki/Suggested-Tags

PR #7619, opened 2020-01-08
Harbor

linuxgurugamer:
Added conflicts JNSQ

Chatted with @JadeOfMaar first to get permission


HebaruSan:
Thanks for taking care of this. I thought about doing it, but opted to go with just what was mentioned in the forum thread.

PR #7634, opened 2020-01-16
Fix SimpleLogistics authors

HebaruSan:
#7631 set this module's author to a string containing a comma. Now it's an array.


HebaruSan:
Added to #7632.

PR #7637, opened 2020-01-18
Add Memorial Wall

severedsolo:


HebaruSan:
> 2 hours ago, severedsolo said:
> Memorial Wall 2.0.0.1 Released
> 
> Updated for Flight Tracker 3.0 (Don't try and use 2.0 it will break)

Should we make this [dependency require version 3.0 or greater](https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#relationships)?

```json
    "depends": [
        {
            "name": "FlightTracker",
            "min_version": "3.0"
        }
    ]
```

From the forum thread, it sounds like that would be important.

severedsolo:
You can do - 3.0 has already been indexed anyway, and this build is specifically marked 1.8.x only so I'm *assuming* CKAN would automatically only index 2.0.0.1 and auto install FlightTracker 3.0 anyway?

It doesn't hurt to be safe though.

HebaruSan:

PR #7638, opened 2020-01-18
Fix ManeuverD version regex

HebaruSan:
## Problem

@sarbian just pushed this module to master without a pull request, and thus with no validation. It has this inflation error:

```
Could not match version ManeuverD-1.0.0.0.zip with find pattern ^ManeuverD-(?<version>\d+\.\d+\.\d+)\.zip$
```

http://status.ksp-ckan.space/

## Cause

The version has four pieces and the regex is only matching three.

## Changes

Now it will match the fourth piece.

### DIFF
--- a/NetKAN/ManeuverD.netkan
+++ b/NetKAN/ManeuverD.netkan
@@ -8,7 +8,7 @@
     "x_netkan_jenkins": {
         "use_filename_version": true
     },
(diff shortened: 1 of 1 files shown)


sarbian:
Crap, sorry about that🙇

HebaruSan:
No problem, easy to fix. But this *was* meant to be a subtle advertisement for the advantages of pull requests; if you submit one, validation scripts will be run on the metadata and we can inspect the output to make sure everything is OK before merging.

PR #7639, opened 2020-01-19
De-meta BeyondHome

HebaruSan:
This mod's metanetkan was just deleted in https://github.com/Gameslinx/BeyondHomePlanetMod/commit/0f3949be15c71dd507ad7e0e5789b6df3e571fa3.

The author seems enthusiastic about CKAN:

thread

Now the metadata from the last copy of the metanetkan is copied into the netkan, restoring the mod. Note that the compatibility is likely to change from 1.7.3 to 1.8.1 with the next release, so we'll hold off on merging this till then to be sure.

### DIFF
--- a/NetKAN/BeyondHome.netkan
+++ b/NetKAN/BeyondHome.netkan
@@ -1,11 +1,39 @@
-{
+{ 
     "spec_version": "v1.4",
-    "identifier": "BeyondHome",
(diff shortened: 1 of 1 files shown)


HebaruSan:
With KSP-CKAN/CKAN#2970, it no longer makes sense to wait for the next release, as we'll get a staging request automatically.

PR #7640, opened 2020-01-20
Update FieldTrainingFacility.netkan

zer0Kerbal:


HebaruSan:
That `install` stanza isn't needed, the default stanza does exactly that (`find` the identifier and install it to GameData).

HebaruSan:
Dependent modules shouldn't have to track the "latest" identifier for a given mod. We can handle this inside ODFC-Refueled itself, will do that in #7642.

PR #7641, opened 2020-01-20
Update FieldTrainingLaboratory.netkan

zer0Kerbal:


HebaruSan:
Dependent modules shouldn't have to track the "latest" identifier for a given mod. We can handle this inside ODFC-Refueled itself, will do that in #7642.

PR #7643, opened 2020-01-20
Update SimpleLogistics.netkan

zer0Kerbal:


HebaruSan:
Those `resources` are already being pulled from SpaceDock. Usually it's easier to maintain them there, since you can just edit the mod and change them. Are you sure you want to have to do pull requests instead?

zer0Kerbal:
> 
> 
> Those `resources` are already being pulled from SpaceDock. Usually it's easier to maintain them there, since you can just edit the mod and change them. Are you sure you want to have to do pull requests instead?

still learning. rather it pulled auto. thank you. good looking out.

HebaruSan:
Cool, thanks for the updates! And sorry for missing that tag.

PR #7647, opened 2020-01-22
Filter JSIPartUtilities files out of RPM install

HebaruSan:
## Problem

https://github.com/KSP-CKAN/NetKAN/pull/7608#issuecomment-577273818

The latest RPM download unintentionally included files from JSIPartUtilities, so those modules now conflict if you try to install both.

## Changes

Now RasterPropMonitor won't install the files from JSIPartUtilities.


HebaruSan:
#rebuild

PR #7648, opened 2020-01-22
Make ALCORIVAPatch depend on ALCOR

HebaruSan:
## Problem

ALCORIVAPatch is an IVA for ALCOR. Not much point in installing it without ALCOR.

## Changes

Now it depends on ALCOR.
This will also allow us to test the fix in #7647.

ckan compat add 1.8


HebaruSan:
#rebuild

PR #7649, opened 2020-01-24
Remove manual staging for hard coded compatibility

HebaruSan:
These netkans had staging enabled due to hard coded compatibility info.

After KSP-CKAN/CKAN#2970, the Inflator will do this automatically, so it no longer needs to be in the metadata.


HebaruSan:
#rebuild

HebaruSan:
#rebuild

HebaruSan:
#rebuild

PR #7665, opened 2020-02-07
Updated 1.8 compatibility

RCrockford:
Updated version compatibility due to changes in RSS.

### DIFF
--- a/NetKAN/RSSTextures2048.netkan
+++ b/NetKAN/RSSTextures2048.netkan
@@ -6,7 +6,7 @@
     "abstract"        : "Textures for Real Solar Systems",
     "license"         : "CC-BY-NC-SA",
     "release_status"  : "stable",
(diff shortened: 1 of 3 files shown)


HebaruSan:
Why? The README.txt says:

```
Real Solar System Texture pack (DDS Edition) for KSP 1.3.1+ (RSS-Kopernicus)
```

There's no trace of any suggestion of a minimum of 1.8 compatibility.

RCrockford:
The lower comments in the readme have the info: 
> v18.0 for RSS 18.0

And RSS 18.0 is 1.8 only. There is a breaking change between RSS 18 and earlier versions as some of the textures changed from PNG to DDS.

RCrockford:
Please also see the updated readme.

HebaruSan:
Alright, thanks for the updates!

PR #7671, opened 2020-02-09
Create RealAntennas.netkan

DRVeyl:
First time adding a mod here.  I believe that the format is correct, and the output from running netkan against this looks reasonable to me.  I just published the official release on my repo.  Please let me know if anything needs to be changed or added!

Looking forward, when I want to release a version of the mod specifically targeting KSP 1.8, what is the right process?  In particular, after publishing a new release on my github repo, should I then go through this process to change the netkan version info?

### DIFF
--- a/NetKAN/RealAntennas.netkan
+++ b/NetKAN/RealAntennas.netkan
@@ -2,7 +2,7 @@
     "spec_version" : 1,
     "identifier"   : "RealAntennas",
     "$kref"        : "#/ckan/github/DRVeyl/RealAntennas",
-    "ksp_version"  : "1.7",
+    "$vref"        : "#/ckan/ksp-avc",
     "license"      : "CC-BY-SA-4.0",
     "resources" : {
         "homepage"     : "https://github.com/DRVeyl/RealAntennas/wiki",
--- a/NetKAN/RealAntennas.netkan
+++ b/NetKAN/RealAntennas.netkan
@@ -1,0 +1,22 @@
+{
+    "spec_version" : 1,
+    "identifier"   : "RealAntennas",
+    "$kref"        : "#/ckan/github/DRVeyl/RealAntennas",
+    "ksp_version"  : "1.7",
+    "license"      : "CC-BY-SA-4.0",
+    "version"      : "1.0",
+    "resources" : {
+        "homepage"     : "https://github.com/DRVeyl/RealAntennas/wiki",
+        "bugtracker"   : "https://github.com/DRVeyl/RealAntennas/issues",
+        "repository"   : "https://github.com/DRVeyl/RealAntennas",
+    },
+    "tags" : [
+        "comms",
+        "plugin"
+    ],
+    "conflicts" : [
+        { "name" : "RemoteTech" },
+        { "name" : "CommNetVisualization" },
+        { "name" : "CommNetConstellation" }
+    ]
+}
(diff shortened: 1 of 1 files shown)


HebaruSan:
Hi and welcome! Not too bad for a first try!
The JSON validator found one minor thing:

```
NetKAN/RealAntennas.netkan:12:4: Error: Strict JSON does not allow a final comma in an object (dictionary) literal
   |  At line 12, column 4, offset 451
   |  Object started at line 8, column 18, offset 232
NetKAN/RealAntennas.netkan: has errors
```

And the `version` property should be omitted so it can be pulled from GitHub as needed automatically for each release. I would also add `"parts"` to the tags.

The easiest way to manage game version compatibility is with a [.version file](http://ksp.cybutek.net/miniavc/Documents/README.htm) in the download. Netkan will parse it and translate it into CKAN properties with no per-release modifications to the .netkan file, if we add this:

```json
    "$vref": "#/ckan/ksp-avc",
```

Short of that, new releases of mods that have hard coded compatibility info will be staged into a pull request instead of indexed automatically, which allows us to check that the info is still correct. As long as the correct version info is on GitHub or your forum thread somewhere, we can update it before it goes out.

DRVeyl:
Thanks!

Mea culpa on the JSON.
Hadn't added parts tag since I'm not really adding any, just patching existing ones, but I defer to you here.

I will create a RealAntennas.version file based on that example shortly.  What is the convention for: should the .version file be placed inside Gamedata\RealAntennas folder, or be top-level in the .zip package?

HebaruSan:
> Hadn't added parts tag since I'm not really adding any, just patching existing ones

Oh, oops! I made uncalled for assumptions when I saw a Parts folder. Good thing I looked at those, because we need to add a ModuleManager dependency to make sure they always work.

> should the .version file be placed inside Gamedata\RealAntennas folder, or be top-level in the .zip package?

CKAN will find the .version file either way. If you want users to receive notifications from KSP-AVC at run time, it should go in the mod's folder.

DRVeyl:
Added the .version file to the folder in the .zip and updated the release on my github.
Removed the KSP_VERSION block in favor of KSP_VERSION_MIN & KSP_VERSION_MAX in the .version file.
You still need to add $vref?

HebaruSan:
Yup, added it just then, looks good.

One last tip about .version files, the `"URL"`property should point to an online copy _of the version file_, which KSP-AVC and CKAN will use to check for updates. Assuming you add it to `GameData/RealAntennas`, this would be:

https://github.com/DRVeyl/RealAntennas/raw/master/GameData/RealAntennas/RealAntennas.version

This is useful in two specific circumstances:

- If you put out a new release, users of KSP-AVC will receive notifications that it is available as soon as you update the online version file
- If you later want to update the compatibility of a release, you can update the online version file, and both KSP-AVC and CKAN will check and respect the new setting (e.g., mods that were originally released for KSP 1.4 commonly used this feature to broaden their version range to 1.4–1.7)

DRVeyl:
> One last tip about .version files, the "URL"property should point to an online copy of the version file, which KSP-AVC and CKAN will use to check for updates. Assuming you add it to GameData/RealAntennas, this would be:

>https://github.com/DRVeyl/RealAntennas/raw/master/GameData/RealAntennas/RealAntennas.version

That certainly makes more sense.
Adjusted, pushed into the repo and again updated the release .zip.

HebaruSan:
#rebuild

HebaruSan:
Looks great! I'm merging this. Thanks for the submission and your patience! :+1:

DRVeyl:
Patience was hardly required.  Many thanks for your help!

PR #7674, opened 2020-02-10
Ose

linuxgurugamer:


linuxgurugamer:
If you want, you can download it from here:
https://github.com/linuxgurugamer/SoundtrackEditorForked/blob/master/Release/GameData/SoundtrackEditorForked/PluginData/Images/sted.png

Put it into the PluginData folder (you will have to make that) right next to the Plugins folder

PR #7675, opened 2020-02-10
Fix NSSC install path

HebaruSan:
ZIP file used to contain NSSC, now NotSoSimpleConstruction

### DIFF
--- a/NetKAN/NSSC.netkan
+++ b/NetKAN/NSSC.netkan
@@ -11,11 +11,26 @@
     ],
     "depends": [
         { "name": "ModuleManager"      },
-        { "name": "SimpleConstruction" },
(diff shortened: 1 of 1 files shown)


HebaruSan:
Hmm, the relationships listed on SpaceDock are fairly extensively changed. I guess they should be updated.

HebaruSan:
```
* NSSC 1.0.2.2 conflicts with SimpleConstruction 1:4.0.2.3
```

... no it doesn't???

HebaruSan:
Ahh, it's the ExtraPlanetaryLaunchpads conflict. SimpleConstruction provides ExtraPlanetaryLaunchpads, and this mod conflicts with it.

PR #7682, opened 2020-02-11
Add Fan blade AOA and rotor torque control from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of CarnationRED, to add [Fan blade AOA and rotor torque control](https://spacedock.info/mod/2327/Fan%20blade%20AOA%20and%20rotor%20torque%20control) to CKAN.

Please direct questions about this pull request to [CarnationRED](https://spacedock.info/profile/CarnationRED).

Mod details:
    name = /mod/2327/Fan%20blade%20AOA%20and%20rotor%20torque%20control
    author = CarnationRED
    abstract =   The mod enables you to control AOA(angle of attack) of fan blades in flight, as well as automatically limit rotor's output to save electric charge
    license = MIT
    Homepage = 
    description =
Hi, the mod is "FanAOAnRotorTorque", which enables you to control AOA(angle of attack) of fan blades in flight, as well as automatically limit rotor's output to save electric charge.

NO MORE annoying manual adjusting fan blade deploy angles

BETTER energy efficiency for rotors

**Usage: press "=" to activate, details see video below**

<iframe width="1542" height="807" src="https://www.youtube.com/embed/cRfwq_FndNE" frameborder="0" allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>

PR #7685, opened 2020-02-11
Moved FanAOAnRotorTorque.netkan

CarnationRED:
I followed the output and moved netkan file in to netkan folder


HebaruSan:
Duplicate of #7682, except with worse tags

PR #7687, opened 2020-02-11
Clean up resource properties

DasSkelett:
I plan a feature in the client to display more resources in the mod info panel.
I wanted to see which resource types are actually used in the wild, and saw that there are `x_curse` and `x_preview` additional to `curse` and `x_screenshot`.
`curse` is proper part of the spec, `x_screenshot` not entirely (it's not mentioned in the list of allowed resources, but occurs in other places), but widely used. The spec could also need an update.

CKAN-meta has more to offer, a PR for it will follow.

## Changes
I replaced every `x_curse` with `curse` and `x_preview` with `x_screenshot`.

before:
```
843 "homepage"
245 "repository"
59  "bugtracker"
17  "x_curse"
17  "kerbalstuff"
11  "x_screenshot"
9   "manual"
8   "spacedock"
8   "license"
3   "x_preview"
```
after:
```
843 "homepage"
245 "repository"
59  "bugtracker"
17  "curse"
17  "kerbalstuff"
14  "x_screenshot"
9   "manual"
8   "license"
8   "spacedock"
```

When writing this description, I realized that I'll probably have to update some spec versions...

### DIFF
--- a/NetKAN/AlternateResourcePanel.netkan
+++ b/NetKAN/AlternateResourcePanel.netkan
@@ -1,5 +1,5 @@
 {
-    "spec_version"   : "v1.4",
+    "spec_version"   : "v1.20",
     "identifier"     : "AlternateResourcePanel",
     "name"           : "Alternate Resource Panel",
     "abstract"       : "An alternate view of vessel resources plugin for Kerbal Space Program",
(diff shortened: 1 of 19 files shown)


DasSkelett:
Hmm, we don't have validation checks in netkan for checking the `spec_version` needed for the different resources, presumably because they have no special handling, not even in the client aside from displaying the `homepage` and `repository` links if available, and don't break anything.

@HebaruSan should I still update the spec version to v1.20 for the `curse` netkans? I lean towards doing so.

HebaruSan:
@DasSkelett, I would leave that up to your discretion, it will work fine either way.

DasSkelett:
Went with updating the `spec_version`, I think it's the cleaner way.

PR #7692, opened 2020-02-13
NEOS - update author information.

mwerle:


HebaruSan:
Current value pulled from GitHub forks:

https://github.com/KSP-CKAN/CKAN-meta/blob/ed44a58d50cf84b95cb77ce237bfd31a6ee17b5c/NehemiahEngineeringOrbitalScience/NehemiahEngineeringOrbitalScience-0.8.2.ckan#L7-L10

```json
    "author": [
        "N3h3miah",
        "mwerle"
    ],
```

mwerle:
Why do some mods have metadata in the CKAN repo, and others in the Netkan repo?  Or is that value auto-generated?

Regardless, "mwerle" is incorrect... where/how should that be edited/overridden?

HebaruSan:
> Regardless, "mwerle" is incorrect... where/how should that be edited/overridden?

It's just your GitHub user name.

Anyway, I wanted to check whether we needed to update other mods. Doesn't look like it, and these changes are fine. Thanks for the update.

mwerle:
Ah ok. Thanks 😊

PR #7694, opened 2020-02-13
SmokeScreen - set compatibility with 1.8 and 1.9

sarbian:
SmokeScreen is compatible with both KSP version 1.8 and 1.9


HebaruSan:
Looks great, thanks for the update.

PR #7701, opened 2020-02-14
Update AVP textures to 1.8 compat

HebaruSan:
See KSP-CKAN/CKAN-meta#1812 and KSP-CKAN/CKAN-meta#1813, these modules have a min game version of 1.2, but the forum thread says 1.8. Now the netkans are updated to match.


HebaruSan:
The 8k pack isn't updated yet. It's not clear yet whether it will be, but if it is, we can update the netkan at that time.

PR #7702, opened 2020-02-14
MechJeb2 - Add ModuleManager as a dependency

sarbian:


HebaruSan:
Confirmed needed for `MechJebNoCommandPod.cfg` file.

PR #7724, opened 2020-02-20
Add ToolbarController and CTB dependency to UNDOCKINATOR

henrybauer:
Hard dependency added in Undockinator 1.1.  Previous versions don't need it.

ckan compat add 1.8

### DIFF
--- a/NetKAN/UNDOCKINATOR.netkan
+++ b/NetKAN/UNDOCKINATOR.netkan
@@ -4,6 +4,17 @@
     "$kref":        "#/ckan/spacedock/984",
     "$vref":        "#/ckan/ksp-avc",
     "license":      "MIT",
+    "x_netkan_override":[
+       {
+          "version":">=1.1",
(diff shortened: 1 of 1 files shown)


DasSkelett:
Changes to the NetKAN file will only affect the currently latest indexed version and all future ones.
This means that we don't need the override if we update the NetKAN just after you releases the new version, and can use a simple `depends`.

henrybauer:
oh cool, I changed it to just "depends" :)

DasSkelett:
The check fails because ToolbarController and ClickThroughBlocker are not released for 1.9 yet, but AFAIK they work fine. User who want to install the new UNDOCKINATOR version will have to mark 1.8 as compatible to be able to install it for now, but knowing LGG the mods will get updated soon.

DasSkelett:
#rebuild

DasSkelett:
Great, thanks for the update!

PR #7734, opened 2020-02-25
Git4

linuxgurugamer:
Updated all netkans for LGG's mods to use github as the source, this to work around the issue with CKAN and Spacedock not honoring the .version file as expected


HebaruSan:
@linuxgurugamer, these changes erroneously deleted the ClickThroughBlocker and ToolbarController dependencies from BetterTimeWarpCont (added in #6361). See #7849.

linuxgurugamer:
Thanks for fixing.  I stopped doing the mass updates a while ago

PR #7743, opened 2020-02-28
Fix PersistentRotation install

HebaruSan:
## Problem

image

## Cause

The outer `PersistentRotation` folder is absent in the latest release.

## Change

Now we just look for `GameData/PersistentRotation`.
~~Also it has a version file, so a vref is added.~~


HebaruSan:
Can't use the version file due to syntax errors. Submitted markusa380/PersistentRotation#37 to fix.

PR #7781, opened 2020-03-21
vref for SimpleLogistics

HebaruSan:
While investigating a hash mismatch for this mod...

https://forum.kerbalspaceprogram.com/index.php?/topic/191045-181-simplelogistics-2032-deployable-serenity-2020-03-20/&do=findComment&comment=3759431

... I noticed it has a version file but no vref. Now a vref is added, which also ought to update the hash.


HebaruSan:
Hashes not updated because the original download is still in the bot's cache.
@techman83 in case you don't get the notification on Discord, we need to purge 17EC5C2F-netkan-SimpleLogistics.zip

PR #7749, opened 2020-03-02
Add Droplet from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of Velctor, to add [Droplet](https://spacedock.info/mod/2356/Droplet) to CKAN.

Please direct questions about this pull request to [Velctor](https://spacedock.info/profile/Velctor).

Mod details:
    name = /mod/2356/Droplet
    author = Velctor
    abstract = Threebody Droplet probe
    license = CC-BY-NC-SA 4.0
    Homepage = 
    description =
This is your mod listing! You can edit it as much as you like before you make it public.

To edit **this** text, you can click on the "**Edit this Mod**" button up there.

By the way, you have a lot of flexibility here. You can embed YouTube videos or screenshots. Be creative.

You can check out the SpaceDock [markdown documentation](/markdown) for tips.

Thanks for hosting your mod on SpaceDock!

PR #7784, opened 2020-03-23
Add Long March mods from SpaceDock

HebaruSan:
These mods were somehow included in the initial commits of #7749 and #7767, in what looks like a SpaceDock bug, see KSP-SpaceDock/SpaceDock#249. I removed them from those PRs so the primary mod of each could be indexed.

- https://spacedock.info/mod/2351/LongMarch-5%EF%BC%88CZ-5%EF%BC%89
- https://spacedock.info/mod/2362/LongMarch-11%EF%BC%88CZ-11%EF%BC%89

Now they have their own PR.


HebaruSan:
Both mods include these files:

- GameData/Wanhu/Wanhu_Parts/WanhuCategory.cfg
- GameData/Wanhu/Wanhu_Parts/WanhuTools.dll

And neither will work without them. So I think we will have to make them conflict with each other.

PR #3950, opened 2016-05-06
Break sunflare from scatterer, add AstronikiSunflareforScatterer

plague006:
Testing this is a pain. The easiest way I found is to plop the .ckans into a test repo. https://github.com/KSP-CKAN/CKAN-meta/tree/AstronikiSunflareforScatterer_test includes those .ckans with `Scatterer` bumped to kspver 1.1.2

PR #7794, opened 2020-03-26
Force default sunflare for BeyondHome

HebaruSan:
## Background

Scatterer overrides the sunflare based on config in the `scatterer/config/Sunflares` folder.

image

The established way to provide a custom sunflare was to replace this folder, including one or more .cfg files defining `Scatterer_sunflare` config nodes. CKAN handles this by splitting the default sunflare into its own `Scatterer-sunflare` module, which is added to the `provides` list of any modules containing their own sunflares (see #3950). The user is then prompted to choose a sunflare when installing Scatterer.

## Problem

Some mods instead choose to patch the default config nodes with ModuleManager. This only works if the default `Scatterer-sunflare` module is installed, which won't be the case if such a mod provides `Scatterer-sunflare` (see #7741).

Now the user is prompted to select a sunflare, but it will only work as intended if they choose the default sunflare.

## Changes

Now:

- `Scatterer-sunflare` provides `Scatterer-sunflare-default`
- `BeyondHome` depends on `Scatterer-sunflare-default`

This should force only the default sunflare to be installed when installing `BeyondHome`. (If we had it to do all over again from the beginning, `Scatterer-sunflare` should probably be named `Scatterer-sunflare-default` and provide `Scatterer-sunflare`, but this should work also.)

Note that the default sunflare will be installed even if the user doesn't choose to install Scatterer. This is similar to how a full sunflare config in a normal module would be installed even if the user didn't install Scatterer. This won't cause any problems, and the sunflare does not itself depend upon Scatterer.

### DIFF
--- a/NetKAN/BeyondHome.netkan
+++ b/NetKAN/BeyondHome.netkan
@@ -8,7 +8,7 @@
     "ksp_version":  "1.8",
     "license":      "restricted",
     "resources": {
-        "homepage":"https://forum.kerbalspaceprogram.com/index.php?/topic/182708-*"
+        "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/182708-*"
     },
     "tags": [
         "config",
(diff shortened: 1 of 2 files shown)


HebaruSan:
#rebuild

HebaruSan:
#rebuild

Issue #7818, opened 2020-04-04
Process Question: Updating a Mod With a New Dependency


DRVeyl:
This is addressed in the guide here (https://github.com/KSP-CKAN/CKAN/wiki/Adding-a-mod-to-the-CKAN#internal-ckan-files), but I'm a little unclear on the follow-up.

The RO team is preparing several of its mods previously released for KSP 1.7 to support KSP 1.8+. The 1.8 releases will gain a dependency on the new ROLib library, currently PR'd.  The existing 1.7 mods "must" not gain that dependency: there is no ROLib for 1.7.

I believe the correct process is to add a .ckan file to each of the upcoming releases like so:
```
{
    "depends": [
	{ "name": "ROLib" }
    ]
}
```
Once the new release of a given mod is indexed by CKAN, we should then update that mod's .netkan to add the dependency and remove this .ckan file for any future release.  Is this correct?

Is the `"spec_version": "v1.16",` line required for this patching?

HebaruSan:
Some caveats:

- An internal .ckan file can only **add** properties; it cannot override or remove properties that are set in your main netkan entry, which means you would need to delete the `depends` property from the main netkan to be able to populate it from the internal ckan file
- If you want the internal ckan to override `depends`, it would have to contain **all** of the dependencies, not just the new one

> Is the `"spec_version": "v1.16",` line required for this patching?

I think `"spec_version"` is required (`NetKAN.Model.Metadata`'s constructor throws an exception if it's missing), but it doesn't have to be `"v1.16"`, could just be `"1"` if you like, unless you use some property or feature that requires a higher spec version (e.g., `"find"` in `"install"` requires `"v1.4"` or higher; [the spec](https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md) annotates these where appropriate).

The easiest way to handle this in this case might be to just make all the updates and uploads and then let us clean it up after the fact if anything gets messed up. It's easy to revert the metadata for the 1.7 mods if they are affected accidentally, and the 1.8+ ones will be updated as soon as the netkans are changed.

DRVeyl:
What is the recommended order?

Release the new mods for 1.8.  Wait for it to get indexed [with the incorrect .netkan dependencies].  Then PR the .netkan update?  On finding a new .netkan file, is the most recent .ckan data automatically updated, or is that manually initiated?

HebaruSan:
[...]

PR #7820, opened 2020-04-05
Update ROEngines.netkan

DRVeyl:
Move .netkan information into here, instead of maintaining in external ROMods repos.

Be more consistent in where ROMods (ROCapsules, etc) stores the netkan info.
Attempt to be more nuanced in depends/recommends/suggests choices.  The hard dependency on RealismOverhaul is very blunt, but many patches here depend on RO's own patching sequence and it's not worth the effort to get away from that.

### DIFF
--- a/NetKAN/ROEngines.netkan
+++ b/NetKAN/ROEngines.netkan
@@ -1,9 +1,44 @@
 {
-    "spec_version": "v1.10",
-    "identifier": "ROEngines",
-    "$kref": "#/ckan/netkan/https://raw.githubusercontent.com/KSP-RO/ROEngines/master/ROEngines.netkan",
-    "x_netkan_license_ok": true,
-    "tags": [
-        "config"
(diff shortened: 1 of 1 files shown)


DRVeyl:
You can see the previous actual contents of the .netkan file at https://github.com/KSP-RO/ROEngines/blob/master/ROEngines.netkan.

DRVeyl:
This is the first change that applies to https://github.com/KSP-CKAN/NetKAN/issues/7818
The 1.8-supporting release that needs the new metadata is now available on GH.

HebaruSan:
Looks good, merging...

HebaruSan:
Good news, the bot did try to index the new version with the old metadata, but it failed with an error due to the change in the folder structure. So it's only going to be indexed with the new metadata once the current bot pass finishes.

http://status.ksp-ckan.space/

image

PR #7823, opened 2020-04-06
Update ROTanks.netkan

DRVeyl:
Update to go with new 1.8 ROTanks release.  Add resources, improve tags, more nuanced depends/recommends/suggests, fix old conflicts reference.


HebaruSan:
Looks fine!

PR #7838, opened 2020-04-12
Add required dependencies

henrybauer:
Now depends on ToolbarController and ClickThroughBlocker


HebaruSan:
Thanks!

For thoroughness, confirmed that it's mentioned in the GitHub release note:

image

... although that version isn't on SpaceDock yet. We might need to do some clean-up after the dust settles.

PR #7847, opened 2020-04-18
RN KK fix conflicts

raidernick:
-Make ROxxx mods that change parts in the base mods conflict. These
should never be installed at the same time as they change the part names
of the existing meshes. This will break saves and craft files. You can
use one or the other, but not both sets of mods at once. It is already
set to conflict with ROcraftFiles, I just needed to make the cahnges
here also.


HebaruSan:
Thanks for the update!

PR #7860, opened 2020-04-25
Fix install for TantaresAndBDBNowinColors

HebaruSan:
Folder structure changed.


HebaruSan:
The error is just the BDB dep.

PR #7866, opened 2020-04-26
ModularLaunchPads craft files

HebaruSan:
There are some other goodies in this ZIP file. Now we'll install the example craft files.

ckan compat add 1.9.1


HebaruSan:
#rebuild

PR #7663, opened 2020-02-06
Fix Spectra's relationships

HebaruSan:
## Problems

A number of issues have arisen with respect to Spectra:

- It depends on EVE, Scatterer, TextureReplacer, and Kopernicus, but in principle a user could choose to install without those other mods and still get some value out of it
- It provides configs for EVE and Scatterer, but this is not reflected in the metadata
- It has a functional conflict with GreenSkullRevived (both attempt to install an EVA visor for TextureReplacer), but this is not discovered until you try to install both and get an error message

See also:

- https://forum.kerbalspaceprogram.com/index.php?/topic/159443-181-no-kopernicus-spectra-v123-visual-compilation-21th-march-19/&do=findComment&comment=3725981
- https://github.com/KSP-CKAN/NetKAN/issues/7662#issuecomment-582543252

## Changes

- Now EVE, Scatterer, TextureReplacer, and Kopernicus are recommended instead of depends
- Now Spectra is marked as providing and conflicting with EnvironmentalVisualEnhancements-Config, Scatterer-config, and Scatterer-sunflare
- Now GreenSkullRevived is explicitly a conflict

Fixes #7662.

PR #7870, opened 2020-04-30
Force default Scatterer-sunflare for Spectra

HebaruSan:
## Background

In #7663, we tried to set up Spectra's relationships to reflect its Scatterer config and sunflare files. However, this was wrong, because Spectra does **not** provide a full config as in #3950, but instead patches the default config as in #7741.

In #7706 we fixed that problem for the Scatterer-config side of things.

## Problem

The sunflare relationship is still set up as if Spectra provided a full sunflare config, but instead it patches the default sunflare config, which then doesn't work.

## Changes

Now Spectra uses the approach we started in #7794 to pull in the default sunflare.

(Replaces KSP-CKAN/CKAN-meta#1905.)

### DIFF
--- a/NetKAN/Spectra.netkan
+++ b/NetKAN/Spectra.netkan
@@ -17,17 +17,16 @@
     ],
     "provides": [
         "PlanetShine-Config",
-        "EnvironmentalVisualEnhancements-Config",
-        "Scatterer-sunflare"
+        "EnvironmentalVisualEnhancements-Config"
     ],
     "conflicts": [
         { "name": "PlanetShine-Config"                     },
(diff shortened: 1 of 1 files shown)


HebaruSan:
Not a real error, will retry in a little while.

```
Running NetKAN for NetKAN/Spectra.netkan
101699 [1] WARN CKAN.Curl (null) - Curl environment not pre-initialised, performing non-threadsafe init.
101720 [1] FATAL CKAN.NetKAN.Program (null) - libcurl
```

HebaruSan:
#rebuild

PR #7876, opened 2020-05-03
Use GitHub version numbers for AstronomersVisualPack

HebaruSan:
In #6494 we stopped using this mod's GitHub version numbers because they were non-sequential, and started using the version file instead because they were forced to be more reasonable.

Now we have a release with an outdated version file, see themaster402/AstronomersVisualPack#16. Instead of being indexed on its own, 4.0.3 has replaced 4.0.2 and will need cleanup.

https://github.com/KSP-CKAN/CKAN-meta/commit/548803b3c76a0198d12b5dd2b234ccb702710ee5

With auto-epoching, we can now handle non-sequential versions more smoothly. This PR reverts to using the GitHub release version numbers.

ckan compat add 1.8


HebaruSan:
#rebuild

PR #7880, opened 2020-05-06
Add LunaMultiplayer

DasSkelett:
So I've got a request by the author of DarkMultiplayer to get LunaMultiplayer indexed on CKAN :P
Permission is given by @gavazquez / Dagger as seen in the screenshot below:

image

This needs https://github.com/LunaMultiplayer/LunaMultiplayer/pull/344 to be merged first, and a new release which includes the version file.

https://github.com/LunaMultiplayer/LunaMultiplayer/releases

### DIFF
--- a/NetKAN/LunaMultiplayer.netkan
+++ b/NetKAN/LunaMultiplayer.netkan
@@ -1,0 +1,17 @@
+{
+    "spec_version": "v1.4",
+    "identifier":   "LunaMultiplayer",
+    "$kref":        "#/ckan/github/LunaMultiplayer/LunaMultiplayer",
+    "$vref":        "#/ckan/ksp-avc",
+    "license":      "MIT",
+    "tags": [
+        "plugin"
+    ],
+    "resources": {
+        "homepage": "https://lunamultiplayer.com/",
+        "manual":   "https://github.com/LunaMultiplayer/LunaMultiplayer/wiki"
+    },
+    "conflicts": [
+        { "name":   "DarkMultiPlayer" }
+    ]
+}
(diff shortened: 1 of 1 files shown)


HebaruSan:
- The `$kref` should probably use `asset_match` to select the Release asset:
  image
- Should we call it LunaMultiplayer-Client, so users are clear that they're only getting the client piece and will need to set up a server separately?

DasSkelett:
> The `$kref` should probably use `asset_match` to select the Release asset:

Yes, totally wanted to do that, but forgot in the process of writing the netkan. Thanks, will take care of that.

> Should we call it LunaMultiplayer-Client, so users are clear that they're only getting the client piece and will need to set up a server separately?

I went with how `DarkMultiPlayer` did it. But yes, it might be better to add `-Client`. In this case we also need to overwrite the `name`:
` "name": "Luna Multiplayer",`

HebaruSan:
> I went with how `DarkMultiPlayer` did it. But yes, it might be better to add `-Client`. In this case we also need to overwrite the `name`:
> ` "name": "Luna Multiplayer",`

Oh, if we already have `DarkMultiplayer`, then maybe it would be better to leave the identifier consistent and set `"name": "Luna Multiplayer Client"`.

DasSkelett:
> Oh, if we already have `DarkMultiplayer`, then maybe it would be better to leave the identifier consistent and set `"name": "Luna Multiplayer Client"`.

Okay, let's go with this. Keeps the backend consistent, but makes the users aware of what they get.

gavazquez:
Thanks for it guys!! I will do a release tomorrow so it's indexed

DasSkelett:
So, Darklight actually tested installing DMP and LMP installed side by side, and it appears to work fine.
Since some people may want to have them both installed to join different servers (although I'd highly recommend separate instances for this), I'm going to remove that conflict since it's not necessary for any technical reasons.

HebaruSan:
#rebuild

PR #7883, opened 2020-05-07
Add TweakScale depends to InterstellarFuelSwitch-Core

HebaruSan:
InterstellarFuelSwitch bundles TweakScale and references it in code:

https://github.com/sswelm/KSP-Interstellar-Extended/blob/413b6815dfefad3023d9d3eca22da269464767db/FuelSwitch/InterstellarFuelSwitch.cs#L8

Discord user Aivech reports this exception on loading:

```
[ERR 18:40:41.076] AssemblyLoader: Exception loading 'InterstellarFuelSwitch': System.Reflection.ReflectionTypeLoadException: Exception of type 'System.Reflection.ReflectionTypeLoadException' was thrown.
  at (wrapper managed-to-native) System.Reflection.Assembly.GetTypes(System.Reflection.Assembly,bool)
  at System.Reflection.Assembly.GetTypes () [0x00000] in <ad04dee02e7e4a85a1299c7ee81c79f6>:0 
  at AssemblyLoader.LoadAssemblies () [0x000e6] in <2c9e31d65a604d1980fb0cb89728fc1e>:0 

Additional information about this exception:

 System.IO.FileNotFoundException: Could not load file or assembly 'Scale_Redist, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies.
File name: 'Scale_Redist, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'
```

image

image

Now it's added as a dependency.

Also, the CommunityResourcePack dependency is replaced with CommunityTechTree, reflecting a change that happened in IFS 3.13.0.5.

Historical versions will need updating in CKAN-meta.

### DIFF
--- a/NetKAN/InterstellarFuelSwitch-Core.netkan
+++ b/NetKAN/InterstellarFuelSwitch-Core.netkan
@@ -25,7 +25,13 @@
             "install_to" : "GameData/InterstellarFuelSwitch"
         }
     ],
-    "provides": [ "DeadskinsTextureSwitch", "DeadskinsMeshSwitch" ],
+    "provides": [
+        "DeadskinsTextureSwitch",
+        "DeadskinsMeshSwitch"
+    ],
+    "depends": [
+        { "name": "TweakScale" }
+    ],
     "recommends": [
         { "name" : "CommunityResourcePack", "min_version" : "0.7.1" },
         { "name" : "TweakScale",            "min_version" : "2.3.6" }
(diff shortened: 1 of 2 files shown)


HebaruSan:
CommunityTechTree is also bundled. I'm looking to see whether it should be a dep of the non-Core mod...

HebaruSan:
#4016:

> IFS is heavily integrated with Tweakscale and it relies on very specific functionality from Tweakscale to work correctly when installed

So that PR should have made this a `depends` instead of a `recommends`

PR #7888, opened 2020-05-09
Adding TweakScale Companion for ReStock+ into CKAN.

Lisias:


HebaruSan:
> `TweakScaleCompanion-ReStockPlus v1.1.0.0 requires RestockPlus 1.0.3 or later but it is not listed in the index, or not available for your version of KSP.`

Identifiers are case sensitive, `Restock` should be `ReStock`. I can fix this for you if you check this box in the lower right:

image

NetKAN/TweakScaleCompanion-ReStockPlus.netkan
```diff
@@ -0,0 +1,31 @@
...
+        }
+    ],
+    "resources": {
+        "homepage":   "https://forum.kerbalspaceprogram.com/index.php?/topic/192216-*",
+        "repository": "https://github.com/net-lisias-ksp/TweakScaleCompanion_ReStockPlus"
+    },
+    "depends" : [
+        { "name" : "ModuleManager", "min_version" : "4.0.3" },
+        { "name" : "TweakScale", "min_version" : "2.4.4" },
+        { "name" : "RestockPlus", "min_version" : "1.0.3" }
```
HebaruSan, line 29: > `TweakScaleCompanion-ReStockPlus v1.1.0.0 requires RestockPlus 1.0.3 or later but it is not listed in the index, or not available for your version of KSP.`

Identifiers are case sensitive, `Restock` should be `ReStock`. I can fix this for you if you check this box in the lower right:

image

NetKAN/TweakScaleCompanion-ReStockPlus.netkan
```diff
@@ -0,0 +1,31 @@
+  
+{
+    "spec_version"     : "v1.2",
+    "$kref"            : "#/ckan/spacedock/2424",
+    "$vref"            : "#/ckan/ksp-avc",
+    "identifier"       : "TweakScaleCompanion-ReStockPlus",
+    "name"             : "TweakScale Companion for the ReStockPlus AddOn",
+    "abstract"         : "Adds TweakScale /L patches for ReStockPlus.",
+    "license"          : [ "GPL-2.0", "restricted" ],
```
HebaruSan, line 9: I assume you chose GPL-2.0 and restricted because some of the contained files use one license and some use the other. That is **not** what a license array means:

https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#license

> The following are both valid, the first describing a mod released under the BSD license, the second under the _user's choice_ of BSD-2-clause or GPL-2.0 licenses.

> If different assets in the mod have different licenses, the _most restrictive_ license should be specified, which may be restricted.

So this metadata as written would mean that users can choose to use everything in the mod as GPL-2.0 or as restricted, at their option.

Lisias: And that's exactly what it means, the Add'On is double licensed - the user can choose the license he wants to adopt.

HebaruSan: OK, sounds good.

Lisias: The restricted license allows the user to embedded the Add'On on packages as long the material is not changed. It's a safe bet to be used for Authors that embed add'ons and don't use GPL as the license.

NetKAN/TweakScaleCompanion-ReStockPlus.netkan
Lisias: I found no checkbox on this page. I will fix the problem on the pull-request.

HebaruSan:
I notice now that there's a file for RestockRigidLegs. Do you want to pull that in as a dependency as well?

Lisias:
It's an optional - and I had missed that (thanks!).

Should I add it as recommended or suggested?

HebaruSan:
Sure. "recommends" will pull it in as checked by default, "suggests" will be unchecked by default (and "depends" makes it mandatory). It's up to you based on how important it is to using your mod.

Lisias:
Suggests will be. Adding it on the next commit.

HebaruSan:
Looks good! Thanks for submitting.

PR #6521, opened 2018-05-01
Add Kabrams Sun Flare's Pack for Scatterer. from SpaceDock

Space-Duck:
This pull request was automatically generated by SpaceDock on behalf of 0_0_1, to add [Kabrams Sun Flare's Pack for Scatterer.](https://spacedock.info/mod/1843/Kabrams%20Sun%20Flare%27s%20Pack%20for%20Scatterer.) to CKAN.

Mod details:
    name = /mod/1843/Kabrams%20Sun%20Flare%27s%20Pack%20for%20Scatterer.
    author = 0_0_1
    abstract = (6) High Quality flares for the Scatterer mod. Including Low, Medium, and High quality presets, for support of up to 4K screen sizes.
    license = Allows useage rights for modding, for the KSP community only.
    Homepage = 
    description = **[Step 0]** Have the Scatterer mod by BlackRock.

> You can get it from:[ https://github.com/LGhassen/Scatterer/releases]( https://github.com/LGhassen/Scatterer/releases/)

---------------------------------------------------------------------------------------------------------------

**[Step 1]** - Choose a quality:

> **High** - For 4K players, and beefy machines. (*If you get banding then use the medium one instead*).

> **Medium** - Recommended for *most* users.

> **Low** - For older machines and laptops, or people with very little RAM.

---------------------------------------------------------------------------------------------------------------

**[Step 2]** - Pick a colour.

> Look in the 'Preview Images' folder and pick one that you like.

---------------------------------------------------------------------------------------------------------------

**[Step 3]** -  Installation:

> Place the file 'Sun.CFG' and the folder 'Sun' to your ..**\GameData\scatterer\config\Sunflares\ **
> 
> *Scatter actually already comes with these two, so just overwrite, or backup first then overwrite them.*

---------------------------------------------------------------------------------------------------------------

Preview

(C) Copyright 2018 Reddit user /u/0-0-1. All rights reserved. (Unless your modding for the KSP community in which case feel free to use the included elements, and just credit as appropriate).

Please direct questions about this pull request to [0_0_1](https://spacedock.info/profile/0_0_1).

PR #7891, opened 2020-05-12
Sort and fix Spectra NetKAN Provides/conflicts/depends/suggests/rec

Avera9eJoe:
I updated Spectra used this time to fix its NetKAN. I moved minor mods from Recommends to Suggests (added suggests), removed PlanetShine-config from Provides (No longer providing), removed GreenSkullRevived and PlanetShine-config from Conflicts (no longer conflicting), and removed Scatterer-sunflare-default from Depends (no longer depends).

I hope this looks in order, the file did not have suggests before this change, I added it from scratch based on the NetKAN of other mods. The goal is for all the mods in "recommends" to be auto-ticked when you install, and the mods in "suggests" to not.

I also reworked x_netkan_override. This might be set up incorrectly, though it doesn't matter as much as the rest. The current release of Spectra (1.3) is backwards compatible all the way to KSP version 1.3. Is this set up correctly?

### DIFF
--- a/NetKAN/Spectra.netkan
+++ b/NetKAN/Spectra.netkan
@@ -19,10 +19,10 @@
         "EnvironmentalVisualEnhancements-Config"
     ],
     "conflicts": [
-        { "name": "EnvironmentalVisualEnhancements-Config" },
+        { "name": "EnvironmentalVisualEnhancements-Config" }
     ],
     "depends": [
-        { "name": "ModuleManager" },
+        { "name": "ModuleManager" }
     ],
     "suggests": [
         { "name": "DistantObject"              },
--- a/NetKAN/Spectra.netkan
+++ b/NetKAN/Spectra.netkan
@@ -22,7 +22,8 @@
         { "name": "EnvironmentalVisualEnhancements-Config" }
     ],
     "depends": [
-        { "name": "ModuleManager" }
+        { "name": "ModuleManager" },
+        { "name": "Scatterer-sunflare-default" }
     ],
     "suggests": [
         { "name": "CollisionFXReUpdated"               },
@@ -58,11 +59,11 @@
         "install_to": "GameData"
     } ],
     "x_netkan_override": [{
-        "version": "1.3",
+        "version": "v1.3",
         "delete":  [ "ksp_version" ],
         "override": {
             "ksp_version_min": "1.3",
-            "ksp_version_max": "1.9.99"
+            "ksp_version_max": "1.9"
         }
     } ]
 }
--- a/NetKAN/Spectra.netkan
+++ b/NetKAN/Spectra.netkan
@@ -16,7 +16,7 @@
         "graphics"
     ],
     "provides": [
-        "EnvironmentalVisualEnhancements-Config"
+        { "name": "EnvironmentalVisualEnhancements-Config" }
     ],
     "conflicts": [
         { "name": "EnvironmentalVisualEnhancements-Config" }
@@ -25,17 +25,27 @@
         { "name": "ModuleManager" }
     ],
     "suggests": [
-        { "name": "DistantObject"              },
-        { "name": "RealPlume-StockConfigs"     },
-        { "name": "ReStock"                    },
-        { "name": "PlanetShine-Config-Default" },
-        { "name": "SpaceplaneCorrections"      }
+        { "name": "CollisionFXReUpdated"               },
+        { "name": "CrewLight"                          },
+        { "name": "EngineLightRelit"                   },
+        { "name": "HafCoSpaceCenter"                   },
+        { "name": "IndicatorLights"                    },
+        { "name": "IndicatorLightsCommunityExtensions" },
+        { "name": "KSPRC-CityLights"                   },
+        { "name": "PlanetShine"                        },
+        { "name": "RealPlume-StockConfigs"             },
+        { "name": "ReentryParticleEffect"              },
+        { "name": "ReStock"                            },
+        { "name": "ShipEffectsContinued"               },
+        { "name": "PlanetShine-Config-Default"         },
+        { "name": "SpaceplaneCorrections"              }
     ],
     "recommends": [
         { "name": "EnvironmentalVisualEnhancements" },
         { "name": "Scatterer"                       },
+        { "name": "Kopernicus"                      },
         { "name": "TextureReplacer"                 },
-        { "name": "Kopernicus"                      }
+        { "name": "DistantObject"                   }
     ],
     "install": [ {
         "file":       "Step 2 - Spectra/GameData/TextureReplacer",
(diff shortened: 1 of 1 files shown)


HebaruSan:
For the version compatibility, would you be interested in adding a KSP-AVC .version file to this mod? Netkan (and KSP-AVC) can parse these to get the compatibility info automatically per-version, which tends to be more maintainable than tracking it in netkan overrides. If that sounds appealing we can help with figuring out the details.

HebaruSan:
Checking the latest download, there's still a dependency on Scatterer-sunflare-default in GameData/Spectra/Spectra_scatterer/Sunflares/Phi/Sun.cg:

```
@Scatterer_sunflare:AFTER[scatterer]
{
```

Avera9eJoe:
Hey! If that's a cleaner solution, that's alright with me. I'm even less familiar with KSP-AVC than than CKAN, though. I recall LinuxGuruGamer's toolbar controller/click-through-blocker removing all AVC, too I think. In honesty it might be easier to just remove the compatibility flag in total.




Sent from Outlook Mobile







On Tue, May 12, 2020 at 12:24 PM -0700, "HebaruSan" <notifications@github.com> wrote:















Checking the latest download, there's still a dependency on Scatterer-sunflare-default in GameData/Spectra/Spectra_scatterer/Sunflares/Phi/Sun.cg:
@Scatterer_sunflare:AFTER[scatterer]
{

Avera9eJoe:
It looks incompatible but it should work fine. It requires the default Sunflare to be installed and usesmodulemanager to overwrite it. The Sunflare doesn't work unlesst the default Sunflare isalso installed, same with base Scatterer.




Sent from Outlook Mobile







On Tue, May 12, 2020 at 12:27 PM -0700, "Joe Hafner" <joe@hafhead.com> wrote:










Hey! If that's a cleaner solution, that's alright with me. I'm even less familiar with KSP-AVC than than CKAN, though. I recall LinuxGuruGamer's toolbar controller/click-through-blocker removing all AVC, too I think. In honesty it might be easier to just remove the compatibility flag in total.




Sent from Outlook Mobile







On Tue, May 12, 2020 at 12:24 PM -0700, "HebaruSan" <notifications@github.com> wrote:















Checking the latest download, there's still a dependency on Scatterer-sunflare-default in GameData/Spectra/Spectra_scatterer/Sunflares/Phi/Sun.cg:
@Scatterer_sunflare:AFTER[scatterer]
{

HebaruSan:
Confirmed that the changes for PlanetShine (the cfg patches the default config) and GreenSkullRevived (no more visor in the TextureReplacer folder) look correct.

> The goal is for all the mods in "recommends" to be auto-ticked when you install, and the mods in "suggests" to not.

Yup, that's exactly how those are meant to be used! :+1:

HebaruSan:
> It requires the default Sunflare to be installed and usesmodulemanager to overwrite it. The Sunflare doesn't work unlesst the default Sunflare isalso installed, same with base Scatterer.

Yeah, that's what I was saying, it requires the default Sunflare to be installed, so that dependency should probably stay.

Avera9eJoe:
Spectra only technically requires ModuleManager since you don't have to run all of its features if you don't want. You don't have to install Scatterer or ScattererSunflare if you want to run a lighter version.




Sent from Outlook Mobile







On Tue, May 12, 2020 at 12:37 PM -0700, "HebaruSan" <notifications@github.com> wrote:
















It requires the default Sunflare to be installed and usesmodulemanager to overwrite it. The Sunflare doesn't work unlesst the default Sunflare isalso installed, same with base Scatterer.



Yeah, that's what I was saying, it requires the default Sunflare to be installed, so that dependency should probably stay.

HebaruSan:
> You don't have to install Scatterer or ScattererSunflare if you want to run a lighter version.

True for Scatterer, but there are reasons for depending on the default sunflare.

There are alternate sunflares available (see #3950 and #6521), and if Spectra doesn't depend on the default sunflare, then the user will be prompted to choose a sunflare when they install Scatterer. If they choose the wrong one, then they would end up with a sunflare that is not compatible with Spectra's sunflare patches. Keeping the depends will ensure that only the right sunflare is installed.

Avera9eJoe:
Hm, that does make sense, but if you mark it as requiring scatterer-sunflare, won't that auto-install scatterer too?

HebaruSan:
> I'm even less familiar with KSP-AVC than than CKAN, but recall LinuxGuruGamer's toolbar controller/click-through-blocker requiring zeroAVC, too though that might be unrelated.

That was specifically for MiniAVC, KSP-AVC is closely related but separate (and fortunately unaffected by that issue).

> If that's a cleaner solution, that's alright with me, though it might be simpler to just remove it entirely?

OK, here's how a version file would look based on the current metadata:

```json
{
    "NAME":     "Spectra",
    "DOWNLOAD": "https://spacedock.info/mod/1505/Spectra/download/1.3",
    "URL":      "<SEE NOTE BELOW>",
    "VERSION": {
        "MAJOR": 1,
        "MINOR": 3
    },
    "KSP_VERSION_MIN": {
        "MAJOR": 1,
        "MINOR": 3
    },
    "KSP_VERSION_MAX": {
        "MAJOR": 1,
        "MINOR": 9
    }
}
```

If you save that in a file called `Spectra.version`, include it the download, and update it when there's a new release, we can set up Netkan to pull the info automatically, no override pull request required.

The `"URL"` property is special. If you store a copy of the version file online somewhere and save its URL in the `"URL"` property, then any changes that you make to it can also apply to existing downloads without changing them. For example, if you someday discover that Spectra 1.3 works on KSP 1.10, just updating the online version file would make that change in CKAN (and KSP-AVC) as well. (This is also how KSP-AVC alerts the user to available upates.) Most mod authors use GitHub to host these files, although some use http://ksp-avc.cybutek.net/ (the workings of which I do not personally understand).

Avera9eJoe:
Nice! Yeah, I'll definitely do that. Its good to know that this is unrelated to MiniAVC too. I'm getting some things setup and I'll probably have another small commit soon.

Avera9eJoe:
Another strange instance, when I uninstall Scatterer, it seems to uninstall Spectra as well because of the dependency with Scatterer Sunflare. I'm not sure how to fix this yet.

HebaruSan:
Hmm, I tried installing Spectra with the current recommendations just now and then uninstalling Scatterer, and it didn't do that:

```
$ ckan.exe remove scatterer
About to remove:

 * scatterer 3:v0.0610
 * Scatterer Default Config 3:v0.0610

Continue? [Y/n] 
```

Is it possible you also selected to uninstall the sunflare?

Avera9eJoe:
That was the culprit, many thanks. Should I close this and start a new commit, or revert the Scatterer Sunflare changes here?

HebaruSan:
I'd say it's better to keep one pull request alive and update it until everything is in order. Go ahead and push a commit here if you're comfortable doing so, and I can help with any glitches that come up.

HebaruSan:
Looks pretty good now; I'll merge it to master, and if you need any further changes just let us know.

PR #7892, opened 2020-05-12
Create KSPRC-CityLights

Avera9eJoe:
Simple mod to add the CityLights from KSPRC to EnvironmentalVisualEnhancements. Compatible with all KSP versions, requires EnvironmentalVisualEnhancements.


KSPRC-CityLights
```diff
@@ -0,0 +1,24 @@
+{
+	"spec_version": "v1.0",
+	"identifier":   "KSPRC-CityLights",
+	"name":         "KSPRC city lights",
+	"abstract":     "Standalone city lights from KSPRC.",
+	"$kref":        "#/ckan/spacedock/690",
+	"ksp_version":  "any",
```
HebaruSan: Hmm, SpaceDock and the forum thread both specify KSP 1.1.2. Are we sure `"any"` is kosher?

image

image

Avera9eJoe: Eh... sort of sure. This will be compatible with all KSP versions that support EnvironmentalVisualEnhancements, and it first released for 1.1.2. I was under the mindset that marking it as compatible for all modes, and requiring EVE would make it only work for all EVE versions. Is that good? 
image
There is a mod in CKAN that's simply the KSPRC textures which is also marked as 'any', and I was using that as an example since they both draw from the same source.

By the way thank you so much for all your help, and I hope my edits haven't been too annoying. I'm not exactly sure how to properly add a file change to the master so I've resorted to forking it and working from there. I think I'm getting the hang of it now, though.

HebaruSan: > EnvironmentalVisualEnhancements which first released for 1.1.2

Nah, we have a version indexed for KSP 0.25-1.0.4:

- https://github.com/KSP-CKAN/CKAN-meta/blob/master/EnvironmentalVisualEnhancements/EnvironmentalVisualEnhancements-7-4.ckan

Avera9eJoe: Ah! That version is probably the original that was created by rbray before Waz took over when he left. This is not compatible for that.

HebaruSan: > probably causing you a bit of a headache

Oh no, no headache whatsoever. You're keeping my brain active, like doing crossword puzzles. :grin: 

> I should fork the master, make my edits to the files I want in my fork, and then if it looks good I click this and give it a useful title

You've already got a fork for the Spectra changes, so you can skip this part. Click this link, click the pencil icon, then make your changes and commit them:

- https://github.com/Avera9eJoe/NetKAN/blob/patch-1/NetKAN/Spectra.netkan

Avera9eJoe:
It looks like this is all set up correctly, although I can't find "KSPRC-CityLights" in the master or in the launcher. Did I miss a step?

Avera9eJoe:
How long should it take for the bot to create the CKAN-meta file?

DasSkelett:
The PR isn't merged yet, but then it happens pretty much immediately.

If I followed this correctly, I think the KSP compatibility needs an update, because it isn't compatible with the very first indexed EVE version.
Changing `"ksp_version": "any"` to `"ksp_version_min": "1.1.2"` will ensure that it can only be installed for Waz' releases.

Avera9eJoe:
Gotcha- on that account then, if I had set up this correctly and fixed the `ksp_version`, is this all done on my end and all I have to do is wait for the PR (PR = Pull Request, yes?) to merge? Are there any steps I missed?

As for the KSP compatibility, yes you would be correct on that account. The ksp_version_min should be 1.1.2. Sadly though I'm not sure how to make that edit now since this is closed. Would I press `Open` in the top left, and fix the version, and press commit/merge from there?

image

Avera9eJoe:
Alright I think that was it. I just fixed the versioning and merged a pull request. Is this all set up to be automatically added to CKAN after you review it or are there more steps?

DasSkelett:
Okay, GitHub didn't detect the merge, kinda expected that.

Closing this PR manually. This is the commit:
https://github.com/KSP-CKAN/NetKAN/commit/ece16ede198fbee9717c93c6bc2899aeec0c9db5

PR #7897, opened 2020-05-14
Update Spectra.netkan versioning

Avera9eJoe:
Removed versioning override, set mod to be compatible with all KSP versions after 1.1.2.


HebaruSan:
Since there's a version file now, let's try setting that up...

HebaruSan:
There you go. Now you can control the compatibility by editing the version file in the ZIP, no more pull requests needed. And if you want to change it for the current release, you can update http://ksp-avc.cybutek.net/version.php?id=665 .

```json
    "ksp_version_min": "1.3.0",
    "ksp_version_max": "1.9.99",
```

Avera9eJoe:
Thank you so much!

PR #7898, opened 2020-05-14
Remove BackgroundResources conflict (will be supported in Kerbalism 3.8)

gotmachine:


HebaruSan:
3.8 is released, so it's safe to merge now...

PR #7905, opened 2020-05-18
Add ModuleManager depends for mods with MM syntax cfgs

HebaruSan:
## Background

I have a new toy.

image

## Problem

It's common for mods to include .cfg files using ModuleManager syntax to make tweaks or enable themselves. Some mods even do _only_ this.

Naturally, this only works if you install ModuleManager, which means such mods should depend on ModuleManager. Many don't, but as of KSP-CKAN/CKAN#3045, we have the ability to detect this.

## Changes

Now these mods depend on ModuleManager. This should clear a bunch of warnings on the status page and ensure that these mods work correctly when installed.

### DIFF
--- a/NetKAN/AnimatedAttachment.netkan
+++ b/NetKAN/AnimatedAttachment.netkan
@@ -10,5 +10,8 @@
     "tags": [
         "plugin",
         "library"
+    ],
+    "depends": [
+        { "name": "ModuleManager" }
     ]
 }
--- a/NetKAN/AntaresCygnus.netkan
+++ b/NetKAN/AntaresCygnus.netkan
@@ -29,6 +29,9 @@
             "install_to": "Ships"
         }
     ],
+    "depends": [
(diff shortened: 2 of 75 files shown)


HebaruSan:
```
Running NetKAN for NetKAN/GrannusExpansionPack-Primary.netkan
1858 [1] WARN CKAN.Curl (null) - Curl environment not pre-initialised, performing non-threadsafe init.
1895 [1] WARN CKAN.NetKAN.Transformers.AvcTransformer (null) - An error occured fetching the remote AVC version file, ignoring: libcurl
Running NetKAN for NetKAN/HeatPump.netkan
1606 [1] WARN CKAN.Curl (null) - Curl environment not pre-initialised, performing non-threadsafe init.
1620 [1] FATAL CKAN.NetKAN.Program (null) - libcurl
```

I don't want to break this up into smaller chunks, and the bot will tell us if there's a problem. Merging...

PR #7907, opened 2020-05-18
Add ModuleManager depends for third pass of mods using MM syntax

HebaruSan:
Next round of #7906. These mods install cfg files with ModuleManager syntax, so now they depend on ModuleManager.

This group is approximately the first half of the ones caught by the SpaceDock-only pass.

PR #7911, opened 2020-05-18
Add ModuleManager depends for fourth pass of mods using MM syntax

HebaruSan:
Next pass of #7907. These mods install cfg files with ModuleManager syntax, so now they depend on ModuleManager. This group is approximately the second half of the ones caught by the SpaceDock-only pass.

### DIFF
--- a/NetKAN/Contares-Common.netkan
+++ b/NetKAN/Contares-Common.netkan
@@ -10,15 +10,15 @@
     "tags": [
         "agency"
     ],
-    "install": [
-        {
-            "find": "Agencies",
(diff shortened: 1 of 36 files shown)


HebaruSan:
Almost made it all the way through; BDB only goes up to 1.8.99:

```
Running ckan install -c built/TantaresAndBDBNowinColors-2.0.ckan
TantaresAndBDBNowinColors 2.0 requires BluedogDB but it is not listed in the index, or not available for your version of KSP.
If you're lucky, you can do a `ckan update` and try again.
```

Merge 'em all and let the bot sort 'em out...

PR #7914, opened 2020-05-19
Add vrefs to third group of mods with version files

HebaruSan:
Next pass of #7913.
This should be the last group without metanetkans, though there are several left that show the warning because of a version file in a mod they bundled rather than a real version file.


HebaruSan:
#rebuild

HebaruSan:
Validation script isn't starting, but the previous build succeeded and the latest change only removed SETI-Greenhouse because its version file has nonsensical values. Merging.

PR #7919, opened 2020-05-21
Fix SpaceXLaunchVehicles install and recommends

HebaruSan:
A new folder was added at some point and isn't being installed, which makes the mod not work correctly. Also the readme recommends several mods.

image

Now the install is complete and the recommended mods are updated.

### DIFF
--- a/NetKAN/SpaceXLaunchVehicles.netkan
+++ b/NetKAN/SpaceXLaunchVehicles.netkan
@@ -9,14 +9,19 @@
     "depends": [
         { "name": "ModuleManager" }
     ],
-    "install": [
-        {
-            "find": "Launchers Pack",
-            "install_to": "GameData"
-        },
-        {
-            "find": "Ships/VAB",
-            "install_to": "Ships"
-        }
-    ]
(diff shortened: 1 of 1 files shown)


HebaruSan:
Confirmed that the previous release did not have this folder, so no historical updates are needed.

DasSkelett:
We have two `Ships/VAB` folders (one contains real-sized crafts, the other one stock-sized), I think the install stanza for this should be changed to `file` instead of `find`
grafik

Would it make sense to split the mod into two mods, a rescaled and a stock-scale version?

HebaruSan:
> I think the install stanza for this should be changed to `file` instead of `find`
> Would it make sense to split the mod into two mods, a rescaled and a stock-scale version?

Makes sense to me. Do you want to start that?

DasSkelett:
> Makes sense to me. Do you want to start that?

Just opened https://github.com/KSP-CKAN/NetKAN/pull/7962 to change `find` to `file`, will extend it with a new netkan for the rescaled version.

PR #7922, opened 2020-05-22
Adding LisiasT's KSP-Recall to CKAN

Lisias:
Thank you.


HebaruSan:
The errors are here:

- https://ci.ksp-ckan.space/job/NetKAN/9669/console

Enable edits by maintainers for help.

NetKAN/KSP-Recall.netkan
```diff
@@ -0,0 +1,29 @@
...
+    "spec_version"     : "v1.2",
+    "$kref"            : "#/ckan/spacedock/2434",
+    "$vref"            : "#/ckan/ksp-avc",
+    "identifier"       : "KSP-Recall",
+    "name"             : "KSP Recall",
+    "abstract"         : "Recall for KSP blunders, mistakes and borks.",
+    "license"          : [ "GPL-2.0", "restricted" ],
+    "release_status"   : "stable",
+    "ksp_version_min"  : "1.9.0",
+    "ksp_version_max"  : "1.9.1",
```
HebaruSan, line 11:
```suggestion
```

These will be set automatically based on the version file, so they're not needed.

NetKAN/KSP-Recall.netkan
```diff
@@ -0,0 +1,29 @@
...
+            "install_to" : "GameData"
+        }
+    ],
+    "resources": {
+        "homepage":   "https://forum.kerbalspaceprogram.com/index.php?/topic/179030-*",
+        "repository": "https://github.com/net-lisias-ksp/KSP-Recall"
+    },
+    "depends" : [
+        { "name" : "ModuleManager", "min_version" : "3.1.1" }
+    ]
```
HebaruSan, line 26:
```suggestion
    ],
    "tags": [
        "plugin"
    ]
```

Not sure what this mod does based on the thread, but I see there's a plugin.

https://github.com/KSP-CKAN/CKAN/wiki/Suggested-Tags

NetKAN/KSP-Recall.netkan
```diff
@@ -0,0 +1,29 @@
...
+        "LisiasT"
+    ],
+    "install" : [
+        {
+            "file" : "GameData/999_KSP-Recall",
+            "install_to" : "GameData"
+        }
+    ],
+    "resources": {
+        "homepage":   "https://forum.kerbalspaceprogram.com/index.php?/topic/179030-*",
```
HebaruSan, line 21:
```suggestion
        "homepage":   "https://forum.kerbalspaceprogram.com/index.php?/topic/192048-*",
```

Current homepage seems to be for TweakScale.

Lisias:
Damn it, I need to remember running NETKAN before submitting.

I'm trying to find where to allow maintainers edit, but this option doesn't appears to me. Parhaps a paid only option?

In a way or another, silly mistakes easily caught by NETKAN should render a kick on my balls, otherwise I will commit this mistake again - sometimes I do personal activities late night, when I mainly work by "muscle memory" - I need to internalize the need of using netkan.exe before submitting pull requests.

It would make your life easier.

Thanks for your efforts.

Lisias:
And I just burnt my dinner again....

HebaruSan:
Well, submitting the PR runs netkan in a container anyway, so if it's easier this way, that's fine.

The last thing I'd say is that the current abstract doesn't give me a strong sense of when I would want to install this. Is it stock bug fixes a la the old StockBugFixModules? Does it help me to fix **my** mistakes and blunders while flying ships? What's the angle?

Lisias:
(I will update the Add'On description to make it clear it also fixes some things without being called by third parties, thanks for that advice - it is not clear enough at the present time.

Lisias:
How many chars is considered acceptable on the abstract?

HebaruSan:
Schema:

```json
        "abstract" : {
            "description" : "Short description of the mod",
            "type"        : "string"
        },
        "description" : {
            "description" : "Longer description of the mod",
            "type"        : "string"
        },
```

Spec:

```markdown
##### abstract

A short, one line description of the mod and what it does.
```

So there isn't a specific character limit, but it should be a quick summary that the user can browse easily.

Lisias:
Updated the abstract. What you think?

HebaruSan:
So it **is** stock bug fixes a la the old StockBugFixModules? Just trying to make sure I understand.

Lisias:
Calling them fixes is a bit too much, because most of the time I will work around the bugs - reimplementing small bits of the feature, and then intercepting KSP events to trigger my reimplementation to overwrite the mishap.

On the specific case of the Resources on the 1.9.x, I detect when the Ship is changed, identify if a part had its resources changed (using the last known status quo for that part) and if yes, update an internal copy of that ResourceList. And then overwrite that list back when the Editor resets the Resource of a part (and when cloning parts using Alt+Click)

Some add'ons may need to call the event themselves, if they do a change without triggering the OnEditorShipChanged (as TweakScale), these ones will need to call the update event manually.

But this is what is being done now. There's at least two more workarounds on the plan - assuming that I manage to implement them, of course! :)

HebaruSan:
OK, everything looks in order now! Thanks for the submission, merging...

HebaruSan:
> I'm trying to find where to allow maintainers edit, but this option doesn't appears to me. Parhaps a paid only option?

It's not paid-only, I don't pay GitHub anything and I don't think the CKAN project does either.
Here's where it appears while you're creating the pull request:

image

PR #7930, opened 2020-05-24
Update PayToPlay.netkan

DarthPointer:


DarthPointer:
Thanks for explaining the way this stuff is done. PR to the proper repo now.

PR #7931, opened 2020-05-24
Remove unnecessary ModuleManager depends

HebaruSan:
I finally got around to having a look at these.

image

This is netkan telling us that a mod depends on ModuleManager but doesn't actually need it. Upon investigating, I found that a few of these were simply using MM syntax that our regexp currently doesn't detect (future PR likely), but most really don't use MM syntax. These latter are updated accordingly.

### DIFF
--- a/NetKAN/BOPN.netkan
+++ b/NetKAN/BOPN.netkan
@@ -11,8 +11,5 @@
     "install": [ {
         "find":       "BetterOrderPartsNumbered",
         "install_to": "GameData"
-    } ],
-    "depends": [
-        { "name": "ModuleManager" }
-    ]
+    } ]
 }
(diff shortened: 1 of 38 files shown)


HebaruSan:
```
Running ckan install -c built/JDLiquidFuelCell-1.ckan
JDLiquidFuelCell 1 requires FirespitterCore but it is not listed in the index, or not available for your version of KSP.
If you're lucky, you can do a `ckan update` and try again.
Try `ckan install --no-recommends` to skip installation of recommended modules.
Or `ckan install --allow-incompatible` to ignore module compatibility.
```

They all pass netkan validation, removing a dep isn't going to break installs. Merging...

PR #7958, opened 2020-06-04
Update Pood's Skyboxes metadata

Poodmund:
Including the "any" value against ksp_version to allow them to show up against all versions of KSP. Users will be dependent on the skybox plugin mod anyway as these are purely texture files so they are KSP version agnostic.


HebaruSan:
Makes sense to me. Thanks for the update.

PR #7959, opened 2020-06-05
Add RealFuels-Stock to CKAN

ValiZockt:
Tested on 1.8.1, It should do following, install RealFuels-Stock with its dependencies MM and RF the mod provides RealPlumeConfigs & RealFuels-Engine-Configs, on install it recommends RealPlume, but conflicts with RealPlume-Stock and the old RFStockalike.

### DIFF
--- a/NetKAN/RealFuels-Stock.netkan
+++ b/NetKAN/RealFuels-Stock.netkan
@@ -3,6 +3,7 @@
     "identifier"   : "RealFuels-Stock",
     "name"         : "RealFuels-Stock",
     "abstract"     : "Adds RealFuels configs to stock and many mods' engines, keeping engine role as designed to fit with stock and stock-aligned mods.",
+    "author"       : [ "Raptor831", "ValiZockt" ],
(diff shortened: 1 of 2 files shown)


HebaruSan:
```
Validating NetKAN/RealFuels-Stock.netkan...
NetKAN/RealFuels-Stock.netkan:40:0: Error: Strict JSON does not allow a final comma in an object (dictionary) literal
   |  At line 40, column 0, offset 1096
   |  Object started at line 1, column 0, offset 0 (AT-START)
NetKAN/RealFuels-Stock.netkan: has errors
```

HebaruSan:
We generally recommend putting your game version compatibility info in the version file rather than the .netkan, so you can change it with future releases without doing another pull request here:

- https://github.com/ValiZockt/RealFuels-Stock/blob/master/GameData/RealFuels-Stock/RFStockalike.version

HebaruSan:
#rebuild

HebaruSan:

PR #7962, opened 2020-06-07
Make sure the right Ships/VAB folder is installed for SpaceXLaunchVehicles

DasSkelett:
## Problem
This mods has multiple `Ships/VAB/` folders in the zip, one in the zip root for "normal" installations, and one under `Rescaled` for real-sized crafts.
Both can potentially get matched by the `"find": "Ships/VAB"` stanza.

While the current CKAN version seems to always pick the correct one, I found out about this because I am currently using a ckan.exe compiled from the branch of https://github.com/KSP-CKAN/CKAN/pull/3064.
That one tries to install both folders, which fails because there are identically named files in them.

## Solution
Change the `"find": "Ships/VAB"` to a `"file": "Ships/VAB"` to make it unambiguous and only install the stock-sized crafts.
I think this needs a historic backport too.

### DIFF
--- a/NetKAN/SpaceXLaunchVehicles-StockSize.netkan
+++ b/NetKAN/SpaceXLaunchVehicles-StockSize.netkan
@@ -1,14 +1,15 @@
 {
-    "spec_version": "v1.4",
-    "identifier":   "SpaceXLaunchVehicles",
-    "name":         "SpaceX Launch Vehicles - Real Size",
+    "spec_version": "v1.18",
+    "identifier":   "SpaceXLaunchVehicles-StockSize",
+    "name":         "SpaceX Launch Vehicles - Stock Size",
     "$kref":        "#/ckan/spacedock/269",
     "license":      "restricted",
     "tags": [
         "parts"
     ],
+    "provides": [ "SpaceXLaunchVehicles" ],
     "conflicts": [
-        { "name": "SpaceXLaunchVehicles-StockSize" }
+        { "name": "SpaceXLaunchVehicles" }
     ],
     "depends": [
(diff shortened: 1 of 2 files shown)


DasSkelett:
Ohh, reading the README, it looks like the default is actually real-sized, and everything in `Rescaled` is actually to make it stock-sized:
> 3) Select if you want to have the parts appear at real sizes or at scaled-down stock-compatible ones:
	3.1) REAL SIZE: Put contents of "Ships/VAB" folder into your "KSPmain/Ships/VAB" folder
	3.2) STOCK SIZE: Put contents of "Rescaled/GameData" into your "KSPmain/GameData" folder;
			Put contents of "Rescaled/Ships/VAB" into your "KSPmain/Ships/VAB" folder

And from the top of `Rescaled/GameData/KK_SpX_rescale.cfg`:
> // This config rescales all KK SpaceX Pack parts to the stock sizes and system

Good I catched that...

DasSkelett:
> GameData/Kartoffelkuchen/GameData/KK_SpX_rescale.cfg

Oh interesting, ckan.exe/netkan.exe from https://github.com/KSP-CKAN/CKAN/pull/3064 automatically strip the second `GameData`, current master doesn't.

Adding an `as` should fix that.

HebaruSan:
> Oh interesting, ckan.exe/netkan.exe from [KSP-CKAN/CKAN#3064](https://github.com/KSP-CKAN/CKAN/pull/3064) automatically strip the second `GameData`, current master doesn't.

Yes, I think that was previously something that only happened for a `file` directive of `GameData` in the root of the ZIP, and that PR makes it more generic.

PR #7513, opened 2019-11-20
Freeze stale mods

HebaruSan:
## Motivation

See KSP-CKAN/NetKAN-Infra#96, we currently crawl a lot of mods that are vanishingly unlikely to ever update again. Now that KSP-CKAN/NetKAN-Infra#97 is merged and has run, we have a timestamp for every mod, and we can use this to freeze the stale ones.

## Changes

Now any mod with a `last_indexed` date earlier than 2017-01-01 is renamed to .frozen. This will take these mods out of the bot's rotation, saving assorted system resources.

Command used to generate the list:

`wget -q -O - http://status.ksp-ckan.space/status/netkan.json | jq -rS '("2017-01-01T00:00:00Z"|fromdate) as $cutoff | to_entries | map(select(.value.last_indexed and (.value.last_indexed[0:19]+"Z"|fromdate)<$cutoff) | .key) | sort_by(ascii_downcase) | .[]'`

Command used to generate the PR (from the NetKAN/NetKAN folder):

`for X in $(wget -q -O - http://status.ksp-ckan.space/status/netkan.json | jq -rS '("2017-01-01T00:00:00Z"|fromdate) as $cutoff | to_entries | map(select(.value.last_indexed and (.value.last_indexed[0:19]+"Z"|fromdate)<$cutoff) | .key) | sort_by(ascii_downcase) | .[]') ; do git mv $X.netkan $X.frozen ; done`

PR #7966, opened 2020-06-09
Unfreeze netkans frozen in error

HebaruSan:
#7960 unfroze EnduranceSpaceExplorationSystem (which was frozen by #7513 along with 325 other mods that supposedly had not had a new release since 2017-01-01) because it had a new release. After merge, I checked whether a backfill was needed since the previous release was in late 2019:

https://spacedock.info/mod/357/Endurance%20(from%20Interstellar)%20Continued...

However, that release was already indexed!

https://github.com/KSP-CKAN/CKAN-meta/commit/8409fec5e6e7eeb4befcd404b7b4fc0c41b7c41c#diff-384b2401dd8f66b3b7518cfd79648eeb

Which means that the `last_indexed` of this module should have been way too recent for the search from #7960 to flag it, so something went wrong in the preparation of that pull request.

I dug into the git history and found about 16 mods whose most recent release was within the AutoFreezer's current 1000-day window to not be frozen. This PR unfreezes them.

Commands run for future reference (the first one takes several minutes):

```bash
# For each module frozen in #7960, find the most recent commit in its CKAN-meta
# dir in which the bot added a file, and print the date and file path
for M in $(git -C ~/Downloads/KSP/NetKAN show da46067 --name-only --pretty= | sed -e 's/^NetKAN\/\|\.frozen$//g' > identifiers) ; do echo $(git -C ~/Downloads/KSP/CKAN-meta log --diff-filter=A --grep='NetKAN generated mods' -1 --pretty=format:'%cI' --name-only $M) ; done | tee commits-netkan-only 

# Sort by date (done separately so I could watch the progress of the above command)
sort -u commits-netkan-only > sorted

# Inspecting the output, 20 lines have a timestamp since 2017-09-13, extract those identifiers
tail -n 20 sorted | sed -e 's/^.* \|\/.*$//g' > not-stale

# Unfreeze them in NetKAN (4 that have already been unfrozen will error out)
for M in $(cat /tmp/ckan-investigation/mass-freeze/not-stale) ; do git mv $M.frozen $M.netkan ; done
```


HebaruSan:
- [ ] Keep tags: KHWearableKISProps
- [ ] Fix version file: AirParkContinued HullBreach ShowAllFuelsContinued
- [ ] Consider adding vref: AirParkContinued CactEyeCommunity HullBreach LightsOut
- [ ] Consider adding ModuleManager depends: CactEyeCommunity HullBreach SolarisHypernautics
- [ ] Clear bogus URLs from SpaceDock: CloakingDevice EveryS NorthKerbinDynamics 
- [ ] Consider removing vref: EngineLighting
- [ ] Consider removing ModuleManager depends: NorthKerbinDynamics

PR #7972, opened 2020-06-13
New fork for PersistentThrust

HebaruSan:
Replaces #7971.

https://spacedock.info/mod/2450/Persistent%20Thrust

image

PR #7973, opened 2020-06-13
Update homepage link for "Persistent Thrust"

Artemonim:
(Should have been done in #7972.)


HebaruSan:
Oops, good catch. In fact, now that this is on SpaceDock, we can delete this metadata and pull the value from SD so the author can control it directly...

HebaruSan:
Thanks for the fix!

PR #7980, opened 2020-06-19
Maneuver Node Splitter

linuxgurugamer:

PR #7981, opened 2020-06-19
Tags for NodeSplitter

HebaruSan:
#7980 added this mod without tags.
Now it has them.

### DIFF
--- a/NetKAN/NodeSplitter.netkan
+++ b/NetKAN/NodeSplitter.netkan
@@ -1,23 +1,20 @@
 {
-  "license": "MIT",
-  "identifier": "NodeSplitter",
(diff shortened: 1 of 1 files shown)


linuxgurugamer:
thanks.  I need t update my shell script for this
Whats the link to the suggested tags?

HebaruSan:
The tags wiki is here:

- https://github.com/KSP-CKAN/CKAN/wiki/Suggested-Tags

linuxgurugamer:
thank you

HebaruSan:
netkan.exe will actually warn about this now (with the URL!), in addition to a few other things. It can be worth checking the pull request validation output even if it succeeds, just in case.

PR #8000, opened 2020-06-28
Add SSCEP NetKAN metadata to the index

Poodmund:
I think the NetKAN file I have remotely hosted is correct. If not, please let me know here and I'll rectify it before this merge.

https://raw.githubusercontent.com/Poodmund/Stock-Scatter-Collider-Enabler-Patch/master/CKAN/SSCEP.netkan

### DIFF
--- a/NetKAN/Stock-Scatter-Collider-Enabler-Patch
+++ b/NetKAN/Stock-Scatter-Collider-Enabler-Patch
@@ -1,0 +1,5 @@
+{
+    "spec_version": "v1.4",
+    "identifier":   "SSCEP",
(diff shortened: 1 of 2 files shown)


HebaruSan:
I usually advocate for meaningful identifiers so I can remember them and tell what they are.
Any chance you could call this EnableScatterColliders?

HebaruSan:
Don't need this for a new mod. Its purpose is to work around versioning hiccups, usually with mods with lots of releases.

```json
    "x_netkan_epoch": "2",
```

Poodmund:
Changed the identifer to 'StockScatterColliderEnablerPatch'. Its important to be explicit in that it ONLY enables the collider meshes for the vanilla instances of ground scatter for the vanilla planets, hence the name.

HebaruSan:
Works now, thanks for the submission! Merging...

Poodmund:
Thanks for the help. :D

PR #8006, opened 2020-07-02
Add TestLite

HebaruSan:
I was browsing [RP-0's mod recommendations wiki], and it mentions a few mods that aren't in CKAN, such as "TestLite", a replacement for TestFlight tailored to the needs of RP-0.

- https://github.com/KSP-RO/RP-0/wiki/TestLiteVsTestFlight

Now TestLite would be available in CKAN, and RP-0 can update their metanetkan to suggest it.

~~**NOTE**: Author permission not yet acquired, submitting this mostly to check out the validation to confirm whether it will work. Please do not merge unless/until @ec429 gives us the go-ahead.~~

Once this is merged, we can start a PR to add it here:

- https://github.com/KSP-RO/RP-0/blob/master/RP-0.netkan

[RP-0's mod recommendations wiki]:  https://github.com/KSP-RO/RP-0/wiki/Recommended-Extra-Mods

### DIFF
--- a/NetKAN/TestLite.netkan
+++ b/NetKAN/TestLite.netkan
@@ -1,0 +1,27 @@
+{
+    "spec_version": "v1.4",
+    "identifier":   "TestLite",
+    "$kref":        "#/ckan/github/ec429/TestLite",
+    "ksp_version_min": "1.6.1",
+    "ksp_version_max": "1.7.3",
+    "license":      "MIT",
+    "x_netkan_github": {
+        "use_source_archive": true
+    },
+    "resources": {
+        "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/182244-*"
+    },
+    "tags": [
+        "plugin"
+    ],
+    "conflicts": [
+        { "name": "TestFlight" }
+    ],
+    "depends": [
+        { "name": "ModuleManager" },
+        { "name": "RealFuels"     }
+    ],
+    "recommends": [
+        { "name": "RealismOverhaul" }
+    ]
+}


HebaruSan:
Asked:

- https://forum.kerbalspaceprogram.com/index.php?/topic/182244-161-testlite-tf-replacement-for-ro/&do=findComment&comment=3813390

ec429:
If this works and requires no maintenance or intervention from me, then I'm fine with it.

(That caveat is because I know from past experience that I'm too lazy and will fail to maintain metadata if it's left to me.  If I _weren't_ too lazy, I'd have made a `.ckan` for it by now.)

HebaruSan:
Cool, thanks! Yeah, I don't think you'll have to do anything. I guess confirming the game version compatibility would be good; right now I have it as 1.6.1 through 1.7.3 based on discussions in the forum thread. Should it be wider/narrower?

ec429:
1.6.1-1.7.3 sounds right for the current release (0.2.2).
But I have locally a build for 1.8.1 which I've been testing and seems to work, so I really ought to get off my backside and release that.  Probably I'll release 0.2.3 for the old versions and 0.3.0 for 1.8.1; idk how netkans handle that.  Would it help if I added a `*.version` file into the GameData folder?  It's been a long time since I knew how that stuff worked...

HebaruSan:
Yeah, a version file would be great for us! It'll work fine either way, though; without a version file, when you upload a new release, the CKAN infrastructure sends us a notification and lets us do manual review to double check the compatibility before indexing. So as long as we can find the correct info on the GitHub releases or the forum thread title, we can keep it up to date.

Merging as it sounds like the compatibility is correct for now... :+1:

PR #8018, opened 2020-07-04
Fix MissionController2 install

HebaruSan:
The folder is renamed in the latest release.


HebaruSan:
Current version already has a missing version file, so might as well merge this and do clean-up if it gets fixed:

image

PR #8019, opened 2020-07-05
Freeze ConfigurableContainers-Core

allista:
The content it provided is merged into AT_Utils, so this package no longer exists.

Or should I simply remove the file?


HebaruSan:
Thanks for the update!

PR #8049, opened 2020-07-16
Add new Contares

HebaruSan:
#8048 froze Contares, which has reincarnated itself as 11 smaller mods.
We should probably check with the author that it's OK to index these. I just want to see whether there are any errors.

- https://forum.kerbalspaceprogram.com/index.php?/topic/171305-110x-csa-contares-core-2015/
- https://github.com/CONTARES/CSA_Contares_CORE
- https://spacedock.info/mod/1712/CSA%20Contares%20%27Core%27

ckan compat add 1.9 1.8

### DIFF
--- a/NetKAN/Contares-Core.netkan
+++ b/NetKAN/Contares-Core.netkan
@@ -1,0 +1,22 @@
+{
+    "spec_version": "v1.4",
+    "identifier":   "Contares-Core",
+    "$kref":        "#/ckan/spacedock/1712",
+    "license":      "CC-BY-NC-4.0",
+    "tags": [
+        "parts"
+    ],
+    "install": [ {
+        "find":       "Contares",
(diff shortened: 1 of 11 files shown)


HebaruSan:
#rebuild

HebaruSan:
#rebuild

HebaruSan:
Some weird stuff here...

- image

This makes it sounds like GitHub is the primary source for these mods, not SpaceDock. However, none of the GitHub repos have releases!

- image

So I guess users are supposed to `git clone` the repos into GameData...? Netkan can handle source archives, but only if there's a release; I don't think we have a means to just grab a snapshot of the HEAD of some repo at a particular point in time.

HebaruSan:
Looks like installation of this is still strange. Closing, will re-open if the situation clarifies.

PR #8061, opened 2020-07-23
Fix new Angel-125 installs

HebaruSan:
#8057, #8058, and #8060 indexed these mods with incorrect install paths (missing the WildBlueTools parent folder). Now they're fixed. As well the license is set to reflect that the assets are All Rights Reserved, and the relationships are updated to match the ZIPs and forum threads.


linuxgurugamer:
Whatever you did, thank you. It seems that you changed the spacing, so every line was listed as being replaced.
The comment seems to explain it all, though.

Thank you

PR #8065, opened 2020-07-26
Add vref where needed

Olympic1:


HebaruSan:
Did you check all of these to make sure it's the right version file? There are a lot of false positives in the list that have no version file of their own but bundle another mod that does have a version file.

HebaruSan:
Also a flag/ribbon pack or agency with version "any" should be left as is.

HebaruSan:
Yeah please don't mess with this.

HebaruSan:
Example, StockishProjectOrion doesn't have its own version file, but it bundles CommunityResourcePack, which does have a version file. Seeing that in the ZIP triggers the warning; that's why these are warnings, not errors, they need to be investigated.

PR #8066, opened 2020-07-26
Add/Remove ModuleManager where needed

Olympic1:


HebaruSan:
Astrogator's metadata is in an internal .ckan file. Please don't change it.

HebaruSan:
At least some of these are false positives. CEDA has a BOM in a cfg which might be throwing off the validator.

HebaruSan:
JSIAdvancedTransparentPods is also a false positive triggered by the `#` character the author used in the localization tokens.

HebaruSan:
JSIAdvancedTransparentPods ships a `.cfg.noload` file with MM syntax. Should keep MM as a dep in case the user opts in to that functionality after install.

HebaruSan:
The STM ribbon pack is a false positive because it contains a .cfg file with invalid (non-KSP) syntax.

As with #8065, these need to be investigated before changing.

PR #7700, opened 2020-02-14
Add RSSTextures16K

HebaruSan:
Fixes #7699.

PR #8070, opened 2020-07-29
Update RSSTextures16K

### LINKED ISSUE #8069: 16k RSS texture pack
With the latest RSS-Textures release we've added our own official 16k texture pack in addition to Vannadin's 16k textures. Actually the new release is already based on @Vannadin 's work which we've used with their permission and added some of our own improvements on top of that. Now the issue is that heightmap handling changed for some of the planets which makes Vannadin's current release incompatible with the latest RSS.  Additionally it looks like this pack also uses the `RSSTextures16K` identifier which we would have preferred to use for official textures. What do you think would be the best approach to go forward with from the CKAN perspective?
Also does CKAN support setting up release compatibility based on dependency versions instead of KSP versions? This might come in handy since both the RSS and RSS-textures currently only work with the latest releases of each other.

HebaruSan:
This from #7700 has been adopted by the official project. Now it's updated to pull from their repo.

Fixes #8069.

### DIFF
--- a/NetKAN/RSSTextures16K.netkan
+++ b/NetKAN/RSSTextures16K.netkan
@@ -1,9 +1,9 @@
 {
     "spec_version": "v1.4",
     "identifier":   "RSSTextures16K",
-    "name":         "Real Solar System Textures - VaNnadin's 16384 x 8192",
+    "name":         "Real Solar System Textures - 16384 x 8192",
     "abstract":     "Textures for Real Solar Systems",
-    "$kref":        "#/ckan/github/Vannadin/VaNnadin-s-16K-RSS-texture-pack",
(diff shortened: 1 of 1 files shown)


HebaruSan:
Not a real error.

```
System.IO.IOException: Disk full. Path /home/jenkins/workspace/NetKAN/dummy_ksp/GameData/RSS-Textures/PluginData/MarsHeight.dds
```

pap1723:
Do you know why this hasn't been merged? We have people accidentally downloading the wrong textures causing a lot of bugs.

HebaruSan:
@pap1723 I was trying to figure out "Also does CKAN support setting up release compatibility based on dependency versions instead of KSP versions? This might come in handy since both the RSS and RSS-textures currently only work with the latest releases of each other.", but got distracted by other things.

PR #8080, opened 2020-08-07
Opt continued

linuxgurugamer:
https://spacedock.info/mod/2494

ckan compat add 1.3


linuxgurugamer:
@HebaruSan Can you take a look at this?  I'm not sure of the syntax to include in the older version (which isn't compatible )

HebaruSan:
#rebuild

HebaruSan:
The relationship between this and OPTSpacePlaneMain is difficult to handle properly. Will try a new PR with a different approach to see how it shakes out...

PR #8081, opened 2020-08-07
New maintainer for OPTSpacePlaneMain

HebaruSan:
This indexes OPTContinued as a continuation of the original OPTSpacePlaneMain because they seem to contain the same files and the relationships are complex and hard to decipher.

Closes #8080.


linuxgurugamer:
Thank you

PR #8089, opened 2020-08-13
CrowdSourcedScience - Moved tags

Olympic1:
Moved the tags to the self-hosted netkan


HebaruSan:
Isn't there a plugin?

Olympic1:
I removed it in the last version

PR #8090, opened 2020-08-15
Updating the description of the Add'On. Stating the current need of Module Manager 4.0.2 at minimum.

Lisias:
Yep, now it work around the heading drifting of parked crafts on KSP 1.8 and newer :) 

(does not fixes the drift induced by the wheels and landing legs - yet)

### DIFF
--- a/NetKAN/KSP-Recall.netkan
+++ b/NetKAN/KSP-Recall.netkan
@@ -4,7 +4,7 @@
     "$vref"            : "#/ckan/ksp-avc",
     "identifier"       : "KSP-Recall",
     "name"             : "KSP Recall",
-    "abstract"         : "KSP Recall - a bag of tricks and stunts to work around KSP bugs while editing Resources on Editor, and more to come...",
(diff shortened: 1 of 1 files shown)


NetKAN/KSP-Recall.netkan
```diff
@@ -4,7 +4,7 @@
     "$vref"            : "#/ckan/ksp-avc",
     "identifier"       : "KSP-Recall",
     "name"             : "KSP Recall",
-    "abstract"         : "KSP Recall - a bag of tricks and stunts to work around KSP bugs while editing Resources on Editor, and more to come...",
+    "abstract"         : "KSP Recall - a bag of tricks and stunts to work around KSP bugs. Currently soves the Resource resetting on KSP 1.9.x, the heading drift on grounded crafts on KSP >= 1.8, and more to come...",
```
HebaruSan, line 7:
```suggestion
    "abstract"         : "KSP Recall - Solves the resource resetting on KSP 1.9.x, the heading drift on grounded crafts on KSP >= 1.8, and more to come...",
```

"Solves" is misspelled, and we might as well let the reader get to the helpful text as quickly as possible.

Lisias: Accepted. Thanks for the tip.

PR #8099, opened 2020-08-19
Update Kopernicus authors

prestja:
I'm sure everyone's heard of the new Kopernicus authors by now - everybody except for CKAN, that is.


DasSkelett:
Thanks for the update and good job y'all on getting Kopernicus updated for 1.9!

PR #8100, opened 2020-08-19
Update Kopernicus URL

prestja:
Sorry for not including this in #8099, totally forgot that there's a new thread.

### DIFF
--- a/NetKAN/Kopernicus.netkan
+++ b/NetKAN/Kopernicus.netkan
@@ -16,7 +16,9 @@
         "KillAshley",
         "Gravitasi",
         "KCreator",
-        "Sigma88"
+        "Sigma88",
+		"R-T-B",
+		"prestja"
     ],
     "resources": {
         "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/181547-*"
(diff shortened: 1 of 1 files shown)


DasSkelett:
Ough, neither did I catch it, thanks.
I'm wondering, should we also overwrite the `bugtracker` resource to point to your repo? It currently points to `https://github.com/Kopernicus/Kopernicus/issues`, but as of now most bug reports and discussions happen in your repo.
(I'd probably keep the `repository` as is, since that's the source of the actual build downloaded with CKAN, but I'm happy to change that too if you think it'd make sense).

prestja:
> Ough, neither did I catch it, thanks.
> I'm wondering, should we also overwrite the `bugtracker` resource to point to your repo? It currently points to `https://github.com/Kopernicus/Kopernicus/issues`, but as of now most bug reports and discussions happen in your repo.
> (I'd probably keep the `repository` as is, since that's the source of the actual build downloaded with CKAN, but I'm happy to change that too if you think it'd make sense).

Keep it pointed at Kopernicus/Kopernicus. We plan on moving all bugtracking back over there.

PR #8134, opened 2020-09-07
Try switching EnduranceSpaceExplorationSystem kref to GitHub

HebaruSan:
See JPLRepo/Endurance#37, the version file syntax error has been fixed on GitHub but not SpaceDock, due to replacing a release being more difficult on SpaceDock. This PR switches the kref so we can see how different the resulting metadata would be and whether it could work.


HebaruSan:
Looks fine. There's no existing V1.12 file to diff with because of the syntax error (duh), but it looks similar enough to V1.11.

PR #8156, opened 2020-09-20
Create Mk33.netkan

Angel-125:
Adds .netkan entry for my latest mod, Mk-33.


HebaruSan:
Thanks for the submission!

PR #8181, opened 2020-10-08
Fix AECS-Motion-Suppressor abstract

HebaruSan:
This mod's abstract has a bunch of ZERO WIDTH NO-BREAK SPACE characters in it:

screenshot

screenshot

Now they're removed.


HebaruSan:
Oops, unintentionally committed other changes. Rebase upcoming.

HebaruSan:
These were also in the SpaceDock abstract, so we'll need a CKAN-meta PR to clean up the legacy data.

PR #8191, opened 2020-10-15
Create LRKerbalism

pehvbot:
KerbalismConfig configuration for the Less Real Than Real(ism) mod.


HebaruSan:
@pehvbot  Should LessRealThanReal suggest this? It's possible to add that in the same PR.

HebaruSan:
It's just a matter of finding the file in the right spot on GitHub and editing; I just added those changes in a new commit.

HebaruSan:
OK, I think it's looking solid now. Thanks for the submission!

PR #8204, opened 2020-10-31
Update KSP version to any

Tekaoh:
This mod is just a MM patch that should be fine with any version of KSP. Setting the version to any makes it easier to find in CKAN and to install without doing a compatibility override.

@linuxgurugamer What do you think? Would this be okay?

### DIFF
--- a/NetKAN/MechJebForAll.netkan
+++ b/NetKAN/MechJebForAll.netkan
@@ -4,6 +4,7 @@
     "name":         "MechJeb and Engineer for all!",
     "$kref":        "#/ckan/spacedock/1854",
     "$vref":        "#/ckan/ksp-avc",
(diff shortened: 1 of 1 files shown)


HebaruSan:
I just once again abused my SD admin powers to update one of LGG's mods to current compatibility, since the entire last page of its forum thread is people pestering about compatibility and CKAN. Should make this PR less necessary once that change propagates into the metadata.

- https://spacedock.info/mod/1854

HebaruSan:
It's up to 1.10.1; Let's let the architecture work as designed for now.

- https://github.com/KSP-CKAN/CKAN-meta/commit/cef708e3714a07a148f230a679b3ce9f95eddfec

PR #8016, opened 2020-07-03
Mod references updated for new features

Avera9eJoe:

PR #8208, opened 2020-11-04
Install mods' TextureReplacer assets into own subdirectories

DasSkelett:
## Problem
Almost all mods using TextureReplacer are trying to put their files into `GameData/TextureReplacer`. 
This leads to installation errors when trying to install mods that have some conflicting files ([which TR can handle](https://github.com/jrodrigv/TextureReplacer/blob/master/TextureReplacer/Replacer.cs#L382-L388)), even if the rest would work perfectly fine side by side (e.g. HumanStuff and Spectra pre #8016).

However the [README](https://github.com/jrodrigv/TextureReplacer#instructions) shows it's also possible to each mod's assets into its own subdirectory, like `GameData/HumanStuff/TextureReplacer`.

## Changes
I've gone through all mods that depend on TextureReplacer, and adjusted their install stanzas to install them into their own subdirectories, if needed.

### DIFF
--- a/NetKAN/Astro-Cosmonauts.frozen
+++ b/NetKAN/Astro-Cosmonauts.frozen
@@ -9,7 +9,7 @@
     "install": [
         {
             "find":"TextureReplacer",
-            "install_to": "GameData"
+            "install_to": "GameData/Astro-Cosmonauts"
         }
     ]
 }
--- a/NetKAN/AustrianSpaceAgency.frozen
+++ b/NetKAN/AustrianSpaceAgency.frozen
@@ -21,8 +21,8 @@
         "install_to": "GameData"
     },
     {
-        "find":"Suits/AustrianSpaceAgency",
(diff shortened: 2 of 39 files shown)


HebaruSan:
I see that there's code in TR for this, but have you tested a small sample (2-3?) of these mods to make sure they work when installed in these locations? I'd hate to have 39 mods broken by something we thought was trivial.

DasSkelett:
As another data point, Poodmund also packages their skyboxes this way, so at least for skyboxes it should work:
CKAN Contents tab Source zip

I'll try out some non-skybox mods to make sure.

DasSkelett:
I've tested HumanStuff, GreenSkullRevived and Milkyway-Skybox-SpaceEngine, all of them work.
GreenSkullRevived + Milkyway-Skybox-SpaceEngine also work if installed at the same time.
TextureReplacer keeps its promises!

HebaruSan (approved):
Awesome. :tada:

PR #8219, opened 2020-11-07
Add Parallax-StockTextures and create texture dependency system

DasSkelett:
Parallax 1.1.0(-PRE) just released and includes textures for stock planets now, while previously only BeyondHome had textures for Parallax.
https://spacedock.info/mod/2539/Parallax

Parallax now depends on the virtual module `Parallax-Textures`. A dependency on Kopernicus is also added to Parallax itself now, it was previously pulled in by BeyondHome:
image

A new module, `Parallax-StockTextures`, providing `Parallax-Textures`, is introduced. It installs the `Parallax_StockTextures` folder into GameData.
BeyondHome also provides `Parallax-Textures` (moved into separate branch to reduce disk usage).

___

ckan compat add 1.9

PR #8220, opened 2020-11-09
Make BeyondHome provide Parallax-Textures

DasSkelett:
See #8219 
BeyondHome comes with its own planet textures for Parallax

ckan compat add 1.9

### DIFF
--- a/NetKAN/BeyondHome.netkan
+++ b/NetKAN/BeyondHome.netkan
@@ -36,7 +36,8 @@
         { "name": "Scatterer"                       }
     ],
     "provides": [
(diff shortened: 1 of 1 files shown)


DasSkelett:
Still not enough disk space, since we also need to download the full 1.6 GB of stock textures.
We could move the Parallax kref to GitHub, where the core and textures are split.
However writing a regex that matches the correct asset will be tricky:
grafik

We'd also have maintain the compatibility information manually. Thoughts?

HebaruSan (approved):
I think we can be reasonably confident that this will work.

PR #8261, opened 2020-12-16
New mod Notes2Log

linuxgurugamer:


HebaruSan:
BTW @linuxgurugamer, we have a new validation system set up, mostly the same as the previous one, but more integrated into GitHub:

- https://github.com/KSP-CKAN/NetKAN/pull/8261/checks?check_run_id=1560652112

It even pins the errors and warnings to the file for ease of access. In this case it caught some problems with the remote version file:

image

PR #8262, opened 2020-12-16
Manually triggerable inflate Action

HebaruSan:
## Motivation

Currently the Inflator is driven by the Scheduler and Webhooks. Sometimes it would be nice to have the option to trigger it manually, for example if a mod author wishes to upload several consecutive releases to GitHub in a short period.

## Background

I have created a simple new GitHub Action based on xKAN-meta_testing. It uses parts of NetKAN-Infra to submit one identifier for inflation:

- https://github.com/HebaruSan/inflate-netkan

## Changes

Now a new option "Inflate a chosen module" appears in the Actions list of the NetKAN repo:

image

If you click it, a "Run workflow" dropdown is available:

image

If you click that, you can type an identifier and click Run workflow:

image image

This will trigger the action, which checks out the NetKAN and CKAN-meta repos and uses them to submit a request to the queue for the Inflator to process normally.

Note that the AWS secrets must be added to the KSP-CKAN/NetKAN repo for the Action to work. I do not have them so I cannot do this or fully test it.

Fixes KSP-CKAN/NetKAN-Infra#110.

### DIFF
--- a/.github/workflows/manual_inflate.yml
+++ b/.github/workflows/manual_inflate.yml
@@ -1,0 +1,28 @@
+name: Inflate a chosen module
+on:
+    workflow_dispatch:
+        inputs:
+            identifier:
+                description: Identifier of module to inflate
+                required: true
+jobs:
+    Inflate_module:
+        runs-on: ubuntu-latest
+        steps:
+            - name: Get NetKAN repo
+              uses: actions/checkout@v2
+              with:
+                  path: NetKAN
+            - name: Get CKAN-meta repo
+              uses: actions/checkout@v2
+              with:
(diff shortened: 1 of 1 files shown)


techman83:
Did we come to a consensus here? It would be neat to offer on demand inflations, actions limit us to repo owners.

HebaruSan:
My consensus is that this provides some value as-is and can be iterated on if the `workflow_dispatch` mechanism becomes more flexible. I looked for ways to extend it and it seemed like we're boxed in.

techman83:
Ok fair enough. If it provides value, I see no reason not to merge it.

PR #8276, opened 2020-12-23
Migrate NFT repos

ChrisAdderley:
To be merged after I finish the migration for these mods.


HebaruSan:
Thanks for the updates!

PR #8277, opened 2020-12-23
Index SpaceDust

ChrisAdderley:


HebaruSan:
Would you mind putting the `tags` in this file rather than the one in your repo? That way we can do mass-updates if we need to (e.g. when we changed "manned" to "crewed", which would have been much more messy if there were instances of "manned" in metanetkans).

ChrisAdderley:
I mean, yes I can, but that does make it harder for me to manage the tags myself.

HebaruSan:
Thanks for making that change, and thanks for the submission!

PR #8279, opened 2020-12-24
Index SystemHeat packages

ChrisAdderley:
___

ckan compat add 1.10


ChrisAdderley:
If I'm understanding this right, this is not working as ModuleManager is not marked for 1.11?

HebaruSan:
Yeah basically. There's a workaround for that, I'll apply it...

HebaruSan:
@sarbian would it be OK to mark the current MM as compatible wtih KSP 1.11? We are seeing a fair number of 1.11-only mods with an MM dependency.

HebaruSan:
Thanks for the submissions! Looks really solid :100:

sarbian:
@HebaruSan Yes it is ok. I will be able to do it sometime tomorrow but feel free to do it before if you have the time.

PR #8282, opened 2020-12-25
Vref for BeamedPowerPlugin

HebaruSan:
image

Let's see how much of a difference it makes.


HebaruSan:
The diffs are the one piece of Jenkins that we still really need to pull forward. It's easy to do if we always clone CKAN-meta, but it would be nice to avoid that.

image

PR #8285, opened 2020-12-27
Updating licensing terms for TweakScale.

Lisias:
Previous versions are still licensed under the WTFPL (un)license.

### DIFF
--- a/NetKAN/TweakScale.netkan
+++ b/NetKAN/TweakScale.netkan
@@ -7,7 +7,7 @@
     "$kref"            : "#/ckan/spacedock/127",
     "$vref"            : "#/ckan/ksp-avc",
     "x_netkan_force_v" : true,
-    "license"          : "WTFPL",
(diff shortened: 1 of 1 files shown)


HebaruSan:
Just to confirm, the user can choose whether to use the mod under the terms of GPL2 or ARR?

Lisias:
Ouch, sorry for the delay!

yes, it's at user discretion. It's not ARR, the alternate license grants some rights not granted by GPL if the user don't want to cope with the demands of the GPL (with some drawbacks, but yet, it't better than not being able to use it).

It's the same on the [KSP-Recall](https://github.com/KSP-CKAN/NetKAN/commit/1cf0a59596b11ee126d9a378719c9f480eee66d4), by the way.

HebaruSan:
OK, thanks for the confirmation. I like to double check when folks submit license arrays, because sometimes what they want to express "the DLL is GPL, the art assets are ARR", which is _not_ what the spec says a license array means. In your case it makes sense.

PR #8286, opened 2020-12-27
Add recommends and supports entries for the Corvus125mTwoKerbalPod.

mwerle:


HebaruSan:
Thanks for the update.

PR #8287, opened 2020-12-27
NF updates

ChrisAdderley:
Repo transfers, netkan fixes


HebaruSan:
Thanks for another round of updates!

PR #8289, opened 2020-12-28
Add TechTreeKompacted

HebaruSan:
Has CKAN badge on SD but no PR. Author confused why not indexed yet on thread.

- https://spacedock.info/mod/2600/TechTree%20Kompacted?ga=%3CGame+3102+%27Kerbal+Space+Program%27%3E
- https://forum.kerbalspaceprogram.com/index.php?/topic/199096-111x-techtree-kompacted-13/


HebaruSan:
The instructions in the ZIP say CommunityTechTree is required, but the forum thread says that is no longer the case:

image

PR #8328, opened 2021-01-24
Update Custom Asteroids dependencies for data configs.

Starstrider42:
Old configs continue to work with CA 1.6+, while new configs (version 1.9+) require CA 1.9+ and KSP 1.10+. The `x_netkan_override` blocks ensure continuity in behavior until Custom Asteroids 1.9 and CA-OPM 1.2 are actually released (tested by running `netkan` locally).

`CustomAsteroids.netkan` itself requires no updates for the new release; its versioning will be handled through Spacedock/AVC metadata.

### DIFF
--- a/NetKAN/CustomAsteroids-Pops-OPM-Outer.netkan
+++ b/NetKAN/CustomAsteroids-Pops-OPM-Outer.netkan
@@ -55,7 +55,7 @@
                     {
                         "name": "CustomAsteroids",
                         "min_version": "v1.6.0",
-                        "max_version": "v1.99.99",
+                        "max_version": "v1.99.99"
                     }
                 ],
                 "comment": "Match to Custom Asteroids v1.9.0, see NetKAN#3841.",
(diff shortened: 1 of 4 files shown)


HebaruSan:
Compatibility updates should be made at the remote version file:

https://raw.githubusercontent.com/Starstrider42/Custom-Asteroids/master/CustomAsteroids.version

Starstrider42:
Yes, I'm still in the process of releasing Custom Asteroids itself. I was hoping this could be merged first so that there isn't an intervening period of CA 1.9 files with the wrong dependencies.

HebaruSan:
OK, let's give this a try. Thanks for the update.

Starstrider42:
FTR, there was an install problem at https://github.com/KSP-CKAN/CKAN-meta/runs/1758705734 for `Pops-OPM-Outer-v1.2.0.ckan`. I think this is because the updated config pack requires KSP 1.10+, but OPM/Kopernicus haven't been updated for that version yet, i.e., it's not a problem with the pack itself.

HebaruSan:
Yup, I concur. Usually any failure in CKAN-meta's validation is down to that sort of dependency-compatibility. Users can force-install OPM in various ways.

PR #8339, opened 2021-01-30
Allow out of order releases for Kopernicus

DasSkelett:
See https://github.com/KSP-CKAN/CKAN-meta/pull/2251, follow-up to #8338 

Since in the current version format for Kopernicus the KSP version comes before the release counter, we will always get a out-of-order auto-epoch for the assets for older KSP versions.
And apparently also for every other asset in the same release, because they share the `TransformOptions`.

We don't want that, so let's enable `x_netkan_allow_out_of_order` here.

Alternative would be swapping the release counter and the KSP version part in the version, but I think in this case it's nicer to keep the current format, which also helps users identifying the zip belonging to the version on CKAN if they want to have a look or something.

Closes https://github.com/KSP-CKAN/CKAN-meta/pull/2251
Closes https://github.com/KSP-CKAN/CKAN-meta/pull/2252

### DIFF
--- a/NetKAN/Kopernicus.netkan
+++ b/NetKAN/Kopernicus.netkan
@@ -10,6 +10,7 @@
         "find": "^(?<version>.+)$",
         "replace": "release-${version}"
     },
+    "x_netkan_allow_out_of_order": true,
     "release_status": "development",
     "license"       : "LGPL-3.0",
     "author": [


DasSkelett:
Looks like our CI currently only shows and installs one of the generated .ckans.

HebaruSan (approved):
:mountain:

DasSkelett:
...which won't be an easy fix, because it currently determines the generated file by timestamp:
https://github.com/KSP-CKAN/xKAN-meta_testing/blob/0dd3e751abce86cb02036407f5a99c4880ca069d/ckan_meta_tester/ckan_meta_tester.py#L113-L117

DasSkelett:
Commits are here:
https://github.com/KSP-CKAN/CKAN-meta/commit/7d20e85316f8999af736e9749b86ca3adfc3c25a
https://github.com/KSP-CKAN/CKAN-meta/commit/bdb22b3ed7d85be4c04fb989d6eb4e3a3b435f53

Multi-asset with version from filename works!

HebaruSan:
Great work! We have needed this for years, it's almost disorienting to finally have it!

PR #8367, opened 2021-02-17
Msb

linuxgurugamer:


HebaruSan:
https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/MSP3000.netkan

:cry:

HebaruSan:
@linuxgurugamer the remote version file doesn't work:

http://ksp.spacetux.net/avc/Material-Science-Pod

PR #8370, opened 2021-02-20
Update Parallax.netkan

Gameslinx:
Switch reference to github

### DIFF
--- a/NetKAN/Parallax-StockTextures.netkan
+++ b/NetKAN/Parallax-StockTextures.netkan
@@ -1,11 +1,14 @@
 {
     "spec_version": "v1.4",
-    "identifier": "Parallax-StockTextures",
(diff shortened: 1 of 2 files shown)


DasSkelett:
Will need an asset_match, finding a regex that works with all those naming schemes won't be easy:
image
image

HebaruSan:
FYI @Gameslinx, we've had to hard-code the game version compatibility in the netkan because there's no other source for it; I went with 1.11 based on the forum thread. This means that if you need to change it or if you release a new version with different compatibility (say, 1.12), we'll need to update the netkan to do that. A version file is often an easier way to handle all this, since the compatibility is specified within the download itself (and can be updated at a remote location you control):

- http://ksp.cybutek.net/kspavc/Documents/README.htm

PR #8371, opened 2021-02-20
Update Parallax-StockTextures.netkan

Gameslinx:
Switch reference to github instead of SpaceDock


HebaruSan:
Combined this with #8370, we can handle both files over there.

PR #8389, opened 2021-03-03
Toolbarc

linuxgurugamer:


HebaruSan:
@linuxgurugamer we now have a way to do this directly without a throw-away PR:

- https://github.com/KSP-CKAN/NetKAN/actions/workflows/manual_inflate.yml

See KSP-CKAN/NetKAN#8262.

DasSkelett:
I think this would've also been inflated via the webhook, or the 30 minute bot run at the latest.

PR #8393, opened 2021-03-05
PayToPlay: new dependency: KSP-Recall

DarthPointer:
I guess it is the best solution as since KSP 1.11 P2P suffers a KSP API bug and KSP-Recall is a "hard recommendation" to fix it.


HebaruSan:
Thanks for the update.

Issue #5295, opened 2017-03-02
New Tantares and New TantaresLV


DerpyFirework:
I've just created the NetKAN files for the new versions of Tantares and TantaresLV (attached)
I created them separate from Tantares and TantaresLV as they are completely redone from the ground up.
[NewTantares & NewTantaresLV.zip](https://github.com/KSP-CKAN/NetKAN/files/815289/NewTantares.NewTantaresLV.zip)

blackk100:
I guess this should go in a pull request.

PR #8402, opened 2021-03-09
Fix Tantares mods

DasSkelett:
## Problem
See the conversation above and below [this comment](https://forum.kerbalspaceprogram.com/index.php?/topic/73686-111x-tantares-stockalike-soyuz-and-mir-2402122021/&do=findComment&comment=3937912) on the Tantares thread.

In early 2017, the SpaceDock entries for `Tantares` and `TantaresLV` stopped receiving updates, while releases were still published in the GitHub repo.
After [an issue](https://github.com/KSP-CKAN/NetKAN/issues/5295) from a user who seemed close to the project, [new netkans have been added](https://github.com/KSP-CKAN/NetKAN/pull/5317) for those two mods, with the identifiers `NewTantares` and `NewTantaresLV`. Ideally only the kref should've been changed instead of creating two completely new modules, but well.
Shortly after the original netkans have been frozen. Some time later the SpaceDock pages started to receive updates again.

Last year we indexed `TantaresSP` after getting an [automated SpaceDock PR](https://github.com/KSP-CKAN/NetKAN/pull/8091).

The "New" prefixes in the names (and identifiers) don't reflect the official mod names, and users got confused, especially after seeing three other mods without "New" prefix, where to of them are really outdated.

The craft files for `NewTantares` and `NewTantaresLV` don't get installed by CKAN, whereas those of `TantaresSP` do.

And the compatibility data isn't fully correct either:
image

## Changes
The NetKAN files for `NewTantares` and `NewTantaresLV` are cleaned up. The `name` is now pulled form the GitHub repo, so we get rid of the "New" prefix. They'll show up as "Tantares" and "Tantares LV" in the GUI now.
The identifiers still have the "New" in them, but that's where all the releases of the mods are now. Instead they got a `"provides": "Tantares[LV]"`.

We are now also installing the crafts for `NewTantares` and `NewTantaresLV`.

The compatibility ranges have been updated to KSP 1.10.1 - 1.11 for all three mods.
Beale said that they plan to include .version files in the next release :crossed_fingers: 

`TantaresSP` has been switched over to GitHub as well, to match the other two mods. This simplifies future maintenance.

The license for `TantaresSP` has been changed to `CC-BY-NC-SA-4.0` (like both other mods), I also updated the SpaceDock page earlier, since the zip file states:
> TantaresSP is distributed under a Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International License.

## Quick Links
* https://spacedock.info/mod/174/Tantares
* https://spacedock.info/mod/176/TantaresLV
* https://spacedock.info/mod/2498/TantaresSP
* https://github.com/Tantares/Tantares
* https://github.com/Tantares/TantaresLV
* https://github.com/Tantares/TantaresSP
* [Forum Thread](https://forum.kerbalspaceprogram.com/index.php?/topic/73686-Tantares)

### DIFF
--- a/NetKAN/NewTantares.netkan
+++ b/NetKAN/NewTantares.netkan
@@ -1,18 +1,22 @@
 {
-    "spec_version" : 1,
-    "identifier"   : "NewTantares",
-    "name"         : "New Tantares",
-    "abstract"     : "Stockalike Soyuz, LK and more!",
-    "author"       : "Beale",
-    "$kref"        : "#/ckan/github/Tantares/Tantares",
-    "ksp_version"  : "1.11",
-    "license"      : "CC-BY-NC-SA-4.0",
+    "spec_version":    "v1.18",
+    "identifier":      "NewTantares",
+    "abstract":        "Stockalike Soviet Spacecraft",
+    "author":          "Beale",
+    "$kref":           "#/ckan/github/Tantares/Tantares",
+    "ksp_version_min": "1.10.1",
+    "ksp_version_max": "1.11",
+    "license":         "CC-BY-NC-SA-4.0",
     "resources": {
-        "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/73686-*"
+        "homepage":  "https://forum.kerbalspaceprogram.com/index.php?/topic/73686-Tantares",
+        "spacedock": "https://spacedock.info/mod/174/Tantares"
     },
     "tags": [
         "parts",
         "crewed"
+    ],
+    "provides": [
+        "Tantares"
     ],
     "depends": [
         { "name": "ModuleManager" }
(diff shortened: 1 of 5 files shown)


HebaruSan:
Can we prune the `v` out of the SP versions, to match up with the already indexed releases? (This would also have the advantage of adding a "Diffing TantaresSP..." section to the checks, so we can see what's different.)
image

HebaruSan (approved):
Changes look good!

Should we consider migrating the old .ckan files to the newer identifiers (in a separate CKAN-meta PR)? That way users would no longer see multiple modules even in the All filter.

DasSkelett:
> Should we consider migrating the old .ckan files to the newer identifiers (in a separate CKAN-meta PR)? That way users would no longer see multiple modules even in the All filter.

Yeah I think we can do that. If someone's still on those versions, it's probably for nostalgia's sake, and they won't need an upgrade path to the newer ones.

PR #8410, opened 2021-03-17
Create ProvenanceAerospace.netkan

DeltaDizzy:
Adding Provenance Aerospace at the request of its author, it adds stockalike Blue Origin parts.


HebaruSan:
Hi @DeltaDizzy, thanks for the submission!

Can you please help us to confirm that this was at the request of the author? It doesn't look very good if there was a misunderstanding and all we have to rely on is "But DeltaDizzy said so". If there was a public conversation about it, that would be great. I don't see anything notable on the forum thread. Thanks!

DylanSemrau:
He has my permission

HebaruSan:
That'll do, thanks!

PR #8418, opened 2021-03-19
Add KSPBurst

dkavolis:
https://github.com/KSPModdingLibs/KSPBurst

### DIFF
--- a/NetKAN/Kerbalism.netkan
+++ b/NetKAN/Kerbalism.netkan
@@ -7,8 +7,8 @@
     "$vref":        "#/ckan/ksp-avc",
     "license":      "Unlicense",
     "resources": {
-        "homepage": "https://kerbalism.github.io/Kerbalism",
-        "repository": "https://github.com/Kerbalism/Kerbalism"
+        "manual": "https://kerbalism.github.io/Kerbalism",
+        "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/190382-*"
     },
     "tags": [
         "plugin",
@@ -17,30 +17,26 @@
         "science",
         "resources"
     ],
-    "x_netkan_staging": true,
-    "x_netkan_staging_reason": "Bundles Harmony. Please check the DLL's version, and if it's 2 or later, remove the provides and add a depends on Harmony2.",
-    "provides": [
-        "Harmony1"
-    ],
     "depends": [
         { "name": "CommunityResourcePack" },
-        { "name": "Kerbalism-Config" }
+        { "name": "Kerbalism-Config"      },
+        { "name": "Harmony2"              }
     ],
     "recommends": [
         { "name": "KerbalChangelog" }
     ],
     "supports": [
-        { "name": "ConnectedLivingSpace" },
-        { "name": "CryoTanks" },
-        { "name": "KerbalAtomics" },
+        { "name": "ConnectedLivingSpace"       },
+        { "name": "CryoTanks"                  },
+        { "name": "KerbalAtomics"              },
         { "name": "KerbalPlanetaryBaseSystems" },
-        { "name": "NearFutureElectrical" },
-        { "name": "NearFutureSolar" },
-        { "name": "NearFutureSpacecraft" },
-        { "name": "SCANsat" }
+        { "name": "NearFutureElectrical"       },
+        { "name": "NearFutureSolar"            },
+        { "name": "NearFutureSpacecraft"       },
+        { "name": "SCANsat"                    }
     ],
     "install": [ {
-        "find": "Kerbalism",
+        "find":       "Kerbalism",
         "install_to": "GameData"
     } ]
 }


DasSkelett:
Hey, thanks for the submission. Can you give some more information on this mod? This seems to be a more complex topic.
From what I've heard, this tool will eventually be used by a few mods like FAR? Does this mod work completely on its own, or do users need to do and/or install anything else?

> The Burst compiler is archived to prevent KSP from trying to load its dynamic libraries. The archive is extracted the first time the mod runs to <path to KSP>/PluginData/KSPBurst@<mod version>-<archive name>.

CKAN won't remove those files when the mod is uninstalled or upgraded. Will the mod automatically clear out older compiler versions? Otherwise your disk fills up with another 700 MiB with every compiler update.
When the mod is uninstalled, it'll leave behind those 700 MiB of files. This applies to both CKAN users and manual installers. What happens if these files are still there with those from GameData having been removed? Can KSP handle this? Is there really no way this could be put into GameData instead?

> plugins_only version does not contain the compiler, use it only if KSPBurst is a hard dependency and download size is an issue.

Does this mean that users can opt for that `plugin_only` zip, and everything will still work perfectly fine, just without the performance enhancements?
Should we index this `plugins_only` zip as an alternative, so users that don't have that much disk space available or don't want to clutter their KSP installation outside of GameData have another option?

&nbsp;

The netkan currently uses a `version_from_asset`, but it doesn't look like this mod would need it, the regex only matches the big zip We should replace it with a `asset_match` and get the version from the Git release tag, which is usually safer than a filename that may change in format any time.

dkavolis:
> From what I've heard, this tool will eventually be used by a few mods like FAR?
Yeah, and RealAntennas

> Does this mod work completely on its own, or do users need to do and/or install anything else?
Yeah but it doesn't do anything on its own, it's a resource for modders who want to use Burst jobs

> CKAN won't remove those files when the mod is uninstalled or upgraded. Will the mod automatically clear out older compiler versions? Otherwise your disk fills up with another 700 MiB with every compiler update.
Every time an archive is extracted a hidden file is created with a list of files and directories extracted. Next time an archive is going to be extracted, the directory is globbed for the hidden files and everything in them is removed so there should only ever be 1 extracted compiler version. If extraction fails for whatever reason all extracted files and directories are immediately removed.

> What happens if these files are still there with those from GameData having been removed? Can KSP handle this? Is there really no way this could be put into GameData instead?
The extracted compiler remains but KSP doesn't do anything with the files. GameData doesn't work since KSP wants to load all the files with .dll extension which are then passed to the compiler. The compiler itself contains a few managed assemblies and multiple native ones.

> Does this mean that users can opt for that plugin_only zip, and everything will still work perfectly fine, just without the performance enhancements?
Yeah

> Should we index this plugins_only zip as an alternative, so users that don't have that much disk space available or don't want to clutter their KSP installation outside of GameData have another option?
That would be great if someone could help with it

> The netkan currently uses a version_from_asset, but it doesn't look like this mod would need it, the regex only matches the big zip We should replace it with a asset_match and get the version from the Git release tag, which is usually safer than a filename that may change in format any time.
Yeah, I just used what worked since I'm not that familiar with netkan spec

DasSkelett:
Thanks. Another question: there's a `bcl.exe` in the compiler zip, which seems to be important according to the README/forum post. Does this package also work on platforms other than Windows? CKAN currently doesn't support OS specific mods.

> That would be great if someone could help with it

That's what we're here for :)

dkavolis:
> Thanks. Another question: there's a bcl.exe in the compiler zip, which seems to be important according to the README/forum post. Does this package also work on platforms other than Windows? CKAN currently doesn't support OS specific mods.

It's a managed executable, it at least definitely works on Windows and Ubuntu (WSL)

DasSkelett:
> It's a managed executable, it at least definitely works on Windows and Ubuntu (WSL).

I think Windows does some magic so you can run exes inside WSL without having mono installed. Can we be sure that it also works in non-WSL environments that don't have an external Mono installed (i.e. just relying on the one KSP ships with)?

I have a system without Mono where I could test it. Does it try to run `bcl.exe` even without any mod installed that makes use of Burst? If not, is there a pre-release/nightly of FAR or RealAntennas available somewhere so that I can trigger it?

dkavolis:
> I have a system without Mono where I could test it. Does it try to run bcl.exe even without any mod installed that makes use of Burst? If not, is there a pre-release/nightly of FAR or RealAntennas available somewhere so that I can trigger it?

It just collects plugin paths and forwards them to the compiler. Alternatively, you could extract the archive and run `bcl.exe --help`

HebaruSan:
> CKAN currently doesn't support OS specific mods.

Well, _mostly_...   image

But yes, it would be good to not have Mac and Linux users surprised if something doesn't work.

dkavolis:
It will probably not run as it is right now, Unity uses `ManagedProgram` to run `bcl.exe` which executes `mono` (with `profile = null` and `setMonoEnvironmentVariables = false`)
https://github.com/Unity-Technologies/UnityCsReference/blob/master/Editor/Mono/Utils/ManagedProgram.cs

DasSkelett:
So I just tested on a system without Mono installed, and it does not work unfortunately.

```
[ERR 17:02:18.109] [KSPBurst]: Burst compiler terminated in an error
[EXC 17:02:18.268] Win32Exception: ApplicationName='/NAS/games/KubuntuGames/steamapps/common/KSP_Testing/PluginData/KSPBurst@1.5.0.0-com.unity.burst@1.5.0/package/.Runtime/bcl.exe', CommandLine='...'
Native error= Cannot find the specified file
        System.Diagnostics.Process.StartWithCreateProcess (System.Diagnostics.ProcessStartInfo startInfo) (at <376e8c39bbab4f1193a569c8dbe4305c>:0)
        System.Diagnostics.Process.Start () (at <376e8c39bbab4f1193a569c8dbe4305c>:0)
        (wrapper remoting-invoke-with-check) System.Diagnostics.Process.Start()
        System.Diagnostics.Process.Start (System.Diagnostics.ProcessStartInfo startInfo) (at <376e8c39bbab4f1193a569c8dbe4305c>:0)
        KSPBurst.KSPBurst.RunBurstCompiler (System.String directory) (at <510e80975a1d412aaad541e2c71d0e41>:0)
        KSPBurst.KSPBurst.Generate () (at <510e80975a1d412aaad541e2c71d0e41>:0)
        System.Threading.Tasks.Task`1[TResult].InnerInvoke () (at <ad04dee02e7e4a85a1299c7ee81c79f6>:0)
        System.Threading.Tasks.Task.Execute () (at <ad04dee02e7e4a85a1299c7ee81c79f6>:0)
        Rethrow as AggregateException: One or more errors occurred.
        UnityEngine.Debug:LogException(Exception)
        KSPBurst.<BurstCompile>d__15:MoveNext()
        UnityEngine.SetupCoroutine:InvokeMoveNext(IEnumerator, IntPtr)
```
Whereas running the same KSP installation from a system with Mono installed (thanks, NAS), the compiler runs successfully.

The file `/NAS/games/KubuntuGames/steamapps/common/KSP_Testing/PluginData/KSPBurst@1.5.0.0-com.unity.burst@1.5.0/package/.Runtime/bcl.exe` itself does exist, so the extraction works (just to rule that out based on the exception message).

@HebaruSan proposed that you could try catching that exception and tell users to install Mono. Since this mod isn't essential and everything works just as fine without the compiler running (except the performance improvements), this should work.

I can send you the full log if you want, but that investigation shout be done outside of this issue.

dkavolis:
[9835081](https://github.com/KSPModdingLibs/KSPBurst/commit/9835081e6ded09b7f1b4fe260848d8591c9316c6) added a log message about potentially missing mono, PopupDialog may be too annoying.

DasSkelett:
Uhhh, that close was not intended and automatically done by GitHub when pushing? Hang on...

DasSkelett:
And now it's not letting me force-push back to the previous commit because the PR is closed.
Sorry for that, working with multiple remote on GitHub repos can be tricky sometimes. Going to open a new PR replacing this one.

PR #8419, opened 2021-03-19
Add KSPBurst

DasSkelett:
Replaces #8418 due to some commit weirdness, and GitHub locking the PR closed and no abillity to force-push to the previous commit.

Quick links:
* https://github.com/KSPModdingLibs/KSPBurst
* https://forum.kerbalspaceprogram.com/index.php?/topic/201112-18-111-kspburst-burst-compiler-for-kerbal-space-program/

KSPBurst is a library providing the Burst compiler tool. Some mods plan to use it in the future to make use of [Unity Jobs](https://docs.unity3d.com/Manual/JobSystem.html), which requires all job code to be located in a single DLL, so this mod allows mods to dynamically register themselves and compiles all code into a single DLL. At least that's my understanding.

Maintained and submitted by @dkavolis

### DIFF
--- a/NetKAN/KSPBurst-Lite.netkan
+++ b/NetKAN/KSPBurst-Lite.netkan
@@ -1,7 +1,7 @@
 {
     "spec_version": "v1.4",
-    "identifier":   "KSPBurst-NOOP",
-    "name":         "KSPBurst no-op version",
+    "identifier":   "KSPBurst-Lite",
+    "name":         "KSPBurst Lite",
     "abstract":     "A stand-in for KSPBurst that satisfies all dependendent mods but doesn't install and run the compiler",
     "$kref":        "#/ckan/github/KSPModdingLibs/KSPBurst/asset_match/^KSPBurst_[0-9.]+_plugins_only\\.zip$",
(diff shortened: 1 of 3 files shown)


dkavolis:
Don't the 2 mods need `conflicts` with each other since they will install to the same location?

DasSkelett:
Any suggestions for a better name (and identifier)? Or should I just name it "KSPBurst - Plugins Only", similar to the zip?

dkavolis:
What about `KSPBurst Lite`?

DasSkelett:
@dkavolis we changed the license to "unknown" given all these bundled DLLs and files that are not MIT. If we'd keep it at "MIT", our infra would automatically mirror all the releases to archive.org, which I'm not happy with without knowing all the rights and restrictions of every license involved.

DasSkelett:
One more issue I see is that users effectively can't switch from the full to the "Lite" version. Once the compiler is extracted, it stays, and the way the code is written currently, it'll still use the extracted version of the compiler.
But again same applies to manual installs to some degree, you can't "disarm" it by just removing the zip from GameData.

Looks good from our side, thanks again!

PR #8430, opened 2021-03-21
Add new plugin : KspBugReport

gotmachine:
Bug reporting plugin, homepage : https://github.com/KSPModdingLibs/KSPBugReport

### DIFF
--- a/NetKAN/KSPBugReport.netkan
+++ b/NetKAN/KSPBugReport.netkan
@@ -1,0 +1,20 @@
+{
+    "spec_version": "v1.18",
+    "identifier":   "KSPBugReport",
+    "abstract":     "Easy to use bug reporting plugin for KSP players",
+    "author":       "Gotmachine",
+    "$kref":        "#/ckan/github/KSPModdingLibs/KSPBugReport",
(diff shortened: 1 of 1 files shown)


HebaruSan:
@DasSkelett, thoughts on including `CKAN/registry.json` in these archives?

DasSkelett:
I doubt any mod author (that isn't on the CKAN team ;-) has ever asked for and looked at the registry.json, but why not, maybe some are going to start doing this when it already comes with the bug report. And it might come in handy if someone uses it to report bugs to us.

HebaruSan:
Well the idea would be to capture the user's mod list. For which it would probably be much better to include the newest file in the `<KSP>/CKAN/history` folder. A few kilobytes instead of 20+ MB.

DasSkelett:
The full list of installed mods it already in `KSP.log`, but it would add the information of which got installed through CKAN and which were done manually. And I agree if it's just for that, the `installed-*.ckan` files would be smaller.

HebaruSan:
Thanks for the submission, and good luck with this one. We might stop by later to submit some CKAN-related enhancements!

PR #8431, opened 2021-03-21
PR branched from old commit to test the diff

HebaruSan:
This is not a real pull request. It is just here to test KSP-CKAN/xKAN-meta_testing#72.

If the checks only look at Astrogator, then it worked. If they check a bunch of other mods, then it didn't work.


HebaruSan:
Looks good!

PR #8449, opened 2021-04-02
Update Kerbalism-Config-RO

Standecco:
the ROKerbalism repo was transferred from my account to the KSP-RO organization; this change reflects that.


HebaruSan:
Thanks for the update!

PR #8450, opened 2021-04-05
Add BoosterGuidance

HebaruSan:
Hi @oyster-catcher, we had a request on the CKAN Discord to add this mod to CKAN.
Would it be OK with you if we did that?

- https://spacedock.info/mod/2587/BoosterGuidance
- https://github.com/oyster-catcher/BoosterGuidance
- https://forum.kerbalspaceprogram.com/index.php?/topic/198760-181-1101-boostguidance-v101/

### DIFF
--- a/NetKAN/BoosterGuidance.netkan
+++ b/NetKAN/BoosterGuidance.netkan
@@ -1,0 +1,14 @@
+{
+    "spec_version": "v1.4",
+    "identifier":   "BoosterGuidance",
+    "$kref":        "#/ckan/spacedock/2587",
+    "$vref":        "#/ckan/ksp-avc",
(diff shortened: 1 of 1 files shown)


HebaruSan:
Tagged here and asked on the forum thread and no response so far. But author was last active in the forum in January, so that doesn't necessarily mean "No". Will wait a bit longer.

oyster-catcher:
Yes sorry. I've been very inactive for a long time. I'm very happy with the mod going on CKAN. Can you do that or do you need my help? like extra meta files, since I don't personally know what's required.

HebaruSan:
It should be there in a couple of minutes. Thanks for the response!

PR #8457, opened 2021-04-07
Adding KSP-Recall as a dependency, as requested by DasSkellet on forum.

Lisias:
https://forum.kerbalspaceprogram.com/index.php?/topic/179030-144/&do=findComment&comment=3949978

(Assuming, of course, I didn't messed up something as usual! :) )


DasSkelett:
Thanks, looking good!
Now users shouldn't be surprised by a missing dependency warning when starting KSP, wondering why CKAN didn't take care of that.

PR #8475, opened 2021-04-14
Make Scatterer-sunflare and Scatterer-config depend on Scatterer

DasSkelett:
Noticed while working on #8474 

`Scatterer-sunflare` and `Scatterer-config` don't depend on core `Scatterer`.
To my knowledge they do nothing without it being installed as well.

In a case like #8474 where the sunflares are a recommendation, or if a user sees sunflares and wants to install them, they might only select "Scatterer Sunflare" and don't realize that they need Scatterer as well, then wondering why it doesn't work.

I haven't found a reason why they don't have the dependency yet, so I guess it is "because nobody noticed it yet".

So this PR adds `Scatterer` to the dependencies of `Scatterer-sunflare` and `Scatterer-config`.

### DIFF
--- a/NetKAN/Scatterer-config.netkan
+++ b/NetKAN/Scatterer-config.netkan
@@ -1,16 +1,19 @@
 {
-    "spec_version": "v1.4",
-    "identifier":   "Scatterer-config",
-    "name":         "Scatterer Default Config",
-    "abstract":     "The default configuration files for scatterer",
-    "$kref":        "#/ckan/spacedock/141",
+    "spec_version":     "v1.4",
+    "identifier":       "Scatterer-config",
+    "name":             "Scatterer Default Config",
+    "abstract":         "The default configuration files for scatterer",
+    "$kref":            "#/ckan/spacedock/141",
     "x_netkan_force_v": true,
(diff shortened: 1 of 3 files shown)


HebaruSan:
I thought there was a reason we did not do this, but I can't recall it at the moment. I'll take a look at the old PRs and see if I can find some discussion...

HebaruSan (approved):
I've looked and searched and queried and scoured, and I can find no trace of a PR or issue where we tried this and either had problems or explained why it's not a good idea. And I can't think of one myself.

I guess if there _is_ some problem, this can be the PR where we document it. :grimacing:

DasSkelett:
> I guess if there _is_ some problem, this can be the PR where we document it. 😬

Just took me a few months, but now I realized a problem: we made a bunch of planet packs depend on Scatterer-sunflare-default in order to force the installation of the default sunflare, as others would be incompatible.
https://github.com/KSP-CKAN/NetKAN/pull/7794 and all the linked pull requests.

This was done under the assumption that Scatterer-sunflare-default wouldn't depend on Scatterer, and thus would only have an effect if the user installed Scatterer additionally on their own. But now we're forcing it as well.

PR #8497, opened 2021-04-22
Made CommunityResourcePack and CommunityTechTree a requirement

sswelm:


HebaruSan:
Hi @sswelm, thanks for the update.
Does this apply to the currently released version, or a new version to be released soon? We would want to know this to make sure the metadata ends up in the correct state.

sswelm:
@HebaruSan   This was exactly already the case for the last 4 months (since 1.26.13) but I forgot to update CKAN. Would be nice if you could fix this retroactively

HebaruSan:
OK thanks, I'll merge this now and update the older versions in a bit.

PR #8500, opened 2021-04-25
Freeze HumanStuff after being deleted from SpaceDock

DasSkelett:
This mod has been deleted from SpaceDock.
We all will miss it.

https://forum.kerbalspaceprogram.com/index.php?/topic/196759-humanstuff-slightly-more-realistic-spacesuits-heads-staff-for-ksp-v181/&do=findComment&comment=3960497

forum comment

bot notification


DasSkelett:
(The mod is still available on [GitHub](https://github.com/aazard/HumanStuff/releases), but the author has explicitly expressed the wish for this mod to be taken down from CKAN in that comment, so we do that)

PR #8515, opened 2021-05-03
Add GalaxiesUnbound and friends

HebaruSan:
Requested by a user on Discord, author sounds enthusiastic:

image

- https://forum.kerbalspaceprogram.com/index.php?/topic/198895-110x-111x-galaxies-unbound-a-stellar-odyssey-102-3-11-jan-2021/
- https://github.com/StarCrusher96/Galaxies-Unbound-A-Stellar-Odyssey
- ~~<https://spacedock.info/mod/2620/Galaxies%20Unbound:%20A%20Stellar%20Odyssey>~~

SpaceDock has these as one big ZIP with nested ZIPs, luckily GitHub has them as separate top level assets, so that's where we're getting it.

### DIFF
--- a/NetKAN/GalaxiesUnbound-AlienWorlds.netkan
+++ b/NetKAN/GalaxiesUnbound-AlienWorlds.netkan
@@ -1,7 +1,7 @@
 spec_version: v1.4
-identifier: GalaxiesUnbound-Lich
-name: Galaxies Unbound - Lich
-$kref: '#/ckan/github/StarCrusher96/Galaxies-Unbound-A-Stellar-Odyssey/asset_match/Lich'
+identifier: GalaxiesUnbound-AlienWorlds
+name: Galaxies Unbound - Alien Worlds Expansion
+$kref: '#/ckan/github/StarCrusher96/Galaxies-Unbound-A-Stellar-Odyssey/asset_match/AlienWorlds'
 license: CC-BY-NC-ND-4.0
 resources:
   homepage: https://forum.kerbalspaceprogram.com/index.php?/topic/198895-*
@@ -11,7 +11,6 @@ tags:
 depends:
   - name: ModuleManager
   - name: GalaxiesUnbound-Core
-  - name: Singularity
 install:
-  - file: GU
+  - find: GU
     install_to: GameData
--- a/NetKAN/GalaxiesUnbound-AlienWorlds.netkan
+++ b/NetKAN/GalaxiesUnbound-AlienWorlds.netkan
@@ -1,6 +1,7 @@
 spec_version: v1.4
 identifier: GalaxiesUnbound-AlienWorlds
 name: Galaxies Unbound - Alien Worlds Expansion
+abstract: Lonely challenging planets around stars
 $kref: '#/ckan/github/StarCrusher96/Galaxies-Unbound-A-Stellar-Odyssey/asset_match/AlienWorlds'
 license: CC-BY-NC-ND-4.0
 resources:
--- a/NetKAN/GalaxiesUnbound-AlphaCentauri.netkan
+++ b/NetKAN/GalaxiesUnbound-AlphaCentauri.netkan
@@ -1,0 +1,16 @@
+spec_version: v1.4
+identifier: GalaxiesUnbound-AlphaCentauri
+name: Galaxies Unbound - Alpha Centauri
+$kref: '#/ckan/github/StarCrusher96/Galaxies-Unbound-A-Stellar-Odyssey/asset_match/AlphaCentauri'
+license: CC-BY-NC-ND-4.0
+resources:
+  homepage: https://forum.kerbalspaceprogram.com/index.php?/topic/198895-*
+tags:
+  - config
+  - planet-pack
+depends:
+  - name: ModuleManager
+  - name: GalaxiesUnbound-Core
+install:
+  - file: GU
+    install_to: GameData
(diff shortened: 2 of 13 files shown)


HebaruSan:
There are two related more downloads available to swap out the home solar system, but they're marked as pre-releases currently:

- https://github.com/StarCrusher96/Galaxies-Unbound-HomeSwitch

HebaruSan:
Version file URL property is wrong, but there's no actual code on the GitHub repo, so we would have to do an issue instead of a pull request. Leaving that for later.

HebaruSan:
Hi @StarCrusher96, this is the pull request that I've created to work on adding your mod(s?) to CKAN. None of this is available to users yet, so we can make changes freely to test out how they work. The "Files changed" tab shows the files that I've created (don't worry about the "Loaded registry with inconsistencies" errors, those are false positives that we've been reluctant to eliminate for historical reasons) and the "Checks" tab has the detailed output from the validation scripts (under the "Test modified netkans" section, the "Inflating" sections show a preview of the version-specific metadata, and the "Installing" sections show a test run of installing each module, including the file paths that are installed).

Currently I have it creating four new modules, one for each of the pieces of GU:

- The main planet pack  (depends on Module Manager and Kopernicus, recommends the other three GU modules)
- The parts  (recommends the main GU planet pack)
- The EVE configs  (depends on EVE and the main GU planet pack)
- The Scatterer configs  (depends on Scatterer and the main GU planet pack)

Eventually we can add the "homeswitch" components as well, but I haven't done that here because they're still in a pre-release phase. You mentioned restructuring the mod, but it's already pretty well suited to CKAN. We can get into the details of the ideas that you had for that when you have time.

Please let us know if you have any questions, and welcome!

StarCrusher96:
Okay being very honest I have NO idea how Github works other than uploading the mod for releases :s
One thing that'll be very important for GU Ckan use is people will need to be able to enable and disable individual star systems. Just as they now manually have to do. Ckan should automate the process.

HebaruSan:
> One thing that'll be very important for GU Ckan use is people will need to be able to enable and disable individual star systems. Just as they now manually have to do. Ckan should automate the process.

OK, interesting, we can look into this. To make sure I'm understanding correctly, are you referring to the step where the user edits the `GU-general_Settings.cfg` file?

cfg

StarCrusher96:
Yes I'm reducing the option to enable system x instead of enable stars or planets. Some people struggle with the option thingy so Poodmund proposed if i one day put GU on CKAN it would be interesting to have people download individual systems.

HebaruSan:
I see, yes, that could be quite nice for users. I assume you would do that by making one ZIP file per system? If so, that would be pretty easy for CKAN to handle.

StarCrusher96:
yeah that would be the idea - still all zip files would merge under 1 general GU folder.

HebaruSan:
Cool, well if you want to do that first, we can put this on hold and come back to it. Let us know if you finish that or change your mind and want to index the current versions as they are. :+1:

HebaruSan:
Looks like the revamp went ahead successfully, the latest release has a core ZIP plus 7 system ZIPs and 2 ZIPs starting with an underscore that I'm not sure about. We can probably come back to this.

https://github.com/StarCrusher96/Galaxies-Unbound-A-Stellar-Odyssey/releases/tag/1.1

HebaruSan:
Hmm, checked the forum thread for install instructions and found this instead:

image

image

image

Maybe not the best time to add this. But there are other posts about how version 1.1 was restructured for CKAN, so it looks weird that we don't have it. :thinking:

HebaruSan:
Install guide found:

https://github.com/StarCrusher96/Galaxies-Unbound-A-Stellar-Odyssey/blob/main/README.md

HebaruSan:
Hmm. The 1.1 release is being edited and updated repeatedly with re-downloads of some parts required instead of making new releases:

image

image

image

image

image

@StarCrusher96, the usual practice for KSP mods is to create new releases when new things are added or old things are changed. This allows users to coherently report to you what they have installed; if I said "I have release 1.1 of Core installed," you would have no idea which of those 6 possible versions I have installed on my disk, which would probably make support an absolute nightmare.

It would also cause headaches with regard to CKAN. Normally CKAN would simply detect a new release and prompt users to upgrade, which is all pretty solidly established code at this point. But when files from a release are replaced in-place like that, you are relying on users to _reinstall_, which is not nearly so well managed because it's not supposed to be needed very often. You would likely have users installing new systems with old cores and so forth.

Holding off on indexing this for now as I don't want CKAN to be the cause of users coming to the mod's forum thread and reporting issues with the wrong sub-release revision of the core ZIP or other such problems. We can take this up again if we have some productive communications and indications of more stability in releases. (But I probably will rework this PR's netkans to reflect the current structure, at least.)

HebaruSan:
Browsing the thread, it looks like the days of "redownload GU_Core" are behind us. :tada: 
There was a 1.2.5 release in January, and 1.3 a few days ago.

HebaruSan:
The "plugin" is just MiniAVC, so no need to track down missing source code.

HebaruSan:
- [ ] Need to add abstracts to every module

HebaruSan:
Only some of these made it through the Inflator; others ran out of disk space:

image

StarCrusher96:
_Core is mandatory :p to make GU work. So the others serve no purpose without _Core.

StarCrusher96:
You should add  Ballisticfox, he's managing the CKAN part of GU in my stead.

HebaruSan:
The problem is in the bot's AWS container, there's not much anyone other than @techman83 can do about that.

StarCrusher96:
Also thanks for doing this

HebaruSan:
The problem with the bot has been fixed, at least for now, and the rest of the modules are now indexed.

PR #8530, opened 2021-05-09
Added B9PartSwitch as recommendation as its used on some of the new m…

sswelm:
…odels


HebaruSan:
Thanks for the update!

PR #8532, opened 2021-05-10
Create StagedAnimation.netkan

NathanKell:
Been a loooong time. I think I reverse engineered the format right.
Desired behavior, since it has no AVC, is to set minversion to 1.8 and max to 1.11.99. Will bump to 1.12 when that comes out since I can't imagine it won't work. Think I got all the rest of the meta right.

### DIFF
--- a/NetKAN/StagedAnimation.netkan
+++ b/NetKAN/StagedAnimation.netkan
@@ -1,0 +1,24 @@
+{
+    "spec_version"    : "v1.4",
+    "identifier"      : "StagedAnimation",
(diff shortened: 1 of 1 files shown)


DasSkelett:
Thanks for the submission, looking very good!
Some links that might help forward engineering in the future:
- https://github.com/KSP-CKAN/CKAN/wiki/Adding-a-mod-to-the-CKAN#advanced
- https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md

PR #8537, opened 2021-05-17
Update RealAntennas.netkan

DRVeyl:
Prep for releasing <https://github.com/DRVeyl/RealAntennas/releases/tag/v2.0>
In pre-release on 16 MAY 2021.  New release will require KSPBurst.  Edited by hand, wanted to make sure this is ready to go for when I officially release 2.0 (expect 17 MAY) and it indexes.

I can update a comment here when that happens.  I'll mark the PR as draft until then.

### DIFF
--- a/NetKAN/RealAntennas.netkan
+++ b/NetKAN/RealAntennas.netkan
@@ -14,7 +14,8 @@
         "plugin"
     ],
     "depends": [
-        { "name": "ModuleManager" }
+        { "name": "ModuleManager" },
+        { "name": "KSPBurst" }
     ],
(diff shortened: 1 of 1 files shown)


HebaruSan:
Cool, thanks for the update! I've subscribed to release notifications on your repo, so if you merge it while I'm awake, we should be able to keep up, with or without a comment. If past experience is any indicator, @DasSkelett may do the same, so you'll be covered from multiple continents.

DRVeyl:
RA released, and 2.0 now indexed.

PR #8538, opened 2021-05-17
Add JNSQ-RealDate and GreenKerbs

HebaruSan:
Two mods on SD yadda yadda... One suit pack and one calendar display config.

https://spacedock.info/mod/2757/Greenkerbs%20V1

https://spacedock.info/mod/2755/JNSQ%20Real%20Date
https://forum.kerbalspaceprogram.com/index.php?/topic/202434-111x-jnsq-real-date-10/

___

ckan compat add 1.8

### DIFF
--- a/NetKAN/GreenKerbs.netkan
+++ b/NetKAN/GreenKerbs.netkan
@@ -1,0 +1,15 @@
+{
+    "spec_version": "v1.4",
+    "identifier":   "GreenKerbs",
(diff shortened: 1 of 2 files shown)


HebaruSan:
Asked a clarification question for JNSQ-RealDate having to do with Kronometer, which is indexed but which JNSQ also bundles.

May be a reason to make some changes to JNSQ.

PR #8555, opened 2021-06-02
Freeze FruitGoose's mods

DasSkelett:
All mods by this author have been deleted from SpaceDock and GitHub, and their forum threads have been edited to contain no text.
I've written the author a forum PM out of interest, but it really looks like the mods won't come back with a rename or something.

Quick links:
- https://spacedock.info/profile/FruitGooseKSP
- https://forum.kerbalspaceprogram.com/index.php?/profile/210116-fruitgoose/
- Some of the forum threads
  - [ORANGE](https://forum.kerbalspaceprogram.com/index.php?/topic/201227--/)
  - [VerticalSpeedIndicator](https://forum.kerbalspaceprogram.com/index.php?/topic/201898-_/)


DasSkelett:
I'm gonna check all the licenses to see if we need to freeze some of them in CKAN-meta or if there's an archive.org fallback.

DasSkelett:
They all seem to be MIT and thus are archived: https://archive.org/details/kspckanmods?query=&sin=&sort=-publicdate&&and[]=creator%3A%22fruitgooseksp%22
No CKAN-meta work needed :+1:

HebaruSan:
That link will also be handy if anyone wants to adopt these, since the source code is archived as well.

PR #8565, opened 2021-06-10
Hard code IFS repo links

HebaruSan:
## Problem

See KSP-CKAN/CKAN#3390, I'd like to use the GitHub API to populate `resources.repository` for mods on SpaceDock, but InterstellarFuelSwitch uses a special link to a particular subdirectory of the main KSPInterstellarExtended repo, which we would like to preserve, and which would be truncated to the main repo with that change.

## Changes

Now IFS's `resources.respository` "deep links" are set in the netkans, so they will be set regardless of how we treat this property in the Inflator.


HebaruSan:
image

PR #8566, opened 2021-06-10
Convert PlanningNode netkan to YAML

HebaruSan:
As of KSP-CKAN/CKAN#3367, netkans are allowed to use YAML format, however so far none do.
Now PlanningNode uses YAML format. This will give us a test case for YAML to make sure it works.
Also the git attributes are updated to count .netkan files as YAML instead of JSON.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,3 @@
 *.ckan linguist-language=JSON linguist-detectable=true
-*.netkan linguist-language=JSON linguist-detectable=true
+*.netkan linguist-language=YAML linguist-detectable=true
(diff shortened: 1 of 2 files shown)


HebaruSan:
The internal .ckan file will also be using YAML format.

https://github.com/HebaruSan/PlanningNode/commit/44fe036885e88229267ee0c9cd2b712d63dd21bf

HebaruSan:
Worked!

https://github.com/KSP-CKAN/CKAN-meta/commit/8e9ad9be40af53ea846aae34efdca47fd899d53f

PR #8576, opened 2021-06-20
StockWaterfallEffects no longer conflicts with ReStock

HebaruSan:
image


HebaruSan:
Confirmed that the current latest release is the one that's compatible:

https://github.com/KnightofStJohn/StockWaterfallEffects/releases/tag/0.5.1

PR #8579, opened 2021-06-21
Fix PlanetShine indentation

DasSkelett:
One of the mods that caused https://github.com/KSP-CKAN/NetKAN-Infra/pull/220

GitHub highlights this syntax error nicely, now that all netkans are marked as YAML (https://github.com/KSP-CKAN/NetKAN/pull/8566), so I think it was a good thing to do in the end.

image

https://github.com/KSP-CKAN/NetKAN/blob/339d2f3dc336ec880be6b587a352d0e0d809d583/NetKAN/PlanetShine.netkan#L7-L9


DasSkelett:
I'll ignore "Line too long" linting errors for JSONs.

PR #8581, opened 2021-06-22
Remove EVE-Config provides from StrangeNewWorlds

DasSkelett:
Just got pinged by the author on the Kopernicus Discord that this mod conflicts with AVP on CKAN, but it shouldn't.
This is due to the `EnvironmentalVisualEnhancements-Config` provides.

This mod only provides EVE configs for its own planets, but not the stock ones, and it doesn't remove the stock planets.
Thus it still needs another set of EVE configs installed alongside (default configs, AVP, ...) to make EVE work to its full extents,

Now the provides is removed, and instead the virtual module `EnvironmentalVisualEnhancements-Config-stock` is added to the recommendations, alongside `EnvironmentalVisualEnhancements`.

Several EVE config packs that provide configs for the stock planet now provide `EnvironmentalVisualEnhancements-Config-stock`.

---

I've also sent the author a notice about the huge `.git` folder in the zip that really doesn't need to be there;
```
$ du -chd2 SNW/ | sort -h
427K    SNW/__MACOSX
427K    SNW/__MACOSX/StrangeNewWorlds
755M    SNW/StrangeNewWorlds/GameData
778M    SNW/StrangeNewWorlds/.git
1.5G    SNW/StrangeNewWorlds
1.5G    SNW/
1.5G    total
```

It's outside the directory we're installing, so we don't need to filter it extra.

### DIFF
--- a/NetKAN/AstronomersVisualPack.netkan
+++ b/NetKAN/AstronomersVisualPack.netkan
@@ -8,10 +8,14 @@
     "license"      : "CC-BY-NC-SA-4.0",
     "resources": {
         "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/160878-*"
-	},
+    },
     "tags": [
         "config",
         "graphics"
+    ],
+    "provides": [
+        "EnvironmentalVisualEnhancements-Config",
+        "EnvironmentalVisualEnhancements-Config-stock"
     ],
     "conflicts": [
         { "name" : "EnvironmentalVisualEnhancements-Config" }
@@ -30,8 +34,5 @@
         { "name": "DistantObject" },
         { "name": "PlanetShine" },
         { "name": "Chatterer" }
-    ],
-    "provides" : [
-        "EnvironmentalVisualEnhancements-Config"
     ]
 }
--- a/NetKAN/EnvironmentalVisualEnhancements-LR.frozen
+++ b/NetKAN/EnvironmentalVisualEnhancements-LR.frozen
@@ -18,7 +18,8 @@
         }
     ],
     "provides": [
-        "EnvironmentalVisualEnhancements-Config"
+        "EnvironmentalVisualEnhancements-Config",
+        "EnvironmentalVisualEnhancements-Config-stock"
     ],
     "depends": [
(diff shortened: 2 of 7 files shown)


HebaruSan:
The EVE config relationships may benefit from some rethinking in general. What you describe seems common, planet packs include configs for their own planets, which can be used alongside other configs for other planets.

There's a `EnvironmentalVisualEnhancements-Config-stock` virtual module, maybe this mod should depend on that?

DasSkelett:
I don't think an `EnvironmentalVisualEnhancements-Config-stock` dependency is a good fit here, because it doesn't actually depend on EVE, and it also doesn't necessarily require the default configs, it also works fine with AVP according to the author (which doesn't provide `EnvironmentalVisualEnhancements-Config-stock`).

But I agree that the current EVE relationship system doesn't really give us all the options we need.

HebaruSan:
> AVP (which doesn't provide `EnvironmentalVisualEnhancements-Config-stock`)

Maybe it should. It's an EVE config for the stock planets, right?

DasSkelett:
> Maybe it should. It's an EVE config for the stock planets, right?

Yeah it is. I can't find any clue on what this virtual module was initially added for, but it doesn't seem to be used anywhere. So we might as well use it to declare that a mod provides configs for the stock planets now.

DasSkelett:
A list of mods that provide EVE configs for the default planets to my knowledge and might benefit of this provides:
```
AstronomersVisualPack
EnvironmentalVisualEnhancements-HR
EnvironmentalVisualEnhancements-LR  # maybe, depending on backports
SciFiVisualEnhancements
Spectra
StockVisualEnhancements
SciFiVisualEnhancements-WeimaranersTweaks
```

DasSkelett:
And for `StrangeNewWorlds` we put in `EnvironmentalVisualEnhancements-Config-stock` as recommendation, alongside `EnvironmentalVisualEnhancements`?

DasSkelett:
Okay, let's give it a try 🤞 
We can look at other planet packs that don't remove the default planets (OPM comes to mind) and have some relationship with EVE in a separate PR, this one's already quite big, download-/cache-wise.

PR #8429, opened 2021-03-21
Add vrefs to mods with unused version files

HebaruSan:
After KSP-CKAN/CKAN#3327 we now have a much more reasonable list of vref warnings:

image

Now these mods have vrefs.

RadioFreeKerbin isn't updated because it has a metanetkan.

___

ckan compat add 1.8

PR #8591, opened 2021-06-25
Add Harmony2 dependency to LMP, update compatibility

DasSkelett:
The new release 0.28.0 now bundles Harmony2 (updated from Harmony1).
It is also released for KSP 1.12.

After talking to the author, we've decided to filter the Harmony DLL, and replace it with a dependency on our indexed `Harmony2` module (the [HarmonyKSP](https://github.com/KSPModdingLibs/HarmonyKSP) "mod").

They author might switch over to bundling HarmonyKSP themself, which would allow us to get rid of the filter, and make sure that both CKAN users and manual installers would have the same distributions of Harmony installed.
Since this wouldn't need a change in our metadata (besides removing the filter stance), I'm going to go ahead with this change already.

Also see https://github.com/LunaMultiplayer/LunaMultiplayer/issues/415

ckan compat add 1.11

### DIFF
--- a/NetKAN/LunaMultiplayer.netkan
+++ b/NetKAN/LunaMultiplayer.netkan
@@ -15,4 +15,4 @@ depends:
 install:
   - find:       LunaMultiplayer
     install_to: GameData
-    filter:     0Harmony.dll 
+    filter:     0Harmony.dll
--- a/NetKAN/LunaMultiplayer.netkan
+++ b/NetKAN/LunaMultiplayer.netkan
@@ -1,21 +1,18 @@
-{
-    "spec_version": "v1.4",
-    "identifier":   "LunaMultiplayer",
-    "name":         "Luna Multiplayer Client",
-    "$kref":        "#/ckan/github/LunaMultiplayer/LunaMultiplayer/asset_match/^LunaMultiplayer-Release\\.zip$",
-    "comment":      "Version file is outdated",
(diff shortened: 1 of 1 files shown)


HebaruSan:
Was dropping the vref intentional?

DasSkelett:
The vref has been dropped a long time ago, in #8281.
Also https://github.com/KSP-CKAN/NetKAN/pull/8429#issuecomment-803603239

It is still broken currently, at the time of the release it wasn't updated at all, a few minutes ago the remote file has been updated in https://github.com/LunaMultiplayer/LunaMultiplayer/pull/431, but still forgot to bump `VERSION`.

So I won't re-add the vref in the foreseeable future, the author is far to unreliable in this regard.

enderger:
Yeah, sorry 'bout that. I've published a PR to fix that (LunaMultiplayer/LunaMultiplayer#432), I'm a bit new to the KSP modding scene and didn't recognize that it needed to get bumped.

DasSkelett:
No worries @enderger, you made a step in the right direction.
We still won't be able to add the vref back to the netkan, since the .version file embedded in the released zip still has the incorrect `VERSION` property. This would require a new release, but even then we probably won't add it back, since it's too unreliable in this mod's case. Manually updating hardcoded compat data in the netkan once a year is easier than having to go back and fixup old metadata, and less disruptive for users.

PR #8597, opened 2021-06-27
Create netkan for RSS-CanaveralHD

NathanKell:


NathanKell:
Oh hell I forgot to, like, make a release. Sec.

NathanKell:
Solved. \o/

HebaruSan:
Hi @NathanKell, thanks for the submission.

Unfortunately I was reading up on license compatibility today, so I have acquired the terrible knowledge that CC-BY-NC-SA assets can't be re-licensed as GPL-3.0:

- https://creativecommons.org/share-your-work/licensing-considerations/compatible-licenses

> BY-NC-SA
> Version 4.0
> Your contributions to adaptations of material under BY-NC-SA 4.0 may only be licensed under:
> 
> - BY-NC-SA 4.0, or a later version of the BY-NC-SA license.
> - Ported versions of the BY-NC-SA license, version 4.0 or later.
> - A license designated as a “BY-NC-SA Compatible License” as defined in BY-NC-SA 4.0.
>  **Currently, no non-CC licenses have been designated as compatible with BY-NC-SA 4.0.** Other licenses may be added to this list at any time according to the established process and criteria. Once a license has been added to this list, it will not be removed.

One of the source mods uses that license:

https://github.com/AnticlockwisePropeller/RSSHiDefCape/blob/master/GameData/RSSHiDefCape/Readme.txt

> License: CC-BY-NC-SA

... so there is a conflict here over whether commercial uses are allowed; RSSHiDefCape says No and RSS-CanveralHD is trying to say Yes (but it can't, for those assets). (KatnisssCapeCanaveral is already GPL-3.0, so that's no problem.)

NathanKell:
Ah, good point. I am not sure whether this is legally sound, but IIRC historically the KSP forum has allowed mods released as GPL to be incorporated into CC-BY-NC-SA mods, treating the latter as more restrictive. If that is ok I'll relicense as CC-BY-NC-SA.

NathanKell:
This is now done, and readme clarified that it only applies to the work done by me/KSP-RO team. Since Katniss's texture and mesh are unchanged, that should be fine regardless (the actually altered assets are the ones from RSSHiDefCape).

NathanKell:
Is this ok now or do I need to take other action?

HebaruSan:
Yup, looks good. Sorry for the delay!

NathanKell:
No worries! Thanks!

Sent by my thumbs, slowly.

PR #8610, opened 2021-06-30
Fixes missing DLL file issue

sswelm:
Interstellar_Redist.dll should be placed in the GameData folder

### DIFF
--- a/NetKAN/InterstellarRedistributable.netkan
+++ b/NetKAN/InterstellarRedistributable.netkan
@@ -2,4 +2,9 @@ spec_version: v1.4
 identifier: InterstellarRedistributable
 $kref: '#/ckan/spacedock/2800'
 license: MIT
-x_via: Automated SpaceDock CKAN submission
+tags:
+  - plugin
+  - library
+install:
+ - file: Interstellar_Redist.dll
+   install_to: GameData
--- a/NetKAN/InterstellarRedistributable.netkan
+++ b/NetKAN/InterstellarRedistributable.netkan
@@ -1,0 +1,5 @@
+spec_version: v1.4
+identifier: InterstellarRedistributable
+$kref: '#/ckan/spacedock/2800'
+license: MIT
+x_via: Automated SpaceDock CKAN submission
--- a/NetKAN/KSPInterstellarExtended.netkan
+++ b/NetKAN/KSPInterstellarExtended.netkan
@@ -1,10 +1,10 @@
 {
-    "spec_version"  : "v1.4",
-    "identifier" : "KSPInterstellarExtended",
-    "$kref" : "#/ckan/spacedock/172",
-    "$vref" : "#/ckan/ksp-avc",
+    "spec_version": "v1.4",
+    "identifier": "KSPInterstellarExtended",
+    "$kref": "#/ckan/spacedock/172",
+    "$vref": "#/ckan/ksp-avc",
     "x_netkan_allow_out_of_order": true,
-    "license" : "GPL-3.0",
+    "license": "GPL-3.0",
     "resources": {
         "homepage":   "https://forum.kerbalspaceprogram.com/index.php?/topic/172026-*",
         "repository": "https://github.com/sswelm/KSP-Interstellar-Extended"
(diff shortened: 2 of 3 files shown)


sswelm:
apparently, CKAN is much more stricks now. Making files that would be fine in the past to pass today

DasSkelett:
Thanks for the update! Yeah, since we now also allow netkans to be written in YAML instead of JSON, we had to change the linter, which complains about more syntax "differences". We might be able to tune it a bit.

Regarding the DLL, does this only affect the latest release from a few hours ago (1.12.13)? Or do we need to backport this fix to some older metadata as well?

DasSkelett:
Okay, the DLL seems to have this location since 1.28.12, in 1.28.10 it's inside `GameData/WarpPlugin/Plugins/`.
I see it was also bundled within `GameData/PhotonSail/Plugins/`, but is no more. Does PhotonSailor no longer require this DLL? Or do we need to update it to depend on KSP-IE? Or should we split out the redistributable DLL into its own module, that KSP-IE and PhotonSailor depend on?

sswelm:
For now this only applies to 1.28.12 and up. Both WarpPgin and PhotonSail will have to make use of the same DLL. I guess this is problematic as CKAN cannot both deploy the same file  (but right now only KSPIE deployes theredistributable) . To fixed it better I need to put the redistributable DLL in a seperate downloadable package and make it a requirement for both KSPIE and PhotonSailor

DasSkelett:
Exactly, CKAN doesn't allow two mods to install the same file. If both mods depend on this file but not each other, we need to put it in a separate module. Releasing it as a separate zip somewhere would be very neat, otherwise PhotonSailor-only users need to download the huge KSP-IE zip just to get a single DLL out of it (or the other way around, KSP-IE-only users would need to download the PhotonSailor zip if we extract it from there).

sswelm:
Alright, I uploaded the package to Spacedock and marked it indexable for CKAN 

see https://github.com/KSP-CKAN/NetKAN/pull/8612

it need to put it directly into GamaData folder. Should anything be changed?

DasSkelett:
The license of KSPInterstellarExtended is GPL-3.0, whereas the redistributable is marked as MIT on SpaceDock right now. Is this intended and correct as is, or should the redistributable be GPL-3.0 as well?

HebaruSan:
> apparently, CKAN is much more strict now. Making files that would be fine in the past to no longer pass today

> Thanks for the update! Yeah, since we now also allow netkans to be written in YAML instead of JSON, we had to change the linter, which complains about more syntax "differences". We might be able to tune it a bit.

For what it's worth, the new YAML linter has both errors and warnings, and the "too many spaces" stuff should all be warnings. Your PR can still get the green checkmark even if it has some extra spaces.

sswelm:
> 
> 
> The license of KSPInterstellarExtended is GPL-3.0, whereas the redistributable is marked as MIT on SpaceDock right now. Is this intended and correct as is, or should the redistributable be GPL-3.0 as well?

no MIT, they both should be MIT

DasSkelett:
Hm, looking at the SpaceDock page, it says "Custom MIT", and looking at the text in the zip (or in the [repo}(https://github.com/sswelm/KSP-Interstellar-Extended/blob/master/License.md)), it's not the text of the one I know as "MIT". Are they fully compatible, i.e. does the custom one not have any restrictions that the real MIT/Expat license doesn't have, especially regarding redistributing?
Just want to make sure, because the choice of the license in the metadata decides whether we mirror the mod to archive.org, and we don't want to do that if we are not allowed to.

sswelm:
technically  its a custom License which is very similar to an MIT

for details see https://github.com/sswelm/KSP-Interstellar-Extended/blob/master/License.md

DasSkelett:
> which is very similar to an MIT/GPL-3.0 license

This doesn't make sense, because the MIT and GPL licenses are very very different. At the very least, the first one is permissive, and the second ones are copyleft. It can't really be similar to both of them.

I'm inclined to change it to "unknown" then, because if it's not MIT or GPL-3.0, it's not MIT or GPL-3.0.

sswelm:
ok fine

DasSkelett:
Okay, thanks. I have to ask about PhotonSailor as well, because our metadata says "CC-BY-NC-3.0", SpaceDock says "GPLv3", and the forum thread says "CC-BY-NC". Which one is correct there?

sswelm:
Ok, I fixed the license to CC-BY-NC-3.0

DasSkelett:
I think we're all set now. Thanks again!
Will backport to KSPIE 1.28.12 in a bit.

sswelm:
Alright, people are complaining they cannot install KSPIE  with any other version than 1.12.1 because the InterstellarRedistributal only supports 1.12.1. This is wrong as it is compatible down to KSP 1.8. Could this be fixed somehow?

sswelm:
I|m attemping to fix it with https://github.com/KSP-CKAN/CKAN-meta/pull/2386

PR #8627, opened 2021-07-06
SmokeScreen for KSP 1.12

### LINKED ISSUE #8618: RealPlume and RealPlume Stock Config Dependency Issue
Background
----------
* Operating System: Windows 10 Version 20H2
* CKAN Version: 1.30.4
* KSP Version: 1.12.1

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?**

Yes

Problem
-------
**Describe the bug**
RealPlume says RealPlume Stock Configs are a dependency which isn't up to date so you can't download either and the same happens in reverse (i.e. try to download Stock Configs, says RealPlume is out of date). This is despite both being listed to support any version of 1.12

**Steps to reproduce**
Click on box to install either RealPlume or RealPlume stock configs

**Expected behavior**
Will say, "RealPlume 2:v13.3.2 depends on RealPlumeConfigs, which is not compatible with the currently installed game version" or, " RealPlumeConfigs depends on RealPlume 2:v13.3.2, which is not compatible with the currently installed game version."

sarbian:
Mark the current SmokeScreen as 1.8 to 1.12 compatible.

___

Fixes #8618.

### DIFF
--- a/NetKAN/SmokeScreen.netkan
+++ b/NetKAN/SmokeScreen.netkan
@@ -10,7 +10,7 @@
     },
     "x_netkan_version_edit": "^SmokeScreen-(?<version>\\d+\\.\\d+\\.\\d+\\.\\d+)\\.zip$",
     "ksp_version_min": "1.8.0",
-    "ksp_version_max": "1.10.90",
+    "ksp_version_max": "1.12.90",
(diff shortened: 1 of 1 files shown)


HebaruSan:
Thanks for the update!

PR #8638, opened 2021-07-12
Fix KerbinSideRemasteredTLA install location

DasSkelett:
Fixup of #8636 
This mod belongs into `GameData/ContractPacks/`, as it is in the zip. Also the ships belong where they are in the zip, they are part of the contracts, _not_ user-playable ships: `craftURL = "ContractPacks/KerbinSideRemasteredTLA/Assets/Ships/" + @/wreckCraft`.

ckan compat add 1.8


HebaruSan:
Oops, sorry about that, and thanks for fixing it!

PR #8640, opened 2021-07-13
 hard code minimum KSP version Interstellar Redistributable

sswelm:
It should be available for install from any KSP version starting from KSP 1.8.1

Closes https://github.com/KSP-CKAN/CKAN-meta/pull/2386


DasSkelett:
Yup, that looks good, thanks for the update!

PR #8642, opened 2021-07-13
De-meta RadioFreeKerbin metanetkan with hardcoded download

HebaruSan:
See #8641, a metanetkan with hard-coded `download` property has downsides and isn't necessary when the mod is hosted on GitHub or SpaceDock (or both!). This one even has a version file.


HebaruSan:
I think it also has an internal ckan file!

PR #8644, opened 2021-07-13
De-meta GPP mods

DasSkelett:
See https://github.com/Galileo88/Galileos-Planet-Pack/pull/69:
The compatibility data for GPP, GPPClouds*, GPPSecondary and GPPTextures are a mess:

This was due to hardcoded compatibility in the metanetkans.
I've had the PR open for several months, but got not a single person of the Galileo team to look at it, even after several pings on the PR itself, or the forum thread. Several team members have been active both on the GitHub repo and the forum thread in the meantime. Maybe they are purposefully ignoring us, I don't know.
In any case, since this is a live problem, with a worse impact than our average "forgot a comma in the version file" upstream issues and causes people to have broken installs, with several people messaging me about this in the last few months, I decided to de-meta these metanetkans and fix them here.

## Changes
The hardcoded compatibilities in all netkan files are removed. That's what the AVC .version file included in the download (and its online copy in the repo) is for, which CKAN will use as source of compatibility data from now on.

~~The .version file got a new `KSP_VERSION_MAX` to make sure the releases don't have unbounded future compatibility. This means a tiny bit of extra maintenance to update the maximum compatibility when new KSP versions drop, but gives an extra safeguard to avoid users installing it even if it is broken for the new release.~~
We can't de-meta version files. Since the version file itself looks good and the unbounded future compat being a minor issue (especially now after the final KSP release), I've decided to keep the vref instead of hardcoding it in our netkans. But I'm going to monitor it, as we can't rely on the GPP team not breaking it.

The zip for GPPTextures doesn't have a version file, so it's now unbounded, which shouldn't be much of a problem since it is dependent on GPP anyway.

To reflect this in CKAN, GPP is added to the dependencies of GPPTextures, as well as GPPCloudsHighRes, GPPCloudsLowRes and GPPSecondary.

The hardcoded `download` link of GPPTextures is replaced with a `kref`+`asset_match`.

Also includes https://github.com/Galileo88/Galileos-Planet-Pack/pull/64

CKAN-meta clean-up needed after merging this

### DIFF
--- a/NetKAN/GPP.netkan
+++ b/NetKAN/GPP.netkan
@@ -22,9 +22,9 @@ install:
     install_to: GameData
     comment: This correctly installs Final Frontier for GPP
 depends:
-  - name: ModuleManager
-  - name: Kopernicus
   - name: GPPTextures
+  - name: Kopernicus
+  - name: ModuleManager
 conflicts:
   - name: StockVisualEnhancements
 recommends:
--- a/NetKAN/GPP.netkan
+++ b/NetKAN/GPP.netkan
@@ -1,9 +1,115 @@
 {
     "spec_version": "v1.4",
-    "identifier":   "GPP",
+    "identifier": "GPP",
+    "$kref": "#/ckan/github/Galileo88/Galileos-Planet-Pack/asset_match/Galileo.*?\\.zip",
+    "$vref": "#/ckan/ksp-avc",
     "x_netkan_epoch": 1,
-    "$kref":        "#/ckan/netkan/https://raw.githubusercontent.com/Galileo88/Galileos-Planet-Pack/master/CKAN/GPP.netkan",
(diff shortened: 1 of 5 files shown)


DasSkelett:
Missing comma, I take it as a sign to convert them to YAML.

DasSkelett:
Ugh, it was inflating all those frozen modules from #8641, another case of https://github.com/KSP-CKAN/xKAN-meta_testing/issues/73 :(

DasSkelett:
Whoops, I think I put the filter for the Grannus Ribbon Pack on the wrong install stanza :grimacing:
It should be on the `Optional Mods/GPP_FinalFrontier/GameData/Nereid` one, not the main `GPP` one:
- https://github.com/Galileo88/Galileos-Planet-Pack/pull/64/files

PR #8646, opened 2021-07-14
Strip date portion of SIMPLEX* and TETRIXTechTree versions

DasSkelett:
These mods have a versioning scheme that doesn't interact well with CKAN's sorting logic.
Basically it consists of a semver-like numbers separated by dots part, and a date part in at the end.
The first part alternates between two and three "subparts", so whenever a release with a third subpart comes out, it's compared against the date of the previous release, and is obviously less, so we get a epoch-bump staging PR.

TETRIXTechTree:
- https://github.com/KSP-CKAN/CKAN-meta/pull/1771
- https://github.com/KSP-CKAN/CKAN-meta/pull/1848
- https://github.com/KSP-CKAN/CKAN-meta/pull/1887
- https://github.com/KSP-CKAN/CKAN-meta/pull/2338

SIMPLEXTechTree:
- https://github.com/KSP-CKAN/CKAN-meta/pull/1847
- https://github.com/KSP-CKAN/CKAN-meta/pull/2339

SIMPLEXKerbalism:
- https://github.com/KSP-CKAN/CKAN-meta/pull/1737
- https://github.com/KSP-CKAN/CKAN-meta/pull/1744
- https://github.com/KSP-CKAN/CKAN-meta/pull/2390

(+ SIMPLEXPropulsion which only has one release so far)

This PR adds an `x_netkan_version_edit` that strips the date part of all version numbers.
As far as I can see, there has never been a release with the first part being the same and only the date bumped. It appears to only act as release date label for these mods – since CKAN already stores and shows them separately, it should be fine to remove them from the version numbers as is.

There was one release (`2.3.B.20200511`, TETRIXTechTree) with a letter instead of a digit as "patch" part of the version number, so I've decided to match all chars in the first part, instead of only digits and dots.
The initial release of SIMPLEXKerbalism also had dots in the date part, which this regex doesn't match. It never appeared again, so I think we are fine, if it _does_ happen again, inflation will fail and we need to revisit the regex.

Examples: https://regex101.com/r/G4U5W0/1

Closes https://github.com/KSP-CKAN/CKAN-meta/pull/2390

ckan compat add 1.11

### DIFF
--- a/NetKAN/SIMPLEXKerbalism.netkan
+++ b/NetKAN/SIMPLEXKerbalism.netkan
@@ -8,6 +8,7 @@
         "replace": "${version}",
         "strict": true
     },
+    "x_netkan_epoch": 2,
     "tags": [
         "config",
         "parts",
--- a/NetKAN/SIMPLEXKerbalism.netkan
+++ b/NetKAN/SIMPLEXKerbalism.netkan
@@ -3,6 +3,11 @@
     "identifier":   "SIMPLEXKerbalism",
     "$kref":        "#/ckan/spacedock/2300",
     "license":      "Unlicense",
+    "x_netkan_version_edit": {
+        "find": "^(?<version>.+)\\.(?<date>[0-9]{4,8})$",
+        "replace": "${version}",
+        "strict": true
+    },
     "tags": [
         "config",
         "parts",
--- a/NetKAN/SimplexPropulsion.netkan
+++ b/NetKAN/SimplexPropulsion.netkan
@@ -3,6 +3,11 @@
     "identifier":   "SimplexPropulsion",
     "$kref":        "#/ckan/spacedock/2112",
     "license":      "CC-BY-NC-SA",
+    "x_netkan_version_edit": {
+        "find": "^(?<version>.+)\\.(?<date>[0-9]{4,8})$",
+        "replace": "${version}",
+        "strict": true
+    },
     "tags": [
         "config"
     ],
--- a/NetKAN/SimplexPropulsion.netkan
+++ b/NetKAN/SimplexPropulsion.netkan
@@ -8,6 +8,7 @@
         "replace": "${version}",
         "strict": true
     },
+    "x_netkan_epoch": 1,
     "tags": [
         "config"
     ],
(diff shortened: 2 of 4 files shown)


DasSkelett:
Hm, they'll generate separate .ckans that show lower as the current releases. We could either epoch them one last time to make them show higher (which will offer users a "fake" update), or freeze the current .ckans, which will only prompt to update once a real new release with a higher minor version number comes out.
Let's go with the first one, ensures that users won't miss any potential real updates.

HebaruSan:
I kind of think this is what epochs are for, authors can make up crazy schemes and have them reflected in CKAN, but we can still sequentialize them properly. I would be comfortable just continuing to merge those auto-epoch PRs.

DasSkelett:
I like the Debian guideline for epochs:
>  **Note that the purpose of epochs is to  allow  us  to  leave  behind  mistakes  in  version
numbering,  and to cope with situations where the version numbering scheme changes.  It is
not intended to cope** with version numbers containing strings of letters which the  package
management system cannot interpret (such as 'ALPHA' or 'pre-'), or **with silly orderings.**

I don't think epoching with every other version is a good way forward. 

> there may be some risk here of causing issues with support if users are mentioning different version numbers than the author expects

I'm aware of that, and that's why I usually try to avoid that. But in this case, the date part doesn't seem to have any meaning in the version number at all, and a version is uniquely identifiable without it. Furthermore, the author uses the short version themself themself:
image
image
image

HebaruSan:
So what **is** Debian's strategy for dealing with silly orderings? I wonder if they have enough control over their packages' version numbers to just ban them and enforce non-silly orderings (unlike CKAN).

HebaruSan:
> the author uses the short version

OK, I retract my objections on that basis.

DasSkelett:
> So what **is** Debian's strategy for dealing with silly orderings? I wonder if they have enough control over their packages' version numbers to just ban them and enforce non-silly orderings (unlike CKAN).

Yeah there definitely is a difference, I'm pretty sure they maintain the version numbers "manually" (with automation scripts and stuff, most likely), as in they can rewrite and reformat the upstream version number to their own needs.
We are definitely limited there, which is not per se a bad thing, as we want to keep the version numbers as close as possible to upstream, but in this case it's possible to automate with the regex, and I think the level of alteration is acceptable.

PR #8648, opened 2021-07-15
Update VenStockRevamp mods' max KSP version to 1.10.1

NathanKell:
These work fine on 1.10 as far as I can tell, and I've seen no reported issues from anyone on RO using them in the last long while that RO's been on 1.10. So I forum-PM'd @kerbas-ad-astra to ask it would be ok to override the max version to 1.10 and was given the OK.

So here we are! It's _probably_ safe to bump them up further, but I can't speak to versions after 1.10 and kerbas is planning to release a new version targeting 1.12 Soon (tm).

### DIFF
--- a/NetKAN/VenStockRevamp-Core.netkan
+++ b/NetKAN/VenStockRevamp-Core.netkan
@@ -24,5 +24,11 @@
         "find"       : "VenStockRevamp",
         "install_to" : "GameData",
         "filter"     : [ "PathPatches", "Squad", "Part Bin" ]
+    } ],
+    "x_netkan_override": [ {
+        "version": "v1.15.1",
+        "override": {
+            "ksp_version_max": "1.10.1"
+        }
     } ]
 }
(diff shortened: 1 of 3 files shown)


HebaruSan:
There's a remote version file. If the author is on board, let's do this over there:

https://github.com/Kerbas-ad-astra/Stock-Revamp/blob/master/GameData/VenStockRevamp/VenStockRevamp.version

NathanKell:
Ah, good point!

HebaruSan:
Version file update merged, metadata updated by bot here:

- https://github.com/KSP-CKAN/CKAN-meta/commit/f5784d01c4bf87c97de0bb132a0a4cbe9243faa3
- https://github.com/KSP-CKAN/CKAN-meta/commit/8c8ac291a6a6f0e022c08429c936be60fee52176
- https://github.com/KSP-CKAN/CKAN-meta/commit/647a95ff0a6daa1c45587586ca5d4ec58b77b074

PR #8665, opened 2021-07-24
Add TUFX to CKAN

NathanKell:
I believe I have done this correctly; it's been a while. In particular, this needs to use a remote version file as hosted on github here ( https://github.com/shadowmage45/TUFX/blob/master/GameData/TUFX/TexturesUnlimitedFX.version ) -- I copied the vref and kref lines from the Ven Stock Revamp core netkan which @HebaruSan pointed out has a remote version link so hopefully that is correct. (The .version in the archive specifies a URL, so I think that will work? The point is that the remote file is good for KSP 1.12 while the release one is older).

(PR'd by permission of @shadowmage45 )

https://github.com/shadowmage45/TUFX
https://forum.kerbalspaceprogram.com/index.php?/topic/192212-19x-tufx-post-processing/

### DIFF
--- a/NetKAN/TUFX.netkan
+++ b/NetKAN/TUFX.netkan
@@ -10,3 +10,5 @@ resources:
 tags:
   - plugin
   - graphics
+depends:
+  - name: ModuleManager
--- a/NetKAN/TUFX.netkan
+++ b/NetKAN/TUFX.netkan
@@ -1,0 +1,20 @@
+{
+    "spec_version": "v1.16",
+    "identifier":   "TUFX",
+    "name":         "TUFX",
+    "abstract":     "Post-processing FX for KSP",
+    "$kref":        "#/ckan/github/shadowmage45/TUFX",
+    "$vref":        "#/ckan/ksp-avc",
+    "license":      "GPL-3.0",
+    "resources": {
+        "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/192212-*"
+    },
+    "tags": [
+        "plugin",
+        "graphics"
+    ],
+    "install": [ {
+        "find":       "TUFX",
+        "install_to": "GameData"
+    } ]
+}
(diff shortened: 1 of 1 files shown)


NathanKell:
Oh and the archive includes ModuleManager but I don't see it actually used, so I did not depends it.

HebaruSan:
Oh nice, we've got a couple of other TUFX configs waiting for TUFX to get indexed (see #8522), mental note to come back to those after this...

NathanKell:
Awesome! :)

HebaruSan:
The readme.md in the ZIP says directly that TUFX won't work without MM:

> TUFX relies upon ModuleManager for database (re)loaded callbacks, and will not function unless 
ModuleManager is installed. This dependency is included in the distribution package and should be
 installed into the root GameData/ folder.

Let's add the ModuleManager dependency.

HebaruSan:
Should the `GameData/TUFX/Profiles/TUFX-Default.cfg` file be split into its own TUFX-Profile-Default module? In other words, if I want to use a different custom profile, would it be a problem to have this default file installed as well?

NathanKell:
Oooh, good spot. I read through the cfgs and saw no patching, but didn't realize the source itself depended. Whoops. :]

HebaruSan:
Kind of looks like it could be harmless to have the default config installed:

image

If the user doesn't want to use it, they can select another profile, sounds like.

NathanKell:
Re: default. It is no problem to have that config installed as well. TUFX loads all configs and you can set what you like ingame.

NathanKell:
Ah, ninja'd.

HebaruSan:
Everything else checks out from what I can see. Thanks for doing the legwork on this. :+1:

NathanKell:
My pleasure!

shadowmage45:
Just dropping by to say that yes, I am fine with TUFX being added to CKAN.  Let her rip!

PR #8666, opened 2021-07-25
Retarget Kerbal Konstructs

NathanKell:
This retargets both KK itself and the CustomPrelaunchChecks dependency until GER-Space (or other main collaborators) are around to restart plugin progress.

- https://github.com/KSP-RO/Kerbal-Konstructs
- https://github.com/KSP-RO/CustomPreLaunchChecks


NathanKell:
(draft until KK release)

NathanKell:
KK release looks good.

DasSkelett:
Was this discussed with the previous authors? Just so we don't end up in a continuous fork switcharound, leading to confusion between users which one is the currently indexed one (a la B9PW...)

NathanKell:
I spoke with @HebaruSan a bunch about this yesterday. We haven't been able to reach GER-Space at all. If they turn up again we're happy to defork.

DasSkelett:
That's exactly what I fear... You fork now, without own forum thread. People will complain on the old forum thread about "incorrect" CKAN metadata (especially if you change the dependencies or whatever one day), nobody knows which fork of the mod the other person is talking about.
Now GER_Space comes back, continues their version, and maybe even decide not to incorporate your new features and bug fixes. If there wasn't chaos before, now there definitely is.

We just had this whole situation with B9PW, and it was a nightmare. I want to do everything to avoid similar things happening again.


If it's not certain that GER_Space will _not_ come back, I'd prefer to handle this with new netkans and provides. This way, we have both of them separate, each team can work on their version however and whenever they want. Users will see a clear distinction, they can choose which one they want (or RO can add an explicit dependency on the RO team's). Preferably also with new forum thread for the new fork, but that's your decision...

HebaruSan:
@NathanKell, maybe we should have had that chat on a public channel of the CKAN Discord, so other members could be aware of it. The "metadata" channel probably would have been a good choice.

DasSkelett:
@HebaruSan you have more insight. How do you feel about it? Is the risk of it backfiring manageable? Is the author gone for good? Or will we end up juggling forks back and forth again?

HebaruSan:
I guess I'm still on the fence. There are real pros and cons on both sides, and the two options seem relatively evenly matched to me; the tie-breaker in my mind was the opportunity to have existing users upgrade to a new, supported fork automatically. But if you feel strongly that a new identifier is the better way, then let's do that.

The point about the forum thread is a good one as well. Users do tend to get confused when the info in a thread title and OP doesn't match the CKAN metadata, as I think is already the case for a few of the KSP-RO adopted mods. @NathanKell, would you consider starting a new forum thread for KK? That way there would be somewhere users could go to try to get support. It would also provide non-CKAN users a way to find out that your fork exists.

NathanKell:
Ah I should join the CKAN discord then. Sorry!

Sure, I can make a thread for that. What adopted mods are you thinking of that don't have correct forum threads? We'll deal with that too.

I am also fine with a new identifier as well if any of you folks feel strongly that that's better.

DasSkelett:
I think what makes me hesitate here is that the original author hasn't been inactive for a long time. They last visited the forum on June 23, which could just be a little modding break due to real life, but basically they could come back in a week or two (where it's much less likely if they've been gone for a year or so).

A new forum thread could change things for me though, we would have two clearly separate mods/forks outside of CKAN with their own places of support and information. If we switch the existing netkan over to the new repo & forum thread, it would be clear which one it is. _If_ GER_space comes back and picks up their own fork (instead of collaborating with the RO team), it would still be a mess to sort out in CKAN (possibly having to split the mods retroactively), but at least we would deal with two distinct mods, which should reduce confusion amongst users.

tl;dr: Making the forks proper forks with own forum threads instead of keeping them as a "unofficial maintenance release collection for users who know where to find it" would improve the situation enough that I'd be okay with re-using the existing identifier as well.

HebaruSan:
> What adopted mods are you thinking of that don't have correct forum threads? We'll deal with that too.

The one that comes to mind most readily is ProceduralParts. I often see people asking whether it is going to be updated, then told to use the KSP-RO fork, then they ask what KSP-RO fork because they don't know about it. In fact, this has just happened in that thread:

image

https://github.com/KSP-RO/ProceduralParts

NathanKell:
Ok, I've reached out one more time (via the forums rather than discord--hopefully that will reach though the other has not). If no reply there either, then we'll make a new thread.

As to PP and PF, yeah, we're dealing with that. Asked Pap to update the PF OP since he's the OP, and we're gonna create a PP thread. Thanks for the reminder.

NathanKell:
This is ready to go now that we reached AlphaAsh and I've started a new thread.

DasSkelett:
Should we update the CustomPrelaunchChecks link also to the new KerbalKonstructs forum thread? Looking at the old one, there pretty much wasn't any activity, probably because it's a library only used by KK; so a separate one might be overkill, but it would be nice not to point to GER_space's.

NathanKell:
Oh shoot yes forgot. Sorry!

Sent by my thumbs, slowly.

DasSkelett:
Okay, updated that URL as well. Sorry for being so persistent about this, but I think ultimately it's better for users, for us _and_ for you if there's clear distinction and a distinct place for new information and support for new forks.

Looks good now, thanks a lot!

NathanKell:
No worries, I 100% agree! Thanks! :)

Sent by my thumbs, slowly.

PR #8669, opened 2021-07-25
Create KSPCommunityFixes.netkan

gotmachine:


HebaruSan:
Hi @gotmachine, thanks for the submission. Would you like a PR on the KSPModdingLibs/KSPCommunityFixes repo to fix that error, or are you already on the case?

gotmachine:
I'm trying to fix it, look like I forgot quotes around the NAME value in the version file

HebaruSan:
Close, there are quotes, but a comma is missing.

HebaruSan:
As always, I must plug one of my favorite validation tools, this can alert you to such problems before release:

https://github.com/DasSkelett/AVC-VersionFileValidator

gotmachine:
Nice one, I will use that in the future.
Meanwhile, the check still fails. Do I need to update the version file in the latest release too ?

HebaruSan:
Right, unfortunately that's one of the weaknesses of the KSP-AVC architecture—the version file inside the release ZIP is the starting point for the whole thing, and if it has a problem, then the whole release is kind of tainted as far as KSP-AVC is concerned. A new release or a fixed re-upload of the same release would fix it.

gotmachine:
Ok, will update the release

gotmachine:
Looks alright now, thanks for the quick help (as usual) :)

PR #8674, opened 2021-07-26
Unfreeze mods adopted by zer0Kerbal

DasSkelett:
"The Gold Standard" has been adopted by @zer0Kerbal. Now that the original author agreed on the adoption (in a forum PM) and zer0Kerbal got access to the SpaceDock entry, we can unfreeze this mod.

While going through the lists of suggests, I noticed that Biomatic has also been adopted by zer0Kerbal, but only after we froze it due to staleness, so we never indexed zer0Kerbal's version.
On CKAN, an optional config file that enables the Biomatic part on all crafts/modules, is split into a separate "BiomaticForAll" module, so it stays disabled when users upgrade the mod. Since it's enabled by default nowadays, I've moved it from suggests to recommends.
There's another optional config file, which makes the part available from the beginning in science/career games. It is disabled by default, having a file extension of `.cfg.OFF`. This doesn't need to be split into a separate mod, as CKAN will keep a renamed copy between updates, only potentially adding a second `.cfg.OFF` again, which has no effect.

Quick links:
- https://forum.kerbalspaceprogram.com/index.php?/topic/203990-v1112-the-gold-standard-minable-gold-and-unobtainium-for-profit-1600-24-jul-2021/
- http://spacedock.info/mod/963
- https://github.com/zer0Kerbal/TheGoldStandard/
- https://forum.kerbalspaceprogram.com/index.php?/topic/191426-173-biomatic-biome-sensor-1320-land-ho-2020-02-03/
- https://spacedock.info/mod/634/Biomatic
- https://github.com/zer0Kerbal/Biomatic/

@zer0Kerbal FYI


zer0Kerbal:
Thank you! 🎈 
BiomaticForAll needs to be off - there is a issue with the code - if multiple ModuleBiomatic's are in physics range - all of them show up at once on the stock command bar. It is an issue I've been struggling with to no avail; LGG has helped me understand what needs to be fixed (split the code up into KSPModule and button code, class code) but that is well beyond my current abilities as a KSP code Modder. Soon™. 😞  😢

DasSkelett:
Hm, so should we freeze BiomaticForAll again, but keep the filter? Or just move it from "recommends" to "suggests" so it's off by default but available for those who feel adventurous?

zer0Kerbal:
`move it from "recommends" to "suggests" so it's off by default but available for those who feel adventurous?`
💭 ^^^this 🌮

PR #8676, opened 2021-07-26
New maintainer for xScienceContinued

linuxgurugamer:

PR #8677, opened 2021-07-26
Fix xScienceContinued dependencies

HebaruSan:
#8676 switched this mod to a new fork, which has some dependencies that weren't added to the netkan. Now they are.

image
image


HebaruSan:
Note, inflation will fail because of the current release is built wrong:

image

PR #8682, opened 2021-07-27
Update SimpleConstruction.netkan

zer0Kerbal:
🎬  2️⃣  

Was processing a cry for help on the forums when I saw the 'Updated' tag. Still the same mod so don't need that. :D adopted with full permission and blessing of original mod author.
📦 〽️ 

also - thought I would update the listings - haven't a clue what 'DeadskinsMeshSwitch' is still doing there. :D

Closes KSP-CKAN/CKAN-meta#2396

### DIFF
--- a/NetKAN/SimpleConstruction.netkan
+++ b/NetKAN/SimpleConstruction.netkan
@@ -1,8 +1,8 @@
 {
     "spec_version": "v1.4",
     "identifier":   "SimpleConstruction",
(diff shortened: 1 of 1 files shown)


HebaruSan:
Thanks for the update.

zer0Kerbal:
> Thanks for the update.

always. thanks for the assist!

PR #8686, opened 2021-07-29
Fix missing author field for Staged Animation

NathanKell:


DasSkelett:
Thanks. With the big number of KSP-RO organization members I guess it makes sense to trim the list to the ones that are the primary maintainers of that mod ^^

Issue #8688, opened 2021-08-01
[Feature] Enhancing Install experience for complex mods/suites (RO/RP-1 use case)


NathanKell:
Opened this in replacement of https://github.com/KSP-CKAN/NetKAN/issues/8615

Right now the install experience for RO/RP-1 is quite complex. We have a lengthy install guide on our wiki to cover how to do it. I've been spending the last month or so trying to get all the mods we depend on available for 1.10, and we're mostly there (there are basically only two remaining mods that are issues). But it's also the case that, even if everything is _available_ on CKAN, the process of selection and install of mods is still quite complex.

The traditional CKAN flow is to select a mod or mods to install. That autoselects dependencies. Then, on apply, the user is presented with a set of recommended mods (which default to checked) and a set of suggested mods (which default to unchecked). Further, any dependencies that can be met multiple ways result in a prompt to select a mod to fulfill the dependency.

Here are the pain points in that process, for RO/RP-1 (at least those pain points I'm thinking about right now):
1. The set of recs and suggests are dependent on which mod(s) the user manually selects. I.e. the experience is different between selecting only RP-1 manually, and selecting both RO and RP-1 manually.
2. There is no in-app guidance offered on which recommendeds Really Should Be selected and which are more optional but still very much recommended. For a large list of recs (given 1 it could be very large!) this is problematic.
3. There is no in-app guidance on selecting which mod to use to fulfill a depends. 
4. MechJeb requires the addition of an additional metadata repository.
5. The download size is very, very large, and if a mod that includes a bunch of assets is updated, the entire mod is redownloaded.

Now, (5) is an issue we're going to address on our side by splitting up plugin/cfg and assets. And hopefully (4) can be addressed with some level of stability in MJ where the RO-specific stuff gets merged into mainline MJ.
(2) I think is a mostly unsolveable problem given the structure of CKAN--I would not suggest adding another level of rec/suggest to get around that; instead my solution will be to have a "simplified install" metapackage that shifts as much as possible into depends.

(1) is tricky. It's desired behavior in normal circumstances. I think the only reasonable answer for RO/RP-1 is that "simplified install" metapackage, and our install guide referencing only that. If users are experienced enough to install individual packages on CKAN, they're experienced enough (I hope) to make reasonable decisions.

Now, as to (3). Here, per the other issue, I think there's some nice low-hanging fruit in the CKAN client, the ability to add a comment that is displayed when a user has to select amongst different mods that fulfill a dependency.

[...]

PR #8595, opened 2021-06-26
Unfreeze RSSVE

HebaruSan:
See comments on #7259, @Theysen wants to get this re-indexed.

- Converted to YAML format
- Switched to KSP-RO team repo
- Removed Scatterer-config and Scatterer sunflare dependencies because Scatterer depends on those anyway
- Updated compatibility to 1.10 and higher
- Removed TextureReplacerReplaced recommendation

PR #8689, opened 2021-08-02
Make RSS recommend RSSVE-Textures

HebaruSan:
Small fix we noticed in the course of #8688.

https://github.com/KSP-CKAN/NetKAN/issues/8688#issuecomment-890592737

> What can we do to make as many of the visual enhancements, that I do think most players install, easiest to get?

https://github.com/KSP-CKAN/NetKAN/issues/8688#issuecomment-890598647

> Should they be recommended or suggested?

https://github.com/KSP-CKAN/NetKAN/issues/8688#issuecomment-890606367

> RSS should rec RSSVE, yes. (And thus by the non-transitive property of recs, RO/RP-1 should too :D ).

https://github.com/KSP-CKAN/NetKAN/issues/8688#issuecomment-890634780

> That may be something easy we can fix in the short term, then; RSS currently only recommends RO, KSCSwitcher, and RSSDateTimeFormatter. Do you want to switch this to a metanetkan hosted on your repo for easier updates?
> 
> https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/RealSolarSystem.netkan

https://github.com/KSP-CKAN/NetKAN/issues/8688#issuecomment-890636971

> Yeah think that probably makes sense, I've just held off touching the RSS metadata pending getting Kerbal Konstructs on 1.10 so I can do everything all at once (I.e. rec RSSVE, RSS-CanaveralHD, and CanaveralPads).

Now `RealSolarSystem` recommends `RSSVE-Textures`, which `RSSVE-LR` and `RSSVE-HR` provide (see #8595 for when we unfroze those modules).

### DIFF
--- a/NetKAN/RealSolarSystem.netkan
+++ b/NetKAN/RealSolarSystem.netkan
@@ -1,38 +1,38 @@
-{
-    "spec_version"    : "v1.12",
-    "identifier"      : "RealSolarSystem",
-    "name"            : "Real Solar System",
-    "abstract"        : "Resizes and rearranges the Kerbal system to more closely resemble the Solar System",
-    "author"          : [ "NathanKell", "eggrobin", "raidernick", "PhineasFreak" ],
-    "$kref"           : "#/ckan/github/KSP-RO/RealSolarSystem",
-    "$vref"           : "#/ckan/ksp-avc",
-    "x_netkan_force_v": true,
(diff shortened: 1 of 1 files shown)


HebaruSan:
image

NathanKell:
Thanks!

PR #8690, opened 2021-08-02
Try to clarify KSPBurst abstracts

HebaruSan:
## Problems

During investigation of #8688, this prompt threw me for a loop:

image

I just want a working install. Will the "Lite" version work, or is it some kind of cheaty bypass? What does any of this even do?

## Changes

I've tried to rewrite the abstracts of those modules to clarify what they're about. The main KSPBurst module now explains that it's about boosting multithreaded performance, and the Lite module says it can do the same job but more slowly. Users should feel more confident choosing between them now.


DasSkelett (approved):
Ich guess it gets the gist, more user friendly than talking about compilers.

PR #8694, opened 2021-08-03
Changed recommends to depends

linuxgurugamer:
Chaned to yaml
Changed recommends to depends

### DIFF
--- a/NetKAN/DavonSupplyMod.netkan
+++ b/NetKAN/DavonSupplyMod.netkan
@@ -1,24 +1,19 @@
-{
-    "spec_version": 1,
-    "x_netkan_epoch": "1",
-    "identifier":   "DavonSupplyMod",
-    "$kref":        "#/ckan/spacedock/364",
-    "$vref":        "#/ckan/ksp-avc",
-    "license":      "GPL-2.0",
-    "resources" : {
-        "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/192101-*"
-    },
-    "tags": [
-        "plugin",
-        "resources"
-    ],
-    "depends": [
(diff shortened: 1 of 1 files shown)


HebaruSan:
@DasSkelett, I guess the API returns None when the body is empty; `.get` would only default to `''` if the `'body'` key is absent:

https://github.com/KSP-CKAN/xKAN-meta_testing/blob/815aed72f2195243964866ccfe0d383662b4ed9c/ckan_meta_tester/__init__.py#L42

```python
            return resp.json().get('body', '')
```

It's probably a good idea to raise at least a warning, possibly an error, when the body is empty, but it should be an explicit check rather than a null reference exception.

HebaruSan:
```
>>> {'a':1, 'b':2}.get('a', '')
1
>>> {'a':1, 'b':2}.get('c', '')
''
>>> {'a':None, 'b':2}.get('a', '')
>>> 
```

DasSkelett:
Archiving stack trace so I don't have to look for it again later.
```
Traceback (most recent call last):
  File "/usr/local/bin/ckanmetatester", line 11, in <module>
    load_entry_point('CkanMetaTester==0.1', 'console_scripts', 'ckanmetatester')()
  File "/usr/local/lib/python3.7/dist-packages/ckan_meta_tester/__init__.py", line 24, in test_metadata
    environ.get('INPUT_DIFF_META_ROOT'))
  File "/usr/local/lib/python3.7/dist-packages/ckan_meta_tester/ckan_meta_tester.py", line 59, in test_metadata
    overwrite_cache = ('#overwrite_cache' in pr_body)
TypeError: argument of type 'NoneType' is not iterable
```

PR #8697, opened 2021-08-05
Continue unfreezing MiniSampleReturnCapsule

HebaruSan:
I meant to start a pull request with https://github.com/KSP-CKAN/NetKAN/commit/c2155c7ec9ca187623e8f7dec699c78aab6f1c78 but accidentally committed it to master instead. The bot reported these warnings:

image

So far these changes are not available to users yet due to auto-epoching, see KSP-CKAN/CKAN-meta#2426.

Now it has tags and an MM dep.

Replaces KSP-CKAN/CKAN-meta#2425.

### DIFF
--- a/NetKAN/MiniSampleReturnCapsule.netkan
+++ b/NetKAN/MiniSampleReturnCapsule.netkan
@@ -3,3 +3,7 @@ spec_version: v1.4
 $kref: '#/ckan/spacedock/831'
 $vref: '#/ckan/ksp-avc'
 license: CC-BY-NC-SA-4.0
+tags:
+  - parts
+depends:
+  - name: ModuleManager


zer0Kerbal:
# Thanks!
### issue 
- it isn't dependent upon MM. 
- Primary license is SimpleBSD-2 - CC only for art assets (in /Assets/)
-

HebaruSan:
- The RemoteTech.cfg file uses MM syntax.
  image
- The `license` field is supposed to use the more restrictive license when there are multiple. NC-SA blocks commercial use and limits what licenses can be used to redistribute, so that wins the restrictive race.

zer0Kerbal:
- True - is an optional patch - only installed if one has RemoteTech installed. Many don't. *shrug*
- understood - removing the CC on SpaceDock and leaving the SimpleBSD-2

am learning. albeit perhaps like a 🐢

PR #8700, opened 2021-08-08
Correct minversion for Scatterer

NathanKell:
There is a netkan version override specifically for 723, but since that was done @LGHassen has released new versions of Scatterer. I believe the correct solution here -- since the forum OP and docs make clear that all these versions are compatible with 1.9+ -- is to apply this override to all versions going forward, until such time as there is a version that is not backwards compatible with 1.9 etc. The issue here appears to be that versioning from Spacedock has only the active KSP version, not a min/max, or such was done when new Scatterers were released, despite other sources maintaining that they are backwards compatible.

### DIFF
--- a/NetKAN/Scatterer-config.netkan
+++ b/NetKAN/Scatterer-config.netkan
@@ -26,7 +26,7 @@
         }
     ],
     "x_netkan_override": [ {
-        "version": "3:v0.0723",
+        "version": ">=3:v0.0723",
         "override": {
             "ksp_version_min": "1.9"
         }
--- a/NetKAN/Scatterer-sunflare.netkan
+++ b/NetKAN/Scatterer-sunflare.netkan
@@ -26,7 +26,7 @@
         "install_to": "GameData/scatterer/config"
     } ],
     "x_netkan_override": [ {
-        "version": ">3:v0.0723",
+        "version": ">=3:v0.0723",
         "override": {
             "ksp_version_min": "1.9"
         }
--- a/NetKAN/Scatterer-sunflare.netkan
+++ b/NetKAN/Scatterer-sunflare.netkan
@@ -26,7 +26,7 @@
         "install_to": "GameData/scatterer/config"
     } ],
     "x_netkan_override": [ {
-        "version": "3:v0.0723",
+        "version": ">3:v0.0723",
         "override": {
             "ksp_version_min": "1.9"
         }
--- a/NetKAN/Scatterer.netkan
+++ b/NetKAN/Scatterer.netkan
@@ -25,7 +25,7 @@
         }
     ],
     "x_netkan_override": [ {
-        "version": "3:v0.0723",
+        "version": ">=3:v0.0723",
         "override": {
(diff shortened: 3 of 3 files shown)


HebaruSan:
Have you confirmed this change with the mod author?

NathanKell:
Given the mod author's own thread on the forums says that this version is compatible with 1.9+, both in the title and in the OP's body, and based on what you did in the prior PR ( https://github.com/KSP-CKAN/NetKAN/pull/8314 ) I figured it was just SpaceDock continuing to SpaceDock (i.e. list only a single compatible version, which has to be the most recent KSP version supported or users with that version of KSP, the normal use case, will not see it as supported when checking on SpaceDock). But I can certainly ask the author to confirm that their OP is still correct.

HebaruSan:
> the mod author's own thread on the forums

How about a link or screenshot fragment so we don't have to go hunting?

NathanKell:
Ah sorry, sure!
image

NathanKell:
Link, also: https://forum.kerbalspaceprogram.com/index.php?/topic/103963-wip19-112-scatterer-atmospheric-scattering-00772-18062021/

NathanKell:
Should I still ask? I'm happy to do so and wait on a response, I was just surprised given the history and the forum OP.

HebaruSan:
Cool, thanks! That works. I'm going to take a look at the validation script output and some of the intermediate versions, but no, you don't need to bother blackrack/LGhassen.

NathanKell:
Ok, cool. Thanks, and sorry for the confusion!

HebaruSan:
OK, previous change was #8314, basically the same idea, the forum thread title also said 1.9 back then. :+1:

NathanKell:
Yeah I switched it to a >= because I presume blackrack will continue making releases, though it merely shifts the versioning burden from "presumptively not backwards compatible" to "presumptively compatible"--so if there ever is, indeed, a compat break this will need revisiting.

LGhassen:
Hi @HebaruSan, I'm about to upload a new version (v0.0824) which will only be compatible with 1.12. I'd modify the files myself and make a PR but didn't quite understand if I can keep `>=3` on the previous version and add the new version, or if I need to explicitely list all versions since 0.0723, so I will leave it up to you.

HebaruSan:
Thanks for letting us know, @LGhassen. #8857 should address that, but unfortunately our pull request validation scripts are choking on SpaceDock's expired SSL certificate. I'll re-run them when @V1TA5 gets around to fixing the server, which hopefully will be very soon.

HebaruSan:
> when @V1TA5 gets around to fixing the server, which hopefully will be very soon.

Heh, it happened almost exactly as I was typing that, maybe 1 minute afterwards. :tada:

HebaruSan:
OK, it should be safe to make your new release now. I'll keep an eye on the metadata to make sure it works as expected.

LGhassen:
Thank you, I will get to it in a few hours. Cheers

PR #8701, opened 2021-08-08
Add RO/RP-1 express install netkans

NathanKell:
## Purpose
This adds "express install" packages (a main package and three choices for graphics level) to allow express installation of RO/RP-1/RSS/etc.

## Motivations
RO/RP-1 users often ask for a simple gamedata zip, or "what mods should I install", or the like--to the point of asking which recommended or suggested mods to install, which graphics mods, etc.

## Changes
This adds a single express install netkan that depends on the core mods for an RP-1 install, and then depends on an RSSGraphics package. That package is provided by three express packages, each reporting only as a detail level, which depend the various RSS-Textures packages, visual enhancements, and Canaveral terrain/pads necessary for a complete experience.

Metanetkans for easy reference:

- https://raw.githubusercontent.com/KSP-RO/RP-1-ExpressInstall/main/RP-1-ExpressInstall.netkan
- https://raw.githubusercontent.com/KSP-RO/RP-1-ExpressInstall-Graphics/main/RP-1-ExpressInstall-Graphics-High.netkan
- https://raw.githubusercontent.com/KSP-RO/RP-1-ExpressInstall-Graphics/main/RP-1-ExpressInstall-Graphics-Medium.netkan
- https://raw.githubusercontent.com/KSP-RO/RP-1-ExpressInstall-Graphics/main/RP-1-ExpressInstall-Graphics-Low.netkan

## Issue ref: #8688

Once https://github.com/KSP-CKAN/CKAN/pull/3426 is merged, I will update the local netkans with comments on the graphics choice.

### DIFF
--- a/NetKAN/RP-1-ExpressInstall-Graphics-High.netkan
+++ b/NetKAN/RP-1-ExpressInstall-Graphics-High.netkan
@@ -1,0 +1,9 @@
+spec_version: v1.12
+identifier: RP-1-ExpressInstall-Graphics-High
+$kref: '#/ckan/netkan/https://github.com/KSP-RO/RP-1-ExpressInstall-Graphics/raw/main/RP-1-ExpressInstall-Graphics-High.netkan'
+x_netkan_license_ok: true
+tags:
+  - config
+  - planet-pack
+  - resources
+  - graphics
--- a/NetKAN/RP-1-ExpressInstall-Graphics-Low.netkan
+++ b/NetKAN/RP-1-ExpressInstall-Graphics-Low.netkan
@@ -1,0 +1,9 @@
+spec_version: v1.12
+identifier: RP-1-ExpressInstall-Graphics-Low
+$kref: '#/ckan/netkan/https://github.com/KSP-RO/RP-1-ExpressInstall-Graphics/raw/main/RP-1-ExpressInstall-Graphics-Low.netkan'
+x_netkan_license_ok: true
+tags:
+  - config
+  - planet-pack
+  - resources
+  - graphics
--- a/NetKAN/RP-1-ExpressInstall-Graphics-Medium.netkan
+++ b/NetKAN/RP-1-ExpressInstall-Graphics-Medium.netkan
@@ -1,0 +1,9 @@
+spec_version: v1.12
+identifier: RP-1-ExpressInstall-Graphics-Medium
+$kref: '#/ckan/netkan/https://github.com/KSP-RO/RP-1-ExpressInstall-Graphics/raw/main/RP-1-ExpressInstall-Graphics-Medium.netkan'
+x_netkan_license_ok: true
+tags:
+  - config
+  - planet-pack
+  - resources
+  - graphics
--- a/NetKAN/RP-1-ExpressInstall.netkan
+++ b/NetKAN/RP-1-ExpressInstall.netkan
@@ -1,0 +1,13 @@
+spec_version: v1.12
+identifier: RP-1-ExpressInstall
+$kref: '#/ckan/netkan/https://github.com/KSP-RO/RP-1-ExpressInstall/raw/main/RP-1-ExpressInstall.netkan'
+x_netkan_license_ok: true
+tags:
+  - plugin
+  - config
+  - parts
+  - career
+  - science
+  - planet-pack
+  - resources
+  - graphics


NathanKell:
Side note: I just learned NetKAN's inflation error reports the line number and error for an internal JSON representation of the YAML spec. I was at first quite surprised to hear of an error on a line, and with a character, that my file did not include!

NathanKell:
I believe this is now correct, though the checks are still failing--this seems to be because the CLI version of CKAN defaults to installing recommended mods, and the main netkan depends on both Ven's New Parts (which recs Ven's Stock Revamp) and on ReStock, and of course VSR and ReStock conflict.

Is there some way to get the checker to run without auto-installing recs?

HebaruSan:
> Side note: I just learned NetKAN's inflation error reports the line number and error for an internal JSON representation of the YAML spec. I was at first quite surprised to hear of an error on a line, and with a character, that my file did not include!

Does it? I don't think that's the case.

HebaruSan:
> I believe this is now correct, though the checks are still failing--this seems to be because the CLI version of CKAN defaults to installing recommended mods, and the main netkan depends on both Ven's New Parts (which recs Ven's Stock Revamp) and on ReStock, and of course VSR and ReStock conflict.
> 
> Is there some way to get the checker to run without auto-installing recs?

The checker is highlighting a real problem. These probably shouldn't recommend conflicting modules.

HebaruSan:
Oops, forgot these were metanetkans, I'll remove those depends, but can you add them to your repo?

HebaruSan:
We'll need to copy this metadata locally and try installing with GUI to make sure the desired work flow happens. The command line tools are already easier for the RP-0 identifier because they don't interactively prompt for recommends or suggests.

HebaruSan:
> We'll need to copy this metadata locally and try installing with GUI to make sure the desired work flow happens. The command line tools are already easier for the RP-0 identifier because they don't interactively prompt for recommends or suggests.

What I'm trying in case anyone wants to replicate it:

1. Make a `CKAN-meta-rp1express` folder somewhere locally (doesn't matter where)
2. Create `RP-1-ExpressInstall-v1.0.ckan`, `RSSGraphics-ExpressInstall-Low-v1.1.ckan`, `RSSGraphics-ExpressInstall-High-v1.1.ckan`, and `RSSGraphics-ExpressInstall-Medium-v1.1.ckan` in that folder with their contents as reported by the validation script
3. Make an archive of that folder: `tar czf rp1express.tar.gz CKAN-meta-rp1express`
4. Add the archive as a CKAN repo: `ckan repo add rp1express 'file:///full/path/to/rp1express.tar.gz'`
5. Launch GUI, click Refresh to load the new repo, start testing
   image

HebaruSan:
My compatible versions are 1.10–1.12. Initial change set:

image

Click Apply, one recommendation appears:

image

Click Continue, prompted to choose RSS textures, I'll install the 2K package:

image

Prompted to choose KSPBurst or KSPBurst Lite, I'll pick the non-lite one:

image

Prompted to choose RSS Graphics, I'll try High to see how it handles a mismatch with the textures:

image

Install fails:

image

So it doesn't look like the `RSSGraphics-ExpressInstall` relationships are working as intended. I was under the impression that that choice was supposed to be presented early and drive most of the rest of it. Probably need to reorder the `depends` list a bit.

HebaruSan:
You don't need to use `any_of` for MechJeb because the dev builds have `"provides": "MechJeb2"`. A simple depends on MechJeb2 will prompt the user to choose if they have the dev repo configured. The `any_of` just makes the dev build show up twice:

image

DasSkelett:
> So it doesn't look like the RSSGraphics-ExpressInstall relationships are working as intended. I was under the impression that that choice was supposed to be presented early and drive most of the rest of it. Probably need to reorder the depends list a bit.

Even reordered it prompts me for a RSSTextures resolution later on.  We might have to fix our RelationshipResolver behavior to recalculate it after TooManyModsProvides selections have been made.
Although thinking about it, doesn't it basically restart the installation from start after a selection has been made? So it should already include them for relationship resolution, no? :thinking:

HebaruSan:
Yeah, this is where it should skip that relationship:

https://github.com/KSP-CKAN/CKAN/blob/a8331592dda3bbfed3e42cdeb5e39f6858d23c5c/Core/Relationships/RelationshipResolver.cs#L381-L385

And this is where that call should already be checking provides:

https://github.com/KSP-CKAN/CKAN/blob/a8331592dda3bbfed3e42cdeb5e39f6858d23c5c/Core/Types/RelationshipDescriptor.cs#L113-L118

So... deeper investigation required.

HebaruSan:
I wonder if the ordering of the `depends` inside the low/med/high modules is affecting it. Depending on RSS before the textures might make it pull in RSSTextures before that texture dependency is applied:

```json
      "depends": [
          {
              "name": "RealSolarSystem"
          },
          {
              "name": "RSSTextures4096"
          }
      ],
```

NathanKell:
> > I believe this is now correct, though the checks are still failing--this seems to be because the CLI version of CKAN defaults to installing recommended mods, and the main netkan depends on both Ven's New Parts (which recs Ven's Stock Revamp) and on ReStock, and of course VSR and ReStock conflict.
> > Is there some way to get the checker to run without auto-installing recs?
> 
> The checker is highlighting a real problem. These probably shouldn't recommend (depend on? not sure yet) conflicting modules.

Perhaps the later bit (where you're attempting to install via GUI) supersedes this but, see [here](https://github.com/KSP-RO/RP-0/blob/master/RP-0.netkan#L30) for RP-0's netkan. The issue arises because the express install netkan depends on Restock, but in RP-0.netkan Ven's is the first of the two any_of recs so it is preferentially installed as the CLI blasts through every single mod's recommendeds.

HebaruSan:
> Ven's is the first of the two any_of recs so it is preferentially installed

I'm looking for code that could do this, and I'm not finding it. The resolution of the `any_of` rec _should_ return both modules, which should then cause the user to be prompted to choose.

NathanKell:
@HebaruSan ah then I misunderstood, so the issue is actually worse in CLI than I thought, i.e. it will turn "any_of" into "all_of"? :D

HebaruSan:
> @HebaruSan ah then I misunderstood, so the issue is actually worse in CLI than I thought, i.e. it will turn "any_of" into "all_of"? :D

No, it should be prompting.

NathanKell:
Oh, even prompting in CLI in the action? Yeah that does seem weird.

Well, I'm happysad I provided a great way to find client bugs, I guess? :\

NathanKell:
( MM dependency added here https://github.com/KSP-RO/RSSGraphics-ExpressInstall/commit/3121b77bf870bc7ddade621d589f96827bae0e5e per note upthread.)

HebaruSan:
> EDIT: Nope, fixing that (putting RSSTextures4096 first) _still_ doesn't suppress the RSSTextures prompt.

I think we're getting tripped up by the relationship resolver working depth-first, essentially. After we pick `RSSGraphics-Whatever` out of the list, we restart the install with two modules in the change set, `RP-1-ExpressInstall` and `RSSGraphics-Whatever`, in that order. `RSSGraphics-Whatever` is able to satisfy relationships _itself_ (this is why we don't immediately get prompted to make the same choice again), but its dependencies are not looked at yet; rather, the relationship resolver tries to satisfy all of `RP-1-ExpressInstall`'s dependencies (and depends-of-depends) _before_ it looks at any of `RSSGraphics-Whatever`'s depends.

So the current `RP-1-ExpressInstall` module approach does not seem to be workable. Rather, it seems that the module the user chooses up front needs to have all the "deep depends" in it, so the new modules could be something like:

- `RP-1-ExpressInstall-Low`
- `RP-1-ExpressInstall-Medium`
- `RP-1-ExpressInstall-High`

Each of those modules would have the `depends` that are currently in the `RSSGraphics` modules (with the detail-specific depends such as the 4K textures as early as possible in the list), and then the `depends` that are currently in `RP-1-ExpressInstall`, **later in the list**. That way it should be possible to pre-satisfy the RSSTextures and other dependencies early on.

Sorry for all the confusion, but this is getting really deep into the functionality of the relationship resolver, and we usually treat it as a black box not to be touched unless we're really sure what we're doing.

HebaruSan:
Alternately, if you really want to have a "main" module branching into "detail level" modules, the "main" module could depend **only** on those "detail level" modules and not on RP-0, RO, RSS, etc.

NathanKell:
Ah, I see what you mean. And you answered the question I was in the middle of asking. It is definitely the case that I want a main module, because as I understand the spec, I need to include an installable with each package or it gets treated as a metapackage and (per an issue from lamont a few years back) `conflicts` won't work.

I'll make that change.

NathanKell:
Also, that route is required as well to take advantage of https://github.com/KSP-CKAN/CKAN/pull/3426 - we need a main package to have a depends with a comment.

NathanKell:
@HebaruSan this change is now done. 🤞

HebaruSan:
OK, I'm taking a look now and I restarted the validation script. We might still need to reorder some of the depends, but the overall structure is a better fit to how the relationship resolver works now.

NathanKell:
Thanks! And thanks for taking the time to dig through this all (especially the deep rabbit holes of CKAN black magic :D )

HebaruSan:
Next test. Initial change set is a lot smaller, as expected:

image

Click Apply. One recommendation again:

image

Then the detail level prompt comes up:

image

I'll pick Low. Womp womp, the next prompt is the textures again:

image

Try putting the 4K/8K/whatever depends as early as possible in the list, before anything that could possibly depend on anything that depends on RSSTextures.

NathanKell:
@HebaruSan  Done, and inflator restarted.

HebaruSan:
I need to take off for an event in a bit, but I'll be back in a few hours. The steps I listed above should still work if you're feeling adventurous.

HebaruSan:
@DasSkelett, if `ckan_meta_tester` published its repo tar.gz file as a downloadable artifact, that could make it a lot easier to do client testing like this.

NathanKell:
Fixed! Thanks. :)

There's one remaining issue, which is related to KSPBurst. That's that instead of the usual setup, where both the full KSPBurst and the Lite KSPBurst `provides` something else, which other mods depend on, instead other mods depend _directly_ on KSPBurst, and KSPBurst-Lite `provides` KSPBurst. I _think_ that means there is no way to force install of the full KSPBust--I switched to `depends`-ing it directly but, of course, you still get a prompt because two packages provide it (one natively and one via `provides`).

NathanKell:
(To be clear, I mean: I did the local testing. It was a pain to get working since I couldn't get CKAN to accept a file:// path from windows cmd, but using WSL Ubuntu I could get it to do so. I probably just wasn't formatting the string right somehow.)

NathanKell:
Actually, given that all the depends got pushed down to the RSSGrarphics stuff, I need to rename all of them because they're not RSSGraphics anymore. Will require renaming the KSP-RO repo, redoing the release, there etc., then updating the PR.

NathanKell:
This seems to be fine now (or at any rate is back to the single error on the CLI regarding the multiple recs colliding).

NathanKell:
@HebaruSan is it safe to add `choice_help_text` now, or do I need to wait for the next release? I do not presume it to _work_ until the next release, but if it's safe to add now I will do so now and not have to remember to do so later. :D

HebaruSan:
> @HebaruSan is it safe to add `choice_help_text` now, or do I need to wait for the next release? I do not presume it to _work_ until the next release, but if it's safe to add now I will do so now and not have to remember to do so later. :D

Yes and no; as @DasSkelett discovered in KSP-CKAN/CKAN#3426, the current client throws an exception if an `any_of` relationship has any additional properties, so it's not safe to use `choice_help_text` in an `any_of` relationship. But if you use it in a normal `provides`-based relationship, then the current client would safely ignore it. Please don't crash everybody's clients. :grimacing:

NathanKell:
Awesome. Yeah, I plan to use it on the initial depends (in the main express-install) for selecting a graphics level, to explain how that works; and then for the KSPBurst depends that each of the sub-mods has (due to that dep being repeated across all 3 submods).

That said that can be done later, so from my perspective I think this is ready to go--the flow is fine from testing locally with a dev repo. Anything else you need from me on this?

HebaruSan:
> I think this is ready to go--the flow is fine from testing locally with a dev repo. Anything else you need from me on this?

I'm planning to get #8704 working and then use it to test the GUI work flow myself again. I'll let you know if we find anything else you need to change.

NathanKell:
> > I think this is ready to go--the flow is fine from testing locally with a dev repo. Anything else you need from me on this?
> 
> I'm planning to get #8704 working and then use it to test the GUI work flow myself again. I'll let you know if we find anything else you need to change.

Ok. Cheers!

HebaruSan:
OK, I tried a low-detail install first. Looks good, I didn't get prompted for anything unnecessary as far as I know. I'm going to try medium and high as well just in case since they're pretty independent now.

HebaruSan:
> There's one remaining issue, which is related to KSPBurst. That's that instead of the usual setup, where both the full KSPBurst and the Lite KSPBurst `provides` something else, which other mods depend on, instead other mods depend _directly_ on KSPBurst, and KSPBurst-Lite `provides` KSPBurst. I _think_ that means there is no way to force install of the full KSPBust--I switched to `depends`-ing it directly but, of course, you still get a prompt because two packages provide it (one natively and one via `provides`).

We can do something about this (thanks to `Scatterer-sunflare-default` for reminding me). If `KSPBurst` provides `KSPBurst-Full` (or whatever name we choose), then you could depend on `KSPBurst-Full` to skip the prompt. I'll start a pull request to add that to `KSPBurst`.

HebaruSan:
OK, @NathanKell, it's looking pretty good now. I think these changes should eliminate the remaining unnecessary prompts:

Low: Depend on KSPBurst-Full
Med & High: Depend on KSPBurst-Full and Scatterer-sunflare-default

NathanKell:
> We can do something about this (thanks to `Scatterer-sunflare-default` for reminding me). If `KSPBurst` provides `KSPBurst-Full` (or whatever name we choose), then you could depend on `KSPBurst-Full` to skip the prompt. I'll start a pull request to add that to `KSPBurst`.

Oh! Yes, if we're willing to edit _other_ netkans than that's a perfect solution! Sweet.
I made the changes to the remote netkans.

HebaruSan:
I ran another few tests, and it looks great in GUI now! :+1: 

I'm going to dig a bit deeper into the errors that the CLI is spitting out in the validation scripts, since it would be helpful for future maintenance if these netkans could pass those tests. But I won't let that hold it up for too much longer, if it starts to look like no simple resolution is possible then I'll be OK with merging, since GUI users will have an easy time.

NathanKell:
Ok! Happy to be a guinea pig. :D That'll give us time to update the wiki

And thanks for all the help as always!

Sent by my thumbs, slowly.

HebaruSan:
CLI findings (using the install-all-recs default behavior of `ckan install RP-1-ExpressInstall` since that's how most CLI users would probably use it):

```
$ ckan install RP-1-ExpressInstall 
Module RP-1-ExpressInstall-Graphics is provided by more than one available module. Please choose one of the following:
1) RP-1-ExpressInstall-Graphics-High (RP-1 Graphics High Res Express Install)
2) RP-1-ExpressInstall-Graphics-Medium (RP-1 Graphics Medium Res Express Install)
3) RP-1-ExpressInstall-Graphics-Low (RP-1 Graphics Low Res Express Install)
Enter a number between 1 and 3 (To cancel press "c" or "n".): 
1
Module MechJeb2 is provided by more than one available module. Please choose one of the following:
1) MechJeb2 (MechJeb 2)
2) MechJeb2-dev (MechJeb 2 - DEV RELEASE)
Enter a number between 1 and 2 (To cancel press "c" or "n".): 
2
Module RSSVE-Textures is provided by more than one available module. Please choose one of the following:
1) RSSVE-HR (RSS Visual Enhancements - High Resolution)
2) RSSVE-LR (RSS Visual Enhancements - Low Resolution)
Enter a number between 1 and 2 (To cancel press "c" or "n".): 
1
Module RemoteTech OR RealAntennas is provided by more than one available module. Please choose one of the following:
1) RemoteTech (RemoteTech)
2) RealAntennas (Real Antennas)
Enter a number between 1 and 2 (To cancel press "c" or "n".): 
2
* VenStockRevamp v1.15.1 conflicts with ReStock 1.4.1
Install canceled. Your files have been returned to their initial state.
```

- ReStock should be before RP-0 to fix the problem with the install failing. The VenStockRevamp recommendation is silently suppressible as long as ReStock is queued up to be installed before that relationship is processed.
- High & Medium: The RSSVE depends should be before RealSolarSystem, since the latter recommends RSSVE-Textures and causes a prompt to choose between RSSVE-HR and RSSVE-LR.
  (For Low, the user will be prompted to choose to resolve that recommendation, but that can't be helped. I've made a note that the CLI user should be able to cancel out of resolving a recommendation without aborting the whole install, seems like that should be a valid choice for recommendations, so hopefully that will get added at some point.)
- RealAntennas should be before RealismOverhaul (and probably whatever is pulling in Kerbalism) so CLI doesn't prompt the user to choose between that and RemoteTech.

NathanKell:
Changes made. Thanks!

HebaruSan:
Well that's exciting!

```
  Error: 383414 [1] ERROR CKAN.CmdLine.ConsoleUser (null) - Oh no! We tried to overwrite a file owned by another mod!
  
  If this problem re-occurs, then it maybe a packaging bug.
  Please report it at:
  
  https://github.com/KSP-CKAN/NetKAN/issues/new
  
  Please including the following information in your report:
  
  File           : GameData/FMRS/FMRS_MM.cfg
  Installing Mod : FMRS v1.1.0.1
  Owning Mod     : FMRSContinued
  CKAN Version   : v1.30.5
```

I'll take a look at those mods, this may be something we can resolve without changing anything on the RP-1 side. Guessing they should at least conflict with one another...

NathanKell:
What the heck is pulling in FMRS? That must be deep in the recs of some rec'd mod. Gonna trawl the action log.

They certainly should conflict, though, yes.

NathanKell:
Also probably need to make  KSC Extended incompatible with RSS, I saw that in the install list too. I'll handle that on the RSS side I guess.

NathanKell:
Oh, found the issue. Canaveral Pads pulls in Tundra Space Center which recs Tundra Exploration which pulls in Kerbal Reusability Expansion and FMRS.

Well, this makes _super_ clear that our install guide needs to point out that if you run CKAN from the command line, it needs to _not_ pull in the recs of all mods down the chain. There's a ton of not-compatible-with-RP-1 stuff in there.

HebaruSan:
That does highlight something you may want to change, though. The original FMRS is only compatible from KSP 0.25 to 1.1.99, so we probably don't want to see it here; and it's only being pulled into the installation because the low/med/high modules have no game version compatibility set, whereas the main module has 1.10.0–1.10.1. Could they all share the same `$vref`?

NathanKell:
Oh, yes, they certainly could. I'll make that change.

That's unrelated to the deeper problem, however, which is somewhat particular to RO/RP-1 (as we were discussing on the issue thread), namely that for most Kerbal contexts, mods are compatible unless indicated otherwise, but for RSS/RO/RP-1, for most mods, mods are _in_ compatible unless indicated otherwise. I.e. it makes perfect sense for KK to rec Tundra Space Center, which recs Tundra Exploration, which recs FMRS and a bunch of other things, and it makes perfect sense for KK to rec a bunch of other mods _that assume your home planet is Kerbin_, and it makes total sense for e.g. Near Future Solar to recommend the rest of Nert's Near Future mods, and so forth--but none of that makes sense in the context of RSS/RO/RP-1.

NathanKell:
> Could they all share the same `$vref`?
Uh, maybe I'm misunderstanding the syntax, but how would I go about specifying that? The existing ksp-avc $vref relies on a version supplied in the archive, is there a way to specify a remote version file in the schema? I can find the CKAN schema but I couldn't find a netkan schema doc.

HebaruSan:
> > Could they all share the same `$vref`?
>
> Uh, maybe I'm misunderstanding the syntax, but how would I go about specifying that? The existing ksp-avc $vref relies on a version supplied in the archive, is there a way to specify a remote version file in the schema? I can find the CKAN schema but I couldn't find a netkan schema doc.

You'd have to put the version file in the archive, yes.

NathanKell:
Done.

HebaruSan:
@NathanKell, please don't put RP/RO/RSS-specific stuff on the CKAN user guide. It's fine to clarify how functionality works, but it's not supposed to be a reference for your project.

NathanKell:
@HebaruSan  Oh holy crap I didn't realize I was editing it. :( :( Thought I was still on our wiki.
Gonna revert all that.

NathanKell:
Yup it _looks_ like it's linking within our wiki because it's in our CKAN101's table of contents, but sure enough it's on yours. Fixing.

NathanKell:
There, force-pushed away all my crap. Again, super sorry about that! Thank you for the quick poke!

HebaruSan:
No worries, thanks for cleaning that up. If you do find things on the CKAN user guide that could be clearer, please do feel free to make edits as long as they are generally applicable to all CKAN users.

I think the validation script has stalled out. I'm going to cancel it and start it again in a little while. Fingers crossed that it should succeed soon!

HebaruSan:
The script just stopped at exactly the same line of installing Low as last time:

```
  26.3 MiB/sec - downloading - 724.2 MiB left - 42%           
```

That makes me think we're hitting a resource limit of the workflow containers that we can't bypass. I believe it is actually working based on local testing, which means it's time to merge!

NathanKell:
WOOHOO! (although sad robot noises at the beleaguered actionrunner :D )

NathanKell:
Per my mistaken PR, in trying the express installer out I still got the prompt to choose KSPBurst version. This is despite KSPBurst-Full being the second line of the `depends`, here: https://github.com/KSP-RO/RP-1-ExpressInstall-Graphics/blob/main/RP-1-ExpressInstall-Graphics-High.netkan#L15

HebaruSan:
I think that means you've got some older versions of the metadata active, from before the depends was updated to -Full. Make sure to remove any testing repos and click Refresh to get the latest official metadata.

NathanKell:
Nope, that was the first thing I did on launching CKAN. :(

NathanKell:
(To double check I closed CKAN, started it again, and refreshed again, same flow.)

HebaruSan:
Check your repositories.

NathanKell:
Only one listed, default, pointing to the official archive.

NathanKell:
image

HebaruSan:
OK fine, but it still isn't possible to pull in KSPBurst-Lite by depending on KSPBurst-Full. Some other weird thing is happening that you're not hinting at. Is anyone else able to reproduce this?

NathanKell:
It is not possible to pull it in from depending on KSPBurst-Full, but _RealAntennas_ depends on just `KSPBurst`, and we depend on RealAntennas. Which I think would make it the same sort of issue as the RSS-Textures thing?

I asked on the RO discord for anyone available to try installing in a blank KSP 1.10 install, hope I'm just cursed. :D

NathanKell:
Ugh, and same with the sunflare. Yep, looks to be exactly the same dependency-resolving issue as before with RSS-Textures, and ordering in the netkan has no effect.

HebaruSan:
> It is not possible to pull it in from depending on KSPBurst-Full, but _RealAntennas_ depends on just `KSPBurst`, and we depend on RealAntennas. Which I think would make it the same sort of issue as the RSS-Textures thing?

If that was the case, it would happen when I try it.

NathanKell:
I have no explanation for why I am enountering that and you are not, unless I somehow have cached metadata that removing the repo it is from did not fix. Gonna clear my CKAN cache.

HebaruSan:
You really were trying all of this in parallel while we were looking at it, though? At which point did this prompt re-appear for you?

NathanKell:
Ah, I think we had a miscommunication then maybe? I had the prompt for KSPBurst, you suggest depending on KSPBurst-Full and moving it above. I did so and you responded that it fixed the GUI here ( https://github.com/KSP-CKAN/NetKAN/pull/8701#issuecomment-895501315 ) , so I did not do a local retest at that point. Once it was merged down and in the main CKAN repo I tested now and got the prompts.

NathanKell:
Ok, this just got weirder. DRVeyl reports:
"I was presented with EER and one other as recommended mods... then the textures [which we all expect]... then the TF configs.  Then it started." With this image:
image

HebaruSan:
Yeah, that one I saw. I assumed you changed something after my last test.

NathanKell:
I do not get that prompt.

Wait, what KSP version were you testing against, and what compat versions? I bet that's it. I am on 1.10 with no compat versions; DRVeyl is on 1.10 but with 1.8 and 1.9 checked.

HebaruSan:
Yup, TestFlight got moved around in the High metanetkan:

image

NathanKell:
Ugh, that would do it. Thought I discarded that. :(

HebaruSan:
> I do not get that prompt.

That's even more evidence that you're not testing with the current metadata. Maybe try a completely fresh game install?

NathanKell:
Will do.

NathanKell:
Ok, a completely fresh install and linking that to CKAN fixed the issue, so now I'm only seeing evidence of my _other_ stupidity 😂 (the TestFlight thing) rather than that weird situation with busted metadata. I fixed the local netkans so that should go away as well.

NathanKell:
Circling back to say--it's working! Thank you again, from all of us--besides being I think a very positive option for our users, it'll make supporting them way easier on us. CHeers!

HebaruSan:
That's awesome and great to hear! :tada: 
Where are you expecting to see the bulk of your support feeback? The KSP-RO Discord maybe? I can stop by to observe now and then to see how it's going rather than making you serve as a go-between for everything.

NathanKell:
Yeah, on the RO discord and sometimes on reddit. The forums are pretty quiet these days. But the issues are often less specific questions _from_ installation (though as I've mentioned we do get the "which of these recommended mods should I _really_ install" :D ), and more we find out a good few messages in that it's actually a broken install in one way or another, not a bug in (whatever submod).

And that's super kind of you! You should know that (AFAIK) next to none of the support load is actually from how to use CKAN--your docs, and our CKAN-101, seems to work pretty well, the only issue there has been walking people through adding the MJ-dev repo (which is why I've been poking at sarbs and lamont to merge dev to release, which they did so yay!). Well, that and the "ok but what should I _really_ pick", so having a one-stop "just click this one thing" answer will be super awesome.

HebaruSan:
I always forget KSP-CKAN/CKAN#2682 when repository shenanigans happen. :roll_eyes: 

I think your CKAN client thought your metadata was up to date, because the Etag was the same as the last time you updated, and so it didn't actually do a refresh. Meanwhile you had removed a testing repo but its metadata was still in your registry, and would remain until the master repo's Etag changed. We don't have a good workaround for this currently; I sometimes edit my `<KSP>/CKAN/registry.json` file and delete the `last_server_etag` property to force a refresh.

NathanKell:
AHA! Yes that makes perfect sense. I'll keep that in mind in future, it perfectly explains what was going on. The fresh install (or nuking the local CKAN folder, which presumably would have fixed it too but I wasn't taking chances, fresh install it was!) sorts that out by dint of utterly annihilating the json rather than tweaking it. :]

PR #8702, opened 2021-08-08
Mechjeb2 - add support for current KSP

sarbian:
The new release of Mechjeb2 supports 1.8 to 1.12.
This should be the last required update of NetKAN for MJ.


DasSkelett:

PR #8596, opened 2021-06-26
Update DeadlyReentry netkan to point at currently-active fork

NathanKell:
Well, currently active not-a-fork, since it's actually the head repo. :D

PR #8703, opened 2021-08-08
 Revert DeadlyReentry source back to Starwaster's fork 

DasSkelett:
This effectively reverts #8596 and #8696 to point the netkan back to https://github.com/Starwaster/DeadlyReentry, after @Starwaster became active again and created a new release.

As the new release has a lower version number than NathanKell's releases, the epoch is incremented to prompt users to "upgrade" to the new release.

A pull request in CKAN-meta will freeze the releases from the KSP-RO repository. Depending on the plans of @NathanKell, they can be moved to a new identifier after the corresponding netkan has been created,.


DasSkelett:
Stray `.git` folder detected, adding a filter.

PR #8704, opened 2021-08-08
Upload repo artifacts for pull requests

HebaruSan:
## Motivation

In #8701 we are trying to test GUI work flows for some metadata changes, which requires copying each .ckan file manually from the validation output to local files, then creating a tar.gz file for CKAN to use. But the metadata tester already has such a file created for its own use.

## Changes

Now `/repo/metadata.tar.gz` is uploaded as an artifact for pull requests, so we can download it and use it with CKAN directly.

### DIFF
--- a/.github/workflows/inflate.yml
+++ b/.github/workflows/inflate.yml
@@ -40,5 +40,12 @@ jobs:
                   source: commits
                   diff meta root: .CKAN-meta
                   pull request url: ${{ github.event.pull_request.url }}
+            - name: Upload inflated repo metadata.tar.gz artifact
+              if: ${{ github.event_name == 'pull_request' }}
+              uses: actions/upload-artifact@v2
+              with:
+                  name: metadata.tar.gz
+                  path: /repo/metadata.tar.gz
+                  retention-days: 7
             - name: Chmod cached files so actions/cache can read them
               run: sudo chmod -R a+r .cache


DasSkelett (approved):
Nice idea :+1:

HebaruSan:
Ugh, GitHub, containers, why must you torment me??

image

(This is from #8701, after some inflations and installations that most definitely used that exact file.)

HebaruSan:
Maybe I need to chmod it? :angry:

HebaruSan:
image

:confused: :confused: :confused: :confused: :confused: :confused: :confused: 
Why you not just do what command says?

DasSkelett:
Ah, I think that's because we're outside the container that's dynamically spawned for the `KSP-CKAN/xKAN-meta_testing` action. We had that with the cache already, which is why we moved it into the working directory using a relative path, which is bind-mounted to the host.

DasSkelett:
Okay, after changing the location in xKAN-meta_testing and the workflow file, we also need to adjust the path in the Dockerfile apparently, because that's where the directory is created:
https://github.com/KSP-CKAN/CKAN/blob/241741fe993ac8e4fed1178368aeccb613615eaa/Dockerfile.metadata#L12

Problem is, because GitHub bind-mounts the workspace folder into the container, I think that files created beforehand in the Dockerfile would be lost. So we might need to create that directory on-demand in the action instead.

DasSkelett:
🎉 
image

https://github.com/KSP-CKAN/NetKAN/pull/8701/checks?check_run_id=3279200969

I'm not sure if we really want to keep it that way, or find another place to create the directory, or move the directory somewhere else.
Going to remove the `mkdir /repo` from the Dockerfile now as well.

DasSkelett:
Small caveat:
image

> There is currently no way to download artifacts after a workflow run finishes in a format other than a zip or to download artifact contents individually. One of the consequences of this limitation is that if a zip is uploaded during a workflow run and then downloaded from the UI, there will be a double zip created.

https://github.com/actions/upload-artifact#zipped-artifact-downloads

So maybe we should do this whole thing with `/ckans` instead, and add it as zip repo...

HebaruSan:
Thanks so much for fixing it!

> So maybe we should do this whole thing with `/ckans` instead, and add it as zip repo...

That sounds good. Maybe we should also put the PR number in the artifact name, in case we want to download multiple of them to the same folder.

PR #8705, opened 2021-08-08
Use metanetkan for RSS

NathanKell:
Convert RSS over to using a locally-managed (on RSS repo) netkan


HebaruSan:
Yup, I remember this coming up. Thanks for the update.

PR #8708, opened 2021-08-10
Unspecified problem relating to KSPBurst

NathanKell:
KSPBurst-Lite does provide/conflict KSPBurst, but KSPBurst itself does not.

### DIFF
--- a/NetKAN/KSPBurst.netkan
+++ b/NetKAN/KSPBurst.netkan
@@ -14,6 +14,8 @@ tags:
   - library
 provides:
   - KSPBurst-Full
+conflicts:
+  - KSPBurst
 install:
   - find: 000_KSPBurst
     install_to: GameData


NathanKell:
@HebaruSan I believe the lack of this is what's causing the KSPBurst chocie prompt to still appear despite depending on KSPBurst-Full.

HebaruSan:
Huh? Conflicts are bi-directional. What actual problem are you seeing?

NathanKell:
Oh, then I am confused. I installed the freshly merged Express Installer and it still prompted for a KSPBurst choice. I had thought that this was the problem, but if not I think it's another case of that relationship resolver issue regarding how things are blobbed together. I'll respond on the Express issue instead of here then.

HebaruSan:
Do you still have some old testing metadata floating around in your CKAN? Make sure that you've removed any custom repos and are just getting the latest stuff from the defaults.

PR #8712, opened 2021-08-12
Update SimpleLogistics.netkan 🦞 🇸🇽 🍭 

zer0Kerbal:
- update spacedock link now that have access to original posting
- update recommends -> suggests
- update suggests
- add supports

Closes https://github.com/KSP-CKAN/CKAN-meta/pull/2435
Closes https://github.com/KSP-CKAN/CKAN-meta/pull/2434

### DIFF
--- a/NetKAN/SimpleLogistics.netkan
+++ b/NetKAN/SimpleLogistics.netkan
@@ -23,7 +23,8 @@
         { "name": "DockingPortDescriptions" },
         { "name": "PicoPortShielded"        },
         { "name": "KerbalChangelog"         },
-        { "name": "Pteron"                 }
+        { "name": "Pteron"                  }
+    ],
     "supports": [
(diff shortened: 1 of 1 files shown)


DasSkelett:
Why switch SpaceDock page after more than a year of using the new one? I feel like that's just going to confuse users.

Are you planning to re-upload 2.0.4.0 as well?

zer0Kerbal:
> Why do you want to switch the SpaceDock page after more than a year of using the new one? I think this might be confusing to users.
> 
> Are you planning to re-upload 2.0.4.0 as well?

yes. finally gained access to the original. have posted a link in place of the readme on the higher (newer) version - and will be replacing the image as well. 

and the newest version 2.0.5.0 as well.

zer0Kerbal:
and the users of CKAN shouldn't notice anything unless they look closely.

PR #8721, opened 2021-08-21
Override TundraSpaceCenter 2.0.1 max KSP version

NathanKell:
I asked damonvv's permission to override the CKAN versioning of TSC and was kindly granted permission. The thread title has "KSP 1.8+" and the mod is just assets and KK statics so it is not version-locked in and of itself, but I wanted to wait until I had a chance to check with damonvv regardless. Having now received the OK, here's the PR.

### DIFF
--- a/NetKAN/TundraSpaceCenter.netkan
+++ b/NetKAN/TundraSpaceCenter.netkan
@@ -8,6 +8,14 @@
         "config",
         "buildings"
     ],
+    "x_netkan_override" : [
+        {
+            "version" : ">=2.0.1",
+            "override" : {
+                "ksp_version_max" : "1.12.99"
+            }
+        }
+    ],
     "depends": [
         { "name": "ModuleManager"    },
         { "name": "KerbalKonstructs" }


DasSkelett:
There's a remote version file here:
https://github.com/damonvv/TundraSpaceCenter/blob/master/GameData/TundraSpaceCenter/Versioning/TundraSpaceCenter.version

It would be preferable if @damonvv updated that one instead, better than maintaining custom overrides on our side. Could you ask damonvv to do that, maybe prepare a PR in that repo?

Or even update the SpaceDock page, which would allow manual installers to see the up to date compatibility on first look as well.

NathanKell:
Oooh that is better indeed. I looked for any repo and did not find it in the thread or the netkan, so thought it was a spacedock-only mod. Yes, it would be better to do it that way.

If memory serves, spacedock supports only one KSP version, which means if it were updated to 1.12 and given precedence over the version file it would prevent install on 1.8, 9, 10, and 11, which seems almost as bad. So remote version it is.

NathanKell:
Ah, it's in the actual .ckan file. Whoops.

NathanKell:
Closed in favor of https://github.com/damonvv/TundraSpaceCenter/pull/2

DasSkelett:
> I looked for any repo and did not find it in the thread or the netkan, so thought it was a spacedock-only mod.

You can check http://status.ksp-ckan.space, it has a link to the remote-avc for TundraSpaceCenter (and all other mods that have one, broken or not).

> given precedence over the version file it would prevent install on 1.8, 9, 10, and 11

SpaceDock doesn't strictly overwrite version file data, it's merged. Otherwise a vref for SpaceDock mods wouldn't make sense, not would changing the remote version file have any effect.
Instead, we take the lower KSP version of the (remote) version file and the SpaceDock mod and set it as `ksp_version_min`, and the higher one of the two as `ksp_version_max`. Of they're both the same, we just set `ksp_version`.

NathanKell:
Ah! Thanks for linking me that, I had it open a ways back and lost the tab. That also warns when there's a naming mismatch too, which is certainly the case for some RO-suite mods that I need to fix at some point.

And yeah that makes so much sense re: versioning. I think I recall running into that issue in some case when there wasn't an independent version file, and that getting warped into "it overrides the version file". Cheers!

DasSkelett:
And the metadata update arrived: https://github.com/KSP-CKAN/CKAN-meta/commit/bf0d838289f40fb5ffc3beece7194112b328c092

Thanks a lot for taking care of it!

damonvv:
Updated both Spacedock and github for TSC v2.0.2 that supports KSP 1.8 up to 1.12. Everything should be correct now 👍

NathanKell:
My pleasure, but mostly thanks to @damonvv for an awesome mod (and you for setting me right 🙂)

PR #8723, opened 2021-08-22
Upload .ckans instead of .repo/metadata.tar.gz as workflow artifact

DasSkelett:
See https://github.com/KSP-CKAN/xKAN-meta_testing/pull/80 and https://github.com/KSP-CKAN/NetKAN/pull/8704#issuecomment-895115027, this uploads the `.ckans` directory instead of the `.repo/metadata.tar.gz` file, to avoid double compression. I hope the filename turns out the way I think it does.

### DIFF
--- a/.github/workflows/inflate.yml
+++ b/.github/workflows/inflate.yml
@@ -41,12 +41,12 @@ jobs:
                   diff meta root: .CKAN-meta
                   pull request url: ${{ github.event.pull_request.url }}
             - name: Chmod cached files so actions/cache and actions/upload-artifact can read them
-              run: sudo chmod -R a+r .cache .repo
+              run: sudo chmod -R a+r .cache
               if: ${{ always() }}
-            - name: Upload inflated repo metadata.tar.gz artifact
+            - name: Upload inflated ckans as metadata repository zip artifact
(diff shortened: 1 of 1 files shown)


DasSkelett:
```
chmod: cannot access '.ckans': No such file or directory
```
While not fatal, would be nice to get rid of this to avoid PRs without .netkan changes to fail.

We could try whether it works without chmodding...

DasSkelett:
Thinking about it, I'm not sure whether we would've seen the same exception with https://github.com/KSP-CKAN/xKAN-meta_testing/pull/80 already merged. But it doesn't hurt to get rid of unneeded chmods.
So far it looks good, but we need the xKAN-meta_testing PR deployed to fully test it.

Drop the last Astrogator commit before merging.

HebaruSan (approved):

DasSkelett:
Upload worked fine, but unfortunately the placeholder didn't get replaced correctly:
image
https://github.com/KSP-CKAN/NetKAN/suites/3563133927/artifacts/85552240

Maybe an environment variable would work. Taking a look at https://github.com/actions/upload-artifact, maybe it had come up in already.

PR #8735, opened 2021-08-28
Make InterstellarTechnologies install only the correct folder

sswelm:
Make InterstellarTechnologies install only the correct folder

### DIFF
--- a/NetKAN/InterstellarTechnologies.netkan
+++ b/NetKAN/InterstellarTechnologies.netkan
@@ -5,6 +5,9 @@ license: CC-BY-NC-SA
 tags:
   - config
   - parts
+install:
+  - find: GameData/InterstellarTechnologies
+  - install_to: GameData
 depends:
   - name: ModuleManager
   - name: FilterExtensions


HebaruSan:
@sswelm, can you please describe the problem you're trying to solve? This change would have no effect, as the default installs stanza looks for a folder named after the identifier and installs it to GameData. In other words, CKAN is already doing this.

https://github.com/KSP-CKAN/CKAN/blob/2a2b12e399818e39534070ccf888e3ac08fd47af/Core/Types/ModuleInstallDescriptor.cs#L208-L221

sswelm:
so what went wrong exactly?

HebaruSan:
The error was because of the syntax here:

```yaml
install:
  - find: GameData/InterstellarTechnologies
  - install_to: GameData
```

That creates two separate objects in the list, one per hyphen, as in:

```json
  "install": [
    {
      "find": "GameData/InterstellarTechnologies"
    },
    {
      "install_to": "GameData"
    }
  ]
```

The right syntax is to omit the second hyphen:

```yaml
install:
  - find: GameData/InterstellarTechnologies
    install_to: GameData
```

PR #8736, opened 2021-08-28
Added InterstellarTechnologies as a recommended mod

sswelm:


HebaruSan:
Thanks for the update! Would you like us to apply this chnage to any previous versions as well?

sswelm:
the current version should be enough

PR #8747, opened 2021-09-03
Merge Flying Saucers and Blueshift .netkan files

Angel-125:


HebaruSan:
We should revisit #8548 now that this is in progress...

HebaruSan:
Blueshift was added in #8357, not sure why it's showing up as a new netkan here...

HebaruSan:
The forum thread lists three dependencies that aren't covered here:

image

@Angel-125, should these be added to the netkan or removed from the forum thread?

PR #3390, opened 2016-03-03
Move Goodspeed (GPO Speed Fuel Pump) to Github

politas:
Per [Forum thread](http://forum.kerbalspaceprogram.com/index.php?/topic/90246-the-comprehensive-kerbal-archive-network-ckan-package-manager-v1160-13-dec-2015/&do=findComment&comment=2439622)

PR #8756, opened 2021-09-12
Get homepage and repository resources from SpaceDock

HebaruSan:
## Motivation

Netkan pulls the homepage and repository from SpaceDock, but many netkans currently override this unnecessarily. It's better to store these values on SpaceDock so non-CKAN users can access them and so authors can edit them more easily.

## Changes

Now `resources.homepage` and `resources.repository` are removed from netkans with a SpaceDock kref, so we will use the upstream author-controlled data instead. Other `resources` properties (`manual`, `license`, etc.) are retained. The `DMTanks` and `InterstellarFuelSwitch` related modules will keep their `repository` properties because they are "deep links" that would be truncated by the changes from #3390.

~~After the first run of the validation workflow finishes, I will check the diffs and abuse my SpaceDock admin privileges to fill in the URLs on SpaceDock for any mods where they're missing.~~ The intention is that by the time this is merged, it will result in no substantial changes for CKAN-meta (slight changes in some URLs are expected).

In addition, end of line characters are standardized with LF, and tab characters are removed from all files already being changed here for other reasons.


HebaruSan:
Well that's confusing.

```
Error: [new-lines] wrong new line character: expected \n
ERROR:root:Test of NetKAN/RealTimeClock2.netkan failed!
```

I assume it's a LF/CRLF issue, but I don't see anything strange about this file locally, possibly because git converts line endings automatically.

HebaruSan:
- [ ] Need to revert changes to the ones with "deep linked" repositories

PR #8760, opened 2021-09-15
Add Kerbal Simpit Revamped mod

PBechon:
A revamp of the original Simpit. Used to create custom controller by enabling communication with an Arduino via USB connection. Forum thread  is probably still pending validation to be public in the KSP forum.

https://github.com/Simpit-team/KerbalSimpitRevamped
https://api.github.com/repos/Simpit-team/KerbalSimpitRevamped/releases
___

Previous maintainer's endorsement of adoption:

https://www.reddit.com/r/KerbalControllers/comments/jmmdqf/kerbal_simpit_would_an_update_be_useful/gay2jyd/?utm_source=reddit&utm_medium=web2x&context=3


HebaruSan:
@DasSkelett, the previous maintainer says:

> I've been hoping somebody would get more involved with the project for a couple of years. It's code that I liked and enjoyed working on, but just haven't had the time or inclination for working on it (or even playing any KSP, really) for quite a while.

Currently this PR creates a new identifier. Should we update the old one (`KerbalSimpit`)?

PBechon:
Since we were not able to coordinate with the previous owner, we used a different repo, a different site for the online documentation and a different forum thread. So in my opinion, it make more sense to have it appear as a separate mod. 

That's my 2 cent, any solution to have a CKAN install of our version is fine with me to be honest.

DasSkelett:
It looks like they are still pretty active on Reddit, did you try it there? Also maybe a ping here helps. @phardy, in this [Reddit comment](https://www.reddit.com/r/KerbalControllers/comments/jmmdqf/kerbal_simpit_would_an_update_be_useful/gay2jyd/?utm_source=reddit&utm_medium=web2x&context=3) you (?) offered to hand over credentials for the readthedocs page for Kerbal Simpit, would you still willing to do that?

Regarding the source repository and forum thread, it probably makes sense to use the new one either way, so you can properly edit them and have full control.

HebaruSan:
OK, one last consideration and the reason I brought up the possibility of keeping the old identifier:

> 2149 [1] WARN CKAN.NetKAN.Validators.PluginsValidator (null) - No plugin matching the identifier, manual installations won't be detected: GameData/KerbalSimpitRevamped/KerbalSimpit.dll, GameData/KerbalSimpitRevamped/WindowsInput.dll

That means that if someone installs by extracting the ZIP manually and then runs CKAN, CKAN will think they have the original `KerbalSimpit`, based on the name of the DLL. Could the DLL name be updated to `KerbalSimpitRevamped.dll`? Re-using the old identifier would solve this as well.

PBechon:
I can change the DLL name to match the new identifier (and provide a setting.cgf.sample instead of a setting.cfg directly). It is better to update the current release or can I do a new release (like 2.0.1) and prevent CKAN from indexing the 2.0.0 ?

HebaruSan:
> It is better to update the current release or can I do a new release (like 2.0.1) and prevent CKAN from indexing the 2.0.0 ?

Either option would work fine. When this is merged, CKAN will index whatever version is the most recent at that time and nothing older. And since it's not indexed yet, replacing the existing download would also not cause problems for CKAN.

PBechon:
Perfect. I will do those modifications and let you know after testing it.

phardy:
Hey folks.

@PBechon @DasSkelett Sure, I'm happy to hand over the readthedocs project. If you'd like to stop polluting this PR with unrelated stuff, then send me a private github message with the readthedocs usernames you want added. I promise I'll actually be paying attention and looking to hand over anything else I can this weekend - I don't want to be involved in working on this project any more, but really don't want to hamper folks trying to make it rock.

phardy:
(and it goes without saying that I also absolutely approve of reusing the old identifier)

PBechon:
Thanks for you support @phardy. For now we were aiming for a revamped version, so we created a new readthedoc project to have full control on it : https://kerbalsimpitrevamped-arduino.readthedocs.io We were able to fully reuse your setting, so it was rather quick. Our handle on readthedocs is `simpit.team` if you want to add us to the original one.

If we take over however, we need to use the same install method for the Arduino lib (for now we were aiming for a manual install). This means also taking over the lib in the Arduino lib manager.

One potential issue though : our new version is not compatible with all version of AGExt except the last one. Is it an issue for the metadata to had this conflict starting from the new version ?

PBechon:
Seems OK for the arduino lib manager for us to take over.

So ok to reuse the Simpit original ID. I also updated our release to match KerbalSimpit folder (instead of installing to KerbalSimpitRevamped folder) and we now install a Setting.cfg.sample file (instead of a Setting.cfg).

Can you replace current Simpit netkan file with the new one ? Or do you want to do another PR ?

PBechon:
This file seems to do the trick for me, to replace the existing Simpit file

```
spec_version: v1.4
identifier: KerbalSimpit
name: Kerbal Simpit
abstract: Communicate with external instrumentation through serial connection.
author:
  - stibbons
  - LRTNZ
  - rogor
$kref: '#/ckan/github/Simpit-team/KerbalSimpitRevamped'
$vref: '#/ckan/ksp-avc'
license: BSD-2-clause
release_status: stable
resources:
  homepage: >-
    https://forum.kerbalspaceprogram.com/index.php?/topic/204852-112x-simpit-revamped-simpit-20/
  manual: https://kerbalsimpitrevamped-arduino.readthedocs.io
tags:
  - plugin
  - app
  - control
recommends:
  - name: AlternateResourcePanel
suggests:
  - name: AGExt
  - name: CommunityResourcePack
conflicts:
  - name: AGExt
    max_version: 1:2.4.1

```

HebaruSan:
> Can you replace current Simpit netkan file with the new one ? Or do you want to do another PR ?

I've pushed a commit to replace the old file in this PR. Looks OK so far...

HebaruSan:
Yup, I think this should work. Thanks for the submission and your patience in working everything out. I'll merge it shortly. Please let us know if you or your users notice any problems.

PR #8779, opened 2021-09-23
Scrap yard

zer0Kerbal:
probably needs some polish 👾


HebaruSan:
> probably needs some polish

Indeed it does. Are you comfortable working the validation errors?

https://github.com/KSP-CKAN/NetKAN/pull/8779/checks

NetKAN/ScrapYard.netkan
```diff
@@ -1,9 +1,39 @@
 spec_version: v1.4
 identifier: ScrapYard
+name: ScrapYard (SYD)
 $kref: '#/ckan/spacedock/1746'
 $vref: '#/ckan/ksp-avc/ScrapYard.version'
-license: MIT
+license: Expat-MIT
```
HebaruSan, line 6: The allowed licenses are listed here:

https://github.com/KSP-CKAN/CKAN/blob/master/CKAN.schema#L444-L477

zer0Kerbal:
> Error: 1418 [1] FATAL CKAN.NetKAN.Program (null) - License Expat-MIT should match spec
that is the official license - not MIT because it isn't MIT - what is collegially referred to as MIT is actually Expat-MIT.

zer0Kerbal (commented):
looks correct, at least after removing trailing spaces and it copy-paste appendix.

NetKAN/ScrapYard.netkan
zer0Kerbal: (https://en.wikipedia.org/wiki/MIT_License)

name it what you will, but legally I am obligated to refer to it as the Expat-MIT license. :shrug:

HebaruSan: The mod literally will crash CKAN if it isn't one of the values on that list.

HebaruSan: > (https://en.wikipedia.org/wiki/MIT_License)
> 
> name it what you will, but legally I am obligated to refer to it as the Expat-MIT license. 

That link does not seem to say that:

image

> The MIT license has the identifier MIT in the SPDX License List

zer0Kerbal: name it what you will. :shrug:

techman83: Your `LICENSE` file specifically references the the MIT License. If it's the X11 variant (it isn't because X11 has a no advertising clause), we are missing that identifier and would use `open-source` instead. The Expat identifier (and is one of our valid identifiers) appears to be largely historic and the commonly accepted short identifier is MIT. 

Beyond that, the identifier is purely metadata and has no legal bearing, apart from ensuring we don't redistribute mods that don't allow redistribution via the Archive.org preservation/mirroring we do as part of the indexing.

* I'm not a lawyer, but wrote most of the initial licensing logic in the original indexing service

DasSkelett: @zer0Kerbal no worries, referring to it as MIT won't take away any rights from you or something.
In CKAN's metadata, the MIT/Expat license needs to be referred to as either `MIT` or `Expat`. That's how it's implemented, and we won't add a third synonym for the same license.

Do you want to do the final clean-up, or should we take care of it?

zer0Kerbal: Probably go with Expat or MIT - either one. it will be changed to Expat/MIT with BY clause { The Expat-MIT License (MIT) with Attribution } which makes it it's own unique beast as far as I can tell. 🦺 

it also has now be localized, at least to us-en

again, mahalo nui loa - onto OyScrap!

techman83: Cool, that's just the regular MIT license then (to which we have both versions of the short identifier). MIT-0 is the identifier for the no attribution variant.

HebaruSan: Unresolved this because the change in the PR still has the invalid string.

DasSkelett:
Pushed a commit to fix the remaining issues to get this moving again. Looks good to me, is there anything else to address?

HebaruSan (approved):
Looks fine now.

PR #8786, opened 2021-09-30
Distant Object is under New Management.

Lisias:
Hi!

Distant Object (Continued) is under New Management, and this pull request redirects the links reflecting the change.

In time… How this `DistantObject-config` stunt works? I found a frozen one for RRS, there's the "default" (aka Stock) one, and so goes on…

Cheers!

(I tried to ping TheDarkBadger here, but I'm unable)

### DIFF
--- a/NetKAN/DistantObject-default.netkan
+++ b/NetKAN/DistantObject-default.netkan
@@ -1,14 +1,14 @@
 {
     "spec_version": "v1.16",
     "identifier":   "DistantObject-default",
-    "name":         "Distant Object Enhancement Continued default config",
+    "name":         "Distant Object Enhancement /L default config",
     "abstract":     "Default planets colors",
-    "$kref":        "#/ckan/github/TheDarkBadger/DistantObject",
+    "$kref":        "#/ckan/github/net-lisias-ksp/DistantObject",
     "$vref":        "#/ckan/ksp-avc",
     "x_netkan_force_v": true,
     "license":      "CC-BY-4.0",
     "resources": {
-        "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/189759-*"
+        "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/205063-*"
     },
     "tags": [
         "config",
--- a/NetKAN/DistantObject.netkan
+++ b/NetKAN/DistantObject.netkan
@@ -1,14 +1,14 @@
 {
     "spec_version": "v1.4",
     "identifier":   "DistantObject",
-    "name":         "Distant Object Enhancement Continued",
+    "name":         "Distant Object Enhancement /L",
     "abstract":     "Visual enhancement mod that makes objects realistically visible over large distances",
-    "$kref":        "#/ckan/github/TheDarkBadger/DistantObject",
+    "$kref":        "#/ckan/github/net-lisias-ksp/DistantObject",
     "$vref":        "#/ckan/ksp-avc",
     "x_netkan_force_v": true,
     "license":      "CC-BY-4.0",
-    "resources" : {
-        "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/189759-*"
+    "resources": {
+        "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/205063-*"
     },
     "tags": [
         "plugin",


HebaruSan:
The inflation errors are because there's a forward slash in the tag, which Netkan inserts into the filename:

image

Not sure what to do about that yet. Could it be removed?

HebaruSan:
> (I tried to ping TheDarkBadger here, but I'm unable)

Actually, @TheDarkBadger just did a release with your changes 3 days ago:

https://github.com/TheDarkBadger/DistantObject/releases

Lisias:
HUmmHUnn

> > (I tried to ping TheDarkBadger here, but I'm unable)
> 
> Actually, @TheDarkBadger just did a release with your changes 3 days ago:
> 
> Looks like it's not abandoned. Should we close this?

He passed DOE to me, you can check on Forum:

https://forum.kerbalspaceprogram.com/index.php?/topic/189759-18-112x-distant-object-enhancement-continued-v2031-29-september-2021/

About the problem with the tag, my building tools use dashes on the tag to hierachisize it (this word exists?). As a matter of fact, TweakScale does it too. My bad, I forgot about this, I should had used the NetKAN from TweakScale as template instead of blindly reusing the current one.

I will rework the NetKAN file using TweakScale as reference and will fix it ASAP.

Lisias:
Humm… I think the checkings ignored the changes on the NetKAN file. It's best to close this and try another pull request, so?

HebaruSan:
No, keep the same pull request.

NetKAN/DistantObject-default.netkan
```diff
@@ -1,14 +1,14 @@
 {
     "spec_version": "v1.16",
     "identifier":   "DistantObject-default",
-    "name":         "Distant Object Enhancement Continued default config",
+    "name":         "Distant Object Enhancement /L default config",
     "abstract":     "Default planets colors",
-    "$kref":        "#/ckan/github/TheDarkBadger/DistantObject",
+    "$kref":        "#/ckan/github/net-lisias-ksp/DistantObject",
```
HebaruSan, line 6: :point_up:

Lisias: Yes, the DistantObject-default thingy.

I don't understand how this "filter" thingy works yet, so I think it's wiser to change this one after I get a grasp on it.

It appears to had been useful in the bast (different configs for Stock, RO, etc), but currently there's only one.

HebaruSan: There's no filter on that line. That's the line that needs to be updated to reference SpaceDock instead of GitHub.

Lisias: Dude, you saved my SAS. I really need to remember avoiding commits to CKAN late night. This was an automatic copy&paste that should not had happened.

There's a filter on the DOE's NetKAN file:

```
    "install": [ {
        "find"        : "DistantObject",
        "install_to"  : "GameData",
        "filter"      : "PlanetColors.cfg"
```

and a dependency on `DistantObject-config`, that is currently provided by `DistantObject-default`. And, as a matter of fact, I could not even find the `DistantObject-default.zip` file anywhere on TheDarkBadger's [repository](https://github.com/TheDarkBadger/DistantObject) - so I'm really reticent on changing this now other than the cosmetic ones.

I'm planning to replace this thing as patches using MM, but this will not happen for a couple releases - so until there, I want the `-default` to pinpoint the TheDarkBadger repo to keep things going (however they do, because I just don't understood it yet).

Later, I want to remove the `DistantObject-config` as a dependency and then freeze the `-default` one.

HebaruSan: `DistantObject-default` just pulls the `PlanetColors.cfg` file out of the same main ZIP file (not `DistantObject-default.zip`) and installs it, and the `filter` property prevents the main `DistantObject` from installing that file. Previously this was so that `AstronomersPack-DistantObjectEnhancement` and `DistantObject-RealSolarSystem` could install their own replacements for `PlanetColors.cfg`, but those modules are both now frozen.

Replacing this mostly-vestigial mess with ModuleManager patches is a good idea. :+1: More maintainable, easier installation for non-CKAN users. In the meantime, though, it's probably wise to make the `$kref` properties the same, since they're supposed to be pulling from the same ZIP file.

Lisias: Now I understand. I will fix them and commit them for review ASAP.

Lisias:
An additional last question, to save commits and time:

the configs are going to stay as CC-BY 4.0, but the code will gradually migrate to GPLv2, Restricted (the same way TweakScale did). Previous code I intent to remain as CC-BY 4.0, only the new code will be licensed at GPv2 (with the Restricted alternative).

How I state this on the license field? Same as TweakScale?

```
    "license"          : [ "GPL-2.0", "restricted" ],
```

By so, how I state that some code are still licensed under CC-BY 4.0?

HebaruSan:
When different files in the download have different licenses, the rule is to choose the more restrictive license. Since GPL requires source code and CC-BY just requires attribution, GPL is probably the more restrictive, and so `[ "GPL-2.0", "restricted" ]` would be what I would expect to see.

Lisias:
Hi! I think we are finally set!

Let me know if you need me to change something else - a release will be published in a few days (before the next weekend) so we have time for any adjustments.

HebaruSan:
Will do, after I re-run the validation checks with the new release...

HebaruSan:
@Lisias, the new version file says it's compatible back to KSP 1.4.5. Are you sure about that? I feel the need to confirm this explicitly because there have been multiple Unity updates since then that broke compatibility in many plugins, which makes this compatibility table surprising:

```
  Version   Game Versions     
  --------  ------------------
  v2.1.1.5  KSP 1.4.5–1.12.2  
  v2.0.3.1  KSP 1.12.0–1.12.99
  v2.0.3.0  KSP 1.12.0–1.12.99
  v2.0.2.0  KSP 1.11.0–1.11.99
  v2.0.1.1  KSP 1.10.0–1.10.99
  v2.0.1.0  KSP 1.10.0–1.10.99
  v2.0.0.2  KSP 1.9.0–1.9.99  
  v2.0.0.1  KSP 1.8.1–1.8.99  
  v2.0.0.0  KSP 1.8.1–1.8.99  
  v1.9.1.1  KSP 1.3.0–1.7.99  
  v1.9.1    KSP 1.3.0–1.4.5   
  v1.9.0    KSP 1.3.0         
```

Lisias:
Ouch, sorry for non answering this… It was a tough week, and when I finally got time to rest, I left this existential plane for some time.

Yes, I confirm this thing works flawlessly from KSP 1.4.5 to newest due judiciously craft of auxiliary DLLs when needed (so the base code runs almost everywhere, and it loads dynamically the needed code for the environment).

PR #8792, opened 2021-10-03
Simple life!

zer0Kerbal:
time to get this one moving. :D
MoarKerbals and MoarKerbalsParts will be heading in for a 🎃 soon. :D

https://spacedock.info/mod/766/SimpleLife
https://github.com/zer0Kerbal/SimpleLife
https://forum.kerbalspaceprogram.com/index.php?/topic/191526-simplelife-under-construction/

___

ckan compat add 1.7


HebaruSan:
- You added a `$vref` but there's no version file, did you mean to add one?
- I see that this bundles `USILifeSupport.dll`, but CKAN has that indexed separately as `USI-LS` , and we're supposed to handle bundled mods as dependencies rather than installing the bundled copy. Could that be done here? (Same goes for `KSPAPIExtensions`)

zer0Kerbal:
just trying to get ahead on this. You can hold until next release which may be as far out as a month or more.

> * You added a `$vref` but there's no version file, did you mean to add one?
> * I see that this bundles `USILifeSupport.dll`, but CKAN has that indexed separately as `USI-LS` , and we're supposed to handle bundled mods as dependencies rather than installing the bundled copy. Could that be done here? (Same goes for `KSPAPIExtensions`)

1. there isn't a version file in the SpaceDock .zip until next release. there is one on [githib](https://raw.githubusercontent.com/zer0Kerbal/SimpleLife/master/SimpleLife.version) - this won't change until next release.
2. I don't know if the current version would work with the current version; whatever is needed. Next version probably will be using another SimpleMod! as its backbone, couple with the suggested MoarKerbals.

HebaruSan:
@zer0Kerbal, if you closed this unintentionally as suggested in the comments of #8806, then you'll need to restore the branch to re-open it.

PR #8794, opened 2021-10-04
Update KerwisChinaAerospacePack.netkan

Icecovery:
New recommendations:
recs

Updated compatibility:
compat


DasSkelett:
Hey, sorry if I'm missing something obvious, are you part of the dev team of this mod? Or are these changes based on observations as a user?

Most resource links are automatically fetched from SpaceDock, no need to hardcode them in the netkan:
https://github.com/KSP-CKAN/CKAN-meta/blob/a3f08f3e00b02fceac24733c3230fb08db8046c8/KerwisChinaAerospacePack/KerwisChinaAerospacePack-1.0.2.ckan#L14-L20

The `ksp_version_max` can be taken from SpaceDock as well.

The license on the forum thread is still CC BY-NC-SA **4.0**, so we shouldn't loose the version specificity there.

What's the reason for the "development" release_status? The mod doesn't seem to be marked WIP or similar.

Please don't remove the tags, every mod should have them.

I'm not sure about the `supports: - name: RealismOverhaul`. It makes the mod show up during the RO installation process, and given that RO has delicate relationships and can break quickly, I'm not sure if we should offer it to users there.

Icecovery:
> are you part of the dev team of this mod? Or are these changes based on observations as a user?

I did not directly involved in the development of this mod, however, I am helping the devs updating this under their consent since English is not their first language and they are not familiar with CKAN.



> Most resource links are automatically fetched from SpaceDock, no need to hardcode them in the netkan: https://github.com/KSP-CKAN/CKAN-meta/blob/a3f08f3e00b02fceac24733c3230fb08db8046c8/KerwisChinaAerospacePack/KerwisChinaAerospacePack-1.0.2.ckan#L14-L20
> 
> The `ksp_version_max` can be taken from SpaceDock as well.
> 
> The license on the forum thread is still CC BY-NC-SA **4.0**, so we shouldn't loose the version specificity there.
> 
> Please don't remove the tags, every mod should have them.

I will fix these in the next commit.



> What's the reason for the "development" release_status? The mod doesn't seem to be marked WIP or similar.

The author told me that it should be in development since there are more parts to be added in future updates.



> I'm not sure about the `supports: - name: RealismOverhaul`. It makes the mod show up during the RO installation process, and given that RO has delicate relationships and can break quickly, I'm not sure if we should offer it to users there.

I didn't think about that, thanks for mentioning it, I will remove it in the next commit.

Icecovery:
> The license on the forum thread is still CC BY-NC-SA **4.0**, so we shouldn't loose the version specificity there.

Checks won't pass the "License should match spec" thing if the version number was added. 
However, the checks will pass if I run `netkan.exe --verbose .\KerwisChinaAerospacePack.netkan` on my local machine.

It would be greatly appreciated if you can help fix this @DasSkelett .

DasSkelett:
> Checks won't pass the "License should match spec" thing if the version number was added.

You need to replace the space with a dash. Just revert it to how it was before your changes:
https://github.com/KSP-CKAN/NetKAN/blob/19d5c820ea6164756c781266a5e52f3a122eae3a/NetKAN/KerwisChinaAerospacePack.netkan#L4

DasSkelett:
> The author told me that it should be in development since there are more parts to be added in future updates.

What type of updates, just general feature updates? `development` in CKAN means that the mod is in its very first stages, with many bugs to be expected and important core features missing. Mod authors usually declare that on the forum thread, and/or use version numbers below 1.0.

Reading the forum thread it doesn't sound like that, but of you have better inside info, I'll trust those.

Icecovery:
> You need to replace the space with a dash. Just revert it to how it was before your changes:

Thanks, I didn't notice it was there before.

DasSkelett:
> I will inform the author about the software versioning convention issue so they will respect that in future projects.

They don't need to change their versioning system for CKAN, please don't bother them for that. That was just a data point in my observation about the stableness of the mod.

Anyway, changes almost look good, I'm just gonna move the `ksp_version_min` into an override to make it work properly (I thought it would merge it with SpaceDock's data as is, but apparently it doesn't)

Icecovery:
> > I will inform the author about the software versioning convention issue so they will respect that in future projects.
> 
> They don't need to change their versioning system for CKAN, please don't bother them for that. That was just a data point in my observation about the stableness of the mod.
> 
> Anyway, changes almost look good, I'm just gonna move the `ksp_version_min` into an override to make it work properly (I thought it would merge it with SpaceDock's data as is, but apparently it doesn't)

Thanks a lot for your help!

DasSkelett (approved):
Looks good now, thanks for the update!

PR #8804, opened 2021-10-08
Install GPPSecondary without GPP

HebaruSan:
## Motivation

See Galileo88/Galileos-Planet-Pack#59; the point of the GPPSecondary module is supposed to be to install GPP as a secondary solar system orbiting the primary one, but as currently implemented it depends on GPP, which natively installs itself as the _primary_ solar system. This means that currently we cannot install GPPSecondary alongside any other primary solar system mod that conflicts with GPP, [such as JNSQ](https://github.com/Galileo88/JNSQ/blob/master/CKAN/JNSQ.netkan).

## Changes

Now new install stanzas for the GPP files are added to GPPSecondary, as are the MM/Kopernicus/etc. dependencies. To make this work, GPP is removed from depends and added to conflicts. After this it will be possible to install JNSQ as your primary solar system and GPP as a secondary solar system.

We'll probably want to give the Galileo team an FYI before merging this. Might also be a good idea to recruit beta testers from the GPP/JNSQ threads. A CKAN-meta PR for the historical versions of GPPSecondary would probably make sense as well.

### How to beta test

1. Launch CKAN
2. Settings &rarr; CKAN Settings
3. Under Metadata Repositories, click New
4. Name: GPP Testing (or whatever you prefer)
   URL: <https://github.com/KSP-CKAN/NetKAN/suites/4006891354/artifacts/101129399>
5. Click Add
6. Close settings
7. Click Refresh
8. Try installing JNSQ with GPPSecondary

Later (after this is merged), you should un-do these changes to return to normal:

1. Launch CKAN
2. Settings &rarr; CKAN Settings
3. Under Metadata Repositories, select the entry you created above
4. Click Delete
5. Close settings
6. Click Refresh

### DIFF
--- a/NetKAN/GPPTextures.netkan
+++ b/NetKAN/GPPTextures.netkan
@@ -8,8 +8,8 @@ resources:
   homepage: https://forum.kerbalspaceprogram.com/index.php?/topic/152136-galileos-planet-pack/
 depends:
   - any_of:
-    - name: GPP
-    - name: GPPSecondary
+      - name: GPP
+      - name: GPPSecondary
 install:
   - find: GPP
     install_to: GameData


DasSkelett:
Sounds sensible to me :+1: 

I wonder if the GPP clouds mods also work with GPPSecondary? The release notes don't seem to mention this, maybe I can find something in the forum thread.

PR #8860, opened 2021-11-11
Conflicts between MinorPlanetsExpansionRevived and MinorPlanetsExpansion

HebaruSan:
While reviewing #8859 I realized that these two mods probably functionally conflict as they add many of the same planets.
Now the conflict is in the metadata.

___

ckan compat add 1.11

### DIFF
--- a/NetKAN/MinorPlanetsExpansionRevived.netkan
+++ b/NetKAN/MinorPlanetsExpansionRevived.netkan
@@ -5,6 +5,8 @@ license: MIT
 tags:
   - config
   - planet-pack
+conflicts:
+  - name: MinorPlanetsExpansion
 depends:
   - name: ModuleManager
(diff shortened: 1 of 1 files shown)


vinix38:
Seems to me the "revived" version is a replacement of an older version, is there a reason not to mark MinorPlanetsExpansion as replaced by MinorPlanetsExpansionRevived ? I noticed this for other mods too, so I'm curious to know

HebaruSan:
Good question, we probably should remember to consider the `replaced_by` property more often. For these two specific mods, though, the latest news is that the author of Revived is abandoning it and says that the author of the original mod is going to be working on it again soon. Once that happens I think we would not want Revived to be replacing the original (or we could even have it the other way around, but that should probably wait until after there is a new release of the original).

PR #8877, opened 2021-11-25
Try to rename ShipManifest's CLSInterfaces.dll

### LINKED ISSUE #8874: Duplicate CLSInterfaces.dll from ConnectedLivingSpace bundled in ShipManifest
Background
----------
* Operating System: Xubuntu 20.04
* CKAN Version: 1.29.0
* KSP Version: 1.12.2.3167

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?**
Manually removed duplicate DLLs when problem appeared previously.

Problem
-------
Installing both of these mods results in an error message about duplicate DLLs, specifically CLSInterfaces.dll which appears in both mods.  Manually removing the copy within ShipManifest (as the message suggests) fixes it, but updating ShipManifest puts it back.

HebaruSan:
## Problem

KSP 1.12 can crash if it finds two DLLs with the same name.
Currently `ConnectedLivingSpace` ships a `CLSInterfaces.dll` file that is _intended_ to be used that way. `ShipManifest` bundles this DLL, so if both mods are installed, the game can crash.

The author is planning to rename this file to eliminate the conflict:

https://forum.kerbalspaceprogram.com/index.php?/topic/109972-111x-connected-living-space-v2006-29-dec-2020/&do=findComment&comment=4058927

## Changes

Now CKAN will install the `CLSInterfaces.dll` from the current release of `ShipManifest` as `SM_CLSInterfaces.dll`. This should fix the problem without breaking anything else.

At the moment I just want to submit this to make sure it works as expected in the validation scripts. If so, it may be suitable to merge, but I want to confirm with the mod author first.

Fixes #8874.

### DIFF
--- a/NetKAN/ShipManifest.netkan
+++ b/NetKAN/ShipManifest.netkan
@@ -32,4 +32,4 @@ install:
   - find: CLSInterfaces.dll
     find_matches_files: true
     install_to: GameData/ShipManifest/Plugins
-    as: SM_CLSInterfaces.dll
+    as: SM_CLSInterface.dll
--- a/NetKAN/ShipManifest.netkan
+++ b/NetKAN/ShipManifest.netkan
@@ -1,4 +1,4 @@
-spec_version: v1.4
+spec_version: v1.18
 identifier: ShipManifest
 name: Ship Manifest
 abstract: Manages Crew, Science, & Resources on a given vessel
@@ -24,3 +24,12 @@ suggests:
   - name: DeepFreeze
   - name: RemoteTech
   - name: ModuleManager
+install:
+  - find: ShipManifest
+    install_to: GameData
+    filter:
+      - CLSInterfaces.dll
+  - find: CLSInterfaces.dll
+    find_matches_files: true
+    install_to: GameData/ShipManifest/Plugins
+    as: SM_CLSInterfaces.dll


DasSkelett (approved):
Papa_Joe just mentioned on the forum thread that it'll be name `SM_Interface.dll`, without `s`.

Apart from that, let's give it a go, a good opportunity to make sure this works as intended.

HebaruSan:
> Papa_Joe just mentioned on the forum thread that it'll be name `SM_Interface.dll`, without `s`.

Ha! I thought that was a typo the first time. Changed it.

georgedorn:
There's at least one more mod with this issue ([Kerbal Health](https://forum.kerbalspaceprogram.com/index.php?/topic/192130-111-connected-living-spaces-adopted-2006-2020-12-29/page/3/&tab=comments#comment-4048421
) and possibly any that Suggest, Support or Depend on Connected Living Space.

I don't use any of those, but if this fix has no found issues after some period of public testing, it might be worth doing a reverse search for candidates and adding more patches.

HebaruSan:
Kerbal Health seems to have already renamed its DLL:

https://github.com/GarwelGarwel/KerbalHealth/commit/9af8ccb108ef5ef767eb8e32dc0aba882f5582d0

image

PR #8881, opened 2021-11-25
GPP/GEP: version_min → min_version

HebaruSan:
## Problem

Right now the latest versions of GPP, GPPSecondary, and GEP conflict with _all_ versions of Scatterer. That's very wrong.

## Cause

In #8880, I tried to add a version-specific conflict to these mods. However, since I very frequently use the `ksp_version_min` property and almost never the `min_version` property, I typed `version_min` instead, which is not a relationship property. The client interprets this as a conflict with no version limits.

## Changes

Now the `version_min` properties are changed to `min_version`.

### DIFF
--- a/NetKAN/GPP.netkan
+++ b/NetKAN/GPP.netkan
@@ -28,7 +28,7 @@ depends:
 conflicts:
   - name: StockVisualEnhancements
   - name: Scatterer
-    version_min: '3:v0.0825b'
+    min_version: '3:v0.0825b'
 recommends:
   - name: Strategia
   - name: FinalFrontier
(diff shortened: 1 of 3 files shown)


HebaruSan:
Should this be an inflation warning, or should we treat the alternate spelling as an alias and support it?

DasSkelett:
> Should this be an inflation warning, or should we treat the alternate spelling as an alias and support it?

I'd vote for not introducing synonyms and showing a warning. It's a potential source of complexity and confusion with no real benefit, correcting the typo is easy enough (once you know it happened...)

PR #8883, opened 2021-11-28
Update MoarKerbalsParts.netkan

zer0Kerbal:
- module manager is not required, even if MoarKerbals is installed


HebaruSan:
That was done to address this inflation warning.

https://github.com/KSP-CKAN/NetKAN/runs/4296218350?check_suite_focus=true

image

CKAN has no way to know whether the patches in those files are needed or not, and the user probably doesn't know either.

PR #8895, opened 2021-12-04
Taerobee

zer0Kerbal:
# Update and rename Taerobee.frozen to Taerobee.netkan
- url's
- Author's ? (Baele and zer0Kerbal)
- version
- license updated to CC BY-ND 3.0 Unported

___

https://github.com/zer0Kerbal/Taerobee
https://spacedock.info/mod/2915/Taerobee


HebaruSan:
FYI, that license change is probably not allowed by the "ShareAlike" clause:

ShareAlike

zer0Kerbal:
The license has been changed to accomodate @Beale's desire - it is now `CC BY-ND 3.0`

HebaruSan:
OK, took a look into it (Beale seems to be @Tantares on GitHub rather than "beale", so we probably wouldn't have gotten a reply from that tagging), and the forum thread went through a few stages...

Abandonment:

image

Deletion:

image

More forceful abandonment w/ ARR:

image

Acceptance:

image

So based on that, one possible interpretation is that Beale as the owner has granted @zer0Kerbal the rights that were previously reserved, including assigning a new license. It would be nice to have that as a formal license rather than having to make up my own dubious legal case (and technically there could still be problems if the adopted content had contributions from other authors), but the main thing is making sure that the original/previous author(s) are happy. :+1:

zer0Kerbal:
I have the documentation to back up your assertations.  Since it was via direct message I will not post unless legally necessary and legally prudent.

HebaruSan:
Sounds good, I'm fine with that. :+1:

HebaruSan:
@zer0Kerbal, the release is missing the ZIP file:

https://github.com/zer0Kerbal/Taerobee/releases/tag/1.3.9.9-adoption

HebaruSan:
> I have the documentation to back up your assertations. Since it was via direct message I will not post unless legally necessary and legally prudent.

For what it's worth, the scenario I want to avoid is:

1. The original author of a CKAN-indexed mod with a restrictive license goes quiet
2. Someone else shows up claiming to have received private permission to adopt
3. We update the netkan to track the new fork
4. Six months later, the original author appears and says, "No, I never gave them permission to do that, they've been violating the license this whole time"

That could be a big mess for CKAN, with a lot of fence-mending and metadata pruning and user announcements to take care of. If we can point to a public announcement by the original author supporting the new maintainer's claims, as we (mostly) can in this case, then that's a pretty solid defense, as few people will knowingly argue against themselves once presented with such receipts. But if we look back and it turns out that we just took the word of someone when it wasn't appropriate to do so, that would look pretty bad.

PR #8898, opened 2021-12-05
Freeze Cashedowt's mods

HebaruSan:
image

I'm not sure how the author made the second one fail, since it seems to still be there:

https://spacedock.info/mod/2902


DasSkelett:
It's the download link that's 404 (https://spacedock.info/mod/2902/Discontinued/download/1) after the redirect (https://spacedock.info/content/Cashedowt_71741/RVE64K_continued/RVE64K_continued-1.zip)

Somehow they managed to delete the file from the disk without deleting the actual version, it appears?

HebaruSan:
I wonder if there's a 500 status in the log somewhere; maybe they _tried_ to delete and an exception was thrown before it was completed.

HebaruSan:
Hmm, this might fail if the mod is renamed, no? And then the db changes would be aborted?

https://github.com/KSP-SpaceDock/SpaceDock/blob/4dee33b09072e018cd3f47954f9393a153e55eef/KerbalStuff/blueprints/mods.py#L392-L393

HebaruSan:
Deleting a `ModVersion` does not seem to remove the file.

Also, there's a `db.commit()` before that `rmtree` call, so scratch what I said about the db changes being aborted. For that to happen, the exception would have to be thrown before the `db.commit()`.

HebaruSan:
KSP-SpaceDock/SpaceDock#316 is adding `ignore_errors=True` to the `rmtree` calls, as well as `os.remove` on deleting a `ModVersion`. Any attempt to address this would probably conflict with that PR, so maybe we should continue this over there.

HebaruSan:
image

HebaruSan:
So, mysterious storage glitch?

DasSkelett:
Or maybe mysterious database glitch with the mod/mod versions stuck in a weird state. Or both

HebaruSan:
Since the `/download/` route was still successfully redirecting to the right `/content/` link, I'm guessing the db is OK.

HebaruSan:
> (https://spacedock.info/content/Cashedowt_71741/RVE64K_continued/RVE64K_continued-1.zip)

One more breadcrumb, that was renamed from `continued` to `Continued` before KSP-CKAN/NetKAN#8878 was submitted.

HebaruSan:
There just aren't that many places where we delete these files. Trying to analyze whether `try_remove_file_and_folder` could be involved...

1. Create a mod, including uploading a ZIP and setting the background/thumbnail
   After this, the files are all in `storage/username_userid/original_modname` together
2. Rename the mod
   Now `Mod.base_path()` no longer matches `Mod.background`, `Mod.thumbnail`, or `ModVersion.download_path`
3. Change the background
   The `storage/username_userid/new_modname` dir is created, the new background is placed there, and `Mod.background` and `Mod.thumbnail` are updated to match the new path
4. Change the background _again_
   1. `update_mod_background` is called
   2. `_update_image` is called to replace the old background
   3. `try_remove_file_and_folder` is called to remove the old background and possibly its folder, which is `storage/username_userid/new_modname`, which might be empty depending on whether the new thumnail is already there, because the ZIP is still in `storage/username_userid/original_modname`
   4. `_update_image` tries to save the new background to `storage/username_userid/new_modname`, which may just have been deleted if it was empty, if so this would fail and throw an exception
   5. `update_mod_background` then calls `try_remove_file_and_folder` to remove the old thumbnail, possibly twice if `Mod.thumbnail` doesn't match the thumbnail that would be generated for `Mod.background` (?)

HebaruSan:
> 1. Rename the mod
> 2. Delete it
> 
> ... since that might call `rmtree` on a folder that doesn't yet exist. But that wouldn't fit the user story, which was that the mod was not deleted till after the problem started.

image

DasSkelett:
> One more breadcrumb, that was renamed from `continued` to `Continued`

The capitalization change made me a bit suspicious, since I know that the storage is backed by a CIFS/SMB mount, and given that it's a Windows-native protocol who knows how it handles case sensitivity. As you've already spotted, I've played with that on alpha a bit, but didn't encounter anything unusual, after uploading a new thumbnail it's two different directories with different files.

PR #8900, opened 2021-12-06
Update LessRealThanReal.netkan

pehvbot:
LRTR v2+ changes the dependency list and requires KSP version 1.12+.  It works with RSS and KSRSS (Kerbal Sized RSS which is not currently on CKAN) so I moved RSS to 'recommends' to allow a simpler install for any KSRSS users.  It now also conflicts with KerbalConstructionTime.

___

ckan compat add 1.10

### DIFF
--- a/NetKAN/LessRealThanReal.netkan
+++ b/NetKAN/LessRealThanReal.netkan
@@ -4,8 +4,7 @@
     "name":         "Less Real Than Real(ism)",
     "abstract":     "Real Progression One without RealismOverhaul",
     "$kref":        "#/ckan/github/pehvbot/LRTR",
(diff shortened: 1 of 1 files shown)


HebaruSan:
What's going on with KerbalConstructionTime? It's already a dependency, but now you're bundling it? CKAN isn't supposed to install bundled dependencies, to ensure compatibility with other mods using the same dependencies.

HebaruSan:
OK, if it's a patched version of the mod, then I guess a conflict is the best we can do, since other mods are unlikely to want to share the same patched version. Thanks for the update, merging...

HebaruSan:
Consider revisiting this after #10158 is completed.

PR #8914, opened 2021-12-11
Update KVASS.netkan

yalov:


HebaruSan:
Thanks for the update.

PR #8922, opened 2021-12-19
Add Conformal Decals and Shabby

HebaruSan:
https://spacedock.info/mod/2451/Conformal%20Decals
https://github.com/drewcassidy/KSP-Conformal-Decals
https://forum.kerbalspaceprogram.com/index.php?/topic/194802-18-111-conformal-decals-028-decals-done-the-right-way/

image

http://taniwha.org/~bill/Shabby.version
https://github.com/taniwha/Shabby

~~We still need to check with @taniwha to see if adding Shabby to CKAN is OK before merging.~~ Submitting the PR just for metadata validation. Should merge KSP-CKAN/CKAN#3494 first to be safe, since the `filter_regexp` in Shabby is how I noticed that issue.

### DIFF
--- a/NetKAN/ConformalDecals.netkan
+++ b/NetKAN/ConformalDecals.netkan
@@ -3,6 +3,8 @@ identifier: ConformalDecals
 $kref: '#/ckan/spacedock/2451'
 $vref: '#/ckan/ksp-avc'
 license: GPL-3.0
+resources:
+  manual: https://github.com/drewcassidy/KSP-Conformal-Decals/wiki
 tags:
   - plugin
   - parts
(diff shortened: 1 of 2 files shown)


HebaruSan:
From reviewing taniwha/Shabby#2, Shabby might not be ready. It is version 0.1.2.0, and the pull request that updated it to use Harmony2 changed the version to 0.2.0, which doesn't exist yet:

http://taniwha.org/~bill/Shabby_v0.2.0.zip

So we also need to wait for a release of Shabby that includes those changes.

HebaruSan:
image

drewcassidy:
I have a build of Shabby 0.2.0 hosted at https://pileof.rocks/KSP/Shabby_v0.2.0.zip. I also added an "unbundled" build to the GitHub releases for Conformal Decals that can be used by CKAN so it doesn't download a bunch of extra files

HebaruSan:
Looks like this should work now. Thanks for contacting us!

PR #8929, opened 2021-12-28
Update FacilityTree.netkan

sunnypunny:
author removed dependency in recent update


HebaruSan:
Thanks! Confirmed in SpaceDock changelog, but it happened a few versions back, so we'll need a pull request in CKAN-meta as well to update the historical versions:

https://spacedock.info/mod/2918/FacilityTree#changelog

image

PR #8936, opened 2022-01-01
Relationship updates for UnKerballedStart

theonegalen:
Please update the CKAN listing if this works

### DIFF
--- a/NetKAN/UnKerballedStart.netkan
+++ b/NetKAN/UnKerballedStart.netkan
@@ -1,135 +1,44 @@
-{
-    "spec_version": "v1.4",
-    "identifier": "UnKerballedStart",
-    "name": "UnKerballed Start",
-    "abstract": "An unmanned start tech tree structured around early probes and aircraft. Manned capsules become available as later tech.",
-    "author": [
-        "SpinkAkron",
-        "theoneGalen"
-    ],
-    "version": "1.3.0.0",
-    "ksp_version_min": "1.12.2",
-    "license": "CC-BY-NC-SA-4.0",
-    "resources": {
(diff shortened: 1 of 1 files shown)


DasSkelett:
Hey, it looks like instead of a .netkan file you pasted an inflated .ckan here, with a lot of hard coded values like mod version, download URL and hash etc. that should be auto-generated by netkan.exe itself.

The primary changes you want to make seem to be with the relationships and tags. Can you re-apply those changes to the raw netkan instead?
Otherwise, if you want, you can tell us what exactly you'd like to have changed, and we can help you out.

HebaruSan:
> re-apply those changes to the raw netkan instead?

Took care of this in the latest commit. Now the problem is the internal .ckan file that you mentioned on Discord.

theonegalen:
So why didn't it give me the error when I inflated it here? I thought I was supposed to post the inflated version to you guys. I reread the instructions multiple times, but it was about 4:30am when I did that.

HebaruSan:
> So why didn't it give me the error when I inflated it here?

The first version of this pull request hard-coded the compatibility (and everything else) in the netkan, which bypasses the logic in question, but you really don't want that.

> I thought I was supposed to post the inflated version to you guys.

No, the inflated .ckan file gets generated automatically by the bot and saved into the CKAN-meta repo. The NetKAN repo is for the pre-inflated .netkan file.

> I reread the instructions multiple times, but it was about 4:30am when I did that.

If you can tell us where you thought you read that, we can try to make it clearer.

HebaruSan:
Looks like the download was replaced and no longer contains a .ckan file. :+1:

PR #8275, opened 2020-12-23
Split out WBI play modes

HebaruSan:
## Problem

See #8270, the WildBlueIndustries mods have something called "play modes" which are sets of configurations that can be swapped out to change how several mods work. They're distributed inside the downloads, with the default play mode's files distributed as .cfg files and the other play modes' files distributed as .txt files. Changing the play mode renames a bunch of files from .cfg to .txt and others from .txt to .cfg.

Currently CKAN just installs all of these files as they are in the ZIP, which is essentially unmaintainable, because the user can use a tool built into the mods to switch play modes, which renames the files, which means CKAN loses track of them and can't remove, upgrade, etc. properly.

## Changes

**NOTE: These changes are almost ready for merge!** So far it looks like this solution will work, but it would be a good idea to validate it with @Angel-125 before merging, and we should wait to test the Heisenberg+USI combo when it is ready.

Now the play modes are split into separate `<Mod>-PlayMode-<PlayModeName>` modules for these mods:

Mod | Play modes
:-- | :--
WildBlueTools | ClassicStock, CRP, Pristine, Simplified, USI
Buffalo | ClassicStock, CRP, USI
DSEV | ClassicStock, CRP, Pristine, USI
MOLE | ClassicStock, CRP, Pristine, USI
Pathfinder | ClassicStock, CRP, Pristine, Simplified, USI
Heisenberg | ClassicStock, CRP, Pristine, Simplified

The USI play mode is from @TheDogKSP's separate mod, also added in this PR. It is handled as a single module providing play modes for multiple mods, since it is distributed in one ZIP.

- Each play mode module installs .cfg files only, renaming upstream files from .txt as necessary.
- Each play mode module `provides` a `<MainMod>-PlayMode` virtual module, which is added to the main mods' `depends` lists. This way CKAN compels you to choose a play mode when installing any of these mods.
- Play mode modules depend on the main `WildBlue-PlayMode-<Whatever>` module for their play mode and conflict with the other play mode modules for the same mod, so you can only have one play mode installed at a time, and it will be the same one for all mods
- When a mod doesn't ship a set of files for a particular play mode, the core WildBlue play mode `provides` that play mode for that mod. For example, if you choose to install Buffalo, you can choose from `Buffalo-PlayMode-ClassicStock`, `Buffalo-PlayMode-CRP`, `WildBlue-PlayMode-USI`, `WildBlue-PlayMode-Pristine`, and `WildBlue-PlayMode-Simplified`. This way all play modes are available for all mods.
[...]

PR #8944, opened 2022-01-03
Split out KerbalKomets play modes

HebaruSan:
This mod has play modes and should have been included in #8275, but I only found out about it while reviewing output from #8943.

Now two new play mode modules are created for CRP and ClassicStock, and the WildBlueTools play modes for Pristine and Simplified provide KerbalKomets play modes as well.


HebaruSan:
Luckily there's ony one version of this mod in the metadata, so we don't need to worry about historical updates.

PR #8954, opened 2022-01-10
Add Extra Docking Ports

coldrifting:
Adds a few new useful docking ports and some parachutes to go with them.


HebaruSan:
Thanks for the submission!

PR #8957, opened 2022-01-12
Update scatterer Sompatibility

LGhassen:
Hello, I'm about to release an update which restores compatibility to 1.9-1.12.x

I'm not sure I did this correctly though?  The [spec page](https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md) explains how to use the operators but there seems to be no mention of what the `3:` preceding the version number does. In doubt I just pushed it like this.

### DIFF
--- a/NetKAN/Scatterer-config.netkan
+++ b/NetKAN/Scatterer-config.netkan
@@ -19,6 +19,10 @@ install:
     filter: Sunflares
 x_netkan_override:
   - version:
+      - '>=3:v0.0829'
+    override:
+      ksp_version_min: '1.9'
+  - version:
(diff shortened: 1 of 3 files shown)


HebaruSan:
Thanks for the update. Looks fine to me so far.

The `3:` is an epoch, which is incremented to keep numerically "out of order" upgrades like `v0.0195` &rarr; `v0.021c` (21 < 195) or `v0.0541` &rarr; `v0.055` (55 < 541) in-order, somewhat explained here:

https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#epoch

LGhassen:
I see, thanks a lot!

PR #8959, opened 2022-01-13
Add missing dependencies to SimpleContractDisplay

linuxgurugamer:
Follow-up of #8958, they have been missed there.

ckan compat add 1.8


linuxgurugamer (commented):
h

DasSkelett:
Merging now so CKAN users aren't left with a completely broken mod due to missing dependencies.
Didn't do an in-dept review, maybe it could use another tag, but I think we got the most important stuff.

PR #8963, opened 2022-01-15
[WIP] ModularLaunchPadsLite.netkan

yalov:
I want to make the MLP Lite entity on CKAN, only for the General stuff, 
@AlphaMensae, any objections ?

___

ckan compat add 1.11 1.12


yalov:
maybe later

PR #8965, opened 2022-01-18
Update Komplexity.netkan

zer0Kerbal:


HebaruSan:
Thanks for the update.

PR #8977, opened 2022-01-24
Add Sandcastle

HebaruSan:
image

image

Setting `resources.remote-avc` while Angel-125/Sandcastle#1 is in progress.


HebaruSan:
I ran `ModAnalyzer` on this, and it added a `B9PartSwitch` depends; omitted that because it's optional.

Issue #9014, opened 2022-02-22
[Mod] PRVE


SpaceODY:
Mod info
--------
* Mod name: PRVE
* Author(s): SpaceODY
* License: Primarily CC-BY-NC-SA 4.0
* Homepage: https://github.com/SpaceODY/PRVE (Currently don't have it on the forums)
* Download: https://github.com/SpaceODY/PRVE/releases

The zip files for each release are a bit tedious in their downloads since different variants need to overwrite the main download to work. I should probably use a :after for the configs in them. If you want anything changed with the setup of it let me know.

SpaceODY:
I'm looking through the guide more now and I'm going to set the mod up differently before I send a new request.

PR #9015, opened 2022-02-22
Add PhotoRealisticVisualEnhancement

HebaruSan:
See #9014, the author @SpaceODY requested this mod to be added, then later closed the request with plans to set things up differently.

https://github.com/SpaceODY/PRVE

### DIFF
--- a/NetKAN/PhotoRealisticVisualEnhancement-64k.netkan
+++ b/NetKAN/PhotoRealisticVisualEnhancement-64k.netkan
@@ -1,5 +1,7 @@
 spec_version: v1.4
 identifier: PhotoRealisticVisualEnhancement-64k
+name: PRVE 64k
+abstract: 64k Earth ground textures for PhotoRealisticVisualEnhancement
 $kref: '#/ckan/github/SpaceODY/PRVE/asset_match/PRVE64k-[0-9.]+\.zip'
 ksp_version: 1.10.1
 license: CC-BY-NC-SA-4.0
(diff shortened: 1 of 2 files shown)


HebaruSan:
- [ ] Probably needs to `provides`/`conflicts` some of the virtual config modules
- [ ] The repo says this is derived from 6 prior mods, so we should check them to make sure the license is OK
- [ ] The Scatterer dependency should be versioned

HebaruSan:
Hmm, I'm guessing it's bad to mix and match versions here:

```
   * scatterer 3:v0.0772 (spacedock.info, 51.6 MiB)
   * Scatterer Default Config 3:v0.0831 (cached)
```

SpaceODY:
I just updated the mod to better support ckan. The base mod download hasn't changed but the 64k addon has. The configs now overwrite other files via module manager. There is a separate gamedata folder for this addon labeled "PRVE-64k" now.

As for the scatterer v0.08 configs leave them off of ckan for now. If people wish to have them they will need to manually download it.

HebaruSan:
All the licenses that were easy to check, check out fine. The situation with KSRSS is painfully complicated, so we'll let them worry about checking that everyone's following all the rules and complain if they find problems. :relieved:

SpaceODY:
Alright, thank you for the help!

HebaruSan:
> Also, whenever I release a new version, how do I go about updating it on ckan?

CKAN will do that for you automagically when you upload a new release to GitHub (within 30 minutes or so). If you want to make changes to the compatibility or dependencies, you can either try editing the netkans yourself in a pull request, or describe the changes in a new issue like you did in #9014.

PR #9023, opened 2022-03-03

themaster402:
Replace deprecated KS3P recommendation with TUFX


HebaruSan:
Thanks for the update!

PR #9040, opened 2022-03-19
Update StagedAnimation ksp max version

NathanKell:
This is what I get for not using a .version


HebaruSan:
Thanks for the update!

PR #9041, opened 2022-03-20
Add Shabby shader bundle loader

drewcassidy:
Add the [Shabby shader bundle loader mod](https://github.com/taniwha/Shabby) by @taniwha
Permission given [here](https://forum.kerbalspaceprogram.com/index.php?/topic/194802-18-112-conformal-decals-0210-decals-done-the-right-way/&do=findComment&comment=4104168)


drewcassidy:
Ah, didn't see #8922

HebaruSan:
Let's do this in #8922 so we can test both mods side by side.

PR #9042, opened 2022-03-20
Add Conformal decals mod

drewcassidy:
Depends on #9041 since Shabby is a dependency for ConformalDecals


HebaruSan:
Let's do this in #8922 so we can test both mods side by side.

PR #9043, opened 2022-03-20
Suggest stock configs for ModuleDepthMask

drewcassidy:
So people aren't confused if they install DepthMask and don't see anything change.
Also mark Stock depth masks as incompatible with restock, since those patches won't work with it

Hope this is ok @kingnoob1377, I only just learned about your mod

### DIFF
--- a/NetKAN/DepthMask.netkan
+++ b/NetKAN/DepthMask.netkan
@@ -5,6 +5,8 @@ license: CC-BY-SA
 tags:
   - plugin
   - library
+suggests:
+  - name: StockDepthMaskConfig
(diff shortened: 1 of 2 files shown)


kingnoob1377:
This is completely fine♥️

HebaruSan:
Thanks for the update. :+1:

HebaruSan:
(And in case anyone else is confused as to how `resources.homepage` is getting set to SpaceDock for the config, it's set in the repo, here:

image
https://github.com/kingnoob1377/Stock_DepthMask_Config)

PR #9059, opened 2022-04-02
Mechjeb 2.14 version for KSP 1.12.x only

sarbian:
With an override because I already released a patch and want to make sure all is ok


HebaruSan:
Thanks for the update! :+1: 
The bot doesn't re-inflate old modules automatically, so we'll update those manually in CKAN-meta after this...

PR #9067, opened 2022-04-08
Update Komplexity.netkan

zer0Kerbal:
* updated to license
* requires either CBK ***or*** CBK-RO
thank you.

___

- https://forum.kerbalspaceprogram.com/index.php?/topic/202749-1123-komplexity-kplx-1000-release-edition-07-apr-2022/
- https://github.com/zer0Kerbal/Komplexity
- https://spacedock.info/mod/2805/Komplexity

### DIFF
--- a/NetKAN/Komplexity.netkan
+++ b/NetKAN/Komplexity.netkan
@@ -9,7 +9,6 @@ tags:
 depends:
   - name: ModuleManager
   - name: CustomBarnKit
-  - name: CustomBarnKit-RO
 supports:
   - name: Strategia
 conflicts:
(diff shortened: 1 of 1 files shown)


HebaruSan:
Thanks for the update, will merge it shortly.

A question regarding the license: Is this a _change_ (GPL-3.0 was correct for older versions), or a _correction_ (GPL-3.0 was incorrect for older versions)? If the latter then we will need to do some updates of the historical metadata.

zer0Kerbal:
> Thanks for the update, will merge it shortly.
> 
> A question regarding the license: Is this a _change_ (GPL-3.0 was correct for older versions), or a _correction_ (GPL-3.0 was incorrect for older versions)? If the latter then we will need to do some updates of the historical metadata.

excellent query - 0.9.9.9-prerelease on changed to the CC from GPL2. No historical changes should be needing to be made. thank you.

PR #9076, opened 2022-04-17
Make ProceduralParts conflict with KSP-Recall

siimav:
net-lisias-ksp/KSP-Recall#41

### DIFF
--- a/NetKAN/ProceduralParts.netkan
+++ b/NetKAN/ProceduralParts.netkan
@@ -23,6 +23,9 @@
         { "name" : "FreedomTex" },
         { "name" : "BasicProceduralTextures" }
     ],
+    "conflicts" : [
+        { "name": "KSP-Recall" }
+    ],
     "x_netkan_override": [
         {
             "version": "v1.2.10",


Lisias:
I strongly advise against this measure before some study on the matter.

Additionally, a simple patch can withdraw any Recall component from a Procedural Part if needed.

I suggest siimav to take a look on https://github.com/net-lisias-ksp/TweakScale/issues/240 before jumping into conclusions.

Lisias:
The problem **COULD NOT BE REPRODUCED** using the steps provided by one reporter.

THERE'S no misbehaviour between Procedural Parts and KSP-Recall.

I suggest this pull request be closed without merging, as the reason provided could not be validated.

I will pursue the real problem on the mentioned issue, as this is probably related to something else.

Lisias:
FYI the problem was finally reproduced and it's being diagnosed.

Lisias:
Diagnose concluded. IMHO Procedural Parts could be better initialised (preventing this mess not only to Recall, but to any PartModule that needs the node's definitions to be valid on OnLoad). But since I could work around the problem on Recall itself, I'm going to the path of less friction to have this issue solved at once.

I recomend this pull request to be closed without merging, as it definitely has no merit.

HebaruSan:
@Lisias, if you would prefer to keep Procedural Parts and KSP-Recall as compatible in the CKAN metadata, would you be willing to commit to testing them together before releasing future versions of KSP-Recall?

Try to look at this from the perspective of the KSP-RO team: Procedural Parts worked the same way for many, many years without problems, and suddenly their users started repeatedly asking for support about a weird new issue (the first GitHub issue I can find is KSP-RO/ProceduralParts#315, but I fould five additional discussions of it on the KSP-RO Discord in the past month). Normally we don't add conflicts just for simple bugs (especially when [a fix is already in the pipeline](https://github.com/net-lisias-ksp/KSP-Recall/commit/5d3056e78b85d2ba6fa462002ead237010d01bfe)) I can understand the KSP-RO team wanting to protect themselves and their users from such issues in the future.

Lisias:
> @Lisias, if you would prefer to keep Procedural Parts and KSP-Recall as compatible in the CKAN metadata, would you be willing to commit to testing them together before releasing future versions of KSP-Recall?

I already did such tests, diagnosed the real cause of the mishap on Procedural Parts but since the faster way out of the mess is a work around, this is currently in "Acceptance Tests" before being release.

There was no problem on KSP-Recall side, it was a mishap on the Part Configs on Procedural Parts that were informing the wrong node definitions for the part.

So, In a nut shell, I will commit on keeping KSP-Recall from stepping on the RO's toes if the RO guys would kindly do me the favor of reporting the problems they find in a way I could reproduce. :) All the reports I had until today were dismissed **after being proven not accurate**. Only one someone finally demonstrated how to reproduce the problem.


> Try to look at this from the perspective of the KSP-RO team: Procedural Parts worked the same way for many, many years without problems, and suddenly their users started repeatedly asking for support about a weird new issue (the first GitHub issue I can find is [KSP-RO/ProceduralParts#315](https://github.com/KSP-RO/ProceduralParts/issues/315), but I fould five additional discussions of it on the KSP-RO Discord in the past month). Normally we don't add conflicts just for simple bugs (especially when [a fix is already in the pipeline](https://github.com/net-lisias-ksp/KSP-Recall/commit/5d3056e78b85d2ba6fa462002ead237010d01bfe)), but I can understand the KSP-RO team wanting to protect themselves and their users from such issues in the future.

I did it. On the other hand, the RO guys could try to look from the perspective of 3rd parties that relies on stablished behaviours of KSP. 

Recall was fooled because it trusts the Part Config is accurate on the OnLoad callback (what was always true until this problem). Had the Config files be using node definitions that matched the default sizes of the parts, none of this would had happened.

Please note that the RO guys closed the issue on the spot and called for a conflict without **ANY DISCUSSION** of the problem with me. 

And I also need to protect KSP-Recall (and TweakScale) users from these problems myself.

It's my understanding that we are all on the same boat.

HebaruSan:
> Had the Config files be using node definitions that matched the default sizes of the parts, none of this would had happened

Is that something that could be submitted in a pull request?

Lisias:
> That's conditional on someone else doing something for you and also not what I was asking (avoiding breaking things versus simple testing before release), so I will take that as a "No." If future KSP-Recall releases will not be tested with Procedural Parts, then maybe a conflict is called for here after all.

KSP-Recall **is not the one breaking things**. You are essentially asking me to be the beta tester for their add'on!!.

The problem **is** on Procedural Parts, and I'm fixing it and submitting a pull request to the maintainer.

In the mean time, the workaround on KSP-Recall will be deployed the same - so even if RO guys refuses the pull request that would fix their artefacts, KSP-Recall will not be one that will bork anymore due Procedural Part's problem.

Please be kind and wait a few hours while I write (also) the fix for Procedural Parts.

However, if the only way to avoid a gratuitous conflict caused by Procedural Part itself, ok. I agree on being their free beta tester and reporting any problem caused by Procedural Parts if I detect a new one.

Lisias:
@HebaruSan 

A patch for Procedural Parts was published here: https://github.com/net-lisias-ksp/KSP-Recall/issues/41#issuecomment-1103055074

it solves the problem **NOW** while I finish the next release of KSP-Recall, where I will work around the problem anyway. I will advise as soon as I apply the Pull Request to the upstream.

Lisias:
A pull request was applied to Procedural Parts.

https://github.com/KSP-RO/ProceduralParts/pull/316

PR #9079, opened 2022-04-20
Update PRVE compat

HebaruSan:
image

Closes KSP-CKAN/CKAN-meta#2683.

___

ckan compat add 1.12


HebaruSan:
The Scatterer dependency's compat is out of date:

image

SpaceODY:
Not sure if it's because it hasn't refreshed yet but the new 1.7.1 version is only available for PRVE and not PRVE64k

HebaruSan:
We're getting an error for that one that I don't understand yet:

image

Hopefully our server guy will have a chance to look into it, could just be that we need more disk space.

HebaruSan:
Should be resolved now, sorry for the delay!

https://github.com/KSP-CKAN/CKAN-meta/commit/2fbc92b8aec53af0067ee4531d45ecbee95e70a3

Issue #9082, opened 2022-04-22
PartWizard compatibility information seems wrong for KSP 1.9.1.2788


damerell:
Background
----------
* Operating System: Windows 7 (yes, I know; no, it isn't)
* CKAN Version: 1.30.4
* KSP Version: 1.9.1.2788

**Have you made any manual changes to your GameData folder (i.e., not via CKAN)?**
No; completely clean install other than using CKAN to install PartWizard and its dependencies. (I did originally observe the issue in a heavily modded 1.9.1).

CKAN offers me PartWizard 1.3.8.3 for KSP 1.9.1. In the VAB, clicking PartWizard's icon tells me "An error has occurred in PartWizard." 1.3.8.2 works as expected.

HebaruSan:
The version file for 1.3.8.3 says:

```json
  "KSP_VERSION_MIN": {
    "MAJOR": 1,
    "MINOR": 8,
    "PATCH": 0
  }
```

And SpaceDock says:

![image](https://user-images.githubusercontent.com/1559108/164584757-2863a8af-1d71-46e1-85d4-a46e547ec621.png)

And that's what's in the metadata:

```json
    "ksp_version_min": "1.8.0",
    "ksp_version_max": "1.12.1",
```

So it seems to be fine. @linuxgurugamer, was the version file wrong for this version?

damerell:
It does not seem to be fine, since 1.8.0 is less than 1.9.1, but I appreciate it has been fixed upstream.

DasSkelett:
`KSP_VERSION_MIN` means the _minimum_ KSP version, so CKAN's data does match what was in the version file at that time.

damerell:
I understand it means the minimum KSP version; however, since 1.8.0 <= 1.9.1, the effect is that a version of the mod that doesn't work gets installed, and this is not fine. (I appreciate that this is an upstream bug).

DasSkelett:
> the effect is that a version of the mod that doesn't work gets installed, and this is not fine.

The "So it seems to be fine." was in context of CKAN's metadata being correct, which it is, so the metadata is fine.
This is our primary area of concern.

That it doesn't work for you is not fine of course. But before changing the metadata we have to coordinate with the author. Because right now there's no indication that it is actually a compatibility problem (that's just your assumption, maybe a correct one, but still just a guess), it might very well be a different bug.

HebaruSan:
[...]

PR #9083, opened 2022-04-22
Roverscience

linuxgurugamer:


HebaruSan:
Please check out the question on #9082 while you're here, if you could.

PR #9094, opened 2022-04-28
Switch ImpossibleInnovations kref to GitHub

HebaruSan:
This mod has one more release on GitHub:

https://github.com/net-lisias-ksp/ImpossibleInnovations/releases

... than it does on SpaceDock:

https://spacedock.info/mod/340/Impossible%20Innovations#changelog

Now we get it from GitHub.


HebaruSan:
Oh right, there was at least one mod with `/` in its version numbers, which breaks the validation script.

image

PR #9105, opened 2022-05-04
Update OblivionAerospace.netkan

zer0Kerbal:
correct license


HebaruSan:
Validation failed:

https://github.com/KSP-CKAN/NetKAN/runs/6282611318

image

List of allowed licenses:

https://github.com/KSP-CKAN/CKAN/blob/4e0b9d8818ce52243f8f32b3453e67257306cd4a/CKAN.schema#L444-L477

(Needs to be in all caps)

PR #9122, opened 2022-05-12
Update DaMichelLtd.netkan

zer0Kerbal:
- agency
  - flags
  - config


HebaruSan:
Redundant with #9116.

zer0Kerbal:
> Oh, you targeted the PR's branch. Sure, that works.

only way it would let me edit it. same goes with other.

PR #9149, opened 2022-06-03
Update and rename ChineseSpacecraftPack.frozen to ChineseSpacecraftPa…

Armo00:
…ck.netkan

Chinese Spacecraft Pack is now back


HebaruSan:
Thanks for the update!

PR #9135, opened 2022-05-23
Antennas

zer0Kerbal:
update homepage
update license

PR #9151, opened 2022-06-04
Update and rename Antennas.frozen to Antennas.netkan

zer0Kerbal:


JonnyOThan:
https://github.com/KSP-CKAN/NetKAN/pull/9135

PR #9154, opened 2022-06-04
Update NTChineseRocketsPack.netkan

Armo00:
Remove the file from ICSP as ICSP is now maintained by another author.


DasSkelett:
Ah yeah, that's been #9149.
Do we need to add a dependency on ChineseSpacecraftPack (ICSP)?

Armo00:
Negative, ChineseSpacecraftPack and NTChineseRocketsPack is not dependent on each other.

PR #9193, opened 2022-07-03
ROCapsules use remote netkan

NathanKell:
We've had a remote netkan for a while but apparently no one ever PR'd NetKAN to update to point to it... :(

### DIFF
--- a/NetKAN/ROCapsules.netkan
+++ b/NetKAN/ROCapsules.netkan
@@ -1,4 +1,4 @@
-spec_version: v1.12
+spec_version: 1
 identifier: ROCapsules
 $kref: '#/ckan/netkan/https://github.com/KSP-RO/ROCapsules/raw/master/ROCapsules.netkan'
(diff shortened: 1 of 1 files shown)


NathanKell:
Once again I forgot to strip the eleventy-million recs, so inflate was gonna take forever.

HebaruSan:
Thanks for the update. :+1: A few points that aren't worth holding it up:

- `resources.homepage` formerly was a link to the KSP-RO Discord, and it's not in the new metanetkan. It might be nice to keep this so users can find you.
- You can use YAML for the remote metanetkan if you like.

NathanKell:
Ah yeah lemme fix the homepage thing, thanks. Also I just copied RSS's metanetkan which is why it's in YAML and was spec version 1.12.
The remote I guess we could convert over, but for now I'll leave in json? It doesn't really matter to me anyway, and I don't think anyone else in RO has strong opnions?

Cheers!

PR #9215, opened 2022-07-19
Kerbalism and official config packs metadata updates

gotmachine:
- Added "depends" on KSPCommunityFixes for Kerbalism core. Technically not a hard dependency but it fixes several game-breaking bugs that particularly affect Kerbalism and I'm getting tired of asking users to install it.
- Moved "depends" on CommunityResourcePack from Kerbalism core to configs. It isn't actually a core dependency and alternate config packs may not want it installed.
- Removed all "supports" from Kerbalism core. Most of them are actually not that well supported, and alternate config packs might have a different view on what is supported.

### DIFF
--- a/NetKAN/Kerbalism-Config-Default.netkan
+++ b/NetKAN/Kerbalism-Config-Default.netkan
@@ -18,7 +18,8 @@
     ],
     "depends": [
         { "name": "ModuleManager" },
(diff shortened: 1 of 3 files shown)


HebaruSan:
Thanks for the update!

PR #9218, opened 2022-07-22
Add ROHeatshields to CKAN

NathanKell:
We finally got around to releasing this, adding the netkan.

___

- https://github.com/KSP-RO/ROHeatshields

### DIFF
--- a/NetKAN/ROHeatshields.netkan
+++ b/NetKAN/ROHeatshields.netkan
@@ -1,0 +1,9 @@
+spec_version: 1
+identifier: ROHeatshields
+$kref: '#/ckan/netkan/https://github.com/KSP-RO/ROCapsules/raw/master/ROHeatshields.netkan'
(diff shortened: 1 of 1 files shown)


DasSkelett:
Is there a place with more information about the mod? There's no README in the zip, the one in the repo has only one line, and I also can't find a forum thread for it. Just to have a rough overview and making sure the tags fit.

NathanKell:
Thanks for the reminder on that; as we've been using the prerelease for a while in the KSP-RO community I forgot that, well, folks might well want to know more.

I've updated the readme to be much clearer (on the repo) and included both readme and license in the root of the zip. There is no forum thread to point at though (as is true for all the other RO* mods e.g. ROTanks or ROCapsules), as most all discussion and support occurs on the RO discord.

HebaruSan:
Thanks for the submission!

PR #9219, opened 2022-07-22
Add KSPCF as dependency to ProceduralParts

StonesmileGit:
As of the latest version of ProceduralParts, KSPCF is needed. Adding as a dependency


DasSkelett:
Thanks for the update!
(You might also want to add the dependency to the forum post, especially since it isn't bundled so people know about it)

PR #9221, opened 2022-07-23
Rename BDArmoryForRunwayProject to BDArmoryPlus

josuenos:
- BDArmory for Runway Project is being re-branded as BDArmory Plus, the Spacedock link is remaining the same.


HebaruSan:
Changing an `identifier` isn't well-supported. Users would not be able to upgrade to newer versions, dependencies would break, etc.

The `name` is already updated from SpaceDock. Do you really need the `identifier` updated as well?

https://github.com/KSP-CKAN/CKAN-meta/commit/2006ed8b1bef89db56f8114694e358c8698515b3

PR #9223, opened 2022-07-24
Updates for PhotoRealisticVisualEnhancement

### LINKED ISSUE #9222: [Mod] PRVE 1.9 Changes
With 1.9 released there are two things I need changed:

1. I forgot to retag the draft from v1.8.1 to v1.9 so I will need the .ckan file for v1.8.1 deleted and a new one for v1.9 with its correct links and version number created (sorry).
image

2. There is a new mod I want added from the GitHub release on 1.9 called "PRVE-LRv1.9". 
Name = "PRVE Low Resolution"
Conflicts: PRVE
Dependencies: same as PRVE

3. Finally, since "PRVE Low Resolution" will conflict with "PRVE", I need "PRVE" to also conflict with "PRVE Low Resolution" (Vice versa)
Thanks!

___

https://github.com/SpaceODY/PRVE

HebaruSan:
https://github.com/SpaceODY/PRVE

- Filename format for 64k changed from `PRVE64k-` to `PRVE-64k`
- New low res variant added starting as copy of regular variant
- Conflicts added between low res and regular. One conflict would be enough for the relationship resolver, but this way it will show up under the Relationships tab for both mods.

Fixes #9222.

### DIFF
--- a/NetKAN/PhotoRealisticVisualEnhancement-64k.netkan
+++ b/NetKAN/PhotoRealisticVisualEnhancement-64k.netkan
@@ -2,7 +2,7 @@ spec_version: v1.4
 identifier: PhotoRealisticVisualEnhancement-64k
 name: PRVE 64k
 abstract: 64k Earth ground textures for PhotoRealisticVisualEnhancement
-$kref: '#/ckan/github/SpaceODY/PRVE/asset_match/PRVE64k-[0-9.]+\.zip'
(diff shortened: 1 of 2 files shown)


HebaruSan:
I'm going to cancel the validation script because I think it ran out of CPU, RAM, and network capacity trying to process these mods. At least the inflation step completed OK.

PR #9231, opened 2022-07-27
Update Spectra-VibrantSunsets.netkan

Avera9eJoe:
Add Scatterer as dependent mod as it is required as well to make this work


HebaruSan:
Thanks for the update!
(Prepare to read that sentence several more times... :raised_eyebrow:)

PR #9237, opened 2022-07-27
Update and rename RVE64kKerbinCloudsforSpectra.netkan to Spectra-RVE6…

Avera9eJoe:
…4kKerbinClouds.netkan

Changes to match other Spectra Add-ons

https://spacedock.info/mod/2471/RVE64k%20Kerbin%20Clouds%20for%20Spectra


Avera9eJoe:
This change would break any existing ties in other `.netkan` files. Thankfully the only `.netkan` that references this is Spectra, and with the update in a few days it will be moved to a `.ckan` file included in the download and thus fix itself

(comment for posterity)

HebaruSan:
It's also worth noting that you're the author (see #8030), so we don't have to worry about anyone else being upset about this change.

HebaruSan:
@Avera9eJoe, you might consider clicking the update button if this still works in current game versions:

image

HebaruSan:
Let's give this a try :+1: .

PR #9236, opened 2022-07-27
Create Spectra-MinmusGeysers.netkan

Avera9eJoe:
Add-on for Spectra

PR #9235, opened 2022-07-27
Create Spectra-MinmusScatterer.netkan

Avera9eJoe:
Add-on for Spectra

PR #9239, opened 2022-07-27
Install whole subfolders of Spectra Extras

HebaruSan:
## Background

Spectra has "Extras" that look like this:

image

## Problem

In #9238 and #9236 and #9235 and #9231 we set up netkans to install these by selecting the `Spectra_EVE`, `Spectra_Scatterer`, and `Spectra_textures` folders and installing them to `GameData/Spectra`. While this gives a correct result, it's somewhat inflexible; if @Avera9eJoe subsequently decides for example to add a `Minmus Geysers/Spectra_Scatterer` folder, we would need to update the netkan to install that. (Also I have some ulterior motives related to internal .ckan files which may come to fruition at a later time.)

## Changes

Now we select the main folder for each Extra and use `as` to rename it to Spectra and install it to GameData. This should still produce a correct install but with more futureproofing for these modules. The `spec_version`s are updated to v1.18 to allow use of `as`.

### DIFF
--- a/NetKAN/Spectra-JoolShineOnLaythe.netkan
+++ b/NetKAN/Spectra-JoolShineOnLaythe.netkan
@@ -1,4 +1,4 @@
-spec_version: v1.4
+spec_version: v1.18
 identifier: Spectra-JoolShineOnLaythe
 name: Jool shine on Laythe for Spectra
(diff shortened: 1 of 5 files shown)


HebaruSan:
Seemed to work! :tada:

PR #9244, opened 2022-07-31
Let Spectra's internal ckan files set its relationships

HebaruSan:
See KSP-CKAN/CKAN#3615, we are planning to move Spectra's relationships to internal .ckan files so @Avera9eJoe can manage them more easily. This PR removes them from the netkans. It should not be merged until there is a new Spectra release containing the corresponding internal .ckan files.

https://spacedock.info/mod/1505/Spectra


HebaruSan:
Confirmed the 1.6.3 release has the internal .ckan files.

PR #9245, opened 2022-08-01
Spectra mod relations and updates

Avera9eJoe:
Removed other mod suggestions as they are now included in a `.ckan` file in the download, updated abstract


HebaruSan:
Thanks for the update. Will take care of this in #9244.

PR #9249, opened 2022-08-04
Spectra patches

Avera9eJoe:
Hoping this is it, and that I won't have to change anything dramatic after this goes through.

Big updates to the internals of Spectra. Scatterer configs have been moved to Extras, and files have been updated so `Scatterer-config` and `Scatterer-sunflare` are no longer required to install—and are now marked as being provided by Spectra. Older configs depended on the default sunflare and configs being installed as well as they were just patches for these, but the configs have been properly updated, and can now be installed as complete replacements for the default configs, if someone wishes to use theme standalone.

Addons have been renamed to prefix the parent mod they rely on, cleaning up their order in CKAN and better reflecting what mod they add effects for.

The recommends/suggests should also be properly working too, though it's hard to check this without making it live first.


HebaruSan (approved):
- [x] The `install` stanzas look good
- [x] The `.ckan` files seem to be valid and applied correctly
- [x] The inflation errors and warnings are fixed

I'm nervous about changing identifiers again, but it might be OK. Let's try it.

PR #9252, opened 2022-08-05
Spectra - fix addon names/identifiers

Avera9eJoe:
Pull 1 of probably 2; there are three addons which still show duplicate versions in CKAN which this pull does not address. I'm still looking into what's creating them, it could be tied to the old 1.6.3 Spectra release, since they're all three labeled latest version 1.6.3.

These three should be removed from appearing in CKAN:
Identifier: `Spectra-KerbinLaytheSnow`
Identifier: `Spectra-MinmusGeysers`
Identifier: `Spectra-MinmusScatterer`

I'm still looking into what's creating these, it might be the .frozen files? Or possibly the 1.6.3 release if that's possible. Spectra v1.6.3 can be completely removed from CKAN, even, as it's an inferior update to 1.6.5. Heck, ALL previous versions of Spectra can be safely deleted, since all releases are backwards compatible to the first.


HebaruSan:
The duplicates were caused by #9249 changing the identifiers and fixed in KSP-CKAN/CKAN-meta#3038.

Avera9eJoe:
Gotcha. I can re-release Spectra this evening as 1.6.6, updating the recommended mods list to use the correct (old) identifiers, that way it appears right in CKAN and doesn't require changing the identifiers/filenames again. It'll just be the title in CKAN that updates

HebaruSan:
No, we've just changed to the new ones, leave it

Avera9eJoe:
Oh, sounds good then. I'm on mobile so I couldn't check properly — I'll peek again when home

PR #9257, opened 2022-08-06
Add DeployableInventory

yalov:
Mod descriptors: inventory - Mods that use stock inventory system

- https://github.com/yalov/DeployableInventory
- https://forum.kerbalspaceprogram.com/index.php?/topic/209148-112-deployable-inventory-beta-2022-08-01/

### DIFF
--- a/NetKAN/DeployableInventory.netkan
+++ b/NetKAN/DeployableInventory.netkan
@@ -9,6 +9,7 @@ resources:
 tags:
   - plugin
   - config
+  - crewed
 depends:
   - name: ModuleManager
   - name: StationPartsExpansionRedux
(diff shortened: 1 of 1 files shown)


HebaruSan:
Thanks for the submission!

yalov:
parts in the DeployableInventory are not crewed.

inventory tag could also be added to the KSP-PartVolume

HebaruSan:
> parts in the DeployableInventory are not crewed.

OK, I'll remove that. Thanks for noting it.

> inventory tag could also be added to the KSP-PartVolume

Could be but hasn't been. See:

- https://github.com/KSP-CKAN/CKAN/wiki/Suggested-Tags#new-tags

PR #9258, opened 2022-08-06
SandcastleCore

yalov:
version of the sandcastle without Parts, just dll.
Suppose to be used without EL, just printer plugin for ScienceLab
The ScienceLab3DPrinter.txt file is renamed to the ScienceLab3DPrinter.cfg by the CKAN

is it ok, @Angel-125 ?

### DIFF
--- a/NetKAN/SandcastleCore.netkan
+++ b/NetKAN/SandcastleCore.netkan
@@ -17,6 +17,7 @@ recommends:
   - name: KSP-PartVolume
 conflicts:
   - name: Sandcastle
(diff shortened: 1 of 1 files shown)


yalov:
added ExtraPlanetaryLaunchpads to the conflict of the SandcastleCore, for clarity:
with the EL user need Sandcastle, but not the SandcastleCore

HebaruSan:
Rather than leaving this open indefinitely, please let us know if you hear from @Angel-125 and we can re-open this. In the meantime, perhaps you can depend on `Sandcastle`, which is already indexed.

PR #9269, opened 2022-08-12
Fix Shabby max_version in conflicts section

Icecovery:
In #9240 I suggested `v0.2.0` should be the minimum version of Shabby that Shaddy supports. 

Now that Shabby 0.3.0.0 has been added to CKAN in #9262, and apparently, CKAN doesn't think 0.3.0.0 is above v0.2.0 and is stopping me from installing Shaddy v2.1 when Shabby 0.3.0.0 is installed. The reverse is also true: when Shaddy v2.1 is installed, I cannot install Shabby 0.3.0.0.

I have therefore changed the max_version of Shabby in the conflicts section from `v0.2.0` to `0.2.0.0` to match the version scheme Shabby uses.

image  

image

I am using CKAN v1.31.0 in case it is me who messed up something with CKAN.

Aso do we have to update the [meta file](https://github.com/KSP-CKAN/CKAN-meta/blob/master/Shaddy/Shaddy-v2.1.ckan)?

### DIFF
--- a/NetKAN/Shaddy.netkan
+++ b/NetKAN/Shaddy.netkan
@@ -8,7 +8,7 @@ tags:
   - graphics
 conflicts:
   - name: Shabby
-    max_version: v0.2.0
+    max_version: 0.2.0.0
 depends:
   - name: ModuleManager
 recommends:


HebaruSan (approved):
Oops, that makes sense. Nice catch and thanks for the fix.
The bot will take care of the file in CKAN-meta after merge (up to 35 minute delay depending on system load).

PR #9278, opened 2022-08-15
Update Scatterer for PRVE to 0.0838

SpaceODY:


HebaruSan:
Hi @SpaceODY, thanks for the update. What do you think about making this relationship version-independent (i.e., removing the `version` property altogether), so we would no longer have to update it every release? Do you feel that future updates of Scatterer are likely to break compatibility again?

SpaceODY:
I was just thinking the exact same thing. Since scatterer probably won't go through the update it did from 0.07 to 0.08 again lets remove the dependency for the scatterer version.

PR #9288, opened 2022-08-20
Add Galaxies Unbound optional Clouds module

HebaruSan:
There's one more module that we missed in #8515:

image

Now it's added.

### DIFF
--- a/NetKAN/GalaxiesUnbound-StockClouds.netkan
+++ b/NetKAN/GalaxiesUnbound-StockClouds.netkan
@@ -1,5 +1,7 @@
 spec_version: v1.4
-identifier: GalaxiesUnbound-Core
+identifier: GalaxiesUnbound-StockClouds
+name: Galaxies Unbound - Stock Clouds
+abstract: Clouds for Galaxies Unbound
 $kref: '#/ckan/github/StarCrusher96/Galaxies-Unbound-A-Stellar-Odyssey/asset_match/Core'
 ksp_version_min: 1.11
 ksp_version_max: 1.12
@@ -8,15 +10,14 @@ resources:
   homepage: https://forum.kerbalspaceprogram.com/index.php?/topic/198895-*
 tags:
   - config
-  - planet-pack
+  - graphics
+provides:
+  - EnvironmentalVisualEnhancements-Config
+conflicts:
+  - name: EnvironmentalVisualEnhancements-Config
 depends:
-  - name: ModuleManager
-  - name: Kopernicus
-recommends:
+  - name: GalaxiesUnbound-Core
   - name: EnvironmentalVisualEnhancements
-  - name: Scatterer
-suggests:
-  - name: GalaxiesUnbound-Clouds
 install:
-  - find: GU
+  - find: GU_StockClouds/GameData/GU
     install_to: GameData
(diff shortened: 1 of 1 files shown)


JonnyOThan:
Hey it looks like the recommendation for the base mod is wrong.  `GalaxiesUnbound-StockClouds` is the identifier for the clouds module, but the base module recommends `GalaxiesUnbound-Clouds`

HebaruSan:
Ahh, good catch, I renamed it and didn't update that relationship.

JonnyOThan:
Hmm, CKAN is still indicating that the clouds aren't indexed, but the names match now.
image

HebaruSan:
Hmm, it works fine for me:

image

It looks like there may be something else wrong with your registry, Kopernicus, EVE, and Scatterer all have further relationships and should have a `[+]` icon next to them for expanding. And the mod versions should be shown as well. Are you on a really old CKAN or something?

JonnyOThan:
I'm on latest ckan.

Interesting, I switched to a different game instance and now it looks correct.  But swapping back to the original one still shows the broken metadata.  I have everything set to auto-refresh, and a manual refresh doesn't fix it either.  So it seems like something stale is getting cached somewhere.

HebaruSan:
Can you try the test build from KSP-CKAN/CKAN#3638? That has as significant rewrite of the relationships tab and I'm curious whether it's still affected. It's possible that whatever the cause is would have been fixed incidentally during refactoring there.

JonnyOThan:
I assume you mean use one of the builds from CI?  Does it matter which one?

HebaruSan:
Yeah, I meant the one under the Artifacts dropdown here:
https://github.com/KSP-CKAN/CKAN/pull/3638/checks

JonnyOThan:
OK, it still shows it as missing by default, but if I click the "reverse relationships" button then it seems to pick it up properly.

HebaruSan:
Thanks, so that probably means the problem is somewhere other than the UI code, possibly in core relationships handling. Have you had any other weird glitches with that game instance? Dependencies missing, mods not installing, etc.?

JonnyOThan:
Not that I can recall but it's been a while since I did any mod management there.

Also, with reverse relationships enabled, the GU stock clouds entry indicates that it conflicts with itself.

JonnyOThan:
[CKAN_old.zip](https://github.com/KSP-CKAN/NetKAN/files/9389512/CKAN_old.zip)
I renamed my CKAN directory for that instance and now it works correctly.  Here's the broken one.

HebaruSan:
Thanks, I can reproduce the issue if I overwrite my registry.json with yours, now to investigate...

HebaruSan:
... the lack of mod versions is because the format is different when they're installed, I just couldn't find that codepath at first...

HebaruSan:
... and the "not indexed" message is because you already have an EVE config installed, AVP, which conflicts with the stockclouds module. Uninstalling AVP makes it look like my screenshot. I'll look for ways to make this clearer, probably in that same pull request...

HebaruSan:
The test build should now be updated with fixes for both of those issues, if you care to re-download and try it out:
https://github.com/KSP-CKAN/CKAN/pull/3638/checks

JonnyOThan:
Yep, all seems well with that build.  Thanks!

PR #9298, opened 2022-08-27
Revert "Adopting of Targetron"

linuxgurugamer:
Reverts KSP-CKAN/NetKAN#9297


linuxgurugamer:
I clicked merge too quickly, so reverted it

PR #9310, opened 2022-08-30
Buffalo2 and WildBlueCore NetKAN Additions

Angel-125:
Adds WildBlueCore and Buffalo 2 to CKAN.

### DIFF
--- a/NetKAN/Buffalo2.netkan
+++ b/NetKAN/Buffalo2.netkan
@@ -1,40 +1,33 @@
-{
-    "spec_version": "v1.4",
-    "identifier":   "Buffalo2",
-    "name":         "Buffalo2",
-    "abstract":     "A next-gen, Restock-alike replacement for the original Buffalo mod",
-    "author":       "Angel-125",
-    "$kref":        "#/ckan/github/Angel-125/Buffalo2",
-    "$vref":        "#/ckan/ksp-avc",
-    "license":      "GPL-3.0",
-    "resources": {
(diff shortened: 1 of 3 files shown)


HebaruSan:
Saw this:

image

And realized Sandcastle is already indexed, so we better make sure they won't conflict. The current release doesn't bundle it, but it's in a pre-release, so presumably the next version as well:

image

image

Updated Sandcastle's `install` to only pick out the Sandcastle folder to avoid a future file conflict. We can add the dependency when it's needed.

Angel-125:
> Saw this:
> 
> image
> 
> And realized Sandcastle is already indexed, so we better make sure they won't conflict. The current release doesn't bundle it, but it's in a pre-release, so presumably the next version as well:
> 
> image
> 
> image
> 
> Updated Sandcastle's `install` to only pick out the Sandcastle folder to avoid a future file conflict. We can add the dependency when it's needed.

WildBlueCore is already indexed? I know WildBlueTools is, but I'd forgotten that WBC was already in CKAN. Great spot, thank you :)

HebaruSan:
> WildBlueCore is already indexed? I know WildBlueTools is, but I'd forgotten that WBC was already in CKAN. Great spot, thank you :)

No, **Sandcastle** is already indexed, sorry for the confusion.

HebaruSan (approved):
Looks good to me, let's give this a try. Thanks for the submission!

PR #9318, opened 2022-09-04
Update FerramAerospaceResearchContinued dependencies

dkavolis:
Next version of FerramAerospaceResearchContinued will depend on KSPBurst.

Also, could you change the indexed versioning to be dependent on the release tag instead of master branch? I'm not sure which field needs to be changed and to what. Thanks.

### DIFF
--- a/NetKAN/FerramAerospaceResearchContinued.netkan
+++ b/NetKAN/FerramAerospaceResearchContinued.netkan
@@ -21,7 +21,8 @@
     "depends" : [
         { "name" : "ModuleManager" },
         { "name" : "ModularFlightIntegrator" },
-        { "name" : "TweakScale-Redist" }
+        { "name" : "TweakScale-Redist" },
+        { "name" : "KSPBurst" }
     ],
     "install" : [
         {


HebaruSan:
> Also, could you change the indexed versioning to be dependent on the release tag instead of master branch? I'm not sure which field needs to be changed and to what. Thanks.

Hi @dkavolis, can you please explain what you mean by this? What problem are you trying to solve? CKAN looks at release tags by default, and this mod is already set up that way, but maybe something else is going on.

dkavolis:
The KSP compatibility versions were indexed when I updated the version file, ideally they should be taken from the release.

HebaruSan:
Oh, that's how version files are supposed to work (in KSP-AVC as well), not something special about how CKAN is indexing this mod. If the `URL` field points to a valid version file, and if the `VERSION` property of the local and remote version files match (both say 0.16.0.5 for this mod currently), then the remote one takes precedence. That way you can update compatibility after release. The idea was you could release a version file when say KSP 1.10 was current with compatibility extending only that far, and then when KSP 1.11 comes out, you can test the current release and just update the remote version file instead of having to make a new release with a new version file. Examples:

- https://github.com/HebaruSan/Astrogator/commit/7504bb7899c02c46d79c2179ec8aea995731c559
- https://github.com/HebaruSan/Astrogator/commit/a29cb3769939f5e08d18d1caef238d223c99772c

If you want to advertise a new version that isn't related to the version people are using, you can update `VERSION` to whatever that release number will be, and CKAN will no longer treat it as replacing the local file.

dkavolis:
Thanks. Let's just update the dependencies then.

HebaruSan:
Sounds good. If we merge this right now, it will change the dependencies of 0.16.0.5, and you said this was for the next version, so I'm assuming we should wait. I have just signed up for release notifications on your repo so I can merge this as soon as the new release is out.

HebaruSan:
... there it is:

https://github.com/dkavolis/Ferram-Aerospace-Research/releases/tag/v0.16.1.0_Marangoni

HebaruSan:
Thanks for the update!

PR #9319, opened 2022-09-05
Add Parallax-Scatter-Textures requirement

Gameslinx:
Parallax_ScatterTextures is required alongside Parallax_StockTextures as of release 2.0.0. If stock textures are installed, these must also be installed. However, installing Parallax will not prompt either the stock textures or scatter textures to download.

https://github.com/Gameslinx/Tessellation

### DIFF
--- a/NetKAN/Parallax-ScatterTextures.netkan
+++ b/NetKAN/Parallax-ScatterTextures.netkan
@@ -1,0 +1,31 @@
+{
+    "spec_version": "v1.4",
+    "identifier":   "Parallax-ScatterTextures",
+    "name":         "Parallax - Scatter Textures",
+    "abstract":     "Parallax textures for the scatters on the stock planets",
+    "author":       "Linx",
+    "$kref":        "#/ckan/github/Gameslinx/Tessellation/asset_match/Scatter",
+    "license":      "CC-BY-NC-ND-4.0",
+    "resources": {
+        "homepage": "https://forum.kerbalspaceprogram.com/index.php?/topic/197024-*"
+    },
+    "tags": [
+        "config",
+        "graphics"
+    ],
+    "depends": [
+        { "name": "ModuleManager" },
+        { "name": "Parallax" },
+        { "name": "Parallax-StockTextures" }
+    ],
+    "provides": [
+        "Parallax-Scatter-Textures"
+    ],
+    "conflicts": [
+        { "name": "Parallax-Scatter-Textures" }
+    ],
+    "install": [{
+        "find":       "Parallax_ScatterTextures",
+        "install_to": "GameData"
+    }]
+}
--- a/NetKAN/Parallax-ScatterTextures.netkan
+++ b/NetKAN/Parallax-ScatterTextures.netkan
@@ -1,31 +1,23 @@
-{
-    "spec_version": "v1.4",
-    "identifier":   "Parallax-ScatterTextures",
-    "name":         "Parallax - Scatter Textures",
-    "abstract":     "Parallax textures for the scatters on the stock planets",
-    "author":       "Linx",
-    "$kref":        "#/ckan/github/Gameslinx/Tessellation/asset_match/Scatter",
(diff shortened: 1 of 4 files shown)


HebaruSan:
There's a problem with the new install stanza:

image

The path is wrong, but I'll have to download the ZIP to check it, and it's 1.1 GB, so that will take a while. @Gameslinx, if you happen to know the right path, please feel free to push an update.

Gameslinx:
My mistake. The file path is <zip file>/Parallax_StockTextures
The Parallax_StockTextures must merge with the Parallax_StockTextures provided by the "Parallax_StockTextures" zip. Confusing, I apologize for that, but I had to overcome a github file size limit

HebaruSan:
Thanks, that makes sense. My browser says I have 3 minutes till it's done downloading, so I'll just wait for that to finish, double check it, and push a fix.

Gameslinx:
Thank you for the swift action. I hope all goes well :) 
Apologies for my lack of knowledge regarding github. Such is the life of someone who codes

HebaruSan:
> If stock textures are installed, these must also be installed.

- From this I am thinking that `Parallax-StockTextures` should depend on `Parallax-ScatterTextures`, but currently it's not updated. Shall I add that?
- `Parallax-ScatterTextures` and `Parallax-Scatter-Textures` are very similar, which could be confusing. Could we rename the concrete module `Parallax-StockScatterTextures` to indicate its connection to stock?

Gameslinx:
1. Yes that is correct, it should depend on the scatter textures as well
2. Yes, that makes more sense

HebaruSan:
Next problem is a file conflict between the texture modules:

image

We can filter that file out of one or the other module. Is there one of them that's a more natural "owner" of it?

HebaruSan:
I need to double check that, it's possible that both of the texture modules' `asset_match` patterns are matching the same file...

HebaruSan:
Now the bot can't read the main texture ZIP:

image

@Gameslinx, is that using LZMA? I think that's what we usually discover as the cause for that error. I've started downloading that file, estimated 17 minutes remaining till I can investigate it...

Gameslinx:
I have no idea! It's whatever Windows uses as its default zipping method

HebaruSan:
OK, I don't see any sign of LZMA, but I think Deflate64 can cause it as well (and the error message specifies one of these files):

image

We can try to add the Deflate64 method to CKAN, but that would only work if the SharpZipLib dependency to supports it, and it would require a new CKAN client release, which wouldn't help users who are using the current CKAN client. Let me see if I can find instructions for how to make it not use Deflate64...

Gameslinx:
I wonder if that's caused by deleting files within a Zip after compressing files. I had to do that in this case, but I also had to do it for the Scatters zip so I'm unsure why it's different here

HebaruSan:
I am finding explanations that Windows switches to Deflate64 if the ZIP file is over a certain size. They're also saying that Deflate64 is a proprietary format owned by PKWare, so it might not be possible to add it to CKAN after all. (Which makes the wisdom of Windows using it by default a bit ... questionable at best.)

The easiest solution appears to be to use something like 7zip or WinZip instead of "send to compressed folder".

Sorry for the inconvenience, I know you already invested a lot of time making and uploading this ZIP and it sucks to have to do it all over again.

Gameslinx:
Damn. I will start reuploading now using 7zip

Gameslinx:
How do I check if a zip is using deflate or deflate64?

HebaruSan:
> How do I check if a zip is using deflate or deflate64?

I opened it via right click → 7-Zip → Open archive, and there was a Method column (I dragged it further to the left to make it easier to screenshot).

Gameslinx:
Cheers. Reuploading a new zip using Deflate rather than Deflate64 now

Gameslinx:
Reuploaded, should work now!

HebaruSan (approved):
The test script succeeded that time. Looks good!

PR #9320, opened 2022-09-05
Create Parallax-ScatterTextures dependency

Gameslinx:
Added for Parallax 2.0.0 release, this file is required alongside Parallax_StockTextures. Explained more in patch-11


HebaruSan:
Thanks; merged this into #9319 so they can be tested together.

PR #9321, opened 2022-09-06
Make Parallax 2.0.0 modules conflict with <2.0.0 releases

DasSkelett:
The core Parallax >=2.0.0 only works with the StockTextures and Textures >=2.0.0, so it makes sense to enforce this.

RP-1 depends on Parallax and Parallax-StockTextures 1.3.1, however making the dependency versioned doesn't work, as the relationship resolver pulls in the newer module of the other one when looking at the dependencies of whichever is listed first.
Then it fails with "Parallax - Stock Planet Textures 2.0.0 required, but an version is in the resolver".
https://github.com/KSP-RO/RP-1-ExpressInstall-Graphics/blob/d60207f63f0ef254095e7d4d3e3342dc1abdbc53/RP-1-ExpressInstall-Graphics-High.netkan#L18-L21

Now we make the latest (2.0.0) and all future Parallax releases conflict with Parallax-StockTextures <=1.3.1,
Parallax-StockTextures with Parallax <=1.3.1, and Parallax-StockScatterTextures with Parallax <=1.3.1.

In local testing this did fix the install problem for RP-1, making the resolver correctly pull in the 1.3.1 releases of the two modules.

It might make sense to add similar conflicts to earlier Parallax releases, but this fixes the immediate problem.

### DIFF
--- a/NetKAN/Parallax-StockScatterTextures.netkan
+++ b/NetKAN/Parallax-StockScatterTextures.netkan
@@ -14,6 +14,8 @@ provides:
   - Parallax-Scatter-Textures
 conflicts:
   - name: Parallax-Scatter-Textures
+  - name: Parallax
+    version_max: '1.3.1'
 depends:
   - name: ModuleManager
   - name: Parallax
--- a/NetKAN/Parallax-StockTextures.netkan
+++ b/NetKAN/Parallax-StockTextures.netkan
@@ -14,6 +14,8 @@ provides:
   - Parallax-Textures
 conflicts:
   - name: Parallax-Textures
+  - name: Parallax
+    version_max: '1.3.1'
 depends:
   - name: ModuleManager
   - name: Parallax
--- a/NetKAN/Parallax.netkan
+++ b/NetKAN/Parallax.netkan
@@ -13,6 +13,9 @@ tags:
   - plugin
   - library
   - graphics
+conflicts:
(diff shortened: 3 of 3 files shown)


HebaruSan (approved):
OK, let's see how this works for the RP-1 folks :skateboard:

infoman:
But does ckan client even support version_max in conflicts? I don't see it being registered in registry.json and github search in this repo does not know about other records with version_max

HebaruSan:
You're right, it should be `max_version`. We got bit by that inconsistency again (`max` at start for mods, at end for game versions).

HebaruSan:
Ahh, this change made Parallax _2.0.0_ uninstallable because it now conflicts with its depends, which helped 1.3.1 as a side effect.

DasSkelett:
Whoops. I  also confirmed this "fixing" the issue multiple times. It somehow causes the latest version of the modules to be considered incompatible.
image

DasSkelett:
#9323 fixed 2.0.0, and installing 1.3.1 through RP-1 still works, but installing 1.3.1 on its own doesn't because Parallax still tries to pull in 2.0.0 of Parallax-StockTextures, which in turn conflicts with the selected Parallax 1.3.1.
So we also need to update all 1.x versions to conflict with 2.x at the very least :/

PR #9329, opened 2022-09-10
Vref for ArtemisConstructionKit

HebaruSan:
image

Presumably added in the latest release, since there are only two on SpaceDock.


HebaruSan:
Hmm, now I'm thinking that was from a bundled mod...

HebaruSan:
Yeah, it's from PlumeParty. Never mind.

PR #9332, opened 2022-09-13
Buffalo 2 requires Kerbal Actuators

Angel-125:
Moved KerbalActuators to the dependency list.


HebaruSan:
Thanks for the update!

PR #9333, opened 2022-09-14
Adds sample craft to Buffalo 2

Angel-125:
image


HebaruSan (approved):
Thanks for the update! :rocket:

PR #9355, opened 2022-09-28
Freeze UnKerballedStart

HebaruSan:
This mod has had an error on the status page since it released a version with a syntax error in the version file on 2022-01-22. theonegalen/UnKerballedStart#42 fixed the syntax error, but that fix has not been released. I tried gently suggesting/requesting a new release in a comment on that PR, but to no effect.

Now it's frozen; presumably this isn't important to the maintainer of the mod since not-indexing the latest version of this mod _for eight months_ hasn't been motivation enough to make a new release. But now we'll again go back to an error-free status page. :relieved:


theonegalen:
fixed version is uploaded on github and spacedock

HebaruSan:
Hi @theonegalen, thanks, that's great!

One note, there's an extra dot in the version number after the `v`:

image

Compare:

image

This can do weird things to the version numbers in CKAN down the line (it's where the `1:` prefixes come from). I thought I'd let you know in case you want to change it first.

theonegalen:
ah, thank you

PR #9366, opened 2022-10-04
Add Custom Categories

coldrifting:
Create and edit part categories in the VAB and SPH editors


HebaruSan (approved):
Nice work! No inflation errors or warnings, and it looks immediately mergeable!

PR #9367, opened 2022-10-05
Unfreeze UnKerballedStart

HebaruSan:
See:  https://github.com/KSP-CKAN/NetKAN/pull/9355#issuecomment-1267791750


HebaruSan:
Noted a version formatting number oddity in the original conversation, will wait a bit in case it gets fixed.

HebaruSan:
Alright, an epoch it shall be!

PR #9399, opened 2022-10-26
Add netkan for kOS-KerbalEngineer

markjfisher:
See https://github.com/markjfisher/kOS-KerbalEngineer for mod, adds very small interface to KerbalEngineer

___

ckan compat add 1.11


HebaruSan (approved):
Thanks for the submission!

PR #9402, opened 2022-10-29
Updates Pathfinder to recommend Buffalo 2 and not depend on Buffalo

Angel-125:
The original Buffalo has been supplanted by Buffalo2 (which is already in CKAN).
This update removes the Buffalo dependency and recommends Buffalo 2 in its place.


HebaruSan:
Thanks for the update.

(The validation script is acting up, but it's not needed here.)

PR #9415, opened 2022-11-03
Update DepthMask netkan file in preparation for next patch

drewcassidy:
I'm about to push a new version of DepthMask with a more standard distribution layout, so this is to update the NetKAN to match

- https://forum.kerbalspaceprogram.com/index.php?/topic/185353-111-112-moduledepthmask-portable-holes-for-your-parts/
- https://github.com/drewcassidy/KSP-DepthMask
- https://spacedock.info/mod/2943/ModuleDepthMask

### DIFF
--- a/NetKAN/DepthMask.netkan
+++ b/NetKAN/DepthMask.netkan
@@ -8,12 +8,6 @@ tags:
   - library
 suggests:
   - name: StockDepthMaskConfig
-install:
-  - find: GameData/DepthMask
-    install_to: GameData
-  - find: README.md
-    find_matches_files: true
-    install_to: GameData/DepthMask
 x_netkan_override:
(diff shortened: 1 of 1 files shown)


drewcassidy:
Its possible I misunderstood how the netkan works, its only "read" when a new update is pushed, right?

HebaruSan:
The netkan is read at updates for mods on SpaceDock, yes, but it is also read periodically in between updates in case anything has changed. So after we merge this pull request, it will be applied to whatever the current version is at that time. What we would usually do is wait until the new version is released, then merge.

drewcassidy:
Understood. will let you know

drewcassidy:
Update pushed

HebaruSan:
Thanks for the update!

PR #9427, opened 2022-11-10
Update kOS-Astrogator dep min version for Astrogator

HebaruSan:
See HebaruSan/Astrogator#32. Version v1.0.0 is now required.

___

ckan compat add 1.11 1.12

### DIFF
--- a/NetKAN/kOS-Astrogator.netkan
+++ b/NetKAN/kOS-Astrogator.netkan
@@ -15,7 +15,7 @@ tags:
   - plugin
 depends:
   - name: Astrogator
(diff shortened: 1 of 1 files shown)


markjfisher:
Thanks for this, completely slipped my mind it would be needed for NetKAN. I've done the version bump in the mod to Astrogator, but not here.

HebaruSan:
In practice, it probably won't matter (since CKAN prefers the latest version anyway), but as long as we know there's a specific version requirement, we might as well have it in the metadata.

PR #9432, opened 2022-11-12
Create BluedogDB Extras

Rodg88:
.netkans for optional patches included in the BDB main download, but not installed automatically

___

ckan compat add 1.11


HebaruSan:
For confirmation that the PR creator is the author of these patches:

https://github.com/CobaltWolf/Bluedog-Design-Bureau/commits/master/BD_Extras%20(No%20Warranty)/GameData/Bluedog_DB_Extras/BDB_CCK

HebaruSan (approved):
Thanks for the submission!

(I checked with the other BDB folks on Discord, and nobody demanded another module or had a suggestion for making sure users who shouldn't install this won't.)

PR #9433, opened 2022-11-12
Create BDB-Methalox.netkan

Rodg88:
.netkan for an optional patch included in the BDB main download, but not installed automatically


HebaruSan:
I'm going to combine these three pull requests so it's easier to confirm that our edits are consistent...

HebaruSan:
Merged into #9432.

PR #9434, opened 2022-11-12
Create BDB-RealNames.netkan

Rodg88:
.netkan for an optional patch included in the BDB main download, but not installed automatically


HebaruSan:
Merged into #9432.

PR #9435, opened 2022-11-12
Filter Scale_Redist.dll out of TweakScale

Lisias:
I have an external tool that installs 999_Scale_Redist.dll on the GameData for people that do manual installation and forget to install the Redist.

Such tool **is not** available on CKAN installs for obvious reasons, so the presence of the "reference dll" is not a problem (it's inside `PluginData`, a folder KSP doesn't touch).

However ZeroMiniAVC recently started to inspect 3rd parties folders for duplicates, and bluntly ignores that anything inside `PluginData` is not loaded by KSP and then started to pesky users about the presence of the "reference DLL".

The goal of this pull request is to prevent this file to be installed by CKAN, avoiding spurious support calls from people complaining about what's a misfeature on the ZeroMiniAVC tool and not a real problem.

### DIFF
--- a/NetKAN/TweakScale.netkan
+++ b/NetKAN/TweakScale.netkan
@@ -23,6 +23,7 @@ tags:
 install:
   - file: GameData/TweakScale
     install_to: GameData
+    filter: Scale_Redist.dll
 depends:
   - name: TweakScale-Redist
   - name: ModuleManager


HebaruSan:
Submitted linuxgurugamer/QuickMods#18 to hopefully do away with that warning on current game versions. It was sketchy to put it in ZeroMiniAVC in the first place since it has nothing to do with MiniAVC, and now it's just being disruptive to no purpose.

HebaruSan (approved):
Thanks for the update and the explanation. :+1:

PR #9466, opened 2022-12-10
Create Wacapella.netkan

harveyt:
This is a mini-mod that balances and combines CNAR and Taerobee using ModuleManager only. Provides parts to build a WAC Corporal sounding rocket.

Tested using netkan.exe and installed on stock 1.12.4 using ckan (as per wiki).

- <https://github.com/harveyt/Wacapella>
- <https://forum.kerbalspaceprogram.com/index.php?/topic/210985-112x-wacapella-early-sounding-rockets-100-9th-dec-2022/>

___

ckan compat add 1.8

### DIFF
--- a/NetKAN/Wacapella.netkan
+++ b/NetKAN/Wacapella.netkan
@@ -1,0 +1,23 @@
+spec_version: v1.4
+identifier: Wacapella
+name: Wacapella
+abstract: Early sounding rockets
+author: 610yesnolovely (Harvey Thompson)
+license: MIT
+resources:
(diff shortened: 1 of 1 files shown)


harveyt:
I know what this is, CNAR is 1.8.1 only and I was testing with "compatible versions" enabled.
CNAR is required, so I guess either:
- Mark this mod as min version 1.8
- OR can one mark CNAR as a specific version (or greater)?

HebaruSan (approved):
Thanks for the submission!

PR #9468, opened 2022-12-12
Update and rename KeridianDynamicsVesselAssembly.frozen to KeridianDy…

zer0Kerbal:
…namics.netkan


HebaruSan:
Please don't change the identifiers of .frozen modules, those are archival. Just add a new one with the new identifier.

linuxgurugamer:
Curiosity question, is there a problem unfreezing a module?

PR #9493, opened 2022-12-25
Updated kref

linuxgurugamer:


HebaruSan:
I think this was done in #9494.

PR #9497, opened 2022-12-25
Update SRBWaterfallEffects.netkan

AdiriKSP:


HebaruSan:
SpaceDock still says:

image

https://spacedock.info/mod/2992/SRB%20Waterfall%20Effects

The forum thread says it "will NOT work without these", before saying it's not a hard requirement, which sounds like a contradiction:

image

https://forum.kerbalspaceprogram.com/index.php?/topic/207196-112x-srb-waterfall-effects/

PR #9503, opened 2023-01-03
FreeIva: add kpbs and kerbalism as supported mods

JonnyOThan:


NetKAN/FreeIva.netkan
```diff
@@ -23,3 +23,5 @@ supports:
   - name: PlanetsideExplorationTechnologies
   - name: StationPartsExpansionRedux-IVAs
   - name: Buffalo2
+  - name: KerbalPlanetaryBaseSystems
+  - name: Kerbalism-Config-Default
```
HebaruSan, line 27: Cross posting from Discord:

> Hmm, did you mean to reference Kerbalism-Config-Default instead of just Kerbalism? And if so, could you please explain why in the pull request description? That looks very weird.

JonnyOThan: Yes, the support is for the greenhouse and gravityring parts that are in the default config, not the core mod.

HebaruSan: OK cool, thanks for the response.

PR #9512, opened 2023-01-11
Update AdjustableModPanel.netkan

zer0Kerbal:
apologies - still learning :octocat:


HebaruSan:
Looks like there's no change in the status of the conflict:

https://forum.kerbalspaceprogram.com/index.php?/topic/147463-112x-end-your-parts-list-nightmare-introducing-the-janitors-closet/

image

@linuxgurugamer, if that's out of date, ccould you please update the OP and let us know? Then we could re-open this and merge it.

zer0Kerbal:
LGG says that JC is incompatible with KAMP then put that in the JC listing not in KAMP. Currently it isn't in JC and I don't want it in KAMP's.

PR #9514, opened 2023-01-11
KSA should depend on RasterPropMonitor-Core, not the full mod

JonnyOThan:


HebaruSan:
Thanks for the update.
(Don't worry about the validation error, that just started happening on all pull requests, looks like some kind of problem with GitHub.)

PR #9515, opened 2023-01-12
Adjustable mod panel

zer0Kerbal:
there you go - remove from the listing that doesn't need/want and put onto the listing that does want it.


HebaruSan:
Let's let this sleeping dog lie. Everything's working fine as-is, but if we merge this, we'll have people re-installing mods for no reason, potentially encountering difficulties with that, etc. There's no need to spend time on this.

zer0Kerbal:
I don't agree.

HebaruSan:
Yes, we already know that. Cheers and good luck to you!

PR #9530, opened 2023-01-20
FreeIVA Updates

JonnyOThan:
Move reviva and b9ps to recommendations, because they're rellevant for stock parts

add newly supported mods


HebaruSan (approved):
Thanks for the update.

PR #9532, opened 2023-01-20
Added a conflicts statement

linuxgurugamer:
This mod should not be installed when Restock is installed


HebaruSan:
Looks like we shouldn't do this: #8576

HebaruSan:
FYI @linuxgurugamer, to view the edit history of any mod, you can find it on the status page:

- http://status.ksp-ckan.space/

... hover it, then click "history":

image

Which in this case leads to the PR linked above.

linuxgurugamer:
Weird.  Well, then I found a problem in another mod, but thats in the mod.
Ok, thanks

PR #9536, opened 2023-01-22
Adding TweskScale Companion to CKAN

Lisias:
I'm kick starting a new TweakScale's related package, TweakScale Companion, with all the Companions available to make life easier to everybody (including me).

All sub-packages are known to behave nicely when the target add'on is not installed, making this an ideal solution for lazy installers that were bugging me for years about how TweakScale was easier to install in the old, buggier times.

However, I had to declare two add'ons as conflict due the reasons specified below:

* `TweakScaleCompanion-ReStockPlus`
    * It's already included on the package.
    * I don't want to supersede it because not everyone that uses Restock is willing to cope with a "everything and the kitchen's sink" distribution model. So people that wants to use only Restock doesn't have to shove all that files in their GameData just for it.
* `AirplanePlusThings`
    * There're already AirplanePlus patches on this package, and they are incompatible with `AirplanePlusThings`
        * Having both of them installed is theoretically possible, but I would like to avoid the burden to support this weird setup. And would be the point?

Every single Companion is available separately on GitHub under their original licenses (compatible with the license used on this one). The [README.md](https://github.com/net-lisias-ksp/TweakScaleCompanion/blob/master/README.md) on the package lists them.

There's , however, a problem that I'm unsure how to solve.

There's a file on an old add'on called `LShipPartsRequired`, part of [Large Boat Parts Pack](https://spacedock.info/mod/167/Large%20Boat%20Parts%20Pack) that have absolutely terrible patches in one file, that thing would patch indiscriminately so many 3rd party parts that I could't managed to counter-patch the mess, so the only viable solution was to rewrite that config file myself and replace the original when it's present.

The name of the file is `GameData/LShipPartsRequired/LBP_tweakableComplete.cfg` on the my zip package, and it needs to replace the original file on the existing KSP installation.

What would be the best way to cope with that on CKAN? Right now, this netkan does nothing about, and if the LBPP is installed a huge TweakScale FATALITIES storm will fall on the user's head - prompting them to ask for help on Forum where I will tell the user to replace manually the file.

### DIFF
--- a/NetKAN/TweakScaleCompanion.netkan
+++ b/NetKAN/TweakScaleCompanion.netkan
@@ -1,0 +1,31 @@
+spec_version: v1.8
+identifier: TweakScaleCompanion
+name: TweakScale Companion
+abstract: >-
+  **TweakScale Companion Program** is a series of "Add'On's Add'Ons",
+  adding TweakScale /L support for third parties but decoupling the
+  target Add'On's life cycle from TweakScale's, easing maintenance
+  and distribution efforts. This package aims to make TweakScale
+  user's life easier at first installs with a (most of the time)
+  updated collection with all the currently available Companions
+  at once.
+author:
+  - LisiasT
+$kref: '#/ckan/spacedock/3202'
+$vref: '#/ckan/ksp-avc'
+x_netkan_force_v: true
+license:
+  - restricted
+release_status: testing
+tags:
+  - plugin
+  - convenience
+install:
+  - file: GameData/TweakScaleCompanion
+    install_to: GameData
+depends:
+  - name: TweakScale
+  - name: ModuleManager
+conflicts:
+  - name: TweakScaleCompanion-ReStockPlus
+  - name: AirplanePlusThings


NetKAN/TweakScaleCompanion.netkan
```diff
@@ -0,0 +1,31 @@
...
+  adding TweakScale /L support for third parties but decoupling the
+  target Add'On's life cycle from TweakScale's, easing maintenance
+  and distribution efforts. This package aims to make TweakScale
+  user's life easier at first installs with a (most of the time)
+  updated collection with all the currently available Companions
+  at once.
+author:
+  - LisiasT
+$kref: '#/ckan/spacedock/3202'
+$vref: '#/ckan/ksp-avc'
```
HebaruSan, line 15:
```suggestion
$vref: '#/ckan/ksp-avc/TweakScaleCompanion.version'
```

@Lisias, the many many version files in the ZIP are causing inflation errors. You can use this syntax to disambiguate.

Lisias:
@HebaruSan , there's something else in need to be fixed?

Thanks.

NetKAN/TweakScaleCompanion.netkan
```diff
@@ -0,0 +1,31 @@
...
+identifier: TweakScaleCompanion
+name: TweakScale Companion
+abstract: >-
+  **TweakScale Companion Program** is a series of "Add'On's Add'Ons",
+  adding TweakScale /L support for third parties but decoupling the
+  target Add'On's life cycle from TweakScale's, easing maintenance
+  and distribution efforts. This package aims to make TweakScale
+  user's life easier at first installs with a (most of the time)
+  updated collection with all the currently available Companions
+  at once.
```
HebaruSan, line 11: The `abstract` should be a **short** description, a brief sentence fragment to help the user understand what the mod is. This is much longer than it really should be.

NetKAN/TweakScaleCompanion.netkan
```diff
@@ -0,0 +1,31 @@
...
+author:
+  - LisiasT
```
HebaruSan, line 13:
```suggestion
```

You don't need this, as it will be retrieved from SpaceDock automatically.

NetKAN/TweakScaleCompanion.netkan
```diff
@@ -0,0 +1,31 @@
...
+$kref: '#/ckan/spacedock/3202'
+$vref: '#/ckan/ksp-avc/TweakScaleCompanion.version'
+x_netkan_force_v: true
+license:
+  - restricted
```
HebaruSan, line 18:
```suggestion
license: restricted
```

No need to make this a list when it's just one item

NetKAN/TweakScaleCompanion.netkan
```diff
@@ -0,0 +1,31 @@
...
+release_status: testing
+tags:
+  - plugin
+  - convenience
+install:
+  - file: GameData/TweakScaleCompanion
+    install_to: GameData
```
HebaruSan, line 25:
```suggestion
```

You don't need this; the default install stanza when it's left out is to `find` the identifier and install it to GameData, which is equivalent to this.

NetKAN/TweakScaleCompanion.netkan
```diff
@@ -0,0 +1,31 @@
...
+  target Add'On's life cycle from TweakScale's, easing maintenance
+  and distribution efforts. This package aims to make TweakScale
+  user's life easier at first installs with a (most of the time)
+  updated collection with all the currently available Companions
+  at once.
+author:
+  - LisiasT
+$kref: '#/ckan/spacedock/3202'
+$vref: '#/ckan/ksp-avc/TweakScaleCompanion.version'
+x_netkan_force_v: true
```
HebaruSan, line 16:
```suggestion
```

You don't need this; it's for when a mod is adopted from a modder who adds a `v` prefix to release nubmers a modder who doesn't, which doesn't apply for new mods.

NetKAN/TweakScaleCompanion.netkan
```diff
@@ -0,0 +1,31 @@
+spec_version: v1.8
+identifier: TweakScaleCompanion
+name: TweakScale Companion
```
HebaruSan, line 3:
```suggestion
```

The `name` will be retrieved from SpaceDock automatically

NetKAN/TweakScaleCompanion.netkan
```diff
@@ -0,0 +1,31 @@
...
+  updated collection with all the currently available Companions
+  at once.
+author:
+  - LisiasT
+$kref: '#/ckan/spacedock/3202'
+$vref: '#/ckan/ksp-avc/TweakScaleCompanion.version'
+x_netkan_force_v: true
+license:
+  - restricted
+release_status: testing
```
HebaruSan, line 19:
```suggestion
release_status: testing
```

This property isn't actually used to drive any functionality, simpler to leave it out

HebaruSan (changes requested):
Added some comments based on review

HebaruSan:
> There's a file on an old add'on called `LShipPartsRequired`, part of [Large Boat Parts Pack](https://spacedock.info/mod/167/Large%20Boat%20Parts%20Pack) that have absolutely terrible patches in one file, that thing would patch indiscriminately so many 3rd party parts that I could't managed to counter-patch the mess, so the only viable solution was to rewrite that config file myself and replace the original when it's present.
> 
> The name of the file is `GameData/LShipPartsRequired/LBP_tweakableComplete.cfg` on the my zip package, and it needs to replace the original file on the existing KSP installation.
> 
> What would be the best way to cope with that on CKAN? Right now, this netkan does nothing about, and if the LBPP is installed a huge TweakScale FATALITIES storm will fall on the user's head - prompting them to ask for help on Forum where I will tell the user to replace manually the file.

Sorry, I didn't notice this when I skimmed the description earlier. Overwriting files from other mods isn't allowed (imagine how annoying and messy it would be if LBPP had done that to _your_ mod). I can think of a few ways to address this:

1. Submit a fix to the LBPP maintainer and get them to release a new version without these problems (ideal solution)
2. If the maintainer is no longer active, a new maintainer could adopt it and do the updates
3. Mark a `conflict` between LBPP and this or TweakScale, if there's no prospect for fixing the underlying issue and you're confident the problem is too severe to ignore
4. Continue as you have it, with an error message and users manually copying files around. Not a good user experience for anyone, but this mod is for KSP 1.4.4, so it's probably not all that common of a problem either.

NetKAN/TweakScaleCompanion.netkan
Lisias: Using default CKAN behaviours instead of specifying it

NetKAN/TweakScaleCompanion.netkan
Lisias: Removing data that could be retrieved from SpaceDock automatically

NetKAN/TweakScaleCompanion.netkan
Lisias: Solved on commit https://github.com/KSP-CKAN/NetKAN/pull/9536/commits/21b0d112e86019e628d56c9abb8aa98b15316372

Lisias: Amended on commit https://github.com/KSP-CKAN/NetKAN/pull/9536/commits/3b8a6c0e0bd55088dbbff7be6a7591f117382feb

Lisias:
> 4. Continue as you have it, with an error message and users manually copying files around. Not a good user experience for anyone, but this mod is for KSP 1.4.4, so it's probably not all that common of a problem either.

The author ir retired, and that file intents exactly to fix thing in order to allow it to be used as is.

I'm taking the 4th option. It's the easiest way out for while.

Lisias:
Hi, @HebaruSan !

There's something else in need to be fixed before merging?

Thanks!

Lisias:
Hi, @HebaruSan !

There's something else in need to be fixed before merging?

Thanks!

Lisias:
Hi, @HebaruSan !

There's something else in need to be fixed before merging?

Thanks!

PR #9544, opened 2023-01-31
Declares `TweakScaleMakingHistoryConfigs` as a conflict.

Lisias:
This thing is old, the patching format is deprecated and I had revamped the patches on TweakScale. There's no way this thing would be usefull nowadays, and I'm starting to get tickets about it again.

Since the very source of the add'on (SpaceDock) is not providing it anymore, I don't see another way to solve this problem but to declare a conflict.

TweakScale works fine on previous versions of KSP, to in essence, TweakscaleMakingHistoryConfigs is really uneeded.


HebaruSan (approved):
Thanks for the update.

Issue #9456, opened 2022-12-04
[Mod] Habtech2


oofergoofer:
Mod info
--------
* Mod name: Habtech2
* Author(s): benjee10
* License: ARR
* Homepage: https://forum.kerbalspaceprogram.com/index.php?/topic/133501-112x-habtech2-stockalike-iss-parts-026/
* <https://github.com/benjee10/HabTech2>
* <https://spacedock.info/mod/2078/HabTech2>

HebaruSan:
Sent Discord PM to author...

HebaruSan:
No reply so far. @oofergoofer, if you're able to get in touch with Benjee10, please let us know and we can re-open this.

PR #9582, opened 2023-02-23
Add missing HabTech2 dependencies

HebaruSan:
#9456 indexed this mod based on the dependencies listed in the SpaceDock description, which were out of date.

- <https://spacedock.info/mod/2078/HabTech2>
- <https://forum.kerbalspaceprogram.com/index.php?/topic/133501-112x-bg-habtech2-stockalike-iss-parts-100-the-final-update/>

Now they're updated to reflect the forum thread (and the newly updated SpaceDock description).

As part of this, `HabTechRobotics` is added as well since it is a bundled dependency.

- <https://github.com/benjee10/htRobotics>

___

ckan compat add 1.12

### DIFF
--- a/NetKAN/HabTech2.netkan
+++ b/NetKAN/HabTech2.netkan
@@ -11,8 +11,11 @@ depends:
   - name: ModuleManager
   - name: B9PartSwitch
   - name: CommunityResourcePack
+  - name: Benjee10-SharedAssets
+  - name: HabTechProps
+  - name: HabTechRobotics
   - name: Waterfall
-  - name: KIS
+  - name: BreakingGround-DLC
 recommends:
   - name: NewTantares
   - name: BluedogDB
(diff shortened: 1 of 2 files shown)


HebaruSan:
Regarding author permission for HabTechRobotics, the download ZIP contains a .netkan file, which I took as an expression of author intention to add the mod to CKAN.

Clayell:
@HebaruSan I believe HabTechRobotics also requires Breaking Ground, as is listed on its [readme](https://github.com/benjee10/htRobotics):

> DEPENDENCIES
- B9PartSwitch
- KSP Breaking Ground DLC

Clayell:
@HebaruSan Additionally, I believe that HabTech2 doesn't require Breaking Ground itself? Only the robotics part of it does, which I believed is covered under HT Robotics.
Of course, HabTech2 does require HT Robotics anyway, so maybe this is redundant.
image

PR #9590, opened 2023-02-28
Update StarshipExpansionProject Metadata

Kari1407:
(author)

Reflecting the mod's most recent update, I would like to update the netkan to reflect the new state of the mod

note: on "recommends" section there is a new netkan specifically for the IVA portion of the mod, which will be PR shortly after this one

- <https://spacedock.info/mod/2959/Starship%20Expansion%20Project%20-%20SEP>
- <https://forum.kerbalspaceprogram.com/index.php?/topic/206555-1101-112x-starship-expansion-project-sep-v101-january-30th-2022/>
- <https://github.com/Kari1407/Starship-Expansion-Project>

PR #9591, opened 2023-02-28
Add Starship Expansion Project IVA

Kari1407:
PR mentioned on #9590


HebaruSan:
Thanks, added this to #9590.

Kari1407:
Hi, no this is not active anymore and I don't have any plans to split the IVA from the main mod in the future when I make new ones. This is effectively dead

PR #9603, opened 2023-03-05
Update FreeIva.netkan

JonnyOThan:
add more supported mods


HebaruSan (approved):
Thanks for the update!

PR #9610, opened 2023-03-09
Added  ksp_version_min and ksp_version_max

sswelm:
I added this to allow CKAN to mark mod this as a compatible with the 1.12.x release


HebaruSan:
Wait, the 1.12.2 is coming from SpaceDock.

image

We can just click "Yes, update automatically", no need to mess with the .netkan file.

image

https://spacedock.info/mod/172/KSP%20Interstellar%20Extended

HebaruSan:
OK, I clicked it and it's updated now.

https://github.com/KSP-CKAN/CKAN-meta/commit/b675e717b5880877bccabb28a299b5a5cbb64270

PR #9631, opened 2023-03-22
ArtemisConstructionKit.netkan - Add PlumeParty as a dependency

Poodmund:
When this Pull Request, to remove PlumeParty out from being bundled within the mod's Benjee10_Orion folder, gets merged and a new distribution is released, this metadata will require updating to require PlumeParty (just the core package, not the Stock configs) as a dependency.

https://github.com/benjee10/Benjee10_Orion/pull/14

___

ckan compat add 1.12


Poodmund:
Ah yes, that would also need doing. 😅

PR #9641, opened 2023-03-30
Update InfernalRoboticsNext.netkan

meirumeiru:
TweakScale or TweakScaleRedist is no longer used in the mod and has been removed completly

there is no incompatibility known with KJR-continued (anymore)

### DIFF
--- a/NetKAN/InfernalRoboticsNext.netkan
+++ b/NetKAN/InfernalRoboticsNext.netkan
@@ -17,20 +17,17 @@
         "plugin"
     ],
     "conflicts"      : [
-        { "name": "InfernalRobotics"         },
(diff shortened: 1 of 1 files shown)


HebaruSan (approved):
Thanks for the update!

FYI, searched the ZIP for "tweak", and there's still one file left that might be a holdover from when TweakScale was bundled:

image

Obviously that makes no functional difference for us here, but I thought I should mention it in case it has been left behind. Cheers!

meirumeiru:
> FYI, searched the ZIP for "tweak", and there's still one file left that might be a holdover from when TweakScale was bundled:

thanks... I will delete all the remainings

PR #9691, opened 2023-05-23
Update InfernalRoboticsNext.netkan

meirumeiru:
new dependency (starting from version 3.1.12) ... I removed some classes and put them into an own mod (for better reusability)


HebaruSan (approved):
Thanks for the update.

PR #9692, opened 2023-05-27
Update RealFuels.netkan to target new home of repo

NathanKell:
Was moved under KSP-RO org


HebaruSan (approved):
Thanks for the update 👍

PR #9697, opened 2023-06-03
Create RP-1.netkan

NathanKell:
This will (a) allow us to maintain a legacy version of RP-1 (using the old identifier) for folks finishing old saves, and (b) unify the identifier with the mod name (and now the repo name).

### DIFF
--- a/NetKAN/RP-1.netkan
+++ b/NetKAN/RP-1.netkan
@@ -1,0 +1,13 @@
+{
+    "spec_version": "v1.4",
+    "identifier": "RP-1",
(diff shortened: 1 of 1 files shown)


NathanKell:
On the off chance discord is wonky, tagging here too: @HebaruSan @DasSkelett 
Thanks guys!

NathanKell:
(Uh, to be clear, I ping because the update is out, and it'd be great if it were on CKAN, since we depend so heavily on CKAN via the express install)

PR #9718, opened 2023-07-02
Update FreeIva supported mods

JonnyOThan:


HebaruSan (approved):
Thanks for the update! 👍

Issue #9728, opened 2023-07-14
[Bug]: Errors unzipping Extraplanetary Launchpads


diabetocles:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Operating System

Windows 10

### CKAN Version

Latest 1.32.0

### Game Version

1.12.5.3190

### Did you make any manual changes to your game folder (i.e., not via CKAN)?

Other than manually installed mods no. Also removed manually installed mods and tested with same results.

### Describe the bug

Upon ckan installing EL it has an error unzipping see below.

**Error message**
CKAN.InvalidModuleFileKraken: ExtraPlanetaryLaunchpads 6.99.3.0: C:\Users\redacted\AppData\Local\CKAN\downloads\downloading\5E0A4EB7-ExtraPlanetaryLaunchpads-6.99.3.0.zip is not a valid ZIP file: Error in step EntryData for ExtraplanetaryLaunchpads/Parts/LandingPad/LandingPad.mu: Exception during test - 'Unknown block type 6'
   at CKAN.NetModuleCache.Store(CkanModule module, String path, IProgress`1 progress, String description, Boolean move, CancellationToken cancelToken)
   at CKAN.NetAsyncModulesDownloader.ModuleDownloadComplete(Uri url, String filename, Exception error, String etag)

**CKAN output**
About to install:

 * Extraplanetary Launchpads 6.99.3.0 (taniwha.org, 16.1 MiB)
Downloading "http://taniwha.org/~bill/ExtraplanetaryLaunchpads_v6.99.3.zip"
Finished downloading ExtraPlanetaryLaunchpads 6.99.3.0, validating and storing to cache...
Trying to install ExtraPlanetaryLaunchpads 6.99.3.0, but it's not downloaded or download is corrupted
Error during installation!
If the above message indicates a download error, please try again. Otherwise, please open an issue for us to investigate.
If you suspect a metadata problem: https://github.com/KSP-CKAN/NetKAN/issues/new/choose
If you suspect a bug in the client: https://github.com/KSP-CKAN/CKAN/issues/new/choose

### Steps to reproduce

Install Extraplanetary launchpad from ckan interface.

### Relevant log output

```shell
**Error message**
[...]

Issue #9578, opened 2023-02-22
Download for EPL is corrupted


quasinaut:
Downloaded via CKAN (http://taniwha.org/~bill/ExtraplanetaryLaunchpads_v6.99.3.zip):
![image](https://user-images.githubusercontent.com/126029878/220490389-a5c63209-9c97-429e-88a8-46f88f01cfb3.png)

Downloaded via web browser (http://taniwha.org/~bill/ExtraplanetaryLaunchpads_v6.99.3.zip):
![image](https://user-images.githubusercontent.com/126029878/220490247-96061978-9513-4b68-b0f3-7f6b0200f74d.png)

Quick fix:
Change extra planetary launchpads to use github (https://github.com/taniwha/Extraplanetary-Launchpads/releases/tag/v6.99.3), where I have no issues with corruption upon unzipping

Feature fix:
Change CKAN to automatically skip files corrupted files

HebaruSan:
This should be reported to the mod author, whose server has been distributing corrupted ZIPs for a while now (maybe a failing disk?):

- <https://github.com/taniwha/Extraplanetary-Launchpads/issues>

The GitHub releases are not usable because the source archive does not contain the plugin. You can unzip them, but the mod won't work.

quasinaut:
Thanks, I did not realize the "release" contained no plugin. To be fair, CKAN could skip corrupted files like windows does but this is a distribution issue

HebaruSan:
Skipping corrupted files would be installing the mod incorrectly. Mod authors tend not to like it if CKAN installs their mods incorrectly, so we go to great lengths to avoid that.

quasinaut:
That makes sense but does that mean via direct download would be incorrect then? It seems like there is no way to get the mod now. Anyways, thanks

HebaruSan:
I don't know what you mean by "via direct download would be incorrect", so I can't answer that.

Yes, it's quite problematic that this author's mods are unreliably available. We're not able to solve that, though.

HebaruSan:
The server problems have been reported to the mod author here:

- taniwha/Extraplanetary-Launchpads#202
- taniwha/Extraplanetary-Launchpads#205

No response so far. If @taniwha isn't active anymore, maybe we should freeze these mods and redirect their downloads to archive.org.

HebaruSan:
Bad sector on firewall noted about 8 months ago:

[...]

PR #9734, opened 2023-07-18
Freeze mods hosted on broken taniwha.org server

HebaruSan:
## Problem

Users increasingly frequently report problems installing Extraplanetary Launchpads, see:

- KSP-CKAN/NetKAN#9578
- KSP-CKAN/CKAN#3855
- KSP-CKAN/NetKAN#9728
- KSP-CKAN/NetKAN#9733

(Side note, it's clear that no one has been taking the "I have searched the existing issues" checkbox seriously. Maybe we should remove it.)

## Cause

taniwha.org sometimes (often?) distributes corrupted ZIP files, which can be confirmed independently of CKAN:

image

This broken server is the only place these mods are hosted by the author. It sounds like the host's disk is failing:

<https://forum.kerbalspaceprogram.com/topic/54284-112-extraplanetary-launchpads-v6993/?do=findComment&comment=4201640>

forum post about bad sector on firewall

CKAN is doing exactly what it's supposed to in this case: detecting the corruption, reporting it, and refusing to install.

## Previous attempts

The ideal solution would have been:

- Author fixes the problems with the host; OR
- Author switches to hosting these mods on a more reliable host (like SpaceDock or GitHub)

Users have raised the issue with the author:

- taniwha/Extraplanetary-Launchpads#202
- taniwha/Extraplanetary-Launchpads#205

So far there has been no response, and the problems with the server continue.

## Changes

Now these mods are frozen, which means the bot will no longer check them for new releases.

For mods with redistributable licenses, a subsequent pull request in CKAN-meta will replace the `download` fields in the indexed versions with links to the Internet Archive, uploaded when the bot was intermittently able to access a valid file (without freezing, the bot would overwrite these with the current broken host). This will eventually allow users to install these mods without getting errors about the corrupted downloads.

If the author becomes active again, hopefully the hosting issues will be resolved before further releases are created.


HebaruSan:
Great, archive.org is timing out now. This is going to be fun to support. Maybe we should just remove these mods from CKAN completely.

PR #9752, opened 2023-08-06
add kOS-Career.netkan

JonnyOThan:


HebaruSan:
Thanks! I got the notification for the new release and re-opened #8151 already.

PR #9758, opened 2023-08-12
Add EmbeddedKOSKirkuits

rkunze:
https://github.com/rkunze/EmbeddedKOSKirkuits
https://forum.kerbalspaceprogram.com/topic/218872-112x-embedded-kos-kirkuits/

### DIFF
--- a/EmbeddedKOSKirkuits.netkan
+++ b/EmbeddedKOSKirkuits.netkan
@@ -1,0 +1,26 @@
+spec_version: v1.27
+identifier: EmbeddedKOSKirkuits
+name: Embedded kOS Kirkuits
+abstract: Adds kOS processors to selected probe cores and command modules
+resources:
+  homepage: https://forum.kerbalspaceprogram.com/topic/218872-*
+$kref: '#/ckan/github/rkunze/EmbeddedKOSKirkuits'
+$vref: '#/ckan/ksp-avc'
+release_status: stable
+license: CC-BY-NC-4.0
+tags:
+  - config
+  - control
+install:
+  - find_regexp: EmbeddedKOSKirkuits-.*
+    install_to: GameData
+    as: EmbeddedKOSKirkuits
+ksp_version: any
+depends:
+  - name: ModuleManager
+  - name: kOS
+recommends:
+  - name: SkyhawkScienceSystem
+x_netkan_github:
+    use_source_archive: true
+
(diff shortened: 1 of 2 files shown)


rkunze:
Oops. Sorry for putting the file at the top level, did not intent to do that. Thanks for fixing.

As for the indentation warning: Is that documented somewhere that you want 4 spaces as indent? I usually use two spaces for YAML to avoid too much blank space at the start of the line (especially for deeply nested structures). I'll try to keep it in mind for the next time.

HebaruSan (approved):
Thanks for the submission!

There aren't specific formatting requirements, we just run everything through yamllint to keep it relatively tidy. It was complaining because there were 4 spaces before `use_source_archive`.

rkunze:
@HebaruSan Thanks for merging!

I found a little oddity in "ckan list" behaviour on my dev instance for EmbeddedKOSKirkuits: Now that the mod is officially on CKAN, I'd expect "ckan list" on my dev instance (where it is installed manually) to show it as an "unmanaged" mod. But it does not show up at all there, even after an explicit "ckan update" and "ckan scan".

My guess from the verbose output of "ckan scan": "ckan scan" only registers manually installed mods that contain a DLL, and EmbeddedKOSKirkuits is purely a set of ModuleManager patches and does _not_ contain a DLL. Supporting evidence: Same behavior with manually installed versions of SkyhawkScienceSystem and SkyhawkKerbalism, which are pure ModuleManager patch sets as well. 

Question: Is this known (and intended) behavior, or should I report a bug against the CKAN client?

rkunze:
Spontaneous (and probably dumb) idea: How about checking the directory listing of `GameData` against known mod IDs? Should be fast enough even for a monster install with several hundred mods, and would at least catch all well-behaved mods that install to `GameData/{mod-id}`...

HebaruSan:
I think there were reasons we didn't go with that solution, but it's been a while now since I've concentrated on it, so I don't have a comprehensive explanation ready to go. At a glance, it seems like a big leap to go from "this directory exists (and might have a file in it)" to "this mod is fully and correctly installed such that mods that depend on it will work properly."

HebaruSan:
If you just want a quick workaround for your dev environment, note that the DLL file doesn't have to be valid. If make a 0-byte file at `GameData/YourMod/PluginData/YourMod.dll`, I think CKAN would detect it and the game wouldn't even try to load it thanks to how it treats `PluginData` folders.

PR #9761, opened 2023-08-15
remove "supports: freeiva" from freeiva mk3 passthrough because it's …

JonnyOThan:
…deprecated

CKAN presents this mod to be installed when you try to install freeiva, but it's deprecated and shouldn't be installed by anyone

- <https://spacedock.info/mod/3193/[DEPRECIATED]%20FreeIVA%20Mk3/Mk2%20Passthrough>


HebaruSan:
Should we just remove it?

HebaruSan:
Let's just delete it, since it says it's deprecated and is basically an empty download...

PR #9762, opened 2023-08-18
Remove unneeded dependencies for ProbeControlRoomRecontrolled

Jules-Bertholet:
These are no longer necessary, per https://github.com/JonnyOThan/KSP-ProbeControlRoom/

### DIFF
--- a/NetKAN/ProbeControlRoomRecontrolled.netkan
+++ b/NetKAN/ProbeControlRoomRecontrolled.netkan
@@ -14,9 +14,7 @@
     } ],
     "depends": [
         { "name": "ModuleManager" },
-        { "name": "RasterPropMonitor-Core" },
-        { "name": "ToolbarController" },
-        { "name": "ClickThroughBlocker" }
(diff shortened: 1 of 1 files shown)


HebaruSan (approved):
- [x] Confirmed that the version we get from SpaceDock matches up with that repo (@JonnyOThan was added as a co-author)
- [x] Confirmed that the ToolbarController and ClickThroughBlocker dependencies are not mentioned

Thanks for another one!

JonnyOThan:
Uh woops, toolbarcontroller is actually a dependency on this one.

I'd appreciate a chance to weigh on on metadata changes to my mods.

HebaruSan:
These are the places I checked:

Forum thread:

image

SpaceDock:

image

GitHub:

image

The ZIP file:

image

I'm sorry if this was an incorrect change, but you gotta give non CKAN users a clue somewhere!

JonnyOThan:
Yeah, just give me a heads up - it's clear that either the metadata or the forum post is wrong.  In this case it was the forum post.

PR #9765, opened 2023-08-22
VesselView depends on VesselView-UI-Toolbar

JonnyOThan:
The intention was that you could use either VV-RPM or VV-Toolbar, but until that gets fixed CKAN should probably just install all 3.

@linuxgurugamer for visibility


JonnyOThan:
hmmm I can't repro the issues that I thought I saw before, so closing this until I have something more solid

PR #1165, opened 2015-05-06
Add SolarSailNavigator from Kerbal Stuff

KerbalStuffBot:
This pull request was automatically generated by Kerbal Stuff on behalf of mrsolarsail, to add [SolarSailNavigator](https://kerbalstuff.com/mod/752/SolarSailNavigator) to CKAN.

Please direct questions about this pull request to [mrsolarsail](https://kerbalstuff.com/profile/mrsolarsail).

PR #9766, opened 2023-08-23
Add PersistentThrustNavigator

### LINKED ISSUE #9488: [Mod] Persistent Thrust Navigator (v1.1.0)
Mod info
--------
* Mod name: PersistentThrustNavigator
* Author(s): JarredEagley
* License: Unlisted (or i just can't find it)
* Homepage: https://github.com/JarredEagley/PersistentThrustNavigator
* original forum post: https://forum.kerbalspaceprogram.com/index.php?/topic/107676-wip13-solarsailnavigator-v109/
* Download: https://github.com/JarredEagley/PersistentThrustNavigator/releases

HebaruSan:
An adoption and significant rewrite/rebranding of SolarSailNavigator (see #1165, #1376, #8477).

- https://github.com/JarredEagley/PersistentThrustNavigator
- https://forum.kerbalspaceprogram.com/topic/107676-*/?do=findComment&comment=4152274

TODO:

- [x] Asked for clarification about the PersistentThrust dependency on the forum thread.
- [x] Needs compatibility

Fixes #9488.

___

ckan compat add 1.10

### DIFF
--- a/NetKAN/PersistentThrustNavigator.netkan
+++ b/NetKAN/PersistentThrustNavigator.netkan
@@ -4,6 +4,8 @@ author:
   - mrsolarsail
   - JarredEagley
 $kref: '#/ckan/github/JarredEagley/PersistentThrustNavigator'
+ksp_version_min: 1.12.2
+ksp_version_max: 1.12.5
 license: LGPL-2.1
 resources:
   homepage: https://forum.kerbalspaceprogram.com/topic/107676-*/?do=findComment&comment=4152274
--- a/NetKAN/PersistentThrustNavigator.netkan
+++ b/NetKAN/PersistentThrustNavigator.netkan
@@ -1,0 +1,16 @@
+spec_version: 1
+identifier: PersistentThrustNavigator
+author:
+  - mrsolarsail
+  - JarredEagley
+$kref: '#/ckan/github/JarredEagley/PersistentThrustNavigator'
+license: LGPL-2.1
+resources:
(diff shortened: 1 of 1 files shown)


HebaruSan:
Hi @JarredEagley, I guess I should try asking follow-up questions here in case you're not frequently active on the forum...

- Which version of PersistentThrust is required? Is it this one?
  https://github.com/sswelm/PersistentThrust
  [https://spacedock.info/mod/2450/Persistent Thrust](https://spacedock.info/mod/2450/Persistent%20Thrust)
  https://forum.kerbalspaceprogram.com/topic/194707-1711811911101-persistent-thrust-extended-175-now-with-background-persistent-thrust/
  (Asking because that's the one on CKAN, so if there's a newer fork, we'll need to update that first...)
- Which game versions are compatible? Just 1.12?

JarredEagley:
Hi @HebaruSan! Yeah I don't check the forum often. I should probably reply there too since it's more accessible.

As for game versions: This version of PTN is compiled for KSP 1.12.2, but I would imagine it probably works in the latest release since KSP doesn't really get updates anymore.

This plugin was built with FreeThinker's persistent thrust mod in mind ([](https://forum.kerbalspaceprogram.com/topic/194707-1711811911101-persistent-thrust-extended-175-now-with-background-persistent-thrust/))
As far as I'm aware that plugin is rarely updated so it should? Still work? Hopefully. Let me know if it doesn't and I can take a look.

HebaruSan:
Cool, thanks for the quick response, it should all be good to go now! 🚀

PR #9793, opened 2023-09-04
website added, recommendation added

meirumeiru:
CameraControllerNext: website added

- <https://forum.kerbalspaceprogram.com/topic/219271-cameracontroller-next/>
- <https://github.com/meirumeiru/CameraController-Next>

InfernalRoboticsNext: recommendation for CameraControllerNext added

### DIFF
--- a/NetKAN/CameraControllerNext.netkan
+++ b/NetKAN/CameraControllerNext.netkan
@@ -6,5 +6,7 @@ author: Rudolf Meier
 $kref: '#/ckan/github/meirumeiru/CameraController-Next'
 $vref: '#/ckan/ksp-avc'
 license: GPL-3.0
+resources:
+  homepage: https://forum.kerbalspaceprogram.com/index.php?/topic/219271-*
 tags:
   - plugin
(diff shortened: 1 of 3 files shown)


HebaruSan:
Hi @meirumeiru, thanks for the update!

A better way of setting the homepage is to edit the website link in the GitHub repo, which will be pulled into that field automatically:

image

&rArr;

image

That way, the author (you!) can control the value at all times without having to submit netkan pull requests. It also ensures that the homepage link is available to non-CKAN users viewing the repository. Would you be interested in doing this for Camera Controller Next?

meirumeiru:
didn't know this... I will set it
other question: what happens if I'd set it for e.g. KJR next (which already has this value in the .netkan file)?

HebaruSan:
> didn't know this... I will set it

Cool! 👍 

> other question: what happens if I'd set it for e.g. KJR next (which already has this value in the .netkan file)?

The .netkan file overrides, but we can easily remove that to allow the setting in the repo to take over.

meirumeiru:
ok then... I have added the website to CameraController-Next, InfernalRobotics-Next and KerbalJointReinforcement-Next

should I close this request and send a new one with the modified changes?

HebaruSan:
> should I close this request

No, we can update this pull request to keep the desired changes and drop the others. One moment...

HebaruSan:
The website for CameraController-Next has already propagated through in https://github.com/KSP-CKAN/CKAN-meta/commit/7399c5bed948375cbc41124191840df082b4d0c9.

HebaruSan (approved):
Thanks for the updates!

PR #9800, opened 2023-09-16
add harmony as a dependency to kos-eva

JonnyOThan:


HebaruSan:
You can disregard that inflation error, looks like a bug from KSP-CKAN/CKAN#3904 that I need to fix...

HebaruSan (approved):
Thanks for the update!

PR #9804, opened 2023-09-19
Suppress recs on CKAN-hosted netkans for RO group mods

NathanKell:
https://github.com/KSP-CKAN/CKAN/pull/3892

### DIFF
--- a/NetKAN/CanaveralPads.netkan
+++ b/NetKAN/CanaveralPads.netkan
@@ -1,4 +1,4 @@
-spec_version: v1.16
+spec_version: v1.26
 identifier: CanaveralPads
 name: CanaveralPads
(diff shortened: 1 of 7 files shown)


NathanKell:
I left spec version at 1.26 where it was, and upgraded it to 1.26 where it was lower than that, but does suppress recs need a higher spec version? If so I'll make that change here and also all the remote netkans.

HebaruSan:
No; `suppress_recommendations` will be part of the v1.34 spec, but its presence does not break previous client, so no change to `spec_version` is needed.

NathanKell:
Thanks! Ok then I think I hit everything here + remote.

HebaruSan (approved):
Thanks for the update! 🎆

Issue #9328, opened 2022-09-10
Mod Compatibility Issue


CyberStar24:
Kerbal Field is now compatible with BDArmory Plus (previously known as BDArmory for RunwayProject)

CyberStar24:
https://forum.kerbalspaceprogram.com/index.php?/topic/209092-19x-112x-bdarmory-plus-bda-2022-09-01/

Under "Weapon Packs for BDA+" Kerbal Field is listed. I have also tested both mods together and it works.

HebaruSan:
Based on that response I'll assume you are not the author. Thanks.

HebaruSan:
Trying to gather indications of the author's intentions...

Forum thread:

![image](https://user-images.githubusercontent.com/1559108/189495696-d3c3e053-8c7b-4639-9b16-dd81e9750243.png)

SpaceDock:

![image](https://user-images.githubusercontent.com/1559108/189495722-66f53982-f81d-4d07-8257-9e65f67478e4.png)

The BDAPlus thread is written by the author of BDAPlus, not this mod's author, so someone else thought they worked together, not the author of this mod.

At this point it looks like CKAN's metadata is an accurate reflection of what this mod's author has stated. I'm sorry for the inconvenience, but we wouldn't want to cause users to report bugs with this mod because of something CKAN did. If you'd still like to have this changed, please contact the author and ask if they can update the dependency list somewhere or make a statement that it's actually supported.

PR #9806, opened 2023-09-20
Make KerbalField depend on the virtual BDArmory identifier

JonnyOThan:
I got permission from the author on the ksp forums for this

image

___

Previous time this came up: #9328

### DIFF
--- a/NetKAN/KerbalField.netkan
+++ b/NetKAN/KerbalField.netkan
@@ -1,15 +1,11 @@
-{
-    "spec_version": "v1.4",
-    "identifier":   "KerbalField",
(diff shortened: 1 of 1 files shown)


HebaruSan:
> I got permission from the author on the ksp forums for this

Was it in a PM? I like to confirm things like this to be on the safe side, so I checked the thread and the author's recent activity, but I didn't find anything.

JonnyOThan:
ah yep it was in a DM: 
image

BingoXXXXX:
I approve

HebaruSan (approved):
Thanks for taking care of this!

PR #9818, opened 2023-10-17
Force-allow KSPWheel on 1.12 while AVC is down

NathanKell:
This works around the issue where the remote version file (which gives this compatibility) is hosted on the AVC server, which is down (as it sometimes is).

### DIFF
--- a/NetKAN/KSPWheel.netkan
+++ b/NetKAN/KSPWheel.netkan
@@ -1,26 +1,17 @@
-{
-    "spec_version": "v1.4",
-    "identifier":   "KSPWheel",
-    "$kref":        "#/ckan/github/shadowmage45/KSPWheel",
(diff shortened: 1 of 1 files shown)


NathanKell:
@HebaruSan I assume this is the better approach than editing the .ckan file?

HebaruSan (approved):
Yeah, if you changed the .ckan in CKAN-meta, the bot would just overwrite it. This should be fine.

NathanKell:
Thanks! :)

HebaruSan:
The change made it to CKAN-meta, should be available to users in 2 minutes or so.

https://github.com/KSP-CKAN/CKAN-meta/commit/8c6e37a563d507aa04f69dc1735166364fac8f61

PR #9820, opened 2023-10-23
Add TweakScale-Redist as dependency to PhotonSailor

JonnyOThan:
This mod breaks without TweakScale-Redist (it's also bundled in the download)

- https://spacedock.info/mod/1895/Photon%20Sailor

### DIFF
--- a/NetKAN/PhotonSailor.netkan
+++ b/NetKAN/PhotonSailor.netkan
@@ -16,7 +16,8 @@
     } ],
     "depends": [
         { "name": "InterstellarRedistributable" },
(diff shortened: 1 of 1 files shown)


HebaruSan (approved):
Thanks for the update! 👍

HebaruSan:
BTW @JonnyOThan, I think that your invitation to the metadata wranglers team was sent at some point. If it's still there, you can accept it to merge stuff like this yourself if you like. I don't know whether those invites expire, so just let me know if you want it re-sent.

PR #9831, opened 2023-11-07
Add ROUtils and RONoCareer

NathanKell:
@HebaruSan for once I actually made the mods (including netkan!) before making the PR. Woot!

(FYI the plan for these are that ROUtils will be a dependency for most KSP-RO mods, but should not be installed on its own; and RONoCareer will go into the express install in an any_of with RP-1 so that folks can choose to install RO but _not_ RP-1, finally. It's purely a stub so the any_of can work, so it doesn't actually do anything other than exist and be findable by CKAN.)

### DIFF
--- a/NetKAN/RONoCareer.netkan
+++ b/NetKAN/RONoCareer.netkan
@@ -1,0 +1,6 @@
+spec_version: v1.4
+identifier: RONoCareer
+$kref: '#/ckan/netkan/https://raw.githubusercontent.com/KSP-RO/RONoCareer/master/RONoCareer.netkan'
+x_netkan_license_ok: true
+tags:
+  - career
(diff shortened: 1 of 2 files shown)


NathanKell:
Well that was fun. Apparently using github.dev created wrong line endings when I did save-as on an existing metanetkan. In the end I had to nuke and recreate both files using the old web interface. (Slightly weird the inflation bot doesn't accept \r\n though!)

NathanKell:
Is there anything else that needs doing here? There's multiple KSP-RO mod releases blocked on this being merged so it'd be awesome if folks could let me know if any changes are needed so this can go in. I'm also happy to merge this myself if folks are busy!

PR #9832, opened 2023-11-07
Update KourageousTourists.netkan

Lisias:


HebaruSan:
Looks like https://ksp.lisias.net/blogs/rants/on-CKAN/ and https://ksp.lisias.net/blogs/rants/on-CKAN/why-we-ditched-CKAN/ are still up.

PR #9834, opened 2023-11-08
adopting ALCOR and pointing it at my new version

JonnyOThan:
note that the new version includes the IVA instead of keeping a separate package, so I've marked the old IVA pack as a conflict

### DIFF
--- a/NetKAN/ALCOR.netkan
+++ b/NetKAN/ALCOR.netkan
@@ -1,7 +1,7 @@
 {
     "spec_version": "v1.4",
     "identifier":   "ALCOR",
-    "$kref":        "#/ckan/spacedock/1203",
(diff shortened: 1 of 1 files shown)


JonnyOThan:
uhhh not really sure what's wrong here :/

HebaruSan:
> uhhh not really sure what's wrong here :/

A C# exception generally means there's a formatting problem for which there's no explicit check. One moment...

HebaruSan (approved):
Thanks for the update! Are you planning to update the recommendations/supports to match the lists on the repo? If not, then this looks ready to merge to me, but I'll hold off just in case you were planning that.

JonnyOThan:
Hmm, let's leave the netkan as-is for now; most of the stuff in the readme on github is inherited from the old version and I haven't cross-checked.

PR #9836, opened 2023-11-09
RPM: change alcor suggestion to the new identifier

JonnyOThan:


HebaruSan (approved):
Thanks for the update.

PR #9860, opened 2023-11-24
Force_v for KerbalKrashSystem.netkan

HebaruSan:
image

A `v` prefix was dropped, which triggered an epoch in KSP-CKAN/CKAN-meta#2129, and the v0.4.7 version is now double indexed:

image

We'll need to do some clean-up in CKAN-meta in a subsequent PR.


HebaruSan:
image

PR #9896, opened 2024-01-06
Update RationalResourcesParts relationships

HebaruSan:
image

FYI to @JadeOfMaar and @JonnyOThan.

### DIFF
--- a/NetKAN/RationalResourcesParts.netkan
+++ b/NetKAN/RationalResourcesParts.netkan
@@ -11,5 +11,5 @@ tags:
   - resources
 depends:
   - name: ModuleManager
+recommends:
   - name: RationalResources
-  - name: RationalResourcesCompanion


HebaruSan:
image

I think this error happens because `soft_resolve` is only set to true for direct recommendations, and the conflict here involves a dependency-of-recommendation:

- RRP recommends:
  - RR recommends:
    - CommunityResourcePack
    - ClassicStockResources depends:
      - WildBlue-PlayMode-ClassicStock conflicts:
        - CommunityResourcePack

The resolver doesn't remember enough state to pop off a recommendation that's already in the process of being resolved, so when WildBlue-PlayMode-ClassicStock's conflict is hit, it just fails; something to return to later. A workaround would be to reorder RationalResources's recommendations so the conflict is already encountered by the time CommunityResourcePack gets checked. Should be fine in practice because GUI doesn't use `with_recommends`.

JadeOfMaar:
Circular, conflicting recommends, eh. Oh boy. Looks like I've earned myself the **Bug Breeder** role 😇

JonnyOThan:
Note that ClassicStockResources requiring WildBlue-PlayMode-ClassicStock is probably incorrect. But there’s also a decent chance that CommunityResourcePack and ClassicStockResources should be in conflict anyway.  https://github.com/KSP-CKAN/NetKAN/issues/9899

Does any_of work for recommendations?

PR #9904, opened 2024-01-15
Split out play modes for ClassicStockResources

### LINKED ISSUE #9899: WBI play modes and classic stock resources needs another look
Relevant background: https://github.com/KSP-CKAN/NetKAN/pull/8275

Currently ClassicStockResources depends on WildBlue-PlayMode-ClassicStock, which seems backwards: https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/ClassicStockResources.netkan

Recently we took steps to make sure that CommunityResourcePack conflicts with wildblue play modes other than WildBlue-PlayMode-CRP and WildBlue-PlayMode-USI (note the latter is maintained by TheDog and not Angel-125.  And it may be abandoned).  https://github.com/KSP-CKAN/NetKAN/pull/9893.  This came out of some bug reports from people trying to use CommunityResourcePack and other play modes.

So, my proposal:

* Flip the dependency: WildBlue-PlayMode-ClassicStock should depend on ClassicStockResources, not the other way around.
* Make ClassicStockResources conflict with CommunityResourcePack
* MAYBE: make WildBlue-PlayMode-Pristine conflict with ClassicStockResources
* MAYBE: make WildBlue-PlayMode-Simplified depend on ClassicStockResources

I'd want to do at least a little digging to look for other mods that interact with these to make sure we're not screwing anything up.

### LINKED ISSUE #9905: Pathfinder installs WBIPlayMode.cfg to GameData root
https://github.com/KSP-CKAN/NetKAN/blob/83fe0b95edb2f6280935ce144ddec9aa14672154/NetKAN/Pathfinder.netkan#L39

Is this intentional?  Shouldn't this be controlled by the play mode selection?  If I select a play mode in-game and then try to install pathfinder (using simplified play mode, the same one I had selected), CKAN throws up this warning: 
image

For reference, other WBI mods like DSEV and Heisenberg *do not* install the WBIPlayMode.cfg file.

HebaruSan:
## Motivations

These mods bundle ClassicStockResources:

- DSEV
- Pathfinder
- MOLE
- Heisenberg
- Buffalo

If you install any of them manually and choose the ClassicStock play mode, the ClassicStockResources configs will be active. However, if you install with CKAN and pick the ClassicStock play mode, ClassicStockResources is **not** installed (unless the user chooses it). This isn't the expected/intended state of the mod in that case.

In addition, most but not all of ClassicStockResources's .cfg files are renamed to .txt if you pick any play mode other than ClassicStock. But in CKAN you can end up with all or none of the .cfg files when you should only have one.

Finally, PathFinder installs `GameData/WBIPlayMode.cfg`, which is bad because it means this file will be overwritten at install and upgrade.

## Changes

- Now the ClassicStock play mode depends on ClassicStockResources, so it will always be installed if you pick the ClassicStock play mode in CKAN
- Now ClassicStockResources-PlayMode-ClassicStock is created to model the state of having ClassicStockResources installed and choosing the ClassicStock play mode. Installing ClassicStockResources without this module will model the state of choosing some other play mode.
  - ClassicStockResources-PlayMode-ClassicStock conflicts with CommunityResourcePack, because these are incompatible. The main module does not conflict, because you can have it installed while using another play mode.
  - Now the mods that bundle ClassicStockResources depend on it, to ensure that they always have access to that one file that doesn't get renamed to .txt in other play modes
- Now `GameData/WBIPlayMode.cfg` is no longer installed by Pathfinder

Fixes #9899 (once the historical PR in CKAN-meta is done).
Fixes #9905.

### DIFF
--- a/NetKAN/Buffalo.netkan
+++ b/NetKAN/Buffalo.netkan
@@ -1,49 +1,42 @@
-{
-    "spec_version": "v1.4",
-    "identifier":   "Buffalo",
-    "name":         "Buffalo",
-    "author":       "Angel-125",
-    "$kref":        "#/ckan/github/Angel-125/Buffalo",
-    "$vref":        "#/ckan/ksp-avc",
-    "x_netkan_staging": true,
-    "x_netkan_staging_reason": "Make sure that no new play modes were added to the Templates folder",
-    "license":      "GPL-3.0",
-    "resources": {
-        "homepage": "http://forum.kerbalspaceprogram.com/index.php?/topic/122617-*"
-    },
-    "tags": [
-        "parts",
-        "crewed"
-    ],
-    "depends": [
-        { "name": "Buffalo-PlayMode" },
-        { "name": "ModuleManager"    },
-        { "name": "WildBlueTools"    },
-        { "name": "KerbalActuators"  }
-    ],
-    "supports": [
-        { "name": "ConnectedLivingSpace" },
-        { "name": "TACLS"                },
-        { "name": "USI-LS"               },
-        { "name": "Snacks"               }
-    ],
-    "recommends": [
-        { "name": "Snacks"                   },
-        { "name": "ExtraPlanetaryLaunchpads" },
-        { "name": "KIS"                      },
-        { "name": "KAS"                      },
-        { "name": "ASET"                     },
-        { "name": "ASETAvionics"             }
-    ],
-    "install": [ {
-        "find":       "Buffalo",
-        "install_to": "GameData/WildBlueIndustries",
-        "filter":     [ "Ships", "Templates" ]
-    }, {
-        "find":       "Buffalo/Templates/Common",
-        "install_to": "GameData/WildBlueIndustries/Buffalo/Templates"
-    }, {
-        "find":       "Ships/SPH",
-        "install_to": "Ships"
-    } ]
-}
+spec_version: v1.4
+identifier: Buffalo
+name: Buffalo
+author: Angel-125
+$kref: '#/ckan/github/Angel-125/Buffalo'
+$vref: '#/ckan/ksp-avc'
+x_netkan_staging: true
+x_netkan_staging_reason: Make sure that no new play modes were added to the Templates folder
+license: GPL-3.0
+resources:
+  homepage: http://forum.kerbalspaceprogram.com/index.php?/topic/122617-*
+tags:
+  - parts
+  - crewed
+depends:
+  - name: Buffalo-PlayMode
+  - name: ModuleManager
+  - name: WildBlueTools
+  - name: ClassicStockResources
+  - name: KerbalActuators
+supports:
+  - name: ConnectedLivingSpace
+  - name: TACLS
+  - name: USI-LS
+  - name: Snacks
+recommends:
+  - name: Snacks
+  - name: ExtraPlanetaryLaunchpads
+  - name: KIS
+  - name: KAS
+  - name: ASET
+  - name: ASETAvionics
+install:
+  - find: Buffalo
+    install_to: GameData/WildBlueIndustries
+    filter:
+      - Ships
+      - Templates
+  - find: Buffalo/Templates/Common
+    install_to: GameData/WildBlueIndustries/Buffalo/Templates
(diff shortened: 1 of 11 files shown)


NetKAN/ClassicStockResources.netkan
```diff
@@ -1,23 +1,19 @@
...
+resources:
+  homepage: https://github.com/Angel-125/ClassicStockResources
+tags:
+  - library
+  - resources
+conflicts:
+  - name: CommunityResourcePack
+depends:
+  - name: ModuleManager
+  - name: WildBlue-PlayMode-ClassicStock
```
JonnyOThan, line 16: we want to remove this one right?

NetKAN/ClassicStockResources.netkan
```diff
@@ -1,23 +1,19 @@
...
+license: GPL-3.0
+resources:
+  homepage: https://github.com/Angel-125/ClassicStockResources
+tags:
+  - library
+  - resources
+conflicts:
+  - name: CommunityResourcePack
+depends:
+  - name: ModuleManager
```
JonnyOThan, line 15: man it's very close to not even needing MM, but there are a couple `:NEEDS[!CommunityResourcePack]` in there

NetKAN/ClassicStockResources.netkan
HebaruSan: I don't think so?

JonnyOThan: I'm pretty sure Simplified play mode requires it

HebaruSan: Can you say why?

HebaruSan: Confirmed that the Simplified play mode disables ClassicStockResources, see https://github.com/KSP-CKAN/NetKAN/issues/9899#issuecomment-1892552659.

JonnyOThan: I still think this is wrong, there are a number of mods that use ClassicStockResources but don't need a WBI play mode.

https://github.com/search?q=repo%3AKSP-CKAN%2FNetKAN%20ClassicStockResources&type=code

HebaruSan: ClassicStockResources is part of the ClassicStock play mode. If a mod "uses" (note, that's not the same as `supports` or `suggests`, which is most of what is linked there) it, then it uses the ClassicStock play mode.

HebaruSan:
image

This is due to taniwha.org being broken (see #9734), which we can't fix because Kethane has no alternate downloads. Not much choice but to ignore this error.

HebaruSan:
image

And now we're getting an error due to Angel-125/DSEV#20. Again we'll probably just have to ignore it, since attempts to reach out on GitHub, Discord, and the forum have failed.

NetKAN/Buffalo.netkan
```diff
@@ -1,49 +1,42 @@
...
+resources:
+  homepage: http://forum.kerbalspaceprogram.com/index.php?/topic/122617-*
+tags:
+  - parts
+  - crewed
+depends:
+  - name: Buffalo-PlayMode
+  - name: ModuleManager
+  - name: WildBlueTools
+  - name: ClassicStockResources
```
JonnyOThan, line 19: Wait I don't think this is correct either, doesn't this prevent you from using any playmode other than classic stock?

HebaruSan: No, the changes evolved a bit as I worked through this, hence the title change.

NetKAN/ClassicStockResources.netkan
JonnyOThan: > ClassicStockResources is part of the ClassicStock play mode

I don't think this is true.  Other playmodes may disable ClassicStockResources, but that's not the same thing.

HebaruSan: Right. All but one .cfg file is part of that play mode. The remaining one is generically available. That's modeled in the latest changes. I'd recommend taking a look at the files, then opening CKAN, clicking Refresh, and taking it for a spin to see how it all works.

PR #9908, opened 2024-01-16
Add XenonISRU

A-T-L-A-S:
- <https://github.com/A-T-L-A-S/XenonGas-ISRU>

### DIFF
--- a/NetKAN/XenonISRU
+++ b/NetKAN/XenonISRU
@@ -1,0 +1,19 @@
+spec_version: v1.27
+identifier: XenonISRU
+name: XenonISRU
+abstract: A tiny MM patch that adds Xenon Gas to Vanilla ISRUs in Kerbal space program
+author:
+  - A.T.L.A.S
+license: MIT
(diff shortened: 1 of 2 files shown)


JonnyOThan:
Oh huh...there's already a mod using the XenonISRU identifier: https://spacedock.info/mod/1010/Xenon%20ISRU

That one seems a little more fleshed out...is there a reason we need another?  If so then you'll need to pick a different identifier.

BTW since that mod's license is public domain, you're totally free to adopt it and replace it with your own version on CKAN (though we'd recommend making an effort to contact the original author first).

HebaruSan:
Hi @A-T-L-A-S, we had some follow-up questions about this mod (see previous post by @JonnyOThan). If you have time to take a look and respond, that would help us decide what to do next. Thanks!

HebaruSan:
Hi @A-T-L-A-S, if you have a chance at some point to address the follow-up questions above, please do, and we can re-open this.

PR #9910, opened 2024-01-23
Rename EVE stock configs to reduce user confusion

JonnyOThan:


HebaruSan (changes requested):
This is changing a string that explains what the module is to one that is opaque and doesn't mean anything to the user. I'm not seeing why we'd do this.

JonnyOThan:
Another example:

IMG_7684

HebaruSan:
Interesting, that user found the current name clear once they read the whole thing.

HebaruSan:
@JonnyOThan, what if we just shorten "Environmental Visual Enhancements" to "EVE" so the rest can be seen more easily?

EVE - Stock Planet Config files

JonnyOThan:
> what if we just shorten "Environmental Visual Enhancements" to "EVE" so the rest can be seen more easily?

Would certainly help I think

HebaruSan (approved):
Alright, let's try this.

JonnyOThan:
How about adding "aka BoulderCo" or something in the abstract?  In any case this is certainly an improvement.

Issue #9044, opened 2022-03-20
[Mod Update] Rocket Sound Enhancement - Default Configs


ensou04:
Mod info
--------
Can the mod be renamed to "Rocket Sound Enhancement - Default"?
GitHub Repository has been moved to https://github.com/ensou04/RocketSoundEnhancementDefault

* Mod name: Rocket Sound Enhancement - Default
* Author(s): ensou04
* License: CC-BY-NC-SA-4.0
* Homepage: https://forum.kerbalspaceprogram.com/index.php?/topic/179579-wip110x-112x-rocket-sound-enhancement-v060-audio-framework-for-complex-sound-effects/
* Download: https://github.com/ensou04/RocketSoundEnhancementDefault/releases/

HebaruSan:
Thanks for letting us know, should be updated for users within about 35 minutes.

ensou04:
thank you!

ensou04:
I just noticed the typo on ckan, it says "Rocket Sound **Enhancment** - Default"

HebaruSan:
Thanks, a fix for that is in c31c575df5d11e1a943840888d813ff59ccc2804 and working its way through the system, shouldn't be more tan 30 minutes.

PR #9918, opened 2024-02-05
Add "config" to RSE-default display name

JonnyOThan:
There's some confusion about what this module is and whether it's necessary, this might help

### DIFF
--- a/NetKAN/RocketSoundEnhancement-Config-Default.netkan
+++ b/NetKAN/RocketSoundEnhancement-Config-Default.netkan
@@ -1,6 +1,6 @@
 spec_version: v1.4
 identifier: RocketSoundEnhancement-Config-Default
-name: Rocket Sound Enhancement - Default
+name: Rocket Sound Enhancement - Default Config
 abstract: RSE sounds for stock engines, wheels, decouplers, and more
 $kref: '#/ckan/github/ensou04/RocketSoundEnhancementDefault'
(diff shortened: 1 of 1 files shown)


HebaruSan:
That was actually specifically requested by the author in #9044, unless there was some misunderstanding there, which is very possible.

JonnyOThan:
As always thanks for the background!  I think your take is correct.  @ensou4 any thoughts here?  This doesn't come up *that* often but I just saw someone on discord that was confused about what to install.

HebaruSan:
Let's leave this alone for now until we can confirm it wouldn't miff the author.

ensou04:
> As always thanks for the background! I think your take is correct. @ensou04 any thoughts here? This doesn't come up _that_ often but I just saw someone on discord that was confused about what to install.

[Rocket Sound Enhancement](https://github.com/ensou04/RocketSoundEnhancement) the Plugin is the brains behind the mod, it doesn't have any library of sounds or configs. Can be installed alone and it'll provide sound muffling and other basic atmospheric sound filtering.

[Rocket Sound Enhancement - Default](https://github.com/ensou04/RocketSoundEnhancementDefault) is the library of sounds and configs that replaces the stock and supported mod sound effects plus additional atmospheric effects. It requires the plugin to function. 

Separated them because they're technically two different mods and had different development cycles, didn't want to constantly be uploading a large library of files every time we updated the plugin.

PR #7264, opened 2019-06-21
Create JNSQ.netkan

Galileo88:
https://raw.githubusercontent.com/Galileo88/JNSQ/master/CKAN/JNSQ.netkan

ckan compat add 1.6

PR #9936, opened 2024-02-18
Add optional JNSQ rescale mods to stock (1x) and real (10x) size

rkunze:
These optional mods are already included in the JNSQ distribution itself, this PR merely makes them installable in CKAN.

- <https://github.com/Galileo88/JNSQ>

### DIFF
--- a/NetKAN/JNSQ-PartRebalancer-10x.netkan
+++ b/NetKAN/JNSQ-PartRebalancer-10x.netkan
@@ -1,17 +1,25 @@
 spec_version: v1.4
 identifier: JNSQ-PartRebalancer-10x
-name: JNSQ PartRebalancer 10x
-abstract: Rebalances the mass of parts and the density of propellants to give rockets more life-like mass ratios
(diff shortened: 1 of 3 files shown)


HebaruSan:
xref: https://forum.kerbalspaceprogram.com/topic/197082-ckan-the-comprehensive-kerbal-archive-network-v1332-laplace-ksp-2-support/?do=findComment&comment=4367242

The wrong line is still wrong and will require re-work.

HebaruSan:
Regarding author permission to add modules, JNSQ was added via author-submitted PR #7264, which doesn't say anything explicit to limit or extend the scope of permission, and there was no discussion at that time of what to do with the contents of the `Optional Mods` folder. I think this plausibly falls within the scope of "adding JNSQ" generally, but @Galileo88 and @OhioBob, if you disagree, let us know and we can remove these.

HebaruSan (approved):
Thanks for the submission!

PR #7541, opened 2019-12-08
Add Grannus Expansion Pack

HebaruSan:
This is a planet pack with some optional components, originally grew out of GPP, now available in standalone mode but also with JNSQ compatibility.

Permission requested here, monitor for responses:

https://forum.kerbalspaceprogram.com/index.php?/topic/169664-173-grannus-expansion-pack-v111-7-dec-2019/&do=findComment&comment=3715014

Note Galileo88/Galileos-Planet-Pack/issues/59, which currently makes a functional GPPSecondary+GEP_Primary install impossible. Need GPPSecondary updated to actually install its needed files.

PR #9938, opened 2024-02-19
Add optional GEP rescale 10x mod

rkunze:
- <https://github.com/OhioBob/Grannus-Expansion-Pack>
- #8330 added GEP rescale (2.5x)
- #7541 added GEP
- #9936 added the part rebalancer 10x that both JNSQ and GEP use


HebaruSan (approved):
Thanks for this submission as well!

PR #9956, opened 2024-03-14
Fix gpp scatterer version

JonnyOThan:
In #8880 and #8881 the scatterer version for GPP was set, but it incorrectly allowed scatterer 0.0852 to be installed.

### DIFF
--- a/NetKAN/GPP.netkan
+++ b/NetKAN/GPP.netkan
@@ -28,7 +28,7 @@ depends:
 conflicts:
   - name: StockVisualEnhancements
   - name: Scatterer
-    min_version: '3:v0.0825b'
+    min_version: '3:v0.0825'
 recommends:
   - name: Strategia
   - name: FinalFrontier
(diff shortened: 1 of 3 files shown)


HebaruSan:
Is it actually incorrect?

image

JonnyOThan:
yes, but it's also not compatible with 0.0825.

HebaruSan:
> GPP is also not compatible with 0.0825

I guess I was hoping to learn the reason behind this claim, since the announcement by the author is ambiguous and doesn't address it.

JonnyOThan:
Hard to say, but considering the 'b' designation and the fact that 0.0852b was released a day after 0.0852, I'm guessing that the GPP authors were merely referencing the most recent publicly available build of scatterer which was not compatible.

JonnyOThan:
Oh shoot do we also need to go edit the historical versions according to the comment in 8880?

JonnyOThan:
Hm I tried installing latest scatterer and then GPP, and it correctly flagged the conflict.  I also tried installing scatterer 0.0852b and then GPP, and it was also marked as conflict.  Maybe installing older versions of GPP would be allowed but it seems like the common pathways are correct at least.

PR #9977, opened 2024-04-06
Declares a conflict with TweakScaleRescaled-Redist

Lisias:
`TweakScaleRescaled-Redist` is being offered as an option when installing `TweakScale`, whats a mistake - while I can guarantee compatibility with the current interfaces, I can't say the same for future new ones.

### DIFF
--- a/NetKAN/TweakScale-Redist.netkan
+++ b/NetKAN/TweakScale-Redist.netkan
@@ -12,6 +12,8 @@ x_netkan_force_v: true
 license:
   - WTFPL
 release_status: stable
+conflicts:
+  - name: TweakScaleRescaled-Redist
 tags:
   - plugin
   - library


HebaruSan:
@JonnyOThan, that's your mod. Does this look right to you?

Lisias:
Interesting. Why nobody called me on this Pull Request?

https://github.com/KSP-CKAN/NetKAN/pull/9972

JonnyOThan:
@Lisias  What are you trying to do here?  TweakscaleRescaled-Redist already has a conflict with Tweakscale-redist so  not sure why this is needed. Did you mean to declare a conflict between Tweakscale and tweakscaleRescaled-redist?

Lisias:
When I installed TweakScale™ via CKAN, CKAN asked me if I wanted to install `TweakScale-Redist` or `TweakScaleRescaled-Redist` , what's an error.

If TweakScale is being installed, the only option available should be `TweakScale-Redist`, and it should not be installed without it.

JonnyOThan:
> When I installed TweakScale™ via CKAN, CKAN asked me if I wanted to install `TweakScale-Redist` or `TweakScaleRescaled-Redist` , what's an error.
> 
> If TweakScale is being installed, the only option available should be `TweakScale-Redist`, and it should not be installed without it, or using the alternative.

That makes sense, but this PR doesn't do that.  You would need to declare that `TweakScale` conflicts with `TweakScaleRescaled-Redist`.

JonnyOThan:
By the way, having two versions of Tweakscale-redist does create a bit of a headache; what would you think about collaborating on that library so that there's only one version?  Have you looked over the changes I made in tweakscalerescaled-redist?

Lisias:
TweakScale 2.5 will introduce a new interface, with retro-compatibility. Merging the Redists will bring over us yet more headaches than we have nowadays.

I'm fixing the pull request.

PR #10022, opened 2024-05-26
RealChute 1.4.9.0 update

ChrisViral:
- Added dependency to [ClickThroughBlocker](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/ClickThroughBlocker.netkan)
- Added dependency to [ToolbarController](https://github.com/KSP-CKAN/NetKAN/blob/master/NetKAN/ToolbarController.netkan)
- ModuleManager moved to recommendation; MM patches are optional and not required for the mod to function

### DIFF
--- a/NetKAN/RealChute.netkan
+++ b/NetKAN/RealChute.netkan
@@ -17,6 +17,10 @@
         "sound"
     ],
     "depends" : [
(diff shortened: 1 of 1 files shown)


HebaruSan (approved):
Thanks for the update! 👍

PR #10056, opened 2024-06-19
Add new dependency for RealAntennas

Capkirk123:
RA v2.3 now uses (and therefore requires) clickthroughblocker


HebaruSan (approved):
Thanks for the update!

Confirmed at: https://github.com/KSP-RO/RealAntennas/commit/8abbafd62a644ff26ba746b1696af74aa8c1ee93

PR #10058, opened 2024-06-20
Update paths of stock scatterer configs for new scatterer version

LGhassen:
Hi, I'm going to push a new scatterer update (0.0876) which changes the paths of the built-in stock configs a bit to separate mod and configs for the stock system.

- <https://spacedock.info/mod/141/Scatterer>

In the past the structure of the mod was as follows, where both the built-in quality presets and the stock planet and sunflare configs where all lumped under Scatterer/config

```
└───GameData
    └───Scatterer
        ├───config
        │   │      config.cfg
        │   │      planetsList.cfg
        │   │      qualityPresets.cfg
        │   ├───Planets
        │   ├───PluginData
        │   │       pluginData.cfg
        │   └───Sunflares
        ├───icon
        └───shaders
```

In the new version the stock planet and sunflare configs are moved to "StockScattererConfigs" which is a separate folders as follow, while the built-in quality presets remain under Scatterer/Config

```
└───GameData
    ├───Scatterer
    │   ├───config
    │   │      config.cfg
    │   │      qualityPresets.cfg
    │   ├───PluginData
    │   │      pluginData.cfg
    │   ├───icon
    │   └───shaders
    └───StockScattererConfigs
        ├───Planets
        │        planetsList.cfg
        └───Sunflares
```

I modified the netkan files for scatterer, scatterer-config and scatterer-sunflare to reflect this, let me now if this looks correct!


LGhassen:
Hmm the automated checks don't pass btw because I haven't pushed the new update yet, not sure if I should push the update first or not, and if this can be set to only apply to the new version?

HebaruSan:
Hi @LGhassen, thanks for the update!

> Hmm the automated checks don't pass btw because I haven't pushed the new update yet, not sure if I should push the update first or not, and if this can be set to only apply to the new version?

It would be best if the update came out before this is merged, so we can make sure we haven't missed anything. Hold on a bit, I'll freeze these netkans in the master branch so it won't update till after merge...

HebaruSan:
OK @LGhassen, I think you can safely make the new release now without having CKAN pick it up. Then we can re-run the validation tools here and confirm that everything looks OK before the updated metadata goes out to users.

LGhassen:
It's released now, thanks @HebaruSan

LGhassen:
I see, these are settings meant for the user to tweak performance, didn't realize that some mods would ship them. I'd be fine with having them in StockScattererConfigs for the version downloaded from CKAN. Is it possible to just do that via the Netkan files?

HebaruSan (approved):
OK, I made an update to move the `config` folder back to `Scatterer-config`.

Thanks for the update!

PR #10064, opened 2024-06-25
Update RSS-Adapter.netkan

Globul-Art:
Add supports modes


HebaruSan (approved):
Thanks for the update!

PR #10075, opened 2024-07-02
Fix VertexHeightOblateAdvanced install path

HebaruSan:
This mod's install path changed from `GameData/000_DuckweedUtils/VertexHeightOblateAdvanced` to `GameData/001_DuckweedUtils/VertexHeightOblateAdvanced`, but the CKAN metadata is "deep linking" to the original path. This was done because there were LICENSE and README.md files in `GameData/000_DuckweedUtils`, which would have caused a file conflict if another mod did the same thing.

Now we use a regex so the mod can pick whatever path it wants, and the author has moved the LICENSE and README.md into the `VertexHeightOblateAdvanced` folder.

### DIFF
--- a/NetKAN/VertexHeightOblateAdvanced.netkan
+++ b/NetKAN/VertexHeightOblateAdvanced.netkan
@@ -11,3 +11,6 @@ depends:
 install:
   - find_regexp: .*DuckweedUtils
     install_to: GameData
+    filter:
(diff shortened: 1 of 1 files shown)


HebaruSan:
Author is planning to migrate LICENSE and README.md to the mod folder...

PR #10076, opened 2024-07-02
Add Shabby dependency, remove license and homepage in favor of github autofill, remove install

DeltaDizzy:
An update using the new Shabby dependency is now out. License and Homepage were removed in favor of it pulling them from the github repo, and now the mod has a standard install structure so I also removed `install`. (Edits mostly generated via webtool)

- <https://github.com/DeltaDizzy/PerSaveScreenshots>


HebaruSan (approved):
Thanks for the update!
(Noting for posterity that this is also fixing an inflation error, so the current version isn't indexed yet.)

PR #10096, opened 2024-07-18
Update Mk2Rebalance.netkan

CensoredUsername:
This now has significant support for Mk2Expansion.


JonnyOThan:
@CensoredUsername would you also like to add “suggests” or “recommends” mk2 expansion?

CensoredUsername:
It is perfectly usable and intended to work with stock, just contains some code to also support Mk2Expansion due to popular request. `ecommends` definitely isn't needed, and from what I read about the meaning of the different descriptions, `suggests` would also not be correct

PR #10130, opened 2024-07-26
Update DansSunflareRSS.netkan

dandoesgithub:
I changed the file structure of my mod, this should fix it for the new version.


HebaruSan (approved):
Thanks for the update!

PR #10139, opened 2024-07-29
Recommend/Suggest RP-1

Capkirk123:
Recommend RP-1, not RP-0 Legacy


HebaruSan (approved):
Thanks for the update.

PR #10148, opened 2024-08-05
Add Deferred

### LINKED ISSUE #10147: Add Deferred
```yaml
spec_version: v1.27
identifier: Deferred
abstract: Adds deferred rendering, improves shading and performance
author:
  - Blackrack
license: GPL-3.0
download: https://github.com/LGhassen/Deferred/releases
$kref: '#/ckan/github/LGhassen/Deferred'
release_status: stable
resources:
  bugtracker: https://github.com/LGhassen/Deferred/issues
  license: https://github.com/LGhassen/Deferred/blob/master/License.md
  manual: https://github.com/LGhassen/Deferred/blob/master/Readme.md
  homepage: https://www.patreon.com/posts/deferred-106557481
  repository: https://github.com/LGhassen/Deferred
  x_screenshot: https://i.imgur.com/kraUQIz.png
tags:
  - plugin
  - graphics
ksp_version_min: 1.12.0
depends:
  - name: Shabby
  - name: Harmony2
x_via: Generated by Metadata Webtool

```

DeltaDizzy:
Closes #10147 

Summary of changes from submission given to blackrack, waiting for response

* Remove the download URL (kref does the same job better)
* Change `ksp_version_min` to `ksp_version` to make the compatibility fully bounded (this is a warn not an error so doesn't have to be included)
* Add explicit install directive (specify it is installing zzz_Deferred)


JonnyOThan:
Hey @LGhassen amd @DeltaDizzy :

Are there known incompatibilities with older versions of other mods? (Tufx, tu, etc).  Might be wise to declare them in here using a conflicts with a max version specifier.

LGhassen:
@DeltaDizzy changes sound good.

For the conflicts as suggested by @JonnyOThan I think this should do it:

`conflicts:
  - name: Parallax
    max_version: '2.0.6'
  - name: TexturesUnlimited
    max_version: '1.5.10.25'
  - name: KerbalKonstructs
    max_version: 'v1.8.6.1'
  - name: RasterPropMonitor
    max_version: '1:v0.31.13.4'
  - name: B9-PWings-Fork
    max_version: '3:0.45.1'`
    
@DeltaDizzy can you add these or do you need me to change the file in the original issue? Also I think you can remove Harmony from dependencies, it's only a depency of Shabby so having it as a dependency should cover it.

JonnyOThan:
I think it’s wise to explicitly declare dependencies (assuming deferred does indeed directly use Harmony) rather than relying on transitive dependencies as those may change over time. Unlikely in this case, but just as a general policy.

DeltaDizzy:
@LGhassen I can change it, thanks!

HebaruSan:
Hi @LGhassen, a question about `Deferred.cfg`. If CKAN installs this, then it will be overwritten every time the user upgrades to a new version of the mod. Will that be an issue?

LGhassen:
At the moment it shouldn't be, there is not enough meaningful settings to warrant separate mods for the configs. Everyone should be fine with the defaults.

HebaruSan (approved):
This looks like what I had in my local branch now.
Thanks for taking this on, @DeltaDizzy! Feel free to merge when ready.

LGhassen:
Thanks a lot for this!

PR #10151, opened 2024-08-06
DSEV: adjust recommendations according to content in cfg files

JonnyOThan:
While working on a bug related to RPM and DSEV, I discovered that it uses ASET Avionics props, and most of the IVAs have been converted to use MAS to drive them (the RPM versions are still there but renamed to .txt)


JonnyOThan:
@Angel-125 this look OK?

HebaruSan (approved):
Looks like it
image

PR #10152, opened 2024-08-06
Install craft files for Pwings

linuxgurugamer:
This is an update to add two test craft files to the stock craft.

This is dependent on the next version of PWings, which will be uploaded soon.  Please do not merge this, I'll do it once the new files are available


HebaruSan:
Hey @linuxgurugamer, the `Ships` folder probably shouldn't be in `GameData` since a user who installs manually might end up with files for which they don't remember the source mod. Any interest in changing that?

PR #10153, opened 2024-08-06
RasterPropMonitor: remove core/patches split

JonnyOThan:
I'm gearing up for a new RPM release and I think it's time we collapsed these.  The fact that RasterPropMonitor-Core is the one that contains RasterPropMonitor.dll causes a lot of problems with autodetection, etc.

This *should* just work without needing to go update all the existing mod metadata, but maybe we want to eventually go do that too?

Is replaced_by desirable here?


HebaruSan:
I will be very surprised if upgrading works smoothly here...

NetKAN/RasterPropMonitor-Core.netkan
```diff
@@ -1,51 +0,0 @@
...
-  - name: QuickIVA
-  - name: KSAIVAUpgrade
-install:
-  - find: JSI/RasterPropMonitor
-    install_to: GameData/JSI
-  - find: JSI/RPMPodPatches/BasicMFD
-    install_to: GameData/JSI/RPMPodPatches
-    comment: >-
-      This folder contains the multi-function display props, which belong in
-      this module along with the other props
```
HebaruSan, line 51: Noting here in case the commit comment is missed:
Please rename to .frozen instead of deleting, the infrastructure tooling looks for this.

PR #10166, opened 2024-08-14
Remove spec_version from all netkans

HebaruSan:
As of KSP-CKAN/CKAN#4155, `spec_version` is now optional and auto-generated by the inflator. KSP-CKAN/NetKAN-Infra#336 and KSP-CKAN/metadata-webtool#49 have removed code that added it to machine-generated netkans, and it has been removed from https://github.com/KSP-CKAN/CKAN/wiki/Adding-a-mod-to-the-CKAN.

Now all existing netkans have `spec_version` removed. Since this is almost every file in this repo, it will also be a sort of stress test of the metadata tester.


HebaruSan:
That was interesting. Since the download cache size limit is set per game instance, and inflation occurs before any game instances are created, there is effectively no download cache size limit for the inflation step. Which means that it will reliably fill the disk after enough modules are inflated.

Furthermore, we don't know the size of the download before we download it! So even if a limit was defined, there would be no way to know how much space is needed before each inflation.

PR #10169, opened 2024-08-16
Changed dependencies TundraExploration.netkan

TundraMods:
I've changed the dependencies from Tundra Exploration so that it needs SmokeScreen.

___

ckan compat add 1.11 1.12


HebaruSan (approved):
Thanks for the update!

FYI, I added a bit at the top to enable CKAN to download this from either GitHub or SpaceDock based on user preferences. Should help improve download speeds, in case anyone is having problems with that.

PR #10202, opened 2024-09-07
Add GitHub krefs to SpaceDock mods

HebaruSan:
SpaceDock has been having a lot of downtime lately. Many mods on SpaceDock are also hosted on GitHub,  and we have the ability to list both hosts for one mod, but so far only KSP2 mods have been consistently set up for this.

Now the applicable KSP1 mods are multi-hosted.

Essentially I re-ran the SpaceDock Adder's multi-hosting logic for already indexed mods to generate the GitHub krefs, then made manual edits after noting a few patterns of bad data (especially re: `x_netkan_github.use_source_archive` for plugins).

### DIFF
--- a/NetKAN/4kSPExpanded.netkan
+++ b/NetKAN/4kSPExpanded.netkan
@@ -1,22 +1,25 @@
-{
-  "license": "MIT",
-  "identifier": "4kSPExpanded",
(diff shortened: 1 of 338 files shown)


HebaruSan:
The remaining errors are for installation, which isn't changed here. All we're changing here is inflation, which succeeded.

PR #10173, opened 2024-08-18
Unfreeze Shabby at new fork, add Resurfaced

HebaruSan:
- <https://github.com/KSPModdingLibs/Shabby>
- <https://github.com/Tantares/Resurfaced>

This will fail until a release is added to the new Shabby repo.

Fixes #10172.

PR #10205, opened 2024-09-09
Update shabby netkan

drewcassidy:
Now under KSPModdingLibs


JonnyOThan:
Hey would you rather use version-from-asset or point to the .version file in the release?  I think the latter might be a little more expressive and less fragile, as long as you keep publishing the .version file in releases.

HebaruSan:
Thanks, this was already in progress in #10173

PR #10206, opened 2024-09-10
Update SillyPhotonDrives.netkan

Triplchii:
Moved CommunityResourcePack to suggests, moved PersistentThrust & NearFutureElectrical to suggests, moved CommunityTechTree to supports, removed NearFutureSolar from suggests, added VABOrganizer to supports.

___

ckan compat add 1.12


HebaruSan (approved):
Thanks for the update!

PR #10211, opened 2024-09-13
Update SupplementaryElectricEngines.netkan

Triplchii:
Added SystemHeat, CommunityTechTree, and VABOrganizer to supports. Added NearFuturePropulsion to suggests. Moved Waterfall to recommends.

___

ckan compat add 1.12


HebaruSan (approved):
Thanks for the update!

PR #10217, opened 2024-09-17
force_v for RasterPropMonitor-Core

HebaruSan:
image

The `v` prefix has been dropped on GitHub:

image

image

Now the bot will add it.

### DIFF
--- a/NetKAN/RasterPropMonitor-Core.netkan
+++ b/NetKAN/RasterPropMonitor-Core.netkan
@@ -1,6 +1,8 @@
 identifier: RasterPropMonitor-Core
 $kref: '#/ckan/github/FirstPersonKSP/RasterPropMonitor'
 $vref: '#/ckan/ksp-avc'
+x_netkan_epoch: '1'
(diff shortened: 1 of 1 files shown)


JonnyOThan:
This seems fine, or I could update KSPBuildTools with an option to include a ‘v’ prefix..but I’m not sure I understand all the nuances here.  Would it be more straightforward if the bot just ignored a leading ‘v’ ?

Or would it make more sense to just bump the epoch?

HebaruSan:
> update KSPBuildTools with an option to include a ‘v’ prefix

That's probably a good option to have in general. A lot of mods have a history of `v1.2.3` formatting because GitHub either suggests it or used to suggest it in the repo creation page (I think because a purely numeric Git tag could be ambiguous, or something?), and consistency is always a plus.

JonnyOThan:
Drew is working on some stuff for creating a release from a git tag, so people who use that option should be fine as long as they're consistent.  It's probably better overall but there's a few more things that need to get ironed out.

PR #10216, opened 2024-09-16
Preventing the use of github by CKAN due development cycle constraints.

Lisias:
All of these are available on [archive org](https://archive.org/details/kspckanmods) anyway.

Thankyou.

PR #10218, opened 2024-09-18
Removing github hosting from lisias' mods. Part2.

Lisias:
I missed these two from my last pull requests (see #10216). Trying to run things on working hours is not exactly the best way to do some things...


Lisias:
For the records: this is a follow up for https://github.com/KSP-CKAN/NetKAN/pull/10216

Lisias:
(deleted superfluous message, my apologies - it's after working hours and I'm tired.)

PR #10219, opened 2024-09-19
Update SillyPhotonDrives.netkan

Triplchii:
Added SystemHeat to supports, added HeatControl to suggests


HebaruSan (approved):
Thanks for the update!

PR #10233, opened 2024-09-29
Update BluedogDB.netkan

zorg2044:
Update Bluedog Design Bureau Netkan to add Resurfaced to Reccomends and MagpieMods to conflicts. (note that Magpie considers itself compatible with BDB)


HebaruSan (approved):
Thanks for the update!

PR #10234, opened 2024-10-01
Update BDB netkan

zorg2044:
One more PR with a few more additions.

Moves RealChute from Supports to Recommends

Adds VABOrganizer to Suggests

Adds the 3 official BDB extras patches to Supports

Thanks

### DIFF
--- a/NetKAN/BluedogDB.netkan
+++ b/NetKAN/BluedogDB.netkan
@@ -16,6 +16,7 @@ depends:
   - name: DeployableEngines
 recommends:
   - name: ContractConfigurator
+  - name: Resurfaced
 suggests:
   - name: NearFutureProps
   - name: SCANsat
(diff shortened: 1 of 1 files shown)


HebaruSan:
Hey @zorg2044, a few problems here...

- You don't need those new `supports` relationships. All they would do is list BluedogDB at the bottom of the recommends screen under "supported by". But all three of those modules already hard-depend on BluedogDB, so they'd just do nothing.
- Your branch has conflicts with the master branch. Can you update it and try again?

zorg2044:
Gotcha. Will make a fresh pr

HebaruSan:
Cool, thanks!

FWIW, when I want to start a PR for an existing netkan, I generally go to http://status.ksp-ckan.space/ and click the name, then use the pencil icon on GitHub and make a fresh branch. Trying to re-use the same branch repeatedly is going to be more trouble than it's worth.

PR #10235, opened 2024-10-01
Update BluedogDB.netkan

zorg2044:
Move RealChute to recommends & add VABOrganizer to suggests


HebaruSan (approved):
Thanks for the update! 🚀

PR #10239, opened 2024-10-06
Add KSPCommunityPartModules to CKAN

SofieBrink:
Hereby my pull request to add KSP Community Part Modules to CKAN, It looked to inflate fine locally so i hope the same goes for the automatic checks here. 

Major thanks to @HebaruSan for all the help setting this up and even pushing a patch!

Let me know if anything else is required!


SofieBrink:
Hmm git registers it as the entire file having changed, i added a newline at the end of the file and changed all line endings from CRLF to LF

HebaruSan (approved):
Thanks for the submission!

PR #10256, opened 2024-10-14
Add conflicts with WildBlue play modes' provided names

HebaruSan:
While wokring on the relationship resolver refactor for KSP-CKAN/CKAN#4108, I noticed that the WildBlue play modes and Heisenberg play modes conflicted with the identifiers they provide, but the other mod-specific play modes don't. This is inconsistent and probably makes the resolver work a bit harder to figure these out.

Now all of the play modes conflict with the identifiers they provide.
In addition, "Heisenberg" is replaced by "Airships" in the play mode names for consistency with the main mod.

### DIFF
--- a/NetKAN/Buffalo-PlayMode-CRP.netkan
+++ b/NetKAN/Buffalo-PlayMode-CRP.netkan
@@ -1,33 +1,27 @@
-{
-    "identifier":   "Buffalo-PlayMode-CRP",
-    "name":         "Buffalo CRP Play Mode",
-    "author":       "Angel-125",
-    "$kref":        "#/ckan/github/Angel-125/Buffalo",
-    "$vref":        "#/ckan/ksp-avc/Buffalo.version",
-    "x_netkan_staging": true,
-    "x_netkan_staging_reason": "Make sure that all .txt files for this play mode are installed as .cfg",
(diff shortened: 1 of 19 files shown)


HebaruSan:
```
Installing Airships Classic Stock Play Mode v2.19.2
  Error: 5787 [1] ERROR CKAN.CmdLine.ConsoleUser (null) - Inconsistencies were found:
  * CommunityResourcePack v112.0.1 (needed for: Heisenberg-PlayMode-ClassicStock v2.19.2) conflicts with WildBlue-PlayMode-ClassicStock v1.90.3 (needed for: Heisenberg-PlayMode-ClassicStock v2.19.2)
  Install canceled. Your files have been returned to their initial state.
  INFO:root:Dummy game instance deleted
  ERROR:root:Install of .ckans/Heisenberg-PlayMode-ClassicStock-v2.19.2.ckan failed!
```

I don't know why CRP is being pulled in as a dependency of Heisenberg-PlayMode-ClassicStock. When I try it in GUI, I get a redundant prompt to choose the WildBlue play mode, which of course only works if I choose the classic stock option, but then the install succeeds. Since we're not changing the relationships for Heisenberg, I'm going to chalk this up to quirks of the headless relationship resolver and merge.

image

PR #10258, opened 2024-10-16
KerbalismModularScienceScienceOnly provides Kerbalism-Config

Qazerowl:
From the mod's description:
"Requires Kerbalism (Kerbalism Code, not the Config) , Module Manager
A Science Only Kerbalism Config."

Current implementation does not work properly: Kerbalism-Config is required by Kerbalism, and installing one with KerbalismModularScienceScienceOnly makes Kerbalism display a "duplicate configs detected" warning.


HebaruSan (approved):
Thanks for the fix!

PR #10259, opened 2024-10-16
Relicense ConformalDecals as Apache-2.0

drewcassidy:
Given recent news, I no longer feel comfortable using GPL. All future Conformal Decals releases will be licensed as Apache-2.0 instead

### DIFF
--- a/NetKAN/ConformalDecals.netkan
+++ b/NetKAN/ConformalDecals.netkan
@@ -8,7 +8,7 @@ install:
 identifier: ConformalDecals
 $kref: '#/ckan/spacedock/2451'
 $vref: '#/ckan/ksp-avc'
-license: GPL-3.0
+license: Apache-2.0
 resources:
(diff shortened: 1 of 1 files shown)


techman83:
@drewcassidy - out of curiosity, are you referring to the WordPress / WPEngine events of late? If so, they're arguing over trademarks, a license change doesn't help here.

drewcassidy:
This has nothing to do with Wordpress

techman83:
Ok cool, I'll be leaving someone else to review, but I am curious still (I think license choice is a good thing).

HebaruSan:
Hi @drewcassidy, thanks for the update.
If we merge this now, it would be applied to the **current** release, which is still GPL-licensed if I understand correctly. Is a new release imminent?

drewcassidy:
Oh, no it’ll be a while.

HebaruSan:
OK, then let's close this for now and then we can re-open it when needed. Thanks!

PR #10263, opened 2024-10-18
KerbalStats: unfreeze and point at spacedock

JonnyOThan:
- <https://spacedock.info/mod/3736/KerbalStats>
- <https://github.com/taniwha/KerbalStats>


HebaruSan:
@JonnyOThan please just replace the old `$kref` rather than adding one. We do not want taniwha.org in the downloads list, especially not the first entry.

PR #10268, opened 2024-10-26
Update ChineseModPatches.netkan

tinygrox:
- <https://github.com/tinygrox/KSP_Chinese_Patches>

add Harmony2 to depends

See <https://github.com/tinygrox/KSP_Chinese_Patches/blob/main/Source/KSP_Chinese_Patches/B9PartSwithPatches.cs> for an example of how this mod uses Harmony.


HebaruSan (approved):
Thanks for the update!

PR #10269, opened 2024-10-26
Remove Deferred marking old Parallax versions as incompatible

LGhassen:
Users of Parallax Continued (manual install) will get an error trying to install Deferred "Deferred conflicts with an unmanaged DLL". Remove this version check in the absence of a better option. These old Parallax versions should be mostly phased out by now.


HebaruSan (approved):
Thanks for the update!

PR #10283, opened 2024-11-09
Remove release_status property from all netkans

HebaruSan:
## Motivation

The spec defines `release_status` as `stable`, `testing`, and `development`, in decreasing order of stability.

Currently no code looks at this value, which means we effectively treat all mods as `stable` today (since all mods will be installed for all users).

KSP-CKAN/CKAN#4159 is requesting that we finally follow through on the implicit promise of this property and make available a setting where users can choose their preferred mod stability. This **cannot** work if _Scatterer_, of all things, is telling us that it is a pre-pre-release only suited for the very most adventurous users.

## Changes

Now all active netkans have non-`stable` values of `release_status` removed so we can start populating it with useful, meaningful metadata such as whether a release is marked as a pre-release on GitHub. No future netkan should set `release_status` unless it is actually a completely raw in-development mod whose developer is seeking early feedback (and probably some additional limitations such as a timeline for a full release and not being a dependency of any other mods).

We will probably need to do this in CKAN-meta as well, for users who insist on trying old game versions that are only compatible with historical mod releases indexed with these bogus values of `release_status`.

### DIFF
--- a/NetKAN/BDDMP.netkan
+++ b/NetKAN/BDDMP.netkan
@@ -1,18 +1,12 @@
-{
-    "identifier":   "BDDMP",
-    "$kref":        "#/ckan/spacedock/253",
-    "release_status": "development",
-    "license":      "CC-BY-SA-2.0",
-    "tags": [
-        "plugin",
-        "combat"
-    ],
-    "depends": [
-        { "name" : "DarkMultiPlayer"   },
-        { "name" : "BDArmoryContinued" }
-    ],
-    "install": [ {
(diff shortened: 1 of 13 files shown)


HebaruSan:
The error is just kOSPropMonitor having inconsistent versions across hosts.  Should still be fine to merge.

HebaruSan:
For cleaning up CKAN-meta:

```bash
find . -path ./.git -prune -o -name \*.ckan -print0 | xargs -0 sed -i '/release_status/d'
```

PR #10287, opened 2024-11-11
StarshipExpansionProject: filter Spaces in case they move it

JonnyOThan:
SEP might consider moving the Spaces folder inside the normal location in GameData.  the StarshipExpansionProjectIVA.netkan already can handle this, but we need to exclude it from this one.

### DIFF
--- a/NetKAN/StarshipExpansionProject.netkan
+++ b/NetKAN/StarshipExpansionProject.netkan
@@ -1,5 +1,9 @@
 identifier: StarshipExpansionProject
 $kref: '#/ckan/github/Kari1407/Starship-Expansion-Project'
+install:
+  - find: StarshipExpansionProject
+    install_to: GameData
+    filter: Spaces
 ---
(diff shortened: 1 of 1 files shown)


SofieBrink:
Can confirm we might do this

HebaruSan (approved):
@JonnyOThan feel free to just merge stuff like this when you're happy with it.

JonnyOThan:
> feel free to just merge stuff like this when you're happy with it.
I always wait to make sure I'm not doing something stupid :)

JonnyOThan:
Uhhh I have no idea how or why I reverted this. Might have been a misclick.

HebaruSan:
I think you just made a branch, no PR, not merged to master.

https://github.com/KSP-CKAN/NetKAN/tree/revert-10287-update/StarshipExpansionProject

GitHub has a button to do that somewhere, so a misclick sounds likely.

PR #10301, opened 2024-11-21
Update HideEmptyFilters.netkan

AlexSkylark:
Added suggested mods


HebaruSan (approved):
Thanks for the update!

PR #10305, opened 2024-11-24
More flexible install for CollisionFXReUpdated

HebaruSan:
The folder name for the current stable version of this mod is CollisionFXReUpdated, but the prerelease just uses CollisionFX. Now the netkan handles both.


HebaruSan:
The installation error is just confirming that the current client won't see prereleases. Looks good.

PR #10319, opened 2024-12-06
Coerce StarshipExpansionProject versioning into consistency

HebaruSan:
This mod's last three versions are:

<https://github.com/Kari1407/Starship-Expansion-Project/releases>

- `v.2.1.0`
- `v.2.2.0`
- `v3.0.0-b1` (a pre-release)

It currently has an out-of-order inflation error because the dot after the v is dropped in the pre-release.

Now we put the dot back.
We also have to set the `abstract` because it isn't set on the GitHub repo.


SofieBrink:
Hey I didn’t see this before you merged it but this fix is the wrong way round. The leading . causes issues with the KPBuildTools versioning input so its being dropped from all further releases in the future.

HebaruSan:
Yes, I assumed that.

PR #10346, opened 2024-12-28
Add supports to deferred

Clayell:
Add supports from https://github.com/LGhassen/Deferred?tab=readme-ov-file#mod-compatibility-status.
This PR is currently a draft, as I am waiting for a response relating to an incompatibility with Procedural Fairings. (https://github.com/KSP-RO/ProceduralFairings/issues/59)

There are certain mods that seem to have been always compatible with the public versions of deferred or where deferred itself was seemingly updated to be compatible with these mods, as the mods themselves do not have any patch notes relating to deferred compatibility.
Additionally, if a mod seems to have had multiple patch notes relating to deferred, I went with the most recent one as the supported one.
I have left out Planetshine, as even though it is technically compatible, as LGhassen states it does not a reason to be used with deferred under normal settings, which is what a majority of users will be using.
I also left out the ASET props mod(s) as its compatibility depends on RPM as stated by the mod author.

### DIFF
--- a/NetKAN/Deferred.netkan
+++ b/NetKAN/Deferred.netkan
@@ -21,6 +21,35 @@ conflicts:
     max_version: 3:0.45.1
 depends:
   - name: Shabby
+supports:
+  - name: TexturesUnlimited
+  min_version: 1.6.0.26
+  - name: Parallax
+  min_version: 2.0.8
+  - name: ConformalDecals
+  min_version: 0.2.14
+  - name: B9-PWings-Fork
+  min_version: 3:0.46.0
+  - name: Scatterer
+  min_version: 3:v0.0878
+  - name: EnvironmentalVisualEnhancements
+  min_version: 3:1.11.7.2
+  - name: TUFX
+  - name: Kopernicus
+  - name: Waterfall
+  - name: SimpleAdjustableFairings
+  - name: KerbalKonstructs
+  min_version: v1.8.9.0
+  - name: EngineLightRelit
+  - name: KVVContinued
+  - name: RealSolarSystem
+  - name: RasterPropMonitor-Core
+  min_version: 1:v1.0.1
+  - name: FreeIva
+  min_version: 0.2.19.0
+  - name: Shaddy
+  - name: NeptuneCamera
+  min_version: v4.3
 install:
   - find: zzz_Deferred
     install_to: GameData


SofieBrink:
Could add tundra if you’d like
https://github.com/TundraMods/TundraExploration/releases/tag/7.1.1

Clayell:
> Could add tundra if you’d like
> 
> https://github.com/TundraMods/TundraExploration/releases/tag/7.1.1

Considering that the compatibility isn't mentioned in the Deferred readme, it would likely be better to make Deferred as compatible with Tundra, rather than the other way around.

SofieBrink:
Yeah the relationship exists the other way already. Fair enough

HebaruSan:
Looks like this has stalled out.
@Clayell, if you manage to get this ready to merge, please let us know and we can re-open this.

JonnyOThan:
Yeah this actually doesn't seem like the correct usage of "supports" anyway

Clayell:
Oh, it is actually fine to merge. I was waiting for a response about procedural fairings which never came, although I suppose that could just come with another PR anyway.

> Yeah this actually doesn't seem like the correct usage of "supports" anyway

What is the correct use of supports?

JonnyOThan:
> What is the correct use of supports?

Supports is like a reverse "suggests."  If mod X supports mod Y, then whenever the user installs mod Y they will be prompted to install mod X as well.  I think it's pretty clear that should not happen for several of these.

In other words, "supports" means that this mod could greatly enhance the experience of users of another mod, but it doesn't require that other mod's author's endorsement.  Details here: https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#supports

Clayell:
Hm, that is different than what I expected, but I still don't entirely see why that is a problem. I don't see a reason for deferred to suggest any of these mods to be installed, especially if they could conflict with each other. Conversely, I don't see the problem with those other mods having "supported by deferred" under their relationships tab. What mods do you think it is "pretty clear" for this to not happen to?

JonnyOThan:
Yes it is subtle, which is why we review these PRs and modify or reject them when they do it wrong. There is no relationship metadata for “compatible with.”  That is the default.

Clayell:
I more meant that its very subtle to the user, who might not get the difference between "recommends" and "suggests". Anyway, that's just a minor gripe that I had, what mods do you think should not be listed as "supports" for deferred?

JonnyOThan:
The user sees 3 groups of related mods when installing: recommends, suggests, and supported (by).  They likely don’t care about the distinction between the last two, though the ui does say “not endorsed by chosen mods.”

For example, I don’t think it’s appropriate to add RPM, SAF or KK as “supports” to deferred, because it means the user will be prompted to install deferred when they select one of those mods.  The graphical mods have at least some connection to deferred but at that point you’d might as well just include all graphics mods, so I’d prefer not to.

again: “supports” doesn’t mean “compatible with.”  Mentally replace it with “enhances” (from the Debian example) if that helps.

Clayell:
> They likely don’t care about the distinction between the last two, though the ui does say “not endorsed by chosen mods.”

If mod developers and CKAN take the time to make the distinction between recommend and suggests, the distinction should probably be made more clear than it currently is. (especially as recommend and suggest are just synonyms in english) My point in this would probably require another issue anyway, so I'm willing to drop it here.

> because it means the user will be prompted to install deferred when they select one of those mods

After thinking about this for a few days, I am in agreement with you that this is not great, and should instead be dealt with by using "suggests", albeit for a different reason. If we want to specifically say which mods are compatible with deferred (as those which are not compatible are definitely going to experience visual bugs), then it is easier for the user if the compatible mods are easily listed under deferred's ckan entry, instead of going to an individual mod and checking to see if it is compatible with deferred.

I do wonder which mods (or, which category of mods) use the "supports" relation in general? As far as I've seen, it looks underused.

JonnyOThan:
> If we want to specifically say which mods are compatible with deferred (as those which are not compatible are definitely going to experience visual bugs), then it is easier for the user if the compatible mods are easily listed under deferred's ckan entry, instead of going to an individual mod and checking to see if it is compatible with deferred.

I think you’re still missing something here.  Recommends, suggests, and supports *DO NOT* mean that mods are compatible. All mods are treated as compatible with each other. When we find things that aren’t, we add conflict relationships. We did that with deferred, and then removed them over time when the issues were fixed.

Clayell:
Alright, I understand now. I have changed it to suggests in my fork.
I will wait until the remaining conflict with PF is solved before opening a new PR.

PR #10376, opened 2025-01-19
Create Skopos

Capkirk123:
Releasing, with permission from EggRobin. I checked, CKAN handles the Greek letters just fine.


HebaruSan (approved):
Thanks for the submission!

Clayell:
The title seems to be fine, but the abstract isn't.
image
(https://github.com/mockingbirdnest/Skopos/issues/30)

HebaruSan:
> The title seems to be fine, but the abstract isn't. 

Interesting. The API returns the unicode text, but it gets converted to question marks in the metadata:

- <https://api.github.com/repos/mockingbirdnest/Skopos>
- <https://github.com/KSP-CKAN/CKAN-meta/blob/master/Skopos/Skopos-v1.0.1.0.ckan>

@Clayell, would you mind writing up an issue for that so we don't forget about it?

PR #10386, opened 2025-01-30
MinorPlanetsExpansion: switch to github

JonnyOThan:
The author has been unable to update this mod on spacedock, and the current release has a gamebreaking bug when installed with parallax.  The fix is available on github, so let's just make that the primary download.

> A Word of Warning: SpaceDock/CKAN Still Carries the Initial 1.1.0 Build
DO NOT Use 1.1.0's Parallax Config - THEY ARE UNSTABLE

https://forum.kerbalspaceprogram.com/topic/192848-112x-planet-pack-minor-planets-expansion/

https://forum.kerbalspaceprogram.com/topic/192848-112x-planet-pack-minor-planets-expansion/?do=findComment&comment=4439323

### DIFF
--- a/NetKAN/MinorPlanetsExpansion.netkan
+++ b/NetKAN/MinorPlanetsExpansion.netkan
@@ -1,6 +1,8 @@
+identifier: MinorPlanetsExpansion
+$kref: '#/ckan/github/ExosLab/Minor-Planets-Expansion'
 ---
 identifier: MinorPlanetsExpansion
-"$kref": "#/ckan/spacedock/2383"
+$kref: '#/ckan/spacedock/2383'
 license: MIT
 tags:
 - config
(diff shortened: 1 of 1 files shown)


HebaruSan:
@JonnyOThan, multi-hosting is for the use case where the mod **is** up to date on both servers.
If SpaceDock isn't being updated, we'll need to remove that kref.

JonnyOThan:
Ah I thought it would take the first entry if they didn't match.  It just fails to inflate?  Then I think the only question is whether we should wait for confirmation from @ExosLab before changing this.

HebaruSan:
Right, the point is to catch cases where the author uploads `1.0` to SpaceDock and uses something else like `v1.0` on GitHub. We need to catch those and adjust the netkans as appropriate rather than indexing separate versions that are actually the same.

JonnyOThan:
I'm kinda leaning towards committing this, it's easily undone if they want.

HebaruSan (approved):
Makes sense to me. The author is advising people to use this specific download, and may not be aware that CKAN can also switch. This seems like all upside.

PR #10390, opened 2025-02-03
Add AdvancedPQSTools.netkan

CharonSSS:
Add [AdvancedPQSTools](https://github.com/CharonSSS/AdvancedPQSTools) as dependency for **RSSOrigin-CelestialsPackCore** and **RSSOrigin-TopoRevampConfigs** once [AdvancedPQSTools is added to ckan](https://github.com/KSP-CKAN/NetKAN/issues/10359)

Add [VertexColorMapEmissive](https://github.com/jamespglaze/VertexColorMapEmissive) and [VertexMitchellNetravaliHeightMap](https://github.com/pkmniako/Kopernicus_VertexMitchellNetravaliHeightMap) as dependencies for **RSSOrigin-TopoRevampConfigs**

Add **RSSOrigin-CelestialsPackCore** as dependency for **RSSOrigin-TopoRevampConfigs**

Also, I want to change the title of **_"RSS-Origin CelestialsPack Core"_** to **_"RSS-Origin Core"_**. Its nametag (RSSOrigin-CelestialsPackCore) can remain unchanged to avoid other fussy dependency, recommendation, suggestion and conflicts changes.

The download link of **_"RSS-Origin CelestialsPack Core"_** (now **_"RSS-Origin Core"_**) are changed to https://github.com/CharonSSS/RSS-Origin/releases/download/v1.1.4/RSS-Origin.Core.zip

- <https://github.com/CharonSSS/AdvancedPQSTools>
- <https://github.com/CharonSSS/RSS-Origin>

Fixes #10359.
Fixes #10360.

PR #10391, opened 2025-02-03
Update RSSOrigin.netkan

CharonSSS:


HebaruSan:
Depends on #10390, will combine them...

PR #10404, opened 2025-02-08
Change mods that "support" b9partswitch to "suggest" instead

JonnyOThan:
b9partswitch is a foundational mod that is rarely installed directly by a user's choice.  When trying to craft a .ckan modpack that has a dependency on B9PartSwitch, I was presented with a choice to install these 4 mods when they have nothing to do with my pack.  This isn't what "supports" is meant to be used for.  If the relationship isn't strong enough to be a dependency or recommendation, then "suggests" is appropriate.


JonnyOThan:
@HebaruSan you good with this?  @JadeOfMaar FYI too (feel free to upgrade that to recommends/depends too if you want)

HebaruSan (approved):
Sure, thanks for taking care of this.

PR #10409, opened 2025-02-16
Add KSPCommunityFixes conflict to PersistentRotationUpgraded

HebaruSan:
See linuxgurugamer/PersistentRotation#13, this mod depends on exploiting a stock bug that is fixed by KSPCommunityFixes, so when both are installed, PersistentRotationUpgraded doesn't work.

@linuxgurugamer acknowledged the problem and said a fix is not currently in the works:

- <https://forum.kerbalspaceprogram.com/topic/197082-ckan-the-comprehensive-kerbal-archive-network-v1332-laplace-ksp-2-support/?do=findComment&comment=4443730>

Now they're marked as conflicting.

### DIFF
--- a/NetKAN/PersistentRotationUpgraded.netkan
+++ b/NetKAN/PersistentRotationUpgraded.netkan
@@ -13,6 +13,7 @@ provides:
   - PersistentRotation
 conflicts:
   - name: PersistentRotation
+  - name: KSPCommunityFixes
 depends:
   - name: ToolbarController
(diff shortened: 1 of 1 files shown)


JonnyOThan:
Gross. We should probably just disable that KSPCF patch when persistent rotation is installed, at least until that bug is fixed.  It feels like that should be done in persistent rotation since undoing it should only be done after it’s fixed.  @linuxgurugamer does that sound reasonable?

PR #10413, opened 2025-02-22
Remove release_status property from all netkans

HebaruSan:
#10283 removed `release_status` from all netkans that set it to anything other than `stable` so they wouldn't be treated as prereleases and KSP-CKAN/CKAN#4260 could add support for prereleases.

Turns out that setting it to `stable` causes problems as well, see #10397 for an example. The author of InfernalRoboticsNext reported that his mod had a prerelease indexed as a normal release in CKAN because of this.

Now they're all purged, including frozen in case they're unfrozen in the future or used as templates by other authors.


HebaruSan:
https://github.com/KSP-CKAN/CKAN-meta/compare/954dc833102a04d101996661356c401756426ef5...d8b703a6b997ce08856b704bb7b912cbfb9025f7

PR #10439, opened 2025-03-12
Add MM dependencies to several mods where they are missing

JonnyOThan:
- ExtraDockingPorts
- PrakasaAeroworks
- SpaceY-Corp
- WarpDrive
- kOS


HebaruSan:
Please leave Astrogator out of this!

JonnyOThan:
Huh...I guess it's only needed if you also have RPM?  Which should then include the mm dependency.  Fine.  I did spot check most of the rest

PR #10459, opened 2025-03-28
Add the metadata files for Parallax Continued

Gameslinx:


HebaruSan:
First suggestion: In the repo, click the gear icon in About and paste the forum link in the website field. That way we can pull it automatically without having to maintain it in the .netkan.

HebaruSan:
The `name` fields are all the same. These are what the user will see, so presumably they should say things like "scatter", "textures", etc. instead.

PR #10477, opened 2025-04-05
KSPBurst: add compiler as a new package

JonnyOThan:
I goofed and didn't include the compiler in the last KSPBurst release.  But really, this is how things should have been organized in the first place.  The compiler package changes rarely so it should be a standalone module.

- <https://github.com/KSPModdingLibs/KSPBurst>

See #8418 and #8419 for history. 

More history: 

* https://github.com/KSP-CKAN/NetKAN/issues/8688 (discussion of complex RO/RP1 install flow)
* https://github.com/KSP-CKAN/NetKAN/pull/8690 (changed KSPBurst-Lite abstract) <-- I think this really is the problem I'm trying to solve.  "same job but more slowly" makes it sound like KSPBurst-Lite improves performance.  It doesn't.
* https://github.com/KSP-CKAN/NetKAN/pull/8701 - RP-1 express install
* https://github.com/KSP-CKAN/NetKAN/pull/8706 - added KSPBurst-Full so mods could require it


HebaruSan:
@JonnyOThan, is this right? It just dumps a ZIP file into GameData?
image

JonnyOThan:
> @JonnyOThan, is this right? It just dumps a ZIP file into GameData?
Yep.  It's a little weird, but that's how it was originally set up.  The idea seems to be to avoid having the game load those files; the mod extracts the zip file into ksproot/plugindata.  But really it should have just been in the mod's plugindata in the first place.  I'll probably get around to changing that eventually but I wanted to fix the broken 1.5.5.2 release first.

HebaruSan:
Won't users of the Lite variant be upset at losing it?

JonnyOThan:
> Won't users of the Lite variant be upset at losing it?

No, the two packages in the 1.5.5.2 release are actually identical, and the only difference between the ones in the past was the presence of the compiler package.  While you can install and run *only* KSPBurst-Lite, it actually means that you're not using any of the burst features and your game isn't as fast as it could be.  There's no point.  This setup means that everyone is forced to download the compiler, but only when it changes rather than for every update to KSPBurst itself.

HebaruSan:
Honestly the old system kind of seems fine in retrospect. What problems did you perceive with it?

JonnyOThan:
> What problems did you perceive with it (other than a broken release)?

1. Users are prompted to choose between KSPBurst and KSPBurst-Lite and they should never pick the latter (this is fixable in metadata)
2. 300mb download every time KSPBurst is updated

HebaruSan:
> they should never pick the latter

Why's that? I'm assuming that .netkan was created for _some_ reason.

JonnyOThan:
> Why's that? I'm assuming that .netkan was created for some reason.

Because it doesn't make their game any faster, but it's required for any mod that uses KSPBurst.  It's not like 300mb is that big either; I'd be fine with just collapsing the whole thing down to one package but if there's two then it should be the compiler vs plugins.

HebaruSan:
> 2. 300mb download every time KSPBurst is updated

> It's not like 300mb is that big either;

Just enjoying the pivot. :)

JonnyOThan:
Well the pivot would be to collapse it down to 1 package.  The change I have here is *nearly* optimal, except for the hypothetical automated build case.  We could always make adjustments later when necessary.

JonnyOThan:
I think I'm liking this idea better actually:

> Or maybe KSPBurst should just be the compiler and it depends on KSPBurst-Lite, which is the plugins?

HebaruSan:
OK, so previously users could choose whether to pay a 300 MB download to speed up the game, and now they're forced to.

It looks like the plugins-only variant has a fairly significant number of downloads (~15% or so in the previous release):

```
$ curl -s https://api.github.com/repos/KSPModdingLibs/KSPBurst/releases | jq -rM '.[]|.assets[]|[.download_count,.name]|@tsv'
3097    KSPBurst_1.5.5.2.zip
110     KSPBurst_1.5.5.2_plugins_only.zip
5       KSPBurst_Compiler.zip
174902  KSPBurst_1.5.5.1.zip
28939   KSPBurst_1.5.5.1_plugins_only.zip
192     KSPBurst_1.5.5.0.zip
62      KSPBurst_1.5.5.0_plugins_only.zip
274     KSPBurst_1.5.4.1.zip
161     KSPBurst_1.5.4.1_plugins_only.zip
1091    KSPBurst_1.5.4.0.zip
723     KSPBurst_1.5.4.0_plugins_only.zip
3116    KSPBurst_1.5.0.0.zip
2832    KSPBurst_1.5.0.0_plugins_only.zip
```

So at least some users prefer the smaller download (or don't understand the choice they're making).

JonnyOThan:
> So at least some users prefer the smaller download (or don't understand the choice they're making).

I'd bet that it's the latter.  I did not know until today that KSPBurst-Lite actually does nothing at all by itself.

HebaruSan:
> KSPBurst-Lite actually does nothing at all by itself

Well, doesn't it at least satisfy the dependency and thereby allow Burst-enabled mods to be installed, even if the compiler isn't there?

HebaruSan:
Anyway, I'm fine with whatever direction you think it makes sense to take this mod. Go nuts. 🚀

JonnyOThan:
```
Inflating NetKAN/KSPBurst.netkan
  Error: 924 [1] FATAL CKAN.NetKAN.Program (null) - Could not cast or convert from System.String to CKAN.ModuleRelationshipDescriptor.
  ```
  that's a new one..

**Comment by JonnyOThan** [comment]
OK I think this works?  Any mod depending on KSPBurst should get the full thing, but modders can optionally depend on KSPBurst-Lite if they want to leave it up to the user.  We kind of already had that where the modder could depend on KSPBurst-Full but I'd prefer that the default is the fast one and you have to opt out if you want to save disk space.

**Comment by HebaruSan** [comment]
- I don't like one mod's name being the same as another mod's identifier
- I don't like relying on `replaced_by`, it has not always been super reliable and the old mod will always still be there in the mod list waiting for someone to install it
- A manual install would be autodetected as the `KSPBurst` module based on the DLL name, and after the user "upgrades" to it, they'd then have to "replace" to get the actual intended install? Sounds very clunky.
- `KSPBurst-Lite` could probably use a better `name` so it doesn't look like we have different variants of the same mod installed at once. What if the names were `KSPBurst Plugins` and `KSPBurst Compiler`?
- Could we document somewhere that the compiler module doesn't actually install the compiler, but just deposits it somewhere for the plugin module to actually install?

**Comment by JonnyOThan** [comment]
> I don't like one mod's name being the same as another mod's identifier

are there downsides of this other than the autodetection problem?  It seems like maybe this is the least bad option (use KSPBurst for the compiler, and -lite for plugins).  

> KSPBurst-Lite could probably use a better name so it doesn't look like we have different variants of the same mod installed at once. What if the names were KSPBurst Plugins and KSPBurst Compiler?

love it

> Could we document somewhere that the compiler module doesn't actually install the compiler, but just deposits it somewhere for the plugin module to actually install?

within the Netkan? Sure. I hope to change this eventually but who knows when.

**Comment by HebaruSan** [comment]
> are there downsides of this other than the autodetection problem? It seems like maybe this is the least bad option (use KSPBurst for the compiler, and -lite for plugins).

I'm thinking simple scenarios such as, "User says, 'I installed KSPBurst'". Which module did they install?

**Comment by JonnyOThan** [comment]
Oh sorry I missed that you meant the display name.  So yeah using KSPBurst for the compiler would be better.

Im rethinking the display name change. I want the compiler package to be the one people identify as the “main” mod and they’re not going to understand what “compiler” means anyway.

**Comment by HebaruSan** [comment]
Trying to evaluate some scenarios...

- **Manual installation**: CKAN will autodetect `KSPBurst`. Upgrading will cause the compiler to be installed and the plugins to be pulled in as a dependency. ✔️ 
- Installing a mod that depends on `KSPBurst`: The choice prompt goes away, and instead you get both the plugins and compiler, which is what the current maintainer prefers. ✔️ 
- **Upgrading from having just a previous version of KSPBurst**: The new compiler will be installed and the plugins will be pulled in via the dependency. ✔️ 
- **Upgrading from having just Lite**: Since it no longer provides KSPBurst, it's possible that some other mod would have an unmet dependency. I don't think that would automatically trigger installation of the missing identifier, so that'll probably put the registry into an inconsistent state till the user manually chooses to install KSPBurst. ❌ 

That 4th one is the thing to worry about. The 15% of users who installed Lite may end up with broken installs. There should be an artifact attached to the checks containing metadata that you could use to test this.

**Comment by JonnyOThan** [comment]
OK I tested it out with RealAntennas.  I got this message when trying to update (at least with 1.35.3.25094; I should probably test the public version as well).

```
Inconsistencies were found:
* Unsatisfied dependency KSPBurst needed for: RealAntennas v2.8.0.0
Error during installation!
```

HebaruSan:
To solve that, can the dependency be two-way? They're only separate so they can be updated independently, right?

JonnyOThan:
> To solve that, can the dependency be two-way? They're only separate so they can be updated independently, right?

Ideally no, because this method does allow a modder to decide to allow users to select between full vs plugins, and it allows for build systems to only install the plugins.

Hm, at least 1k downloads for each of those mods is from the netkan bot right?  So the actual percentage is less.  And that also assumes they are disjoint sets.  I think I'm OK with this?

HebaruSan:
> Hm, at least 1k downloads for each of those mods is from the netkan bot right?

No, the bot caches its downloads.

JonnyOThan:
OK I confirmed that CKAN 1.33.2 has the same behavior when trying to update.  I think this is ready?

HebaruSan:
Is this actually true?

> Further, a user (or automated build system) may choose to install just the plugins to save space (but burst will not be functional).

If the user installs a mod depends on `KSPBurst`, then they're forced to install both, right?

HebaruSan:
Shoot. Missed scenarios:

- **Reinstall KSPBurst after this PR changes the existing metadata**: Latest version of both mods should get installed. ✔️ 
- **Reinstall KSPBurst-Lite after this PR changes the existing metadata**: Same as the upgrade scenario, I think. Inconsistent registry due to removal of `provides`. ❌ 

Just noting these for the record (and in hopes of remembering these cases next time). No significant change from previously.

HebaruSan (commented):
Feel free to merge if you're prepared to support the users who get confused.

JonnyOThan:
> If the user installs a mod depends on KSPBurst, then they're forced to install both, right?

Yes, and that’s intentional. If a mod author wants to, they can configure their metadata to allow either one at which point the user can choose.

I’ll probably merge this tomorrow since I won’t have much time for support tonight.

PR #10476, opened 2025-04-04
Fixes #10474: KerbinSideRemasteredGAP: add dependencies on StockalikeStructures and KerbinsideCore

JonnyOThan:
Per author's request on the forum: https://forum.kerbalspaceprogram.com/topic/197082-ckan-the-comprehensive-kerbal-archive-network-v1332-laplace-ksp-2-support/?do=findComment&comment=4454740

However...  Kerbinside Core is only marked as compatible with KSP 1.5.1.  In order for someone to install your mod from CKAN, they will have to get CKAN to install that (ideally by force-installing it from the versions tab, but some people change ckan's compatibility settings to do this which often results in them unintentionally getting old versions of dependencies that break their game).

KerbinSideRemasteredGAP already depends on KerbinSideRemastered.  Is it even feasible to install both KerbinSideRemastered and Kerbinside Core together?

As a policy, we do not change KSP version compatibility for a mod without the author's consent.  So there's a few options:

Get [@Ger_space](https://forum.kerbalspaceprogram.com/profile/167349-ger_space/)'s blessing to change the version compatibility of Kerbinside Core
Find some other way to get the assets you need
Adopt the mod yourself (but you'd still need to get permission from the author since it's ARR)

ckan compat add 1.5.1

PR #10478, opened 2025-04-05
KerbinSideRemasteredGAP: rename internal GameData folder

JonnyOThan:
hopefully this is a temporary fix

Xref: #10476

___

ckan compat add 1.5

### DIFF
--- a/NetKAN/KerbinSideRemasteredGAP.netkan
+++ b/NetKAN/KerbinSideRemasteredGAP.netkan
@@ -18,5 +18,3 @@ install:
   - find: ContractPacks
     install_to: GameData
     filter: Config Replacements
-  - find: ForParallaxContinued/GameData/ContractPacks
-    install_to: GameData
--- a/NetKAN/KerbinSideRemasteredGAP.netkan
+++ b/NetKAN/KerbinSideRemasteredGAP.netkan
@@ -18,5 +18,5 @@ install:
   - find: ContractPacks
     install_to: GameData
     filter: Config Replacements
-  - find: ForParallaxContinued/GameData
-    install_to: ForParallaxContinued/_GameData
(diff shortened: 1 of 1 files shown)


JonnyOThan:
Hm, there are other files in Config Replacements that shouldn't be filtered out.

HebaruSan:
Which ones? I only see stuff that should be filtered out.

JonnyOThan:
Nothing in `Config Replacements` is "active" but a user will need it to do some manual reconfiguring.  We just need to get rid of the GameData folder so the bot is happy.  Hopefully we can help the author find a better way to do this in the long run.

HebaruSan:
I don't think we should make the user worry about that.
If any of those are important, we can break out optional modules.

HebaruSan:
Wait. Are the contents of `Config Replacements` intended to overwrite files _from other mods_??

@caerfinon, what is the deal with that folder?

HebaruSan:
Oh yeah. Confirmed in `0-README.txt`:

```
// The files here are replacement configurations for other mods. 
// You will need to copy them to the directory specified to enable them
// Once copied to the new directory remove .txt extension (delete any previous copy of the file already there) 
```

That's not gonna work out...

@caerfinon, do you know how to use ModuleManager? If not, this would be a very good time to learn...

JonnyOThan:
Perhaps we should just exclude that entire folder. The user can use the “open zip file” feature if they need to get to it.

HebaruSan:
That was my thinking as well (hence the current diff).

HebaruSan:
Replaced by https://github.com/KSP-CKAN/CKAN-meta/commit/51d36b40679bfa7c068d4dff566bcb664988e2c8.

PR #10481, opened 2025-04-06
ASETAvionics: transfer to FirstPersonKSP org

JonnyOThan:
- <https://github.com/FirstPersonKSP/ASET_Avionics>
- <https://spacedock.info/mod/3239/ASET%20Consolidated%20Avionics%20Pack>

### DIFF
--- a/NetKAN/ASETAvionics.netkan
+++ b/NetKAN/ASETAvionics.netkan
@@ -9,7 +9,6 @@ author:
   - alexustas
   - StoneBlue
   - JonnyOThan
-abstract: IVA props for aircraft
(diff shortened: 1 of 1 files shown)


HebaruSan:
The `abstract` is also pulled from the repo description, which is a nicer place to put it:

image

JonnyOThan:
I was wondering why I'd never seen this error before....I just assumed it was a new requirement (but then again, it would have started going off for all mods..)

HebaruSan:
For the version in `master`, the abstract is coming from SpaceDock, because the version numbers match up.

PR #10493, opened 2025-04-11
WildBlueTools Retirement

Angel-125:
Updates Netkan entries for Wild Blue mods to end their dependency on Wild Blue Tools. Also indirectly removes the need for WBI Play Modes; if you have Community Resource Pack installed, you get CRP resources. If you have Classic Stock Resources installed, you get Classic Stock resources. If you have both installed, you get both.

- <https://github.com/Angel-125/DSEV>
- <https://github.com/Angel-125/Airships>
- <https://github.com/Angel-125/MOLE>


HebaruSan (approved):
Thanks for the updates!

PR #10521, opened 2025-04-28
Add netkan file for ISO 7010 Decals

DIYLabsED:
- <https://github.com/DIYLabsED/ksp-iso-7010-decals>
- <https://spacedock.info/mod/3875/ISO%207010%20Decals>


HebaruSan:
> Anything else that I need to change?

Nothing else at the moment, but there might be more later once the validation script gets past the version number thing. Let us know when it's ready!

HebaruSan (approved):

PR #10538, opened 2025-05-11
Multi-hosting for APotato

HebaruSan:
KSP-CKAN/CKAN-meta#3355 included a GitHub repo, but KSP-CKAN/NetKAN#10537 did not.
Now the netkan has both it and SpaceDock.


averageksp:
Thanks for doing that, I couldn't get it to work 👍

PR #10542, opened 2025-05-11
Add GitHub hosting or Munk's Fixes

averageksp:
# Changes
Added GitHub hosting


HebaruSan:
This isn't being packaged the same way on the different servers, so multi-hosting isn't possible.

PR #10572, opened 2025-05-30
Update NightShift requirements

coldrifting:
Version 3.0 of NightShift now requires Shabby to be installed.
By the way, is there a way for this requirement to be version dependent, or is it better to just apply this to all versions?

### DIFF
--- a/NetKAN/NightShift.netkan
+++ b/NetKAN/NightShift.netkan
@@ -14,3 +14,5 @@ tags:
   - graphics
 conflicts:
   - name: LightsOutRelit
+depends:
+  - name: Shabby


averageksp:
> By the way, is there a way for this requirement to be version dependent, or is it better to just apply this to all versions?

I think you do not need to, if you look into [here](https://github.com/KSP-CKAN/CKAN-meta/tree/7eabf638fa0fdf1ba02376c5a7d1ca00e7940ef3/NightShift) you will see that each version of your mod has it's own .ckan file with it's own requirements. So, if someone installs an older version of your mod the person won't get asked to download some of the missing dependency _as far as I am aware of_ :)

HebaruSan:
> By the way, is there a way for this requirement to be version dependent, or is it better to just apply this to all versions?

@averageksp is correct that changing the .netkan file only affects the current / most recent version, but there also is a way to have a change like this take effect as soon as a release is made, without any pull requests:

<https://github.com/KSP-CKAN/CKAN/wiki/Adding-a-mod-to-the-CKAN#internal-ckan-files>

To sum up, you could put a file called `NightShift.ckan` into your mod's ZIP file containing this:

```yaml
depends:
  - name: Shabby
```

... and that would automatically be added to the mod's metadata in CKAN. Then when you release a new version with different dependencies, all you would have to do is edit the copy of that file in the mod's ZIP, and it would be indexed that way from the start.

Issue #10253, opened 2024-10-14
[Mod]: Remove temporarily Celestial Harmony from CKAN due to many internal changes


ProximaCentauri-star:
### Mod Name

Celestial Harmony

### Author(s)

ProximaCentauri-star

### License

All Rights Reserved

### Homepage link

https://github.com/ProximaCentauri-star/Celestial-Harmony

### Download link

https://github.com/ProximaCentauri-star/Celestial-Harmony/releases/

ProximaCentauri-star:
Right now, dependencies were changed and the most is currently listed as a scatterer sunflare config for some reaons

HebaruSan:
Hi @ProximaCentauri-star, thanks for getting in touch. If it's simply a matter of the dependencies changing, we can update the metadata to install the correct dependencies. Would you like to try that option?

HebaruSan:
> the mod is currently listed as a scatterer sunflare config for some reasons

The reason is that `GameData/CelestialHarmony/Scatterer/Config/Sunflares/Sunflares.cfg` contains a `Scatterer_sunflare` config node, which means this mod defines its own sunflare:

![image](https://github.com/user-attachments/assets/0c5f4e74-2140-4f1b-a94d-9139e99ae20c)

ProximaCentauri-star:
> > the mod is currently listed as a scatterer sunflare config for some reasons
> 
> The reason is that `GameData/CelestialHarmony/Scatterer/Config/Sunflares/Sunflares.cfg` contains a `Scatterer_sunflare` config node, which means this mod defines its own sunflare:
> 
[...]

PR #10580, opened 2025-05-31
Add new CelestialHarmony submodule

HebaruSan:
There are two assets attached to the latest release of CelestialHarmony:

Image

Image

As a default, the bot has chosen to use the first one in the list, `_Chaos.zip`, but `_Harmonia.zip` is meant to be the core module.

See also comment from author at https://github.com/KSP-CKAN/NetKAN/issues/10253#issuecomment-2925159856.

Now `CelestialHarmony` has `asset_match/Harmonia`, and a new `CelestialHarmony-Chaos` module is added with `asset_match/Chaos`.


ProximaCentauri-star:
Thanks a bunch!

PR #10585, opened 2025-06-05
GitHub only hosting for KheledrosSystem

averageksp:
## Changes
- Added GitHub hosting to the KheledrosSystem Mod as per Author's Request
- Removed SpaceDock hosting


averageksp:
GitHub has some outages at the moment :/

PR #10589, opened 2025-06-06
Add KAC to recommends of LTP

Clayell:
The "Add Alarm" button for this mod requires KAC to be installed. It'll grey itself out if KAC isn't installed.

(this is from my [LTP remake](https://github.com/Clayell/LunarTransferPlanner), but the original still uses KAC in [about the same way](https://github.com/KSP-RO/LunarTransferPlanner/blob/master/LunarTransferPlanner.cs#L453-L467))
image

### DIFF
--- a/NetKAN/LunarTransferPlanner.netkan
+++ b/NetKAN/LunarTransferPlanner.netkan
@@ -7,3 +7,5 @@ license: MIT
 tags:
   - plugin
   - information
+recommends:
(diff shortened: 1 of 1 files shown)


HebaruSan (approved):
Makes sense to me, thanks!

FYI to @RCrockford. I'm going to merge this; let us know if you want it reverted.

PR #10608, opened 2025-06-27
Better manual inflate action

HebaruSan:
## Motivation

See KSP-CKAN/NetKAN-Infra#355; the manual inflate workflow can be improved, and that PR created a way to do that.

## Changes

Now the `manual_inflate` workflow uses a new `manual_inflate` action to run the new `inflate-netkan` command in the `kspckan/netkan` container.


HebaruSan:
It works!

image

PR #10702, opened 2025-08-25
Add AnimatedAttachment to KerbalJointReinforcementContinued conflicts

siimav:
This mod is known to prevent KJR from working.


HebaruSan (commented):
Thanks for the update! 🦖

PR #10714, opened 2025-08-29
KerbalPowersNaval updates for #10713

Spartwo:
sets up naval like interstellar and adds to kpcore suggestions


HebaruSan (approved):
Thanks for the updates.

PR #10717, opened 2025-08-31
Update KC metadata

AMPW-german:
Adding generic identifiers for KC configs and config expansions


AMPW-german:
It should be fixed now, I hadn't seen the comma in the release and forgot to update the main branch for the config.

AMPW-german:
The "KC-EL-Default" folder is in a GameData folder, I only uploaded an outdated zip when I made the release which contained the old file structure ("GameData/KerbalColonies/FacilityConfigs/EL-Default")

averageksp (approved):

PR #10718, opened 2025-08-31
Handle bundled ScaledDecorator dep in Promised Worlds

averageksp:
Soon the ScaledDecorator dependency will be bundled in Promised Worlds
Don't merge this for now.

- <https://github.com/PromisedWorlds/PromisedWorlds>


averageksp:
@HebaruSan can I merge this? v1.1 is about to be released

PR #10721, opened 2025-09-08
fishy anti missile spam: point at newer version on spacedock

JonnyOThan:
I got in touch with the current maintainer on discord and got permission:

> Neon — 10:00 AM
Hello there, Neon here Im planning to keep updating Fishy weapons, Fishy anti missile spam is outdated and incompatible with fishy weapons due to modules being build on Fishy anti missile spam. Sadly the original author is no longer with us due to personal issues in the background, Im currently in control of Diamant Gotts stuff and planning on updating. Diamant Gott tried to list the original Fishy weapons on CKAN, but it never fully worked with CKAN for whatever reason. This discord account Im currently using was once his not to get confused with the naming originally it was Diamant Gott#9255 before I took over (if you look on my discord badge).
JonnyOThan — 10:02 AM
thanks for the info.  Sorry I should have made it clear - I'm one of the CKAN metadata maintainers so I can help with that.  Off the bat, I think it would be an improvement if we just made the ckan entry for fishy weapons refer to "FAMS continued" instead of the older one, right?
Neon — 10:06 AM
sure do you mean update Fishy anti missile spam to FAMS continued or make a new entry for fishy weapons as FAMS continue?
JonnyOThan — 10:12 AM
I'd probably just update the current entry
if it's intended to be a newer version, and it's by the same author/maintainer anyway, that's usually cleaner
Neon — 10:15 AM
alright thats great, Im currently still holding onto a new patch but still needs a lot of work


JonnyOThan:
Ugh.  These things are a mess.  Maybe I shouldn't have tried...

I just noticed that the FAMS Continued version has 2 .version files in it and neither one is really correct.

PR #10424, opened 2025-03-02
Remove SpaceDock hosting for Lt_Duckweed's mods

HebaruSan:
## Problem

@jamespglaze has reported difficulties uploading to SpaceDock and wishes to have his mods indexed from GitHub only.

- <https://github.com/jamespglaze/BetterKerbol>
- <https://github.com/jamespglaze/QuackPack>
- <https://github.com/jamespglaze/VertexColorMapEmissive>
- <https://github.com/jamespglaze/VertexHeightOblateAdvanced>

## Changes

Now this author's mods are retrieved only from GitHub.

PR #10740, opened 2025-09-14
Drop SpaceDock support for PromisedWorlds

averageksp:
SpaceDock is very slow and mostly just cancels your upload.
This Pull request removes SpaceDock support for PromisedWorlds.

### DIFF
--- a/NetKAN/PromisedWorldsCore.netkan
+++ b/NetKAN/PromisedWorldsCore.netkan
@@ -1,11 +1,6 @@
 identifier: PromisedWorldsCore
 $kref: '#/ckan/github/PromisedWorlds/PromisedWorlds/asset_match/Core'
 $vref: '#/ckan/ksp-avc'
----
-identifier: PromisedWorldsCore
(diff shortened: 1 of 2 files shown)


JonnyOThan:
@averageksp  This doesn't make sense.  CKAN would try the download hosts in the order they're listed in the .netkan file.  We know that github has been timing out large downloads recently.  Is there documented issues with spacedock too?

HebaruSan:
SpaceDock has had issues with uploading for a quite while now, see KSP-SpaceDock/SpaceDock#464.

It's mostly large mods that are affected. Various authors have previously given up on being able to upload to SpaceDock, see #10006, #10424, #10495.

averageksp:
> @averageksp This doesn't make sense. CKAN would try the download hosts in the order they're listed in the .netkan file. We know that github has been timing out large downloads recently. Is there documented issues with spacedock too?

Well multi-hosting was great, but it always took a few hours to upload which wasn't ideal when trying to make an update. GitHub only took 10minutes to upload iirc

PR #10747, opened 2025-09-16
Unfreese KittopiaTech and point to new repo

HebaruSan:
Rolling back #6624 because the new repo has releases now.

- <https://github.com/Kopernicus/KittopiaTech>
- <https://github.com/Kopernicus/KittopiaTech-Legacy>


HebaruSan:
Oops, looks like this never got finished, since this post was a year after the release:

image

PR #10750, opened 2025-09-18
Add CatEarNavballFrame

levitato5594:
This is a highly important and serious mod that will impact billions of KSP players.


levitato5594:
wow I'm dumb

PR #10751, opened 2025-09-18
Add CatEarNavballFrame

levitato5594:
With the netkan in the correct folder this time.

Note: I do not have a forum thread.


JonnyOThan:
@levitato5594 what does this have to do with 1ThousandSpecialPlanetPack?

levitato5594:
Ah alright, that can be removed then. The navball mod was made for Kram, the creator of that planet pack, but there is no in game relationship.

PR #10855, opened 2025-12-16
Update SpaceY-Lifters.netkan

linuxgurugamer:
Added line to support PatchManager


HebaruSan:
@linuxgurugamer can you explain your thinking here please? If I'm installing PatchManager, why would I want to be prompted about SpaceY-Lifters?

PR #10870, opened 2025-12-23
Update Kopernicus.netkan

R-T-B:
Add KSPTextureLoader dependency


HebaruSan (approved):
Thanks for the update! 🐎

PR #10920, opened 2026-01-27
Set x_netkan_check_parent_downloads to false for TIP

Clayell:
Depends on https://github.com/KSP-CKAN/NetKAN-Infra/pull/361


Clayell:
Leaving this comment to remind myself to ask ballisticfox if they want Advanced PQS Tools to have this setting set to false as well (since its forked off RSS)
image

PR #10922, opened 2026-01-29
Fix BD TURD installing to root gamedata

Spartwo:


HebaruSan (approved):
Thanks for the update! 🤹

PR #10938, opened 2026-02-11
Use non-source asset for EasyISP

Tempus-superest:
- <https://github.com/Tempus-superest/EasyISP>

This pull request updates the `EasyISP.netkan` metadata file to correct the source location and remove "use_source_archive: true". 

The initial GitHub release did not have proper release tags set up, and now releases occur with properly versioned tags. This was preventing CKAN from displaying download counts in the CKAN UI.

Spacedock references were added for user benefit, but Spacedock should not be the authoritative source. GitHub release automation will upload to Spacedock


HebaruSan (approved):
Thanks for the updates! 🕸️

PR #10962, opened 2026-03-05
Fix HomeWorldFallback depends

Oprissmian:
This mod makes it so that if planet packs using Kopernicus' HomeWorldName option are uninstalled that the HomeWorldName is reset to Kerbin.


HebaruSan:
This was added in #10758, but the Kopernicus dependency was missed. We can use this PR to add that.

PR #10975, opened 2026-03-13
Add V1.0.0  KOSpopupStop

AdmiralRadish:
KOSpopupStop

- <https://github.com/AdmiralRadish/KOSpopupStop>


HebaruSan:
@AdmiralRadish, I'm trying to figure out why this is a separate mod instead of a bugfix or enhancement submitted to kOS. It sounds like it's just tweaking one thing very specific to that one mod. Can you explain what it is to help clarify that?

PR #11000, opened 2026-03-30
Make ContractConfiguratorExtended conflict with original ContractConfigurator

HebaruSan:
#10844 added a new fork of ContractConfigurator, without marking it as such.

- <https://spacedock.info/mod/4073/Contract%20Configurator%20Extended>

Now it ~~provides and~~ conflicts with that mod.
(`provides` removed because we probably don't want anyone installing this by accident.)

Also the tags are updated to match.


defoma:
I saw that, I'm going to update CCE into an addon for the RO CC.

PR #11002, opened 2026-03-31
IR-Canadarm

meirumeiru:
IR-Canadarm, new mod

- <https://github.com/meirumeiru/IR-Canadarm>


HebaruSan (approved):
Thanks for the submission! 🦍

PR #11005, opened 2026-04-03
Add Kerbalism NFE Nuclear Patch

bimo1d:
Adds a NetKAN entry for KerbalismNFEFRpatch.

KerbalismNFEFRpatch is a compatibility patch between Kerbalism and Near Future Electrical.
It restores stable background electric generation for NFE fission reactors on unloaded vessels and keeps reactor-related resource flows synchronized in background simulation.

- <https://github.com/bimo1d/Kerbalism-NFENuclearPatch>


HebaruSan:
Hi @bimo1d, thanks for the submission!

CKAN will pull the version from the tag of this release, which is `beta`:

image

Looks like you wanted that to be v0.1a instead:

image

If that's correct, then I'd recommend editing the release to set the tag to v0.1a.

averageksp (approved):

PR #11007, opened 2026-04-03
Add BurstPQS dependency to AdvancedPQSTools

ballisticfox:
The latest version of AdvancedPQSTools now depends on BurstPQS.

Replaces KSP-CKAN/CKAN-meta#3384.


HebaruSan (approved):
Thanks for the update! 🕸️

PR #11021, opened 2026-04-11
remove install and relationship metadata from RPM

JonnyOThan:
This is now in the internal .ckan files of the prerelease versions, and the current stable releases won't be re-inflated so they should be fine

- <https://github.com/FirstPersonKSP/RasterPropMonitor>


JonnyOThan:
@HebaruSan something weird is going on here.  As far as I can tell, the .ckan files in RPM are set up correctly:

https://github.com/FirstPersonKSP/RasterPropMonitor/blob/master/GameData/JSI/RasterPropMonitor/RasterPropMonitor-Core.ckan

https://github.com/FirstPersonKSP/RasterPropMonitor/blob/master/GameData/JSI/RPMPodPatches/RasterPropMonitor.ckan

The error in RasterPropMonitor-Core seems to indicate that netkan.exe autogenerated an install stanza:

>   Error: 5600 [1] FATAL CKAN.NetKAN.Program (null) - Bad metadata detected for module RasterPropMonitor-Core 1.0.2: No files found matching find="RasterPropMonitor-Core" to install!

And the produced RasterPropMonitor.ckan seems to be using the install stanza from the internal RasterPropMonitor-Core.ckan:

https://github.com/KSP-CKAN/NetKAN/actions/runs/24293275091/job/70934011336#step:6:122

For now, I can just leave the install instructions in the netkan file in this repo.

HebaruSan:
Guessing those errors come up when it tries to inflate the empty section of your internal .ckan file; you have the multiple-document splitter without another document:

image

What happens if you fix that?

JonnyOThan:
> Guessing those errors come up when it tries to inflate the empty section of your internal .ckan file; you have the multiple-document splitter without another document:

Ah ok, I generated that from an online json to yaml converter, and I figured it was some kind of yaml marker rather than 2 full objects.

PR #11042, opened 2026-04-17
Updated the metadata to promote RealFuels from a recommendation to a hard dependency.

Iftn1:
Reason:
The current version of EngineAnalyzer now strictly requires RealFuels to function properly.


HebaruSan (approved):
Thanks for the update.

PR #11044, opened 2026-04-18
Add dependency on ClickThroughBlocker

FrittenFritz:


HebaruSan (approved):
Thanks for the update! 🚅

PR #11047, opened 2026-04-18
Don't index prereleases for LunaMultiplayer

HebaruSan:
## Problems

- <https://github.com/LunaMultiplayer/LunaMultiplayer>

This mod recently started uploading nightly builds by replacing the same upload over and over, which has wreaked a bit of havoc with constant hash changes:

- <https://github.com/KSP-CKAN/CKAN-meta/commits/master/LunaMultiplayer/LunaMultiplayer-0.29.1-Draft.ckan>

Then the latest stable release got deleted or replaced or _something,_ with a ZIP file that was missing files and caused inflation errors, with a timestamp that made it more recent than the prerelease, so it looked like nothing was inflating even though the nightly builds were still updating, and we received several PRs attempting to patch that up in CKAN-meta:

- KSP-CKAN/CKAN-meta#3389
- KSP-CKAN/CKAN-meta#3390
- KSP-CKAN/CKAN-meta#3391
- KSP-CKAN/CKAN-meta#3392

It seems pretty clear that the desired end state of this is for these prereleases to not be in CKAN, since the previous one got deleted.

## Changes

Now the bot will ignore prereleases for this mod. This will avoid the problem of constant hash updates as well as make sure the 0.29.0 release is considered the latest regardless of any future shenanigans, so the bot will handle the hash of the restored release automatically.

### DIFF
--- a/NetKAN/LunaMultiplayer.netkan
+++ b/NetKAN/LunaMultiplayer.netkan
@@ -1,12 +1,13 @@
 identifier:   LunaMultiplayer
 name:         Luna Multiplayer Client
 $kref:        '#/ckan/github/LunaMultiplayer/LunaMultiplayer/asset_match/^LunaMultiplayer-Client-Release\.zip$'
-comment:      Version file is outdated
-ksp_version:  '1.12'
+x_netkan_github:
+  prereleases: false
+$vref:        '#/ckan/ksp-avc'
 license:      MIT
 resources:
-  homepage: 'https://lunamultiplayer.com/'
-  manual:   'https://github.com/LunaMultiplayer/LunaMultiplayer/wiki'
+  homepage: https://lunamultiplayer.com/
+  manual:   https://github.com/LunaMultiplayer/LunaMultiplayer/wiki
 tags:
   - plugin
 depends:


DrewBanyai:
The desire was only to remove a single pre-release, 0.29.0-Draft, which was added by accident earlier this week

HebaruSan:
Would you be open to considering this change, as well? The current pre-release shows six automatic hash changes so far today. If you're going to be replacing an asset that frequently, users are going to hit download validation errors pretty often between the change on GitHub and when the CKAN bot updates the metadata (and it'll only get worse when the bot stalls or has downtime). If we exclude prereleases, the user experience would be much more reliable.

DrewBanyai:
If integrating this doesn't remove the current 0.29.1-Draft that's in there now, that would be fine. Once the team is awake tomorrow I can find whichever one of them set up this system and figure out how we can remedy the situation.

HebaruSan:
OK, for now let's leave this in limbo and reconvene after your team has a chance to discuss and reach a decision.

HebaruSan:
Looks like the more recent 0.29.2 version took a more expected trajectory, initially releasing as a pre-release, then converting to a regular release a few days later:

- <https://github.com/KSP-CKAN/CKAN-meta/commits/master/LunaMultiplayer/LunaMultiplayer-0.29.2.ckan>

That should work out fine. Closing as no longer needed.

PR #11053, opened 2026-04-19
Add HotReloadKSP

Phantomical:
I have a new mod! Now it's time to add it to CKAN :)

This one allows you to reload DLLs while the game is running. It automatically handles all the details of actually replacing any instantiated versions of its types (i.e. part/vessel/sceneario modules, or arbitrary MonoBehaviors). It also provides a convenient UI to reload the assembly - either manually or automatically when the dll is modified.

I think everything should be right in the netkan - we'll see how CI turns out. The mod has no dependencies so I think I have a good chance here.

### DIFF
--- a/NetKAN/HotReloadKSP.netkan
+++ b/NetKAN/HotReloadKSP.netkan
@@ -1,0 +1,8 @@
+identifier: HotReloadKSP
+author: steamroller
+$kref: '#/ckan/github/Phantomical/HotReloadKSP'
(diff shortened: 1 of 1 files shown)


Phantomical:
Alright, with v0.1.2 this should be good to go. I have cleaned up the accidental deps left in the internal CKAN file, it just needs to be re-inflated.

HebaruSan (approved):
Thanks for the submission! 🦞

PR #11080, opened 2026-04-28
Fix G3MagnetBoots install

HebaruSan:
image


HebaruSan:
Oof, a bundled `0Harmony.dll`...

PR #11081, opened 2026-04-28
SystemHeat: Copy metadata from metanetkans

JonnyOThan:
eventually we will switch this mod to use embedded .ckan files, and then we can remove any redundant info from here.

Note this also adds SystemHeat-IonEngines and SystemHeat-CryoTanks which were previously missing

- <https://github.com/KSPModStewards/SystemHeat>
- <https://forum.kerbalspaceprogram.com/topic/193909-112x-systemheat-a-replacement-for-the-coreheat-system-july-21/>


HebaruSan (approved):
I'm done making edits. 🌐

PR #11111, opened 2026-05-05
Update Panocean canal to be compatible with Kerbinside (Original, not Remastered)

munktron239:
The original Kerbinside mod doesn't actually conflict with this mod, Caerfinon just wanted to make sure people didn't install the old version. But if you do want to use the old Kerbinside, CKAN won't let you install Panocean.

None of the KK statics added by this mod conflict with anything modified by KS (the nearest base is hundreds of kms away from the canal, actually)

I actually brought this up last year (https://forum.kerbalspaceprogram.com/topic/204511-1125the-panocean-canal-v1022025-06-01/page/2/) but never got around to making a PR.

Although if @caerfinon wishes, I can close this PR if he isn't ok with the change being made

### DIFF
--- a/NetKAN/PanoceanCanal.netkan
+++ b/NetKAN/PanoceanCanal.netkan
@@ -4,8 +4,6 @@ license: MIT
 tags:
   - config
   - buildings
-conflicts:
-  - name: KerbinSide
 depends:
   - name: KerbalKonstructs
   - name: KerbinSideCore


munktron239:
also just noticed this is pr #11111 - neat coincidence! and on the same day CKAN hit 3500 mods!
the universe is just so cool sometimes

HebaruSan:
This conflict was added because of these ALL CAPS admonitions with * * two asterisks before and after * * on the forum thread, repo, and SpaceDock page (still present today as they were in 2021):

<https://forum.kerbalspaceprogram.com/topic/204511-1125the-panocean-canal-v1022025-06-01/>

image

<https://github.com/caerfinon/PanoceanCanal>

image

<https://spacedock.info/mod/2835/Panocean%20Canal>

image

---

Based on that, I thought while reviewing #8725 that installing Kerbin Side Complete alongside this mod would cause problems, and that therefore a `conflicts` relationship should be applied between the two. Since Kerbin Side Complete _Continued_ was not mentioned as a superior alternative, I interpreted that to mean that it also suffered from the same underlying conflict that caused this text to be included.

However, reviewing the forum thread, it's clear that @caerfinon does _not_ understand that this mod does in fact currently have its own direct `conflicts` relationship with Kerbin Side Complete:

image

That means that the `conflicts` relationship is not expected or intended by the mod author, so I think I'm in agreement that it is OK to remove. If the * * ALL CAPS WARNINGS * * could be clarified somehow, I would feel 💯% comfortable with it. (That way we would also not get another user stopping by in the future to complain that we're not obeying the author's intentions by allowing the two to be installed side by side.)

munktron239:
I mostly just made this PR as a thing for me, because i like all of the old bases added by the original KS and was frustrated that I had to uninstall Panocean to use them

HebaruSan (approved):
Alright, you've convinced me. ✈️

caerfinon:
No objections from me.

PR #11118, opened 2026-05-06
RN remove craft file installs

raidernick:
-Stop CKAN from installing the stock craft files as they are broken and conflict with RO installs


HebaruSan (commented):
Thanks for the update! 🐡

PR #11123, opened 2026-05-08
Update KerbalFX dependencies

bimo1d:
Hi,

KerbalFX modules now has new dependencies.


HebaruSan:
Hi @bimo1d, thanks for the update!

There's not much point in putting the dependencies in the `description` when they're already in the `depends` property. It's not like users are going to read it and make different decisions, because they _can't_ install the mod without the dependencies. Would you mind if I removed that?

bimo1d:
Yes, ofc, that makes sense.

And also about BlastFX inflate you can just rename visual effects as FX, if necessary.
Thanks!

HebaruSan (commented):
Thanks for the update! 🦭

PR #11145, opened 2026-05-17
Update Enhanced Astronaut Complex suggested mods

ItchyBrother:
Updates EnhancedAstronautComplex.netkan to list related companion mods under suggests.

This changes the optional companion mod relationship metadata for Enhanced Astronaut Complex:

- CrewQueueTwo
- EarnYourStripes
- FlightTracker
- ContractConfigurator

These are optional/nice-to-have companion mods, so suggests is intentional.

After merge, could the current EnhancedAstronautComplex v1.3.0 metadata be re-inflated or regenerated in CKAN-meta so the suggestions appear in CKAN without waiting for the next EAC release?


HebaruSan (approved):
Thanks for the update. 🐒 

> After merge, could the current EnhancedAstronautComplex v1.3.0 metadata be re-inflated or regenerated in CKAN-meta so the suggestions appear in CKAN without waiting for the next EAC release?

Yup, that's how it works.

PR #11167, opened 2026-05-21
Update Enhanced Astronaut Complex dependency and suggested relationships.

ItchyBrother:
Updates Enhanced Astronaut Complex NetKAN metadata for EAC 1.3.1.

Changes:
- Adds Harmony as a dependency.
- Adds DeepFreeze as a suggested companion mod.

### DIFF
--- a/NetKAN/EnhancedAstronautComplex.netkan
+++ b/NetKAN/EnhancedAstronautComplex.netkan
@@ -6,11 +6,15 @@ license: CC-BY-NC-ND-4.0
 tags:
   - plugin
   - crewed
+depends:
+  - name: Harmony2
 suggests:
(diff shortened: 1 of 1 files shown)


HebaruSan (approved):
Thanks for the update. 📶 

FYI, I rebased and squashed your branch, since it was getting really confusing with old commits from previous PRs, etc. In future it's a good idea to keep your fork's `master` branch in sync with the main repo's `master` branch rather than adding new commits there, and then create new "feature" branches a la `git checkout -b update/eac` to make changes.

ItchyBrother:
Thank you... this is still all new to me...hopefully me trying to figure it out hasn't caused more work for you.

PR #11172, opened 2026-05-22
Satisfy BDModularMissileParts dependency with any fork of BDArmory

tsingshitao-nuke:
BDModularMissileParts also works with BDAplus, but the old dependency causes it to only be installable manually.

- <https://github.com/jrodrigv/BDModularMissileParts>

---

ksp compat add 1.12 1.11 1.10 1.9


HebaruSan:
Hey @tsingshitao-nuke, please fill in the description with an explanation of this change and why you think it makes sense.

tsingshitao-nuke:
> Hey @tsingshitao-nuke, please fill in the description with an explanation of this change and why you think it makes sense.

sorry i forget the  description

HebaruSan:
I'm concerned that the description says "the old dependency causes it to only be installable manually", but this does not seem to be true; I just installed this mod with CKAN without issues. When the description says something that's not correct, that tends to be a red flag that the overall idea might be misguided.

tsingshitao-nuke:
Sorry, what I meant was: when you install BDAplus, because its old dependency "BDAcontinued" conflicts with BDAplus, you need to install it manually.
The error occurs because this mod's .version file says its max KSP version is 1.9.1, but Procedural Parts' dependency "ROutils" was published for 1.12.5, so CKAN thinks they are incompatible.

> I'm concerned that the description says "the old dependency causes it to only be installable manually", but this does not seem to be true; I just installed this mod with CKAN without issues. When the description says something that's not correct, that tends to be a red flag that the overall idea might be misguided.

HebaruSan:
Trying to guess whether this is actually OK, I found a previous precedent #9293 in which a mod that depended on BDArmoryContinued was updated to allow either fork. But that correlated with a release of the mod specifically announcing the support, which has not happened here; the BDModularMissileParts repo still points specifically to the BDArmoryContinued repo in its dependency list.

I'll see if I can contact @jrodrigv to weigh in on this...

HebaruSan (approved):
Heard back from @jrodrigv, and he says he is willing to take your word for it.
Thanks for the update!

PR #11175, opened 2026-05-23
Update SpaceWeatherAndAtmosphericOrbitalDecay.netkan

Aebestach:
Kerbalism is currently listed as a dependency, but it would be more appropriate as a recommended installation. Please merge this change, thank you very much.


HebaruSan (commented):
Thanks for the update! 🌋

PR #11185, opened 2026-05-25
Add SerializationFix

Phantomical:
Just a little mod that makes `[Serializable]` attributes work for types defined in mod dlls.

It has no dependencies. I think I got the ckan right, but we'll see what inflate thinks.

- <https://github.com/Phantomical/KSPSerializationFix>


HebaruSan (approved):
Thanks! 🥣

PR #11191, opened 2026-05-27
Add Chemical Propulsion - Solids

CharleRoger:
This is a new(ish) extra modlet for Chemical Propulsion. The main metadata file is embedded within the mod, like the other Chemical Propulsion extras.


HebaruSan (approved):
Thanks for the submission! 🧪

PR #11196, opened 2026-05-29
KSPArchipelago: Add Kerbal-Konstructs suggested mod

nickdavies:
This mod is not required but highly recommended

Also included in https://github.com/KSP-CKAN/NetKAN/pull/11193 so if that PR lands this one can be abandoned


HebaruSan (approved):
Thanks for the update.

(#11196 is still waiting on a reply acknowledging the risks of the approach suggested in that PR.)

PR #11203, opened 2026-05-31
update conformal decals forge url

drewcassidy:
The length of the old domain name was bugging me, and I can never make up my mind, so I moved to a new domain.

Old one will redirect to the new one for the foreseeable future so this is not urgent


HebaruSan:
FYI, SpaceDock still has the old URL:

image

PR #11218, opened 2026-06-04
Updated Enhanced  Astronaut Complex Suggest to Support

ItchyBrother:
Mod listed under suggest should be supports


HebaruSan:
Hi @ItchyBrother, are you aware of how `supports` works in CKAN, what it does specifically?

ItchyBrother:
Specifically?  Not all the details, but my understanding it is tells users basically that the mod you are installed will play nice with this mod without conflict.

HebaruSan:
No, that's not what it means. You can check the spec for the formal definition:

https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#supports

Basically it's for when your mod adds functionality for a mod. In this case, it looks like it would make more sense to keep these as `suggests`.

HebaruSan:
Looks like all the changes are gone; closing for now.

ItchyBrother:
Thank you...sorry for the confusion.

PR #11227, opened 2026-06-08
Make Kerbalism-Config-RO conflict with BluedogDB

JonnyOThan:
This combination creates a fatal error at game startup, and I see it *constantly* in support forums.


HebaruSan:
@JonnyOThan is this moving forward, or should we close it?

Phantomical:
It is now september. @JonnyOThan is this PR still relevant?

PR #11235, opened 2026-06-10
Add Low End Volumetric Clouds Mod

hkha-3:
Add built from scratch volumetric clouds mod for low end pcs based on True Volumetric Clouds Mod


hkha-3:
Hi I would like to say that this mod is broken. I am very sorry. I am trying to fix it and hopefully 1.0.1 will fix the problem. The game hangs most likely due to race condition with the camera.

HebaruSan:
OK, let us know if there's progress! 👍

hkha-3:
There is progress I am first testing it now.

hkha-3:
I mean I can't today it is 10:38 pm here in the UK

PR #11236, opened 2026-06-11
update(neos): add conflict with WildBlueTools

mwerle:
Since both Nehemiah Engineering Orbital Science and Wild Blue Tools define the "ExposureTime" resource, and since the definitions are not identical, experiments can break in interesting ways when both mods are installed.

To this end, add a conflict to avoid people from running into this issue until it can be resolved.


mwerle:
While it won't resolve it, it will mitigate https://github.com/mwerle/OrbitalMaterialScience/issues/49, https://github.com/mwerle/OrbitalMaterialScience/issues/53, and https://github.com/Angel-125/WildBlueTools/issues/53

HebaruSan (approved):
Thanks for the update! 🪀

PR #10715, opened 2025-08-30
Add TURD configurations from Github sources

Spartwo:
- TURD Ckans
  - https://github.com/KerbalPowers/TURD-StockRecolour
  - https://github.com/KerbalPowers/TURD-MH-BG
  - https://github.com/KerbalPowers/TURD-KerbalFoundries
  - https://github.com/KerbalPowers/TURD-BDArmory
- Adjust authors and display names to originals
- Aviator arsenal ckan file
  - https://spacedock.info/mod/2593/Aviator%20Arsenal%20Continued
  - https://github.com/tetryds/AviatorArsenal
- Tie BD armory into what should be the most modern version ( BDArmory => BDArmoryForRunwayProject) in Armory Dependencies

PR #11271, opened 2026-06-22
OPT TURD Configs

Spartwo:
- <https://github.com/KerbalPowers/TURD-OPT>

### DIFF
--- a/NetKAN/TURD-OPT.netkan
+++ b/NetKAN/TURD-OPT.netkan
@@ -1,7 +1,8 @@
 identifier: TURD-OPT
 name: TURD - OPT Continued
-abstract: Collection of Textures Unlimited recolours for Orbit Portal Techologies (OPT) Continued
-author:  
+abstract: >-
+  Textures Unlimited recolours for Orbit Portal Techologies (OPT) Continued
+author:
   - Manwith Noname
   - Spartwo
 $kref: '#/ckan/github/KerbalPowers/TURD-OPT'
--- a/NetKAN/TURD-OPT.netkan
+++ b/NetKAN/TURD-OPT.netkan
@@ -1,7 +1,9 @@
 identifier: TURD-OPT
 name: TURD - OPT Continued
 abstract: Collection of Textures Unlimited recolours for Orbit Portal Techologies (OPT) Continued
-author: Manwith Noname, Spartwo
+author:  
+  - Manwith Noname
+  - Spartwo
 $kref: '#/ckan/github/KerbalPowers/TURD-OPT'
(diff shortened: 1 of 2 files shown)


NetKAN/TURD-OPT.netkan
```diff
@@ -0,0 +1,18 @@
...
+  - Manwith Noname
+  - Spartwo
+$kref: '#/ckan/github/KerbalPowers/TURD-OPT'
+license: CC-BY-NC-SA-4.0
+tags:
+  - config
+  - graphics
+depends:
+  - name: ModuleManager
+  - name: TexturesUnlimited
```
HebaruSan, line 15: This should depend on ~~TURD and~~ OPT, right?

Spartwo: Unfortunately JadeOfMaar has been doing 5 years of great additions solely in the pre-release branch so the ckan for opt continued would be missing practically everything.

As for TURD, it wasn't included as a dependency in the last glut of paint mods that got accepted so I didn't do so here.

HebaruSan: Wait, so this depends on a (version of a) mod that isn't even in CKAN?

HebaruSan: > As for TURD, it wasn't included as a dependency in the last glut of paint mods that got accepted so I didn't do so here.

Which ones were those, if you don't mind sharing? That sounds like some bad metadata that should be fixed.

Spartwo: https://github.com/KSP-CKAN/NetKAN/pull/10715 is the prior one

Below is the specific version of OPT, its just not going to pull many updates off the github page that I don't want to give the wrong impression to folks downloading the paint.
image

HebaruSan: Thanks, #10715 should not have been merged in that state. I will go clean it up.
Please add the needed dependencies here.

HebaruSan: Oh right, "TURD" is the one where there isn't actually any mod by that name, but rather a family of modlets of which this is one example; the `TexturesUnlimited` dependency covers that side of things.

(Explaining this sort of thing is the reason why pull request descriptions exist and should be filled in. Otherwise reviewing is much slower.)

Spartwo: Textures unlimited is the mechanical dependency yes. I've added OPT continued into the depdency list, hopefully the updates will be released as the main version one day.

HebaruSan (approved):
Thanks for the submission.

PR #11288, opened 2026-06-28
remote tech blackout fix

NovaSpem:
Add Remote Tech BlackoutFix mod.

Realistic radio signal fading for RemoteTech during atmospheric entry.
Depends on RemoteTech and Firefly.

GitHub: https://github.com/NovaSpem/RemoteTechBlackoutFix


HebaruSan:
Added to #11289.

PR #11298, opened 2026-07-01
Add OuterParallax dependency on VertexHeightOblateAdvanced

HebaruSan:
image

Confirmed that this is present in what look like an attempt at internal .ckan files (but with a .yaml extension, so the bot won't use them):

- <https://github.com/jthero3/OuterParallaxContinued/tree/main/CKAN>

FYI to @jthero3.


jthero3:
Darn, I was hoping I got the syntax right. Thanks for fixing it for me though!

PR #11319, opened 2026-07-06
Register NoFairingLimits in CKAN

TheCasualLadHere:
Adds NoFairingLimits mod to CKAN registry.

GitHub: https://github.com/TheCasualLadHere/NoFairingLimits


TheCasualLadHere:
Register NoFairingLimits in CKAN

Phantomical:
This currently installs a file directly under `GameData`. Can you add a top-level directory to the mod zip? That way we can install this correctly.

PR #11373, opened 2026-07-18
Add kOS-Kerbcast (KerbcastKos) from SpaceDock

jonpepler:
Adds [kOS-Kerbcast](https://spacedock.info/mod/4366/Kerbcast), a kOS camera-control addon. It lets a you enumerate vessel Kerbcast cameras and control field of view, pan, and target tracking.

- SpaceDock: https://spacedock.info/mod/4419/kOS-Kerbcast (id 4419)
- Identifier `KerbcastKos` (the install folder and AVC name; the SpaceDock display name is kOS-Kerbcast)

### DIFF
--- a/NetKAN/KerbcastKos.netkan
+++ b/NetKAN/KerbcastKos.netkan
@@ -1,0 +1,11 @@
+{
+    "spec_version": "v1.34",
+    "identifier": "KerbcastKos",
+    "$kref": "#/ckan/spacedock/4419",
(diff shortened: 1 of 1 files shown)


Phantomical:
Your spacedock listing says that says that the mod is MIT but `LICENSE` and github seem to be GPL-3.0. I have updated the config to follow GPL since that looks right, but could you confirm which one is correct?

PR #11381, opened 2026-07-20
Eac 1.5.0 split packages

ItchyBrother:


ItchyBrother:
Hi! Just checking whether anything else is needed from me for this PR. The validation checks are passing, and both v1.5.0 release assets are published. Thanks for reviewing!

PR #11383, opened 2026-07-22
ExoInstruments: require BetterTimeWarpCont, recommend EVE Redux + Spectra

batistitooo:
BetterTimeWarpCont is now required rather than merely recommended, since the mod's time-warp integration is core to its observation-scheduling workflow.

Also adds EnvironmentalVisualEnhancements and Spectra as recommendations: the mod's RC20 astrograph reads live cloud cover from EVE when installed, and Spectra is the stock-system EVE config pack needed for that to actually show anything.


HebaruSan (approved):
Thanks for the update! 🖌️

PR #11428, opened 2026-08-08
Modified for Kerbalism version 3.41

Aebestach:


Clayell:
@Aebestach What are your thoughts on updating the [existing](https://github.com/KSP-CKAN/NetKAN/blob/4a301f5be18c0f4bde5cfbe3557e239e25c00e66/NetKAN/Kerbalism-SystemHeat.frozen) `Kerbalism-SystemHeat` netkan instead of making a new one? That way existing users of `Kerbalism-SystemHeat` on CKAN could update to this directly instead of having to uninstall and install the new stuff. You could probably make either KerbalismSystemHeatCompat or KerbalismSystemHeatCore replace the existing Kerbalism-SystemHeat, don't think there would be a difference.

Aebestach:
> @Aebestach What are your thoughts on updating the [existing](https://github.com/KSP-CKAN/NetKAN/blob/4a301f5be18c0f4bde5cfbe3557e239e25c00e66/NetKAN/Kerbalism-SystemHeat.frozen) `Kerbalism-SystemHeat` netkan instead of making a new one? That way existing users of `Kerbalism-SystemHeat` on CKAN could update to this directly instead of having to uninstall and install the new stuff. You could probably make either KerbalismSystemHeatCompat or KerbalismSystemHeatCore replace the existing Kerbalism-SystemHeat, don't think there would be a difference.

@Clayell This plan sounds more reasonable. I'll fix it in a bit.

PR #11434, opened 2026-08-09
Edit KerbalismCompanionCalculator

Aebestach:
With ValiZockt's permission, I've taken over the continued maintenance of KCC.


Phantomical:
I noticed you made a spacedock listing for KCC, so I pulled that in and got a bunch of metadata from there. Feel free to make a PR to change this if you don't want that attached to the CKAN listing.

PR #11438, opened 2026-08-09
Add TWP2

Clayell:
@Nazfib 

Repo: https://github.com/Nazfib/TransferWindowPlanner2

~~There's probably a min version on mechjeb-dev, but it wasn't listed anyway. Nazfib could you provide this?~~
The double license is also somewhat tricky.


Clayell:
Pretty sure there isn't a way to make the dev one work unfortunately, since MechJeb-dev is on a different metadata repository. Maybe sarbian could include it in theirs?

Nazfib:
Author here.

TWP2 integrates very tightly with some internal APIs in MechJeb. This means that, whenever changes are made in those specific parts of MJ, a new version of TWP2 needs to be released — usually this is just a recompile, but occasionally small code changes are needed. This is also the reason for the separate `dev` version. As of today, the compatible versions of MJ are v2.15.0—v2.15.3 for the regular version of TWP2, and v2.16.0-1695 and later for the dev version.

Since the compatible version ranges of TWP2 and MJ will be changing somewhat often, I assume that it will be easier to maintain a separate internal ckan file in each release zip. As I understand it, I can set the compatible MJ version(s) for each release of TWP2 in there.

Is there any way to have both the dev versions and the regular versions under the same CKAN entry, differentiated only by their compatible versions of MJ? I can make changes to my release process to make this work; fully separate GitHub releases with different version numbers for regular/dev versions, for example.

Regarding the license; TWP2 itself is released under the MIT license. The repo and zip files additionally contain the license texts for some dependencies, as required by their licenses.

Clayell:
> TWP2 integrates very tightly with some internal APIs in MechJeb. This means that, whenever changes are made in those specific parts of MJ, a new version of TWP2 needs to be released — usually this is just a recompile, but occasionally small code changes are needed. This is also the reason for the separate dev version. As of today, the compatible versions of MJ are v2.15.0—v2.15.3 for the regular version of TWP2, and v2.16.0-1695 and later for the dev version.
> 
> Since the compatible version ranges of TWP2 and MJ will be changing somewhat often, I assume that it will be easier to maintain a separate internal ckan file in each release zip. As I understand it, I can set the compatible MJ version(s) for each release of TWP2 in there.

Yep, an internal CKAN makes sense for that. Shouldn't be too hard, I'll make the necessary changes here and make a PR in TWP2. (but only for the non-dev version, because as I stated earlier, the dev version can't really exist on the normal ckan-meta repository)

I'll also note that the kref is fairly fragile to how the release naming is set up, so I hope you're doing it via some automated mechanism rather than by hand. (and I hope you don't change it)

> Is there any way to have both the dev versions and the regular versions under the same CKAN entry, differentiated only by their compatible versions of MJ? I can make changes to my release process to make this work; fully separate GitHub releases with different version numbers for regular/dev versions, for example.

AFAIK no. You can't have 2 different versions of the same CKAN entry.

> Regarding the license; TWP2 itself is released under the MIT license. The repo and zip files additionally contain the license texts for some dependencies, as required by their licenses.

I left it as MIT + CC-BY-4.0 since the mod includes assets from both licenses. The LGPL-3.0 might be technically required by the license but its for ClickThroughBlocker and no other mod includes the license of this mod when listing it as a dependency so I left it out. (it would probably be required if you were actually modifying ClickThroughBlocker in some way, but you're just using it as intended) Same situation as MechJebLib.

Nazfib:
> > Is there any way to have both the dev versions and the regular versions under the same CKAN entry, differentiated only by their compatible versions of MJ? I can make changes to my release process to make this work; fully separate GitHub releases with different version numbers for regular/dev versions, for example.
> 
> AFAIK no. You can't have 2 different versions of the same CKAN entry.
> 
What I was thinking here, is a v0.7.x release series for DevJeb, and a v0.6.x series for regular 'Jeb. The question is whether it would cause problems if (for example) v6.2.0 releases, then v7.2.0, then v6.2.1 — each with an internal CKAN file to determine which versions of MJ it's compatible with.

> > Regarding the license; TWP2 itself is released under the MIT license. The repo and zip files additionally contain the license texts for some dependencies, as required by their licenses.
> 
> I left it as MIT + CC-BY-4.0 since the mod includes assets from both licenses. The LGPL-3.0 might be technically required by the license but its for ClickThroughBlocker and no other mod includes the license of this mod when listing it as a dependency so I left it out. (it would probably be required if you were actually modifying ClickThroughBlocker in some way, but you're just using it as intended) Same situation as MechJebLib.

TWP2's license is MIT only. There is some CC-BY-4.0 work incorporated in it; that does not change the license of TWP2 itself, the only requirement of -BY is attribution and that a link to or copy of the license text is included.

Clayell:
> What I was thinking here, is a v0.7.x release series for DevJeb, and a v0.6.x series for regular 'Jeb. The question is whether it would cause problems if (for example) v6.2.0 releases, then v7.2.0, then v6.2.1 — each with an internal CKAN file to determine which versions of MJ it's compatible with.

I think 2 different CKAN entries would make way more sense for the users. Technically you could make this work by marking all v7 releases as pre-release and all v6 releases as normal, but I think it would just cause too much trouble and confusion. Additionally, the dev version would never be able to specify its dependencies properly as long as the dev Mechjeb is not in the normal CKAN-meta repo.

> TWP2's license is MIT only. There is some CC-BY-4.0 work incorporated in it; that does not change the license of TWP2 itself, the only requirement of -BY is attribution and that a link to or copy of the license text is included.

I've moved the license stuff to https://github.com/Nazfib/TransferWindowPlanner2/pull/11, so I'll leave deciding that up to you (you should be able to edit the PR).

PR #11439, opened 2026-08-09
Add MPE x Parallax Continued Texture-load Patch

Greencus:


Clayell:
The name of the file needs to match the identifier. Another option is to correct the spelling on Spacedock and then request another addition to CKAN.

Greencus:
mb should be good now

Clayell:
`Error: 23669 [1] ERROR CKAN.CmdLine.ConsoleUser (null) - Oh no! We tried to overwrite a file owned by another mod!`
Don't try to overwrite MPE's files. Put the MM patch (or whatever it is) in a new file.
Also, if you're trying to do patches after MPE, it may be better to put it in a different folder from MPE so you can do `:AFTER[MPE]`

Greencus:
Is there no way to have it overwrite files at all? The problem is the patch is just a config change to set some textures to diffuse. I don't believe I can modify the cfg using another file, it would have to be just via replacing the file, no? Is there no way to do something like that I suppose?

Clayell:
A module manager patch can do exactly that (assuming the origin file isn't in PluginData)

Greencus:
oh shit i was not aware of this ok i will make a module manager patch thank you so much for explaining

Greencus:
ok now i changed it to be a module manager cfg file that overwrites the original files, it should work now.

Clayell:
image
image

You're still failing the same check.

Greencus:
i apologize i didnt remember the file in netkan needed the folder to be changed i had just changed the spacedock upload

PR #11452, opened 2026-08-13
Add Thunderbolt

Aebestach:
Add Thunderbolt to CKAN.

GitHub: https://github.com/Aebestach/Thunderbolt
SpaceDock: https://spacedock.info/mod/4469/Thunderbolt
Forum: https://forum.kerbalspaceprogram.com/topic/231558-112x-thunderbolt/

The release zip also contains BlastFX; both GameData/Thunderbolt and GameData/BlastFX are installed.


Phantomical:
Two questions:
* Does this require volumetric clouds? V3 still isn't on CKAN so that would block this from getting added.
* Is this intentionally meant to be two mods? BlastFX looks like it should be a separate mod?

I can make it so there are two CKAN mods that come from one zip, but I'd need to know if the split is intentional.

Aebestach:
1. Volumetric clouds are an "optional" dependency, not a mandatory one.
2. BlastFX is a standalone mod, but I do not intend to release it separately; for now, I only hope to bundle it with Thunderbolt.

PR #11463, opened 2026-08-19
Add Low End Volumetric Clouds mod official continue

hkha-4:
Hi I am hkha-3 I lost my account, so I republished my mod as an Official Continue


Phantomical (changes requested):
The release artifact for this seems to be a zip nested inside another zip? That will need to be fixed before this can be added.

hkha-4:
I have fixed the problem. My compilation process is done by an action. The action was zipping it then github zipped the artifact for download

NetKAN/LowEndVolumetricCloudsMod.netkan
```diff
@@ -0,0 +1,7 @@
+identifier: LowEndVolumetricCloudsMod-Official-Continue
```
Phantomical: I can't commit so here's one thing you need to fix first
```suggestion
identifier: LowEndVolumetricCloudsMod
```

Phantomical (changes requested):

PR #11464, opened 2026-08-20
Harmony2 Required

Angel-125:
Adds NetKAN dependency to SunkWorks and Sandcastle for Harmony2.


HebaruSan (approved):
Thanks for the update. 🎵

PR #11468, opened 2026-08-20
IkosAegis

TAz00:
Vessel PIN locking for LMP

### DIFF
--- a/NetKAN/IkosAegis.netkan
+++ b/NetKAN/IkosAegis.netkan
@@ -1,0 +1,45 @@
+{
+    "//_spec": "v1.27 is the oldest version supporting every feature used here - `tags` needs v1.27, `find` needs v1.4, `filter` is ancient. The spec asks for the oldest that works, so this drops to v1.4 if `tags` is ever removed.",
+    "spec_version": "v1.27",
+
+    "//_identifier": "Deliberately identical to the GameData folder name, the ModuleManager :FOR[] token and the assembly filename. CKAN also identifies manual installs by everything before the first dot of a DLL name under GameData, so IkosAegis.dll makes that a fourth place the same string appears.",
+    "identifier": "IkosAegis",
+    "name": "IkosAegis",
+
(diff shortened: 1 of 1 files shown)


NetKAN/IkosAegis.netkan
```diff
@@ -0,0 +1,45 @@
...
+        {
+            "find": "IkosAegis",
+            "install_to": "GameData",
+            "filter": [ "Thumbs.db", ".DS_Store" ],
+            "filter_regexp": [ "\\.pdb$" ]
+        }
+    ],
+
+    "tags": [ "plugin", "career", "sandbox" ]
+}
```
Phantomical: I can't push to this branch so here's the fixed config as a suggestion

```suggestion
identifier: IkosAegis
name: IkosAegis
abstract: Put a PIN keypad on your ships - lock a craft's controls behind a code.
description: >-
  Adds a numeric keypad lock to every command part - probe cores and crewed pods alike. Set
  a code in the VAB, engage the lock, and the craft will not pitch, roll, yaw, throttle,
  stage or fire an action group until somebody types the code back in. Crew cannot EVA out,
  nobody can board in, and the vessel cannot be recovered or terminated without the PIN
  unless it is parked on a launch pad or runway. One craft has one lock and one PIN however
  many command parts it carries. The two Aegis buttons stay reachable in the part menu while
  locked, so a craft can always be opened again. Attached by ModuleManager, so it covers
  modded command parts as well as stock ones.
author:
  - drebsdorf
$kref: '#/ckan/github/taz00/IkosAegis'
$vref: '#/ckan/ksp-avc'
license: MIT
resources:
  homepage: https://github.com/taz00/IkosAegis
tags:
  - plugin
  - control
depends:
  - name: ModuleManager
  - name: Harmony2
install:
  - find: IkosAegis
    install_to: GameData
    filter:
      - Thumbs.db
      - .DS_Store
    filter_regexp: \.pdb
```

TAz00: Cool, looks good to me, appreciate it

PR #11470, opened 2026-08-20
Add Injection Velocity Indicator

Pzixel:
Adds Injection Velocity Indicator, a KSP 1.12.5 plugin that adds target-relative speed to the stock celestial-body closest-approach tooltip. Reason: high eccentricity orbits (such as moho) are tricky to deal with, you can see the closest approach but have no idea what is the relative speed, which might mean that some upfront burning can actually save a lot on arrival. This helps plotting without using advanced tools such as mechjeb autoplanner.

- Source: GitHub releases (`v1.0.0` published)
- Compatibility: KSP-AVC metadata, KSP 1.12.5
- Install: `GameData/InjectionVelocityIndicator`
- Required dependency: `Harmony2`
- The release does not bundle `0Harmony.dll` or modify another mod's folder

Validated with CKAN-NetKAN 1.36.5.26198. A clean CKAN 1.36.4 test install auto-installed Harmony2 into `GameData/000_Harmony` and installed this mod into its own folder.

### DIFF
--- a/NetKAN/InjectionVelocityIndicator.netkan
+++ b/NetKAN/InjectionVelocityIndicator.netkan
@@ -5,17 +5,8 @@ author: Pzixel
 $kref: '#/ckan/github/Pzixel/injection-velocity-indicator'
 $vref: '#/ckan/ksp-avc'
 license: MIT
-release_status: stable
 tags:
   - plugin
   - information
 depends:
   - name: Harmony2
-resources:
-  homepage: https://github.com/Pzixel/injection-velocity-indicator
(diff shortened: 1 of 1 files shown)


cheese3660:
This is in the legacy json format? Netkans are generally written with the YAML format

HebaruSan:
> This is in the legacy json format? Netkans are generally written with the YAML format

The LLMs that people have been using lately probably train on all of them, and the older ones are all still JSON.

Pzixel:
Oh, sorry, I wasn't aware JSON was deprecated. Now it is fixed, I looked up a couple of merged PRs for reference.

PR #11477, opened 2026-08-22
Add/aviation cockpits recolour

Spartwo:
Manual pull version as the spacedock generated version has bugs. This adds turd paint mask configs for aviation cockpits(mod itself is not on ckan)

Original Request:
https://github.com/KSP-CKAN/NetKAN/pull/11476


HebaruSan:
> (mod itself is not on ckan)

We'll need to add it first, then. Where is it?

Spartwo:
> > (mod itself is not on ckan)
> 
> We'll need to add it first, then. Where is it?

https://spacedock.info/mod/658/Aviation%20Cockpits

HebaruSan:
Ahh, that mod is pretty old, so it may be difficult to get a hold of the author for permission.
Please let us know if you're able to contact them, and we can re-open this then.

PR #11485, opened 2026-08-25
Add Heat Tile Painter

peachyGT:
Adds Heat Tile Painter 0.24.1 by peachyG.

- KSP 1.12.5
- GitHub release with KSP-AVC metadata
- Depends on ModuleManager
- Installs GameData/HeatTileOverlayPrototype
- Forum: https://forum.kerbalspaceprogram.com/topic/231652-1125-heat-tile-painter-0241-very-simple-way-to-add-heat-shielding-to-sstosspaceplanes/
- MIT License

Validated locally with netkan.exe v1.36.5.26198; inflation succeeds and the generated SHA-256 matches the published release asset. The validator reports a nonfatal manual-install detection warning because the legacy assembly and folder name remain HeatTileOverlayPrototype for saved-craft compatibility.

### DIFF
--- a/NetKAN/HeatTilePainter.netkan
+++ b/NetKAN/HeatTilePainter.netkan
@@ -1,0 +1,27 @@
+spec_version: v1.26
+identifier: HeatTilePainter
+name: Heat Tile Painter
+abstract: Paint heat-tile patterns or flat colors onto craft with editable layers and optional heat protection.
+author: peachyG
+license: MIT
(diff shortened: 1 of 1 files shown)


Phantomical:
One thing to double check before I merge this: the mod zip has the mod folder in `GameData` as `HeatTileOverlayPrototype`. Is that the folder name you want?

peachyGT:
> One thing to double check before I merge this: the mod zip has the mod folder in `GameData` as `HeatTileOverlayPrototype`. Is that the folder name you want?

Hi! Yes, that's the correct mod folder. Thanks!

PR #11497, opened 2026-08-30
Fix Kerbal Proportions release source

NickDeVeau:
Kerbal Proportions is published to GitHub first and mirrored to SpaceDock. The current two-document metadata fails whenever the mirrors temporarily have different latest versions (currently: `Generated 2 modules but only 1 requested`).

This makes the GitHub release asset authoritative, reads version compatibility from the packaged KSP-AVC file, and keeps SpaceDock in `resources`. It also preserves the package install rule and public author name used by the project.

### DIFF
--- a/NetKAN/KerbalProportions.netkan
+++ b/NetKAN/KerbalProportions.netkan
@@ -1,11 +1,23 @@
+spec_version: v1.4
 identifier: KerbalProportions
-$kref: '#/ckan/github/NickDeVeau/Kerbal-Proportions'
-x_netkan_version_edit: ^v?(?<version>.+)$
----
-identifier: KerbalProportions
-$kref: '#/ckan/spacedock/4513'
+name: Kerbal Proportions
+abstract: Live visual rig and proportion editor for EVA and IVA Kerbals
+author: Biggus
+license: MIT
+$kref: '#/ckan/github/NickDeVeau/Kerbal-Proportions/asset_match/^KerbalProportions-v.+\.zip$'
 $vref: '#/ckan/ksp-avc'
+x_netkan_github:
+  use_source_archive: false
+  prereleases: false
 tags:
   - plugin
   - graphics
   - crewed
+install:
+  - find: KerbalProportions
+    install_to: GameData
+resources:
+  homepage: https://github.com/NickDeVeau/Kerbal-Proportions
+  spacedock: https://spacedock.info/mod/4513
+  repository: https://github.com/NickDeVeau/Kerbal-Proportions
+  bugtracker: https://github.com/NickDeVeau/Kerbal-Proportions/issues


NetKAN/KerbalProportions.netkan
```diff
@@ -1,11 +1,23 @@
+spec_version: v1.4
```
Phantomical, line 1: This should not be in the netkan
```suggestion
```

NetKAN/KerbalProportions.netkan
```diff
@@ -1,11 +1,23 @@
...
-$kref: '#/ckan/spacedock/4513'
+name: Kerbal Proportions
+abstract: Live visual rig and proportion editor for EVA and IVA Kerbals
+author: Biggus
+license: MIT
+$kref: '#/ckan/github/NickDeVeau/Kerbal-Proportions/asset_match/^KerbalProportions-v.+\.zip$'
 $vref: '#/ckan/ksp-avc'
+x_netkan_github:
+  use_source_archive: false
+  prereleases: false
```
Phantomical, line 11: This should generally not be in a netkan
```suggestion
```

NetKAN/KerbalProportions.netkan
```diff
@@ -1,11 +1,23 @@
...
-$kref: '#/ckan/github/NickDeVeau/Kerbal-Proportions'
-x_netkan_version_edit: ^v?(?<version>.+)$
----
-identifier: KerbalProportions
-$kref: '#/ckan/spacedock/4513'
+name: Kerbal Proportions
+abstract: Live visual rig and proportion editor for EVA and IVA Kerbals
+author: Biggus
+license: MIT
+$kref: '#/ckan/github/NickDeVeau/Kerbal-Proportions/asset_match/^KerbalProportions-v.+\.zip$'
```
Phantomical, line 7: If you don't have multiple zip assets (and don't plan to in the near future) then asset match is not needed
```suggestion
$kref: '#/ckan/github/NickDeVeau/Kerbal-Proportions'
```

NetKAN/KerbalProportions.netkan
```diff
@@ -1,11 +1,23 @@
...
   - graphics
   - crewed
+install:
+  - find: KerbalProportions
+    install_to: GameData
+resources:
+  homepage: https://github.com/NickDeVeau/Kerbal-Proportions
+  spacedock: https://spacedock.info/mod/4513
+  repository: https://github.com/NickDeVeau/Kerbal-Proportions
+  bugtracker: https://github.com/NickDeVeau/Kerbal-Proportions/issues
```
Phantomical, line 23: These get automatically set from the github repo and should not be included here if not needed.

NetKAN/KerbalProportions.netkan
```diff
@@ -1,11 +1,23 @@
...
+x_netkan_github:
+  use_source_archive: false
+  prereleases: false
 tags:
   - plugin
   - graphics
   - crewed
+install:
+  - find: KerbalProportions
+    install_to: GameData
```
Phantomical, line 18: This gets automatically added from the name, better to remove it here. It is only needed if the name and install folder are different or if you need to filter out files.

Phantomical (commented):
I'm not really sure why you're doing this TBH. Are you intentionally publishing releases to github and not spacedock? If you aren't then the netkan bot will automatically make the new release available once you have published to both.

PR #11500, opened 2026-08-31
Added KerbalAdvancedDefenseReborn

narni-tchereskov:
added on behalf of Iftn via netkan file

Author: Iftn

* https://github.com/Iftn1/Kerbal-Advanced-Defense-Reborn

Refer to autogenerated netkan for further info:
https://github.com/KSP-CKAN/NetKAN/pull/11494

### DIFF
--- a/NetKAN/KerbalAdvancedDefenseReborn.netkan
+++ b/NetKAN/KerbalAdvancedDefenseReborn.netkan
@@ -1,0 +1,15 @@
+identifier: KerbalAdvancedDefenseReborn
+$kref: '#/ckan/github/Iftn1/Kerbal-Advanced-Defense-Reborn'
+---
+identifier: KerbalAdvancedDefenseReborn
+$kref: '#/ckan/spacedock/4522'
+$vref: '#/ckan/ksp-avc'
+tags:
+  - parts
+  - combat
+depends:
+  - name: ModuleManager
+  - name: BDArmoryForRunwayProject
+install:
+  - find: GameData
+    install_to: GameData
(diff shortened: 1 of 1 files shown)


NetKAN/KerbalAdvancedDefenseReborn.netkan
```diff
@@ -0,0 +1,16 @@
...
+$vref: '#/ckan/ksp-avc'
+tags:
+  - parts
+  - combat
+depends:
+  - name: ModuleManager
+  - name: BDArmoryForRunwayProject
+install:
+  - find: GameData
+    install_to: GameData
```
Phantomical: The file structure inside the [release zip here](https://github.com/Iftn1/Kerbal-Advanced-Defense-Reborn/releases/tag/V0.3.5) is
```
Kerbal-Advanced-Defense-Reborn-main
└───KerbalAdvancedDefenseReborn
```
but this section is expecting it to be
```
GameData
└───KerbalAdvancedDefenseReborn
```

@lftn1 we can either change it here to use what you have now, or you can update the zip to have `GameData` at the root. My recommendation is to update the zip. That way it will be less confusing for people installing manually.

narni-tchereskov: Will changing the current .zip folder structure in the release also require a change in SpaceDock? It seems to have the same issue there.

Phantomical: Yeah both would need to be updated. If they aren't the netkan bot won't index it.

Iftn1: I have changed the folder name inside the zip file on Github and SpaceDock.
image

PR #11522, opened 2026-09-07
Add DimensionSync

sparr:
First-time submission, of my own mod.

DimensionSync is an editor plugin that keeps the dimensions of connected parts in step: resizing a procedural part carries the change along the stack, and a wing segment's tip dimensions carry to the root of the segment outboard of it. It knows about ProceduralParts, ROLib, Procedural Fairings, SSTU, and B9 Procedural Wings out of the box, and its field catalogue is ModuleManager-patchable so other mods can be added without a code change.

- Repo: https://github.com/sparr/ksp-mod-DimensionSync
- Release: https://github.com/sparr/ksp-mod-DimensionSync/releases/tag/v0.2.0
- License: MIT
- KSP 1.12.0–1.12.99

Inflated locally against the v0.2.0 release with `kspckan/metadata` before opening this: exit 0, no warnings, installs and uninstalls cleanly in a dummy instance.

Two notes in case they save a review round-trip:

- `release_status: development` is deliberate rather than a leftover. The mod is tested and works, but it is young and I would rather set expectations low since I am confident there are use cases I haven't tried and edge misbehaviors I haven't triggered. Happy to drop the field if you would prefer it left to default.
- The `suggests` are the mods it has explicit support for, plus `Harmony2`, which it uses optionally to see other mods' writes.

The content and description of this PR were approximately 90% authored by Claude Opus 5.

### DIFF
--- a/NetKAN/DimensionSync.netkan
+++ b/NetKAN/DimensionSync.netkan
@@ -15,7 +15,7 @@ tags:
 recommends:
   - name: ModuleManager
 depends:
-  - Harmony2
+  - name: Harmony2
 suggests:
   - name: ProceduralParts
   - name: ProceduralFairings
--- a/NetKAN/DimensionSync.netkan
+++ b/NetKAN/DimensionSync.netkan
@@ -1,0 +1,26 @@
+identifier: DimensionSync
+name: Dimension Sync
+abstract: >-
+  Keeps the dimensions of connected parts in step: resize one procedural part
+  and the change travels along the stack, or from one wing segment's tip to the
+  next segment's root.
+author: sparr
+$kref: '#/ckan/github/sparr/ksp-mod-DimensionSync'
+$vref: '#/ckan/ksp-avc'
+license: MIT
+tags:
+  - plugin
+  - editor
+  - convenience
+recommends:
+  - name: ModuleManager
+depends:
+  - Harmony2
+suggests:
+  - name: ProceduralParts
+  - name: ProceduralFairings
+  - name: B9-PWings-Fork
+  - name: ROLib
+install:
+  - find: DimensionSync
+    install_to: GameData


NetKAN/DimensionSync.netkan
```diff
@@ -0,0 +1,26 @@
...
+name: Dimension Sync
+abstract: >-
+  Keeps the dimensions of connected parts in step: resize one procedural part
+  and the change travels along the stack, or from one wing segment's tip to the
+  next segment's root.
+author: sparr
+$kref: '#/ckan/github/sparr/ksp-mod-DimensionSync'
+$vref: '#/ckan/ksp-avc'
+license: MIT
+release_status: development
```
Phantomical: Leaving this in will just mean that the mod doesn't get indexed by CKAN.

NetKAN/DimensionSync.netkan
```diff
@@ -0,0 +1,26 @@
...
+  - editor
+  - convenience
+recommends:
+  - name: ModuleManager
+suggests:
+  - name: ProceduralParts
+  - name: ProceduralFairings
+  - name: B9-PWings-Fork
+  - name: ROLib
+  - name: Harmony2
```
Phantomical: Harmony2 should just be a depends. Basically nobody installing your mod will know what Harmony2 does, and it does not change any game behaviour.

NetKAN/DimensionSync.netkan
sparr: At all? Or just not shown by default?

NetKAN/DimensionSync.netkan
sparr: Having it or not here does produce a behavior change. This mod is more aware of other mods' changes to parts with Harmony than without.

NetKAN/DimensionSync.netkan
Phantomical: It won't be shown to anyone unless they have their stability tolerance set to `development`. In any case we don't want you explicitly setting `release_status` in netkans. If you want the mod to be marked as a prerelease then you can make your github releases prereleases (or just document it in the readme).

NetKAN/DimensionSync.netkan
```diff
@@ -0,0 +1,26 @@
...
+$vref: '#/ckan/ksp-avc'
+license: MIT
+tags:
+  - plugin
+  - editor
+  - convenience
+recommends:
+  - name: ModuleManager
+depends:
+  - Harmony2
```
Phantomical:
```suggestion
  - name: Harmony2
```

Phantomical (changes requested):
One small typo, otherwise this LGTM

PR #11538, opened 2026-09-10
Fix ContractConfigurator-KSP2ForScience craft installs

HebaruSan:
This mod installs some craft files. In #11348, the bot set them up to be installed to `Ships/VAB`, but contract packs' craft files are supposed to be installed where the mod puts them in GameData.

However, after the mod was indexed, the author updated it to look in `Ships/VAB`! But this broke the `install` stanza that looked for them in GameData.

image

Now the GameData filter and craft-specific installs are removed, and a new one is added for `Ships/VAB`.


HebaruSan:
Oh. Huh.

image

PR #11539, opened 2026-09-11
Add Tilt'Em Continued

ballisticfox:
Add Tilt'Em Continued, pointing at the new repo here: https://github.com/ballisticfox/TiltEm-Continued


NetKAN/TiltEm-Continued.netkan
```diff
@@ -0,0 +1,22 @@
...
+x_netkan_github:
+  asset_match: '^TiltEm-Continued_v\d+\.\d+\.\d+\.zip$'
+install:
+  - find: TiltEm
+    install_to: GameData
+author:
+  - ballisticfox
+  - dagger
+resources:
+  homepage: 'https://forum.kerbalspaceprogram.com/index.php?/topic/231670-*'
```
Phantomical: You can set this as the `website` link on the github repo and it will automatically be added as the homepage here. I usually recommend that people do that.

NetKAN/TiltEm-Continued.netkan
```diff
@@ -0,0 +1,22 @@
+identifier: TiltEm-Continued
+abstract: >-
+  Working axial tilts for KSP!
+$kref: '#/ckan/github/ballisticfox/TiltEm-Continued'
```
Phantomical: Do you have a `.version` file in the zip? (I haven't had a chance to look yet). If not you will want to specify a supported version range for KSP. If you do then you should add a `$vref` here for it.

ballisticfox: Nope, I do not. I've switched to using `version_from_asset` and specified `ksp_version_min` and `ksp_version_max` values.

NetKAN/TiltEm-Continued.netkan
ballisticfox: Added to the repository, does this mean I can just drop the homepage definition there?

Phantomical: Yep! The netkan will pick it up automatically.

NetKAN/TiltEm-Continued.netkan
Phantomical: `version_from_asset` isn't needed unless you plan to have multiple zip release artifacts. Otherwise I think it is usually preferable to get the version from the github tag.

PR #11551, opened 2026-09-14
Update SpinDrive relationships

QuantumsHevy:
Corrects the relationships for [SpinDrive](https://github.com/NatoKerbal/Spin-Drive). Includes the mods that the mod has support _for_ but does not necessarily add meaningful content _to_, hence only the `recommends` tags. Verified all identifiers against what shows up within the CKAN application.

I would like to ask before merge, what does `$vref: '#/ckan/ksp-avc'` do? I've seen it in other `.netkan` files and I'm wondering if it should be included under SpinDrive's identifiers, too.


HebaruSan (approved):
Thanks for the update! 🌻 

> I would like to ask before merge, what does $vref: '#/ckan/ksp-avc' do? I've seen it in other .netkan files and I'm wondering if it should be included under SpinDrive's identifiers, too.

The official documentation for that is in the spec:

- <https://github.com/KSP-CKAN/CKAN/blob/master/Spec.md#ckanksp-avcpathavcfilenameversion>

To sum up, it makes the bot look for a `.version` file in the ZIP and import its metadata (usually game version compatibility info) into the resulting `.ckan` file. Based on the lack of a warning about this on [the status page](https://status.ksp-ckan.space/), it looks like your mod doesn't have a `.version` file and so wouldn't need a `$vref` either.

PR #11553, opened 2026-09-14
ExoInstruments: no hard dependencies, version file, current description

batistitooo:
Brings the metadata in line with ExoInstruments v0.5.0 (https://github.com/batistitooo/ExoInstruments/releases/tag/v0.5.0) and with the netkan the mod's repository carries.

- No hard dependencies. The mod has no Module Manager patches and no Kerbal Konstructs statics, and it reaches BetterTimeWarpContinued through reflection only when it is installed, falling back to stock warp rates otherwise. BetterTimeWarpCont and EnvironmentalVisualEnhancements move to recommends, Spectra to suggests.
- `$vref: '#/ckan/ksp-avc'`: the release zip carries ExoInstruments.version, which gives KSP 1.8.0 to 1.12.5.
- The abstract and description describe the mod as it is now: an observatory photographing the sky onto real detector models, alongside the exoplanet survey.
- An explicit install stanza for GameData/ExoInstruments.


Phantomical:
You might want to consider using an internal ckan file in the future. Just put it in the release zip under `GameData/ExoInstruments` ([look at BackgroundThrust's  for an example](https://github.com/Phantomical/BackgroundThrust/blob/main/BackgroundThrust.ckan)).

PR #11567, opened 2026-09-17
Add SEP-KAS1-Patch

celino:
ModuleManager patch that makes the Surface Experiment Pack plug work with KAS 1.x. The SEP entry is frozen at 2.7.1 and upstream has been unmaintained since 2019; the same fix was submitted upstream to CobaltWolf/Surface-Experiment-Pack.

Depends on SurfaceExperimentPack, KAS ≥ 1.0, KIS, ModuleManager. Release: https://github.com/celino/SEP-KAS1-Patch/releases/tag/v1.0.0


celino:
Thanks for the cleanup, @Phantomical. I reproduced the Inflate failure locally with `kspckan/metadata` + `ckanmetatester`: the inflation itself passes. The install fails on the fake 1.12.5 install with `Unsatisfied dependency SurfaceExperimentPack (KSP 1.0.5–1.6.1)`, because CKAN-meta still caps SEP v2.7.1 at 1.6.1.

I opened KSP-CKAN/CKAN-meta#3420 to raise that to 1.12.99 (SEP runs on 1.12.5; the plug is the only broken part, and that's what this patch fixes). Once it's merged, a CI re-run here should pass.

